PR #279, opened 2015-01-13
Exclude repos with missing deps as goodrepos.

aplanas:
A good repo now needs to be compiled and also no missing packages.

### DIFF
--- a/create_test_Factory_dvd-1.testcase
+++ b/create_test_Factory_dvd-1.testcase
@@ -21,6 +21,7 @@ job lock name cups
 job lock name ghostscript
 job lock name graphviz
 job lock name gtk2-branding-upstream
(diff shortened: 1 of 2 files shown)


aplanas:
Under review in packagelists.

nilxam:
if this already deployed on packagelists, then I wonder "ERROR -- An exception happends while checking a group [list index out of range]" message is related or not...

aplanas:
For some reason I didn't receive a github email to this comment.  I amended the commit, changing the order when the test cycle is done.  Now there is the possibility that goodrepos become empty.

aplanas:
Do not merge this. Is completely crap.

coolo:
someone had to say it :)

PR #280, opened 2015-01-16
staging accept: create new links for packages when a new .spec file appears

DimStar77:
I miss some good way to test that! so this is all very theoretical here...

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -2,7 +2,7 @@ import re
 from xml.etree import cElementTree as ET
 
 from osc.core import change_request_state
-from osc.core import http_GET, http_PUT
+from osc.core import http_GET, http_PUT, http_DELETE
 from datetime import date
 from osclib.comments import CommentAPI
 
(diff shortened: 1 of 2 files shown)


osclib/accept_command.py
```diff
@@ -56,9 +56,9 @@ def perform(self, project):
             msg = 'Accepting staging review for {}'.format(req['package'])
             print(msg)
 
-        for req in requests:
```
DimStar77: I needed to merge this loop in the above one, as otherwise I do not have information about the package name.
get_package_for_request_id works based on the prj metadata, so this could potentially be an option, but sounds like an overkill (as I don't see much benefit of first accepting staging review for each package, and then the package, as opposed to doing the two accepts in one go, then moving to the next package)

coolo:
see docs/testing - it's possible but generally hard.

DimStar77:
I will change this a bit based on the last submit and use some of the helper methods introduced there (load_from_file)

osclib/accept_command.py
```diff
@@ -80,12 +81,48 @@ def accept_other_new(self):
...
+        filelist = self.api.get_filelist_for_package(pkgname=pkgname, project=project, extension='spec')
+        removedspecs = set(oldspeclist) - set(filelist)
+        for spec in removedspecs:
+            # Deleting all the packages that no longer have a .spec file
+            url=makeurl(['source', project, spec[:-5]])
+            print "Deleting package %s from project %s" % (spec[:-5], project)
+            try:
+                http_DELETE(url)
+            except:
+                # the package link was not yet created, which was likely a mistake from earlier
```
coolo: we discussed that before - don't do that

osclib/stagingapi.py
```diff
@@ -147,6 +147,35 @@ def get_package_information(self, project, pkgname, rev=None):
...
+        try:
+            content = http_GET(url)
+            for entry in ET.parse(content).getroot().findall('entry'):
+                filelist.append(entry.attrib['name'])
+        except urllib2.HTTPError, err:
+            if err.code == 404:
+                # The package we were supposed to query does not exist
+                # we can pass this up and return the empty filelist
+                pass
+
```
coolo: this pass seems to be pretty useless - or the if around it - you catch the exception in all cases without further comment.

You might want to rethrow the err in case of !404

PR #281, opened 2015-01-19
Progress5958

DimStar77:
Please give a review and let me know if I do something completely silly - or if I miss something

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -907,6 +907,50 @@ class StagingAPI(object):
             return False
         return True
 
+    def package_version(self, project, package):
+        """
+        Return the version of a package, None in case the package does not exist
+        The first non-commented Version: tag found is used.
+        :param project: the project the package resides in
+        :param package: the package to check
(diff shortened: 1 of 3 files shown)


osclib/stagingapi.py
```diff
@@ -888,21 +888,70 @@ def build_switch_staging_project(self, target_project):
...
+        if not self.item_exists(project, package):
+            return None
+
+        version = None
+
+        specfile = self.api.load_file_content(project, package, '{}.spec'.format(package))
+        if specfile:
+            try:
+                version = re.findall("Version:(.*)",specfile,re.MULTILINE)[0].strip()
+            except:
```
coolo: not sure how findall works, but you want to anchor the Version: at line start.

coolo:
LGTM

osclib/stagingapi.py
DimStar77: Good catch... Or at least it must not be behind a # (which is implicit if it is at the beginning of a line).

Will update the PR in a bit

PR #283, opened 2015-01-20
Fix saving kiwi

DimStar77:


coolo:
I merged that locally and will push it once it works :)

PR #288, opened 2015-02-09
Locking: Let's get some locking mechanism so we hopefully step less on

DimStar77:
each others toes.

That should give us at least some locking mechanism... it's a global lock over openSUSE:Factory:Staging, as some tasks are just too intrusive.


aplanas:
Integrated in PR https://github.com/openSUSE/osc-plugin-factory/pull/289

PR #290, opened 2015-02-11
Detect if a cycle provides a new package, if not, ignore the cycle.

aplanas:


coolo:
I can't help reviewing this

DimStar77:
I think the code changes themselves make sense, but the warning should be put into some comment on the staging project;

as the repo-checker runs in a loop over everything, there is close to no way anybody would see this warning if it happens (which would mean we could as well just not warn at all).

aplanas:
@DimStar77 You are right.

* Remove the comment when there are not new packages in the cycle.

PR #294, opened 2015-02-19
Refactor configuration parameters into configuration file.

aplanas:


aplanas:
Still a WIP, but I make an effor to remove all the configuration (openSUSE, Factory, by_group, by_user, product, rings...) from the code and put it into the configuration file.

PR #301, opened 2015-02-27
staging: Add --debug parameter, which allows to easily have more debug output

DimStar77:
On the way to tracking the issues lately seen, I need more debug; and I don't really feel like ripping it out at the end again... this helper might be useful in the future too.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -73,6 +73,8 @@ def _full_project_name(self, project):
               help='use the old check algorithm')
 @cmdln.option('-v', '--version', action='store_true',
               help='show version of the plugin')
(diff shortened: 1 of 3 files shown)


osc-staging.py
```diff
@@ -73,6 +73,8 @@ def _full_project_name(self, project):
               help='use the old check algorithm')
 @cmdln.option('-v', '--version', action='store_true',
               help='show version of the plugin')
+@cmdln.option('--debug', action='store_true',
+              help='enable debug output')
 def do_staging(self, subcmd, opts, *args):
```
aplanas: We can use the -d option from osc itself. The debug variable can be tested from osc.conf.config

PR #302, opened 2015-03-02
given the correct ring_dvd path

nilxam:
looks like self.crings is indicating the correct path of rings dvd prj, self.project is "openSUSE:Factory" I think

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -685,7 +685,7 @@ class StagingAPI(object):
         if project.endswith(':DVD'):
             return project  # not yet
 
-        ring_dvd = '{}:2-TestDVD'.format(self.project)
+        ring_dvd = '{}:2-TestDVD'.format(self.crings)
         if self.ring_packages.get(pkg) == ring_dvd:
             return project + ":DVD"
 


DimStar77:
I was first with that :package: 
https://github.com/DimStar77/osc-plugin-factory/commit/3dcdb9709f617f6ec4a5bd0db90d92e14189a022

nilxam:
aha, https://github.com/openSUSE/osc-plugin-factory/pull/301 had 2 commits included, sorry for  I was missed that :P

let's closing

nilxam:
after discuss with DimStar, reopened this PR then, this PR is shorter and will/can quickly merge than #301

nilxam:
@aplanas if we all agree this is the correct fix(I guess we are :) , could you review and merge this PR? :)

aplanas:
Uau! @DimStar77 good catch.  I introduced this bug here:
https://github.com/aplanas/osc-plugin-factory/commit/c9d1e06ca80f231a87703ec24cbe5c6dd9386ed7#diff-f426bb4d7b189a98f1b6d86f95d7919bR688

PR #312, opened 2015-03-19
Add PowerPC:ToTest

k0da:

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -47,6 +47,9 @@ class ToTestBase(object):
     def iso_prefix(self):
         return self.project
 
+    def arch(self):
+        return self.arch
+
     def binaries_of_product(self, project, product):
         url = self.api.makeurl(['build', project, 'images', 'local', product])
         try:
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -240,26 +243,30 @@ def factory_snapshottable(self):
 
         if not self.all_repos_done('openSUSE:%s' % self.project):
             return False
+        if self.arch == "x86_64":
+            for product in ['_product:openSUSE-ftp-ftp-i586_x86_64',
```
coolo: shouldn't this list of products come out of the ToTest class instead of depending on the arch?

k0da:
ok updated.

@DimStar77 what do you think about getting rid of arch specific livecd products?

DimStar77:
@k0da The LiveCDs are still one of the quickest way to get anything tested and demoed (their quality is not the best ever, but the effect of showing something is still there).

We could probably away with only one arch instead of multiple, but I don't think we should sacrifice it completely at this moment.

coolo:
he talks about the .i586 in the package name

DimStar77:
Ah; that actually makes sense: looking at openSUSE:Factory:Live/kiwi-image-livecd-kde.x86_64 for example, it's a unmodified link to openSUSE:Factory:Live/kiwi-image-livecd-kde.i586 (minus the other .spec files)... so splitting it in two does not really make a lot of sense

k0da:
@DimStar77 so we're fine then? could you please review my changes?

DimStar77:
Seems all right to me

coolo:
will you also fix the Live project?

k0da:
@DimStar77 ^^

DimStar77:
I can do the changing of the names (dropping the .x86_64 packages linking) and renaming the .i586 to not have a suffix; but please only after the 0318 moves over to QA

coolo:
that's what I thought

PR #317, opened 2015-03-20
Adjust Ring2 dvd for PowerPC

k0da:


nilxam:
ok, doesn't affect x86_64 :)

PR #318, opened 2015-03-21
don't test for arch in places that aren't exactly arch specific

coolo:


coolo:
this fixes livecd release

PR #319, opened 2015-03-24
ppc fixes

k0da:


totest-manager.py
```diff
@@ -373,7 +376,7 @@ def arch(self):
 
     # for Factory we check the version of the release package
     def current_version(self):
-        url = self.api.makeurl(['build', 'openSUSE:%s' % self.project, 'standard', 'x86_64',
+        url = self.api.makeurl(['build', 'openSUSE:%s' % self.project, 'standard', self.arch(),
```
nilxam: I can't see the point for this line change, ToTestFactory.arch() is always return x86_64 no? as ToTestFactoryPowerPC.arch() always return ppc64le.

coolo: makes it easier to do copy&paste right though

k0da: Right. I changed it to ToTestFactoryPowerPC as well

PR #322, opened 2015-03-25
Add new review bot that checks that request descriptions contain certain tags (bnc#, boo#, etc.)

antlarr:
Add new review bot that checks that request descriptions contain certain tags (bnc#, boo#, etc.) and also a test script that tests different type of requests including and not including some of those tags.


tests/checktags_tests.py
```diff
@@ -0,0 +1,495 @@
...
+        self.assertFalse(result['state_accepted'])
+
+        # second time request is in state new so we can accept
+        # first time request is in in review
+        self.checker.set_request_ids(['770001'])
+        self.checker.check_requests()
+
+        self.assertFalse(result['state_accepted'])
+
+    def test_non_maintainer_submit(self):
```
coolo: kind of obvious what you copied from :)

coolo:
The request description is not what we care about - we want to check the (possibly many) .changes file diff.

tests/checktags_tests.py
antlarr: hehe, never tried to hide it. I copied and adapted some tests from maintenance_tests and factory_source_tests :)

antlarr:
ok, I'm working on the .changes file(s) diff now

antlarr:
aplanas, thanks for the suggestion. It's nicer with super indeed. I already added it but will commit it tomorrow morning together with the modifications to check the tags in the .changes files diff instead of the request description, which I've just finished implementing.

coolo:
LGTM

coolo:
https://build.suse.de/request/show/53266 uses "bnc #NR" and that should be acceptable too I guess. At least OBS parses it.

coolo:
on the other hand our documentation is clear: "no whitespace between prefix and number"

PR #323, opened 2015-03-27
allow groups to be reviewer too

coolo:


ReviewBot.py
```diff
@@ -214,14 +214,13 @@ def _get_linktarget(self, src_project, src_package):
...
-    def can_accept_review(self, request_id, user):
-        """return True if there is a new review for the specified user"""
+    def can_accept_review(self, request_id):
+        """return True if there is a new review for the specified reviewer"""
         states = set()
         url = osc.core.makeurl(self.apiurl, ('request', str(request_id)))
         try:
             root = ET.parse(osc.core.http_GET(url)).getroot()
-            states = set([review.get('state') for review in root.findall('review') if review.get('by_user') == user])
+            states = set([review.get('state') for review in root.findall('review') if review.get('by_user') == self.review_user])
```
lnussel: there is no check by group here

lnussel:
travis failed ...

coolo:
@aplanas I'm trying for 3 weeks now to find time, but I'm unable (next week again on business trip ;(

Can you please twist this to pass?

aplanas:
@coolo: sure, let me check.

coolo:
I see. But this was on purpose as I'm abusing the staging review group for this. So I do one round of possibly-declines with check_tags and then do the normal staging list. So I don't want the check_tags to decline. I guess I should do this either in an extra commit (with different tests) or add another reviewer

coolo:
ok, I now split the reviews into two. So the sle-changelog-checker group has a defined task - and we can merge this. Fine with you @lnussel

PR #327, opened 2015-04-02
Avoid double quoting on http_GET

DimStar77:
http_GET is supposed to retreive the url as per request. url is
generally constructed using makeurl, which already encodes it.

In the rare case where we construct the url ourselves (openQA api
requests), we take care of quoting before passing it on to http_GET

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -78,7 +78,6 @@ class StagingAPI(object):
         return makeurl(self.apiurl, l, query)
 
     def retried_GET(self, url):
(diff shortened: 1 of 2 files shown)


coolo:
you only want to quote the query parts not the full URL IMO

DimStar77:
Fair point, I will update this - my change was a like for like replacement (the quoting caused some issues that any _product:* resulted in error 400, invalid productid)

lnussel:
I wonder why the code doesn't just use osc.core.makeurl which handles the quoting already

coolo:
good point, I will redo the whole thing

PR #328, opened 2015-04-04
Unset nullglob after counting RPMs

nilxam:
If nullglub sets, seems it will affect the following osc api call for finding matched projects as like as the error below,

osc api: takes exactly 1 argument (0 given)

thus unset nullglob after counting RPMs.


lnussel:
that means quoting is incorrect on that osc api call

nilxam:
hmm...but I can't see any quoting problem in that osc api call,
projects=$(osc api /search/project/id?match="starts-with(@name,\"openSUSE:$target:Staging\")" | grep name | cut -d\' -f2)

lnussel:
see pr#329

PR #329, opened 2015-04-07
fix quoting with enabled nullglob

lnussel:


nilxam:
ok, and I tested it, thanks!

PR #330, opened 2015-04-07
Make number of expected jobs architecture specific

k0da:
90 is too high number of expected jobs for PowerPC.

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -47,6 +47,9 @@ class ToTestBase(object):
     def iso_prefix(self):
         return self.project
 
+    def jobs_num(self):
+        return 90
+
     def binaries_of_product(self, project, product):
         url = self.api.makeurl(['build', project, 'images', 'local', product])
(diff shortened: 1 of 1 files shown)


k0da:
@aplanas The change works for me.

aplanas:
@k0da I know why it works for you, but still is wrong: ToTest132, or any new class that derives from ToTestBase will not overwrite `jobs_num` and will use the one in the base class, that is returning a random string instead of a number.

So please:
* return a good estimation in the base class (5, 10? IDK)
* or raise an exception in the base class and provide some value in the overwritten version of every ToTestBase class (missing ToTest132 that sometime in the future will be renamed)

coolo:
^what aplanas said

k0da:
updated. jobs_num is set to 90 globally and overrided only for Factory:PowerPC project

aplanas:
Thanks! : )

PR #331, opened 2015-04-08
only pass the group to the API

coolo:
needs deploying openqa.oo though

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -24,6 +24,7 @@ PLUGINDIR = os.path.expanduser(os.path.dirname(os.path.realpath(__file__)))
 sys.path.append(PLUGINDIR)
 from osclib.conf import Config
 from osclib.stagingapi import StagingAPI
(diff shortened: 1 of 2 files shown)


totest-manager.py
```diff
@@ -410,11 +403,8 @@ class ToTestFactoryPowerPC(ToTestBase):
     def __init__(self, project, dryrun):
         ToTestBase.__init__(self, project, dryrun)
 
-    def openqa_version(self):
-        return 'Tumbleweed'
-
     def openqa_group(self):
-        return 'PowerPC'
+        return 'Tumbleweed PowerPC'
```
nilxam: isn't it 'openSUSE Tumbleweed PowerPC'? I thought that is the correct group name we had for powerpc

coolo: I have some reputation to keep - and that includes that I break power :)

nilxam:
LGTM

PR #332, opened 2015-04-08
totest manager for arm

lnussel:


lnussel:
that would require to duplicate current_version()

aplanas:
True. The same problem is in the powerpc code. In this case maybe makes more sense to inherit the PPC one also from this ToTestFactory : (

PR #333, opened 2015-04-09
sync binaries to seperate directory per architecture

nilxam:
the solv file also will be $dir-$arch.solv


coolo:
why not pass the fixed dir? I find it strange that the function is getting a dir it's not using as is

PR #334, opened 2015-04-12
implement exception.list support for some special case

nilxam:


coolo:
I'm not a fan of this. It would be better if this came out of a change within I itself - e.g. by changing patterns-openSUSE to replace kde by plasma

nilxam:
hmm...a question in my mind, how to deal with

job install provides pattern() = kde4_basis
job install provides pattern() = kde4_multimedia
job install provides pattern() = kde4_utilities

without touch testcase? even those branding thingy?

coolo:
One way is to change patterns-openSUSE to create a subpackage patterns-staging-dvd1 marked as internal and not providing pattern() but requring the right ones. This way a SR to it can modify the package list. 

We just didn't have this case yet, that doesn't mean we have to solve it with hacks IMO.

PR #336, opened 2015-04-15
totest-manager --restart-tests: restart all failed tests that are not li...

DimStar77:
...sted as 'known failures'

It uses a wrapper script 'openqa' that has to exist in $PATH (I have it in ~/bin ) due to the need for having APIKEY and APISECRET present.

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -36,9 +36,10 @@ QA_PASSED = 3
 class ToTestBase(object):
     """Base class to store the basic interface"""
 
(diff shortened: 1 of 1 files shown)


coolo:
I'm skeptical too. Your user story would be better support by totestmanager outputting something that is machine parsable for your external tool to read and act upon.

DimStar77:
ok, I was considering parsing output as well.... the information is already present and potentially can be parsed.

PR #341, opened 2015-05-05
Add machinery repository to Devel whitelist

thardeck:
Machinery is not considered a Devel repository but we would like to keep
our regular path.


coolo:
I think it requires a /

thardeck:
Thx for the review, I will close the pr for now since during today's meeting we weren't sure anymore that we want to submit the package to Factory.

PR #343, opened 2015-05-07
Change the testcase in order to match Plasma 5

nilxam:


nilxam:
open for review and discuss, don't merge it until Staging:I got accept and we all agree this testcase change.

PR #345, opened 2015-05-15
`find_request_package` ignore declined requests.

aplanas:


osclib/request_finder.py
```diff
@@ -83,23 +83,15 @@ def find_request_package(self, package):
...
 
             self.srs[request] = {'project': self.api.project, 'state': state}
+            requests.append(request)
 
-            if last_rq:
-                if self.srs[last_rq]['state'] == 'declined':
-                    # ignore previous requests if they are declined
-                    # if they are the last one, it's fine to return them
-                    del self.srs[last_rq]
-                else:
```
coolo: looks like the code just doesn't work - but I agree to remove it

PR #350, opened 2015-05-27
Add update_crawler.py script

aplanas:
This scripts can analyze two repositories, one that contains
updates, and the other a linked distribution.  Check the last
updated packages and send (if there is not any) a submit request
with the updated package.

### DIFF
--- a/update_crawler.py
+++ b/update_crawler.py
@@ -1,0 +1,182 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2015 SUSE Linux GmbH
+#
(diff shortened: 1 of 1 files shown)


update_crawler.py
```diff
@@ -0,0 +1,182 @@
...
+                        help='project where to submit the updates (default: %s)' % OPENSUSE,
+                        default=OPENSUSE)
+
+    args = parser.parse_args()
+
+    # Set logging configuration
+    logging.basicConfig(level=logging.DEBUG if args.debug
+                        else logging.INFO)
+
+    sys.exit(main(args))
```
lnussel: main doesn't have a return value?

PR #353, opened 2015-06-08
Check the version of the package.

aplanas:
Also adds a new dependency to `python-semantic_version`

### DIFF
--- a/update_crawler.py
+++ b/update_crawler.py
@@ -28,6 +28,7 @@ from xml.etree import cElementTree as ET
 
 import osc.conf
 import osc.core
+import rpm
 
 from osclib.memoize import memoize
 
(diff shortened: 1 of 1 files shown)


aplanas:
Do not merge yet, I will use a rpm's bind instead `python-semantic_version`

aplanas:
* Remove sematic_version dependency and use rpm binding `labelCompare`

update_crawler.py
```diff
@@ -59,13 +90,21 @@ def get_source_packages(self, project, expand=False):
...
 
     def get_source_verifymd5(self, project, package):
         """ Return the verifymd5 of a source package."""
         root = ET.fromstring(self._get_source_package(project, package))
         return root.get('verifymd5')
 
+    def get_source_version(self, project, package):
+        """ Return the version of a source package."""
+        root = ET.fromstring(self._get_source_package(project, package))
+        return stringToVersion(root.find('version').text)
```
lnussel: are you sure this is needed? Looks like obs returns version and release in separate tags

aplanas:
* Use the `version` and `release` returned from OBS

PR #357, opened 2015-06-11
Add exception from changes tag requirement

krig:
Don't require an entry in the changelog if

1. The package is already submitted to Factory and has a changelog
2. The package is a new submission, not an update

### DIFF
--- a/check_tags_in_requests.py
+++ b/check_tags_in_requests.py
@@ -32,6 +32,9 @@ import urllib2
 
 import ReviewBot
 
+import check_source_in_factory
+
+
 class TagChecker(ReviewBot.ReviewBot):
     """ simple bot that checks that a submit request has corrrect tags specified
     """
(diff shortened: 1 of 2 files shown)


krig:
I need some help with fixing the failing tests... since the TagChecker now performs additional remote queries the mock returns don't match.

nilxam:
If I understand it correct, you're missing the dummy data in checktags_tests.py as check_source_submission() trying to open the url eg. /source/....

krig:
Yes, I'm just hoping that someone else (tm) will show up and help me fill in correct dummy data :)

check_tags_in_requests.py
```diff
@@ -106,8 +109,34 @@ def checkTagInRequest(self, req, a):
 
         return True
 
+    def isNewPackage(self, tgt_project, tgt_package):
+        try:
+            osc.core.show_devel_project(self.apiurl, tgt_project, tgt_package)
+        except HTTPError:
```
lnussel: throws HTTPError undefined for me. Also, show_package_meta would be more suitable here

check_tags_in_requests.py
```diff
@@ -106,8 +109,34 @@ def checkTagInRequest(self, req, a):
...
+
+    def checkTagNotRequired(self, req, a):
+        # A tag is not required only if the package is
+        # already in Factory with the same revision,
+        # and the package is being introduced, not updated
+        factory_checker = check_source_in_factory.FactorySourceChecker(apiurl=self.apiurl,
+                                                                       dryrun=self.dryrun,
+                                                                       logger=self.logger,
+                                                                       user=self.review_user,
+                                                                       group=self.review_group)
```
lnussel: you need to pass the project name of Factory, ie make it a comand line parameter for the changelog checker. openSUSE:Factory is not openSUSE:Factory in ibs where this bot is meant to run.

check_tags_in_requests.py
```diff
@@ -106,8 +109,34 @@ def checkTagInRequest(self, req, a):
...
+        factory_ok = factory_checker.check_source_submission(a.src_project, a.src_package, a.src_rev,
```
lnussel: you  may want to check that first and return right away to avoid the potentially expensive factory source check.

krig:
Thanks for the comments! I have fixed the issues you noted.

krig:
There, the tests seem to be passing. @lnussel, let me know if there is something not quite right still. :)

Issue #358, opened 2015-06-17
check_tags_in_requests.py confused by .changes files in tarballs


lnussel:
check_tags_in_requests.py uses osc.core.request_diff to request a diff for the request it's processing. That api call however always also diffs contents of tarballs which may confuse the bot if tarballs also contain .changes files. See also https://github.com/openSUSE/open-build-service/issues/949

lnussel:
@krig also helped debugging this issue

lnussel:
Independent of whether the feature of not diffing tarballs gets implemented server side, the bot uses the diff call at the wrong place. It requests the diff for the whole request for each action. If the bot acts per action it needs to manually diff itself. If it wants to use the server side diff it should override check_one_request() instead.

lnussel:
so I guess it's fixed

PR #359, opened 2015-06-17
Rely on server-side issue tag check

krig:
I think this should fix the problem. It grabs the first diff of a file ending in .changes in the submit diff.

I haven't fixed the tests, so the tests are failing since it now tries to get the server diff.

Fixes: #358

### DIFF
--- a/check_source_in_factory.py
+++ b/check_source_in_factory.py
@@ -54,6 +54,11 @@ class FactorySourceChecker(ReviewBot.ReviewBot):
     def check_source_submission(self, src_project, src_package, src_rev, target_project, target_package):
         self.logger.info("%s/%s@%s -> %s/%s"%(src_project, src_package, src_rev, target_project, target_package))
         src_srcinfo = self.get_sourceinfo(src_project, src_package, src_rev)
+        if src_srcinfo is None:
+            # source package does not exist?
+            # handle here to avoid crashing on the next line
+            self.logger.info("Could not get source info for %s/%s@%s" % (src_project, src_package, src_rev))
+            return False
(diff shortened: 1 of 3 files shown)


krig:
Hmm, this doesn't seem to return what I expect from the public OBS:

https://api.opensuse.org/source/openSUSE:Factory/nano?opackage=nano&oproject=editors&cmd=diff&onlyissues=1&orev=25&view=xml

Gives me

```
<directory name="nano" rev="38" vrev="1" srcmd5="fc6b5b47f112848a1eb6fb8660b7800b">
  <entry name="nano-2.4.1.tar.gz" md5="1c612b478f976abf8ef926480c7a3684" size="1890805" mtime="1429179243" />
  <entry name="nano-2.4.1.tar.gz.asc" md5="a5b725ee67cef5595f7327a84fc80d54" size="181" mtime="1429179243" />
  <entry name="nano.changes" md5="759a3cfc6513ce1c68c61c9c82fdbb72" size="24363" mtime="1429179243" />
  <entry name="nano.keyring" md5="284e50e7b4d0b494dcc7111b704980ec" size="1548" mtime="1427233842" />
  <entry name="nano.spec" md5="3cbbe5d98eee721d3a1aa02d8d23dd00" size="2989" mtime="1429179243" />
</directory>
```

krig:
Ah, or maybe I have to use POST, not GET.

coolo:
yeah, all cmd= are POST

krig:
OK, so I think the code is fine now. However, this breaks all of the tests since the method for checking issues moves to the server... I guess just throwing all the tests away and writing new ones is what is needed :(

PR #362, opened 2015-06-27
Add the appropriate kernel to testcase

nilxam:
Due to pattern_base's change, add the appropriate kernel to testcase. kernel-desktop for x86_64 and kernel-default for ppc64le.

If I understand https://github.com/yast/yast-yast2/blob/master/library/system/src/modules/Kernel.rb correct, the kernel flavor should be right here.


coolo:
kernel-default would do for both actually - the difference between desktop and default shouldn't really matter within VMs.

PR #364, opened 2015-07-03
ports: make sure to not add ports over ports

DimStar77:
Without this, we have code like this:

project = self.api.project
for ports in ports_prjs:
            project = project + ':' + ports

Which results in
1) openSUSE:Factory:PowerPC
2) openSUSE:Factory:PowerPC:ARM

the 2nd obviously wrong.


lnussel:
do'h!

PR #366, opened 2015-07-07
Fixes the logic error in checkTagNotRequired()

nilxam:
Maybe I'm wrong, but this logic should be the properly one I assume

use ''if not is_new: '' once the package is not a new package it will return False and quit the function, then I guess the following factorySourceChecker will not execute.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -194,7 +194,7 @@ class ReviewBot(object):
         url = osc.core.makeurl(apiurl, ('source', project, package), query=query)
         try:
             return ET.parse(osc.core.http_GET(url)).getroot()
-        except urllib2.HTTPError:
+        except (urllib2.HTTPError, urllib2.URLError):
             return None
 
     def get_originproject(self, project, package, rev=None):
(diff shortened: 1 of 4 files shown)


nilxam:
the question here is: should we give the submission have a chance be accept if it isn't a new package and without a valid bsc#, fate#, etc. but its changes has already in Factory? eg. a dependency package update?

lnussel:
that's a policy decission SLE release managers have to take

coolo:
If it's coming from Factory, we don't want any additional checks. 

But if it's not from Factory, we require bnc# for updates. I don't see any reason why we would exclude new packages from that. The is_new is only there to avoid checking Factory for updates IMO.

coolo:
So your change is wrong. To fix the use case you want to cover, we need to always do the factory check and skip the is_new completely.

nilxam:
now the logic is
1) A tag is not required only if the package is already in Factory with the same revision, and the package is being introduced, not updated
2) A new package must be have a issue tag then it can accept to review, sle-release-manager can decide would accept it or not, just don't declined by changelog-checker if it has the issues tag

PS. I cannot get rid of is_new here because of check_source_submission() hard to handle that empty target source while through _check_factory()

lnussel:
Max Lin wrote:
> PS. I cannot get rid of is_new here because of
> check_source_submission() hard to handle that empty target source
> while through _check_factory()

You mean there are requests that have no target package specified? If
that's the case you can just pass the source package name.

PR #379, opened 2015-08-05
don't submit sub-spec files

coolo:


update_crawler.py
```diff
@@ -109,7 +110,7 @@ def get_update_candidates(self):
     def _submitrequest(self, src_project, src_package, dst_project,
                        dst_package, msg):
         """Create a submit request."""
-        states = ['new', 'review', 'declined']
+        states = ['new', 'review', 'declined', 'revoked']
```
aplanas: Why we want to search for revoked ones?

aplanas: Ah ok, I get it now.

aplanas:
+1

PR #380, opened 2015-08-06
Execute based on permission

DimStar77:
I for one can start 'osc staging adi', it gives me an overview of "what's missing" and it accepts the 'ready stagings"; but as I'm not member of factory-staging, the creation of new staging projects, and moving requests to them, fails with a permission denied.

This change set performs a check if the user starting the command is member of factory-staging and only conditionally executes the 2nd part.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -27,6 +27,7 @@ from osc import conf
 from osc import oscerr
 from osc.core import change_review_state
 from osc.core import delete_package
+from osc.core import get_group
 from osc.core import get_request
 from osc.core import make_meta_url
 from osc.core import makeurl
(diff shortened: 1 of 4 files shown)


osclib/stagingapi.py
```diff
@@ -33,6 +33,7 @@
 from osc.core import http_GET
 from osc.core import http_POST
 from osc.core import http_PUT
+from osc.core import get_group
```
aplanas: Usually is expected to be alphabetically ordered

aplanas:
Missing code in the OBS mockup.

After the tests are OK, +1 from me

DimStar77:
I needed to switch APIURL in obs.py to http://localhost instead of https://localhost... this now matches what is in tests/fixture/oscrc

changing it all to https have me connection errors, so it seems the obsmockup server is not started 'secured'

PR #381, opened 2015-08-08
simplify update_crawler

coolo:
Just checking verifymd5 for everything in sle12-picks - if sle12 sp1
comes with a version downgrade, we should see it


aplanas:
Uhm I think that you are replacing a generic solution with a more specific solution. Not a good change IMHO

coolo:
I'm fixing the core responsibility of this tool, yes. See https://progress.opensuse.org/issues/8498

PR #382, opened 2015-08-12
manager42: some packages might be to new to already have five revisi…

DimStar77:
…ons, even in Factory

### DIFF
--- a/manager_42.py
+++ b/manager_42.py
@@ -203,7 +203,7 @@ class UpdateCrawler(object):
         for rev in his.findall('revision'):
             revs.append(rev.find('srcmd5').text)
         revs.reverse()
-        for i in xrange(5): # check last 5 commits
+        for i in range(min(len(revs), 5)): # check last 5 commits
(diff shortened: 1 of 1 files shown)


manager_42.py
```diff
@@ -203,7 +203,7 @@ def check_factory_sources(self, package, verifymd5):
         for rev in his.findall('revision'):
             revs.append(rev.find('srcmd5').text)
         revs.reverse()
-        for i in xrange(5): # check last 5 commits
+        for i in xrange(len(revs) if len(revs) < 5 else 5): # check last 5 commits
```
aplanas: like `min(len(revs), 5)`? in this either case use `range`

DimStar77:
I get the idea of using min (no idea why I didn't use it :) ) but why switch to range instead of xrange? the result should be the same

aplanas:
Mostly two reasons:

* `xrange` for small numbers is useless (even a bit slower)
* In Python 3 there is only `range`

PR #384, opened 2015-08-12
misc fixes

lnussel:


manager_42.py
```diff
@@ -262,19 +290,20 @@ def find_invalid_links(self, prj):
...
-                    http_DELETE(url)
-                    self.packages[project].remove(package)
+                     # XXX: why was this code here?
+#                    # TODO: detach only if actually a link to the deleted package
+#                    url = makeurl(self.apiurl, ['source', 'openSUSE:42', package], { 'opackage': package, 'oproject': 'openSUSE:42', 'cmd': 'copy', 'expand': '1'} )
+#                    try:
+#                        http_POST(url)
+#                    except urllib2.HTTPError, err:
+#                        pass
+                    self.remove_packages(project, [packages])
```
nilxam: typo? it should be 'package' rather than 'packages'

lnussel: indeed

PR #387, opened 2015-08-17
fix memoize key computation

lnussel:


lnussel:
I'm not sure what is_method is meant for. I was considering to add a parameter to tell memoize in case it's called on a method without self. most of the existing users seem to pass self

coolo:
is this still relevant?

coolo:
I close this after 3 months

PR #388, opened 2015-08-17
Re-enable detach linked entry code

nilxam:
Still need it to handle the request accepted from the corresponded
place, ie. from self.project_preference_order. But now detach the linked
entry and add requestid following with copy cmd, therefore, the
following get_latest_request() in crawl will not fail.


lnussel:
shouldn't be really needed, only as safeguard if something fails when recreating the package

PR #406, opened 2015-08-27
Add --by-develproject argument to adi command

nilxam:
If with --by-develproject, adi command will check devel project of source package before selecting requests to adi staging.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -64,6 +64,8 @@ def _full_project_name(self, project):
 
 @cmdln.option('--move', action='store_true',
               help='force the selection to become a move')
+@cmdln.option('--by-develproject', action='store_true',
+              help='sort the packages by devel project')
 @cmdln.option('-f', '--from', dest='from_', metavar='FROMPROJECT',
               help='manually specify different source project during request moving')
 @cmdln.option('-p', '--project', dest='project', metavar='PROJECT', default='Factory',
@@ -172,4 +174,4 @@ def do_staging(self, subcmd, opts, *args):
         elif cmd == 'list':
             ListCommand(api).perform()
         elif cmd == 'adi':
-            AdiCommand(api).perform(args[1:], opts.move)
+            AdiCommand(api).perform(args[1:], move=opts.move, by_dp=opts.by_develproject)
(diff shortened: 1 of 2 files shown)


osc-staging.py
```diff
@@ -172,4 +174,9 @@ def do_staging(self, subcmd, opts, *args):
         elif cmd == 'list':
             ListCommand(api).perform()
         elif cmd == 'adi':
-            AdiCommand(api).perform(args[1:], opts.move)
+            if opts.move:
+                AdiCommand(api).perform(args[2:], opts.move)
+            elif opts.by_develproject:
+                AdiCommand(api).perform(args[2:], by_dp=opts.by_develproject)
+            else:
```
coolo: I wonder why you move from args[1:] to args[2:] - and I think you can just have move=opts.move, by_dp=opts.by_develproject in one line without ifs

nilxam: if with --move argument then args[1] is "--move", doesn't it? if use args[1:] ie. packages actually it's including "--move"?

coolo: args only contains the arguments not filtered out, i.e. --move moved into opts.move and out of args. same with by_develproject

nilxam: ok, so I misunderstand args and opts here, updated then.

osclib/adi_command.py
```diff
@@ -38,7 +39,15 @@ def check_adi_projects(self):
         for p in self.api.get_adi_projects():
             self.check_adi_project(p)
 
-    def create_new_adi(self, wanted_requests):
+    def get_devel_project(self, project, package):
+        m = show_package_meta(self.apiurl, project, package)
```
coolo: self has no apiurl, self.api.apiurl is the proper form

osclib/adi_command.py
```diff
@@ -38,7 +39,15 @@ def check_adi_projects(self):
...
+        node = ET.fromstring(''.join(m)).find('devel')
```
coolo: ET is not imported in the adi command

coolo:
beside those 2 show stoppers it seems to work properly

PR #407, opened 2015-08-28
retry 5xx - especially needed for view=info calls

coolo:


manager_42.py
```diff
@@ -76,12 +76,22 @@ def __init__(self, from_prj, caching = True):
...
         if self.caching:
             return self._cached_GET(url)
-        return http_GET(url).read()
+        return self.retried_GET(url).read()
+
+    def retried_GET(self, url):
+        try:
+            return http_GET(url)
+        except urllib2.HTTPError, e:
+            if e.code / 100 == 5:
```
aplanas: `if 500 <= e.code <= 599` is usually more readable, and in this case more correct.  In Python 3 `/` will produce funny errors:

    Python 3.4.3 (default, Mar 27 2015, 02:30:53) [GCC] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> 503 / 100 == 5
    False

coolo:
see dccb5d3

PR #414, opened 2015-09-07
correspond to new yast2_i openQA change

nilxam:


nilxam:
btw I'm not so sure why we did '''job lock name yast2-nfs-client''' in the past

coolo:
We wanted to get rid of the broken yast module with the broken nfs-utils require

PR #423, opened 2015-10-01
Do not superseding requests with list command

nilxam:
Do not superseding requests with list command, but if had the permission
of staging, can executing with --supersede argument, or even against
specific requests.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -66,6 +66,8 @@ def _full_project_name(self, project):
               help='force the selection to become a move')
 @cmdln.option('--by-develproject', action='store_true',
               help='sort the packages by devel project')
+@cmdln.option('--supersede', action='store_true',
+              help='superseding requests. please make sure you have staging permissions')
 @cmdln.option('-f', '--from', dest='from_', metavar='FROMPROJECT',
               help='manually specify different source project during request moving')
 @cmdln.option('-p', '--project', dest='project', metavar='PROJECT', default='Factory',
@@ -126,7 +128,9 @@ def do_staging(self, subcmd, opts, *args):
         min_args, max_args = 1, None
     elif cmd == 'adi':
         min_args, max_args = None, None
-    elif cmd in ('list', 'cleanup_rings'):
+    elif cmd == 'list':
+        min_args, max_args = 0, None
+    elif cmd == 'cleanup_rings':
         min_args, max_args = 0, 0
     else:
         raise oscerr.WrongArgs('Unknown command: %s' % cmd)
@@ -172,6 +176,6 @@ def do_staging(self, subcmd, opts, *args):
         elif cmd == 'cleanup_rings':
             CleanupRings(api).perform()
         elif cmd == 'list':
-            ListCommand(api).perform()
+            ListCommand(api).perform(args[1:], supersede=opts.supersede)
         elif cmd == 'adi':
             AdiCommand(api).perform(args[1:], move=opts.move, by_dp=opts.by_develproject)
--- a/osclib/list_command.py
+++ b/osclib/list_command.py
@@ -5,13 +5,20 @@ class ListCommand:
     def __init__(self, api):
         self.api = api
 
-    def perform(self):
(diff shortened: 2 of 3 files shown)


nilxam:
next step might be check permission before allow to run with --supersede

osc-staging.py
```diff
@@ -66,6 +66,8 @@ def _full_project_name(self, project):
               help='force the selection to become a move')
 @cmdln.option('--by-develproject', action='store_true',
               help='sort the packages by devel project')
+@cmdln.option('--supersede', action='store_true',
+              help='superseding the requests. please make sure you have the permission of staging')
```
aplanas: `supersede the request. Needs staging permissions` ?? But not sure, my English sucks

osclib/list_command.py
```diff
@@ -5,13 +5,17 @@ class ListCommand:
     def __init__(self, api):
         self.api = api
 
-    def perform(self):
+    def perform(self, packages=[], supersede=False):
```
aplanas: Do not use mutable objects as default parameters:

```
def perform(self, packages=None, supersede=False):
    if not packages:
        packages = []
```

osclib/list_command.py
```diff
@@ -5,13 +5,17 @@ class ListCommand:
...
         """
         Perform the list command
         """
 
-        # First dispatch all possible requests
-        self.api.dispatch_open_requests()
+        if supersede is True:
+            if len(packages):
```
aplanas: `if packages`

osclib/stagingapi.py
```diff
@@ -331,7 +331,7 @@ def accept_non_ring_request(self, request):
             self.do_change_review_state(request_id, 'accepted', message=message,
                                         by_group=self.cstaging_group)
 
-    def supseded_request(self, request):
+    def supseded_request(self, request, target_pkgs=[]):
```
aplanas: same here, do not use mutable objects as a default parameter

osclib/stagingapi.py
```diff
@@ -356,20 +356,25 @@ def supseded_request(self, request):
             msg = msg.format(request_id, action)
             logging.info(msg)
 
+        pkg_do_supersede = True
+        if len(target_pkgs):
```
aplanas: `if target_pkgs`

osclib/list_command.py
```diff
@@ -5,13 +5,17 @@ class ListCommand:
...
 
-    def perform(self):
+    def perform(self, packages=[], supersede=False):
         """
         Perform the list command
         """
 
-        # First dispatch all possible requests
-        self.api.dispatch_open_requests()
+        if supersede is True:
```
aplanas: `if supersede`

osclib/stagingapi.py
```diff
@@ -356,20 +356,25 @@ def supseded_request(self, request):
...
+            if target_package not in target_pkgs:
+                pkg_do_supersede = False
+
         # If the package is currently tracked then we do the replacement
         stage_info = self.packages_staged.get(target_package, {'prj': '', 'rq_id': 0})
-        if stage_info['rq_id'] != 0 and int(stage_info['rq_id']) != request_id:
+        if  pkg_do_supersede is True and stage_info['rq_id'] != 0 and int(stage_info['rq_id']) != request_id:
```
aplanas: I find this line suspicious, `stage_info['rq_id']` looks like that can have strings (converted to int()) or 0 value.  I would suggest this:

`if pkg_do_supersede and stage_info['rq_id'] and int(stage_info['rq_id']) != request_id:`

A better approach is to move `stage_info['rq_id'] and int(stage_info['rq_id']) != request_id` into another helper that try to convert into a int and compare.

osclib/stagingapi.py
```diff
@@ -356,20 +356,25 @@ def supseded_request(self, request):
...
             return stage_info
         return None
 
-    def update_superseded_request(self, request):
+    def update_superseded_request(self, request, target_pkgs=[]):
```
aplanas: same

PR #427, opened 2015-10-07
Whitelist devel:languages:erlang as a devel project

DimStar77:


osc-check_source.py
```diff
@@ -252,6 +252,7 @@ def _checker_one_request(self, rq, opts):
             self._devel_projects['devel:languages:D/'] = 'd'
             self._devel_projects['Novell:NTS'] = 'support'
             self._devel_projects['Apache:Shibboleth/'] = 'Shibboleth'
+            self._devel_projects['devel:languages:erlang'] = 'erlang'
```
coolo: the final / is important!

PR #436, opened 2015-11-02
accept delete request in changelog-checker

nilxam:


ReviewBot.py
```diff
@@ -180,6 +180,10 @@ def check_action_maintenance_release(self, req, a):
     def check_action_submit(self, req, a):
         return self.check_source_submission(a.src_project, a.src_package, a.src_rev, a.tgt_project, a.tgt_package)
 
+    def check_action_delete(self, req, a):
+        """ default implemention does nothing """
+        return None
```
aplanas: `raise NotImplementedError` is better than `return None`, because the last one is actually doing something if is not implemented: is returning None, that is evaluated a `False`

nilxam: but https://github.com/openSUSE/osc-plugin-factory/blob/master/ReviewBot.py#L150 is looking for an returned value esp. None.

aplanas: Oh sorry : (. You are right.

lnussel: the default action could be omitted in ReviewBot.py though

nilxam:
updated(removed dummy implementation part) and tested.

PR #440, opened 2015-11-19
ttm: fixup previous commit.\n\nMoving around things is not a good idea

DimStar77:
Fixing up.

moving the write_version calls where I originally had. For update_totest it would work, as we pass the snapshot version in, but in the publish_factory_totest we don't.

We can fix it either as proposed here (I added guards to write_version to not really do it in case of dryrun) or we change publish_factory_totest to also receive a version argument, just for this case...


DimStar77:
There are more fixes needed (http_PUT is not yet known) - @coolo fixed this up on the machine running ttm; so this PR is obsolete / incomplete

PR #444, opened 2015-11-26
staging accept: check if /totest is properly synced before changing /…

DimStar77:
…standard

With the syncing of /standard to /totest, we have to be sure not to checkin to /standard before the /totest sync is completed.

So add a check before checking in, stopping the user from doing the too early checkins.

Some APIs were changed/enhanced.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1014,6 +1014,13 @@ class StagingAPI(object):
                 pass
         return version
 
+    def get_binary_version(self, project, rpm, repository='standard', arch='x86_64'):
+        """
+        Return the version of a built rpm file
+        """
+        url = self.makeurl(['build', project, repository, arch, '_repository', "%s?view=fileinfo" % rpm])
+        return ET.parse(http_GET(url)).getroot().find('version').text
+
     def load_file_content(self, project, package, filename):
         """
         Load the content of a file and return the content as data. If the package is a link, it will be expanded
@@ -1101,6 +1108,15 @@ class StagingAPI(object):
                 results.append(pkg.attrib['package'])
 
         return results
+
+    def is_repo_dirty(self, project, repository):
+        url = self.makeurl(['build', project, '_result?code=broken&repository=%s' % repository])
+        root = ET.parse(http_GET(url)).getroot()
+        for repo in root.findall('result'):
+            repostate = repo.get('state', 'missing')
+            if repostate not in ['unpublished', 'published'] or repo.get('dirty', 'false') == 'true':
+                return True
+        return False
 
     def check_pkgs(self, rebuild_list):
         url = self.makeurl(['source', self.project])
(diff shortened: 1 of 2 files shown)


osclib/stagingapi.py
```diff
@@ -187,14 +190,21 @@ def get_filelist_for_package(self, pkgname, project, extension=None):
             'extension': extension
         }
 
-        if extension:
-            url = self.makeurl(['source', project, pkgname], query=query)
+        if binaries:
+	    url = self.makeurl(['build', project, repository, arch, pkgname])
+            xmlnode='binary'
+            xmlattrib='filename'
```
lnussel: I'd probably move this to a separate function to improve readability

osclib/stagingapi.py
```diff
@@ -1014,14 +1024,28 @@ def package_version(self, project, package):
                 pass
         return version
 
-    def load_file_content(self, project, package, filename):
+    def guess_binary_version(self, project, package, repository='standard', arch='x86_64'):
+        """
+        Return the most probable version of the package in question
+        """
+        buildinfo = self.load_file_content(project, package, '_buildenv', binary=True, repository=repository, arch=arch)
```
lnussel: $ osc api /build/openSUSE:Factory/totest/x86_64/_product:openSUSE-release/openSUSE-release-20151128-1.2.src.rpm?view=fileinfo_ext
<fileinfo filename="openSUSE-release-20151128-1.2.src.rpm">
  <name>openSUSE-release</name>
  <version>20151128</version>

lnussel:
what about letting the totest manager handle the logic when all totest publishing is done? it could set some lock flag in the dashboard so the code here would only need to check for this flag

osclib/stagingapi.py
DimStar77: On Mon, 2015-11-30 at 01:43 -0800, Ludwig Nussel wrote:
> osc api /build/openSUSE:Factory/totest/x86_64/_product:openSUSE-
> release/openSUSE-release-20151128-1.2.src.rpm?view=fileinfo_ext

Sure - when I already know the file name... so this needs to be
preceeded by:

osc api /build/openSUSE:Factory/totest/x86_64/_product:openSUSE-release

get thruogh the file list, find the .src.rpm, then pass the other api
call...

it's an option.

coolo: You have to know your api ;)

```
osc api /build/openSUSE:Factory/totest/x86_64/_repository/openSUSE-release.rpm?view=fileinfo
```

DimStar77: On Mon, 2015-11-30 at 01:55 -0800, Stephan Kulow wrote:
> 
> You have to know your api ;)
> osc api /build/openSUSE:Factory/totest/x86_64/_repository/openSUSE-
> release.rpm?view=fileinfo
> —
Thanks... I simplified using this API.

PR #447, opened 2015-12-22
staging accept: introduce --force

DimStar77:
Allow to skip full validation of /totest being fully synced.
This can be helpful (and is only meant to be used) when you run two
consequtive staging accept LETTER commands in a row.

* The first one would see /totest fully sync, not dirty
* the second call sees /totest as dirty, as the scheduler would first have
  to process all data

In order to not have to wait for this to happen, the factory-maintainer CAN
override this with osc staging accept --force

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -88,6 +88,11 @@ def do_staging(self, subcmd, opts, *args):
     "accept" will accept all requests in
         openSUSE:Factory:Staging:<LETTER> (into Factory)
 
+    "acheck" will check if it's safe to accept new staging projects
+        As openSUSE:Factory is syncing the right package versions between
+        /standard, /totest and /snapshot, it's important that the projects
+        are clean prior to a checkin round.
+
     "check" will check if all packages are links without changes
 
     "cleanup_rings" will try to cleanup rings content and print
@@ -132,7 +137,7 @@ def do_staging(self, subcmd, opts, *args):
         min_args, max_args = None, None
     elif cmd in ('list', 'accept'):
         min_args, max_args = 0, None
-    elif cmd == 'cleanup_rings':
+    elif cmd in ('cleanup_rings', 'acheck'):
         min_args, max_args = 0, 0
     else:
         raise oscerr.WrongArgs('Unknown command: %s' % cmd)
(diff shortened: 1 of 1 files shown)


DimStar77:
Any word on this?

acheck is just for 'my' convenience / --force is needed if two staging accept commands want to be run in short time one after the other (as /totest will be flagged dirty right after checkin until the scheduler touched it)

osc-staging.py
```diff
@@ -130,7 +132,7 @@ def do_staging(self, subcmd, opts, *args):
         min_args, max_args = None, None
     elif cmd in ('list', 'accept'):
         min_args, max_args = 0, None
-    elif cmd == 'cleanup_rings':
+    elif cmd in ('cleanup_rings', 'acheck'):
```
aplanas: if acheck is a new command, provide info in the `Usage` section

osc-staging.py
```diff
@@ -155,12 +157,18 @@ def do_staging(self, subcmd, opts, *args):
         elif cmd == 'freeze':
             for prj in args[1:]:
                 FreezeCommand(api).perform(api.prj_from_letter(prj))
+        elif cmd == 'acheck':
+            # Is it safe to accept? Meaning: /totest contains what it should and is not dirty
+            version_openqa = api.load_file_content("%s:Staging" % api.project, "dashboard", "version_totest")
+            version_totest = api.get_binary_version(api.project, "openSUSE-release.rpm", repository="totest", arch="x86_64")
+            totest_dirty   = api.is_repo_dirty(api.project, 'totest')
```
aplanas: extra white spaces before the assignment (do not indent assignments in Python)

osc-staging.py
```diff
@@ -155,12 +157,18 @@ def do_staging(self, subcmd, opts, *args):
         elif cmd == 'freeze':
             for prj in args[1:]:
                 FreezeCommand(api).perform(api.prj_from_letter(prj))
+        elif cmd == 'acheck':
+            # Is it safe to accept? Meaning: /totest contains what it should and is not dirty
+            version_openqa = api.load_file_content("%s:Staging" % api.project, "dashboard", "version_totest")
```
aplanas: I have the feeling that this will appear more times, maybe is a good idea to create a method in the StagingAPI or something else to capture the logic (even with error management) to fetch info from the dashboard

lnussel:
@aplanas ok now?

aplanas:
+1

PR #458, opened 2016-02-03
frozenage: show how long a project was last frozen

DimStar77:
With multiple people working on staging projects, it's not always clear to everybody
when things happened and it's helpful to easily  be able to gather the information


aplanas:
+1

PR #459, opened 2016-02-05
don't use recursive retry

lnussel:
don't use recursion and don't hammer server

I couldn't test that one as the servers are back..

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -81,35 +81,30 @@ class StagingAPI(object):
         query = [] if not query else query
         return makeurl(self.apiurl, l, query)
 
+    def _retried_request(self, url, func):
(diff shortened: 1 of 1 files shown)


osclib/stagingapi.py
```diff
@@ -81,35 +81,30 @@ def makeurl(self, l, query=None):
         query = [] if not query else query
         return makeurl(self.apiurl, l, query)
 
+    def _retried_request(self, url, func):
+        retry_sleep_seconds = 1
+        while True:
+            try:
+                return func(url)
+            except urllib2.HTTPError, e:
+                if 500 <= e.code <= 599:
```
aplanas: Nice! I would add a max number of retries, probably as a default parameter

PR #461, opened 2016-02-11
totest-manager: exclude images_staging as used by ARM

lnussel:


totest-manager.py
```diff
@@ -206,7 +208,7 @@ def all_repos_done(self, project, codes=None):
         for repo in root.findall('result'):
             # ignore ports. 'factory' is used by arm for repos that are not
             # meant to use the totest manager.
-            if repo.get('repository') in ('ports', 'factory'):
+            if repo.get('repository') in ('ports', 'factory', 'images_staging'):
```
aplanas: But actually we are excluding this for all `arch`. Only ARM is using `images_staging` repo?

lnussel: yes, only arm uses it so it doesn't hurt

aplanas:
+1

PR #469, opened 2016-03-09
Implement configurable review mode

lnussel:
The new parameter --review-mode allows to specify how the review bot should handle obs reviews

normal: accept review if check is good, decline if bad
no: don't accept or decline requests
accept: always accept review, even if check fails
fallback-onfail: accept review and set other user or group as review if check fails
fallback-always: always accept review and set other user or group as reviewer afterwards

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -68,7 +68,7 @@ class ReviewBot(object):
 
     @review_mode.setter
     def review_mode(self, value):
-        if value not in REVIEW_CHOICES:
+        if value not in self.REVIEW_CHOICES:
             raise Exception("invalid review option: %s"%value)
         self._review_mode = value
 
@@ -365,6 +365,15 @@ class CommandLineInterface(cmdln.Cmdln):
 
         self.checker = self.setup_checker()
 
+        if self.options.review_mode:
+            self.checker.review_mode = self.options.review_mode
+
+        if self.options.fallback_user:
+            self.checker.fallback_user = self.options.fallback_user
+
+        if self.options.fallback_group:
+            self.checker.fallback_group = self.options.fallback_group
+
     def setup_checker(self):
         """ reimplement this """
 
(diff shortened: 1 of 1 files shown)


ReviewBot.py
```diff
@@ -57,6 +58,19 @@ def __init__(self, apiurl = None, dryrun = False, logger = None, user = None, gr
         self.review_group = group
         self.requests = []
         self.review_messages = ReviewBot.DEFAULT_REVIEW_MESSAGES
+        self._review_mode = 'normal'
+        self.fallback_user = None
+        self.fallback_group = None
+
+    @property
```
aplanas: only as curiosity, why to create a property from the beginning?

ReviewBot.py
```diff
@@ -57,6 +58,19 @@ def __init__(self, apiurl = None, dryrun = False, logger = None, user = None, gr
...
+    def review_mode(self):
+        return self._review_mode
+
+    @review_mode.setter
```
aplanas: aaaaand here is the answer, sorry for the noise

ReviewBot.py
```diff
@@ -57,6 +58,19 @@ def __init__(self, apiurl = None, dryrun = False, logger = None, user = None, gr
...
+    def review_mode(self, value):
+        if value not in REVIEW_CHOICES:
+            raise Exception("invalid review option: %s", value)
```
aplanas: `raise Exception('invalid review option: %s' % value)`

ReviewBot.py
```diff
@@ -68,18 +82,32 @@ def set_request_ids(self, ids):
             self.requests.append(req)
 
     def check_requests(self):
+
+        by_user = self.fallback_user if self.fallback_user else None
```
aplanas: `by_user = self.fallback_user` is enough (the same for the next line)

aplanas:
basically +1

PR #470, opened 2016-03-17
findfileconflicts: Be prepared for -x86_64 biarch package introduction

DimStar77:
https://progress.opensuse.org/issues/11268


coolo:
looks reasonable to me

mlschroe:
Is this for syslinux. Then it doesn't look reasonable, it should check the arch instead of the package name. s390x conflicts with x86_64.

lnussel:
so what now?

PR #472, opened 2016-03-22
Revert "make sure to only release the 'images' repo"

DimStar77:
openSUSE:Factory:Live has no 'images' repository to release, but a standard

This reverts commit 8f4e97598fc255b1283661c80444307f80e6ae4f.


nilxam:
Okay, let's find a better way to handle this situation.

PR #473, opened 2016-03-22
Skip obsoleted jobs

k0da:


coolo:
aehm, if there are obsoleted jobs, how can you release it? It means there is a newer build

k0da:
New results are there, but ttm counts obosoleted results. Makeing overal to fail.

PR #474, opened 2016-03-23
Add the possible repo candidates

nilxam:
Add the possible repo candidates to repos_to_check, those candidates
need to have binaries for check.


nilxam:
self merge, yeah, I'm bad! :)

PR #475, opened 2016-03-23
Fixes previous commit

nilxam:


nilxam:
suddenly we got coverage running :)

PR #476, opened 2016-03-23
ttm: Safeguard publishing of :ToTest

DimStar77:
Try to be smart enough to only publish :ToTest if the version we passed
to openQA matches what is found in :ToTest.

This avoids releasing untested snapshots in case the release process
had trouble syncing from openSUSE:Factory/standard to :ToTest (e.g.
partial releases).


lnussel:
I think what's still missing is catching errors when releasing while talking to OBS

PR #483, opened 2016-04-11
Show package also in adi command

nilxam:
it is look like

    openSUSE:Factory:Staging:adi:60 is ready
    cedilla [381552]


DimStar77:
:+1:

PR #484, opened 2016-04-12
repochecker: Adjust the decline message for missing binaries

DimStar77:
Some users are confused as the old message "package never built" does
not have to be true in many cases, especially on longer-lasting
stagings. In this case, the binaries might have been built but have
since been replaced by newer ones - the effect is the same: repo-checker
can't receive them in order to perform the full checks.

### DIFF
--- a/osclib/checkrepo.py
+++ b/osclib/checkrepo.py
@@ -798,7 +798,7 @@ class CheckRepo(object):
                 if arch.attrib['result'] == 'building':
                     r_foundbuilding = repo_name
                 if arch.attrib['result'] == 'outdated':
-                    msg = "%s's sources were changed after submissions and the old sources never built. Please resubmit" % request.src_package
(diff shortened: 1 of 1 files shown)


osclib/checkrepo.py
```diff
@@ -798,7 +798,7 @@ def is_buildsuccess(self, request):
                 if arch.attrib['result'] == 'building':
                     r_foundbuilding = repo_name
                 if arch.attrib['result'] == 'outdated':
-                    msg = "%s's sources were changed after submissions and the old sources never built. Please resubmit" % request.src_package
+                    msg = "%s's sources were changed after submissionsr: the relevant binaries are not available (never built or binaries replaced). Please resubmit" % request.src_package
```
aplanas: typo? `submissionsr`

nilxam:
:+1:

PR #485, opened 2016-04-18
New staging repair command

nilxam:
Mainly fix the request review if the request got reopened.

### DIFF
--- a/osclib/repair_command.py
+++ b/osclib/repair_command.py
@@ -1,0 +1,55 @@
+from __future__ import print_function
+
+import re
+
+from osc import oscerr
+from osc.core import change_review_state
+from osc.core import get_request
+from osclib.request_finder import RequestFinder
+
+
+class RepairCommand(object):
+
+    def __init__(self, api):
+        self.api = api
+
+    def repair(self, request):
+        reviews = []
+        reqid = str(request)
+        req = get_request(self.api.apiurl, reqid)
+
+        if not req: 
+            raise oscerr.WrongArgs('Request {} not found'.format(reqid))
+
+        if req.state.name != 'review':
+            print('Request "{}" is not in review state'.format(reqid))
+            return
+
+        reviews = [r.by_project for r in req.reviews if ':Staging:' in str(r.by_project) and r.state == 'new']
+
+        if reviews:
+            if len(reviews) > 1:
+                raise oscerr.WrongArgs('Request {} had multiple review opened by different staging project'.format(reqid))
+        else:
+            raise oscerr.WrongArgs('Request {} is not for staging project'.format(reqid))
+
+        staging_project = reviews[0]
+        data = self.api.get_prj_pseudometa(staging_project)
+        for request in data['requests']:
+            if request['id'] == reqid:
+                print('Request "{}" had the good setup in "{}"'.format(reqid, staging_project))
+                return
+
+        # a bad request setup found
+        print('Repairing "{}"'.format(reqid))
+        change_review_state(self.api.apiurl, reqid, newstate='accepted', message='Re-evaluation needed', by_project=staging_project)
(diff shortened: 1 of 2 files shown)


osclib/repair_command.py
```diff
@@ -0,0 +1,59 @@
...
+    def repair(self, request):
+        reviews = []
+        reqid = str(request)
+        req = get_request(self.api.apiurl, reqid)
+
+        if not req: 
+            raise oscerr.WrongArgs('Request {} not found'.format(reqid))
+
+        if req.state.name != 'review':
+            print('Request "{}" is not in review state'.format(reqid))
```
aplanas: Nice, but to avoid some corner cases is better to add `from __future__ import with_statement` in the first line

osclib/repair_command.py
```diff
@@ -0,0 +1,59 @@
...
+            return
+
+        for review in req.reviews:
+            if re.match(r'.*:Staging:(.*)$', str(review.by_project)) and review.state == 'new':
```
aplanas: Oh I remember this one : )). Because for now you are not getting the group regex (adi:1, D, etc) is more clear like this: `if ':Staging:' in str(review.by_project) and ...`

osclib/repair_command.py
```diff
@@ -0,0 +1,59 @@
...
+                reviews.append(review.by_project)
```
aplanas: only as a curiosity, the last three lines can be written as:

```Python
review = [r.by_project for r in req.review if ':Staging:' in str(r.by_project) and r.state == 'new']
```

Not sure that we need this `str()`

osclib/repair_command.py
```diff
@@ -0,0 +1,59 @@
...
+
+        if reviews:
+            if len(reviews) > 1:
+                raise oscerr.WrongArgs('Request {} had multiple review opened by different staging project'.format(reqid))
+        else:
+            print('Request "{}" is not for staging project'.format(reqid))
```
aplanas: This is correct, but a more clear way is:

```Python
if not reviews:
    print('Request "{}" is not for staging project'.format(reqid))
if len(reviews) > 1:
   raise oscerr.WrongArgs('Request {} had multiple review opened by different staging project'.format(reqid))
```

osclib/repair_command.py
```diff
@@ -0,0 +1,59 @@
...
+
+        staging_project = reviews[0]
```
aplanas: Ops, if `not reviews` is `True` this will cause an exception, maybe we need to `raise oscerr.WrongArgs` too in line 36?

osclib/repair_command.py
```diff
@@ -0,0 +1,59 @@
...
+    def perform(self, packages):
+        """
+        Repair request in staging project or move it out
+        :param packages: packages/requests to repair in staging projects
+        """
+
+        for reqid in RequestFinder.find_sr(packages, self.api):
+            self.repair(reqid)
+
+
```
aplanas: extra line : p

aplanas:
Some useless comments and one bug. Once the bug is fix: +1 from my side

osclib/repair_command.py
nilxam: I'll need to use group() later when I add an --readd option to it:)

osclib/repair_command.py
nilxam: indeed. it should not continuing.

osclib/repair_command.py
aplanas: This import is still missing

osclib/repair_command.py
```diff
@@ -0,0 +1,55 @@
+import re
+
+from osc import oscerr
+from osc.core import change_review_state
+from osc.core import get_request
+from osclib.request_finder import RequestFinder
+
+def _is_int(x):
```
aplanas: Actually this is not used, isn't?

aplanas:
+1

PR #488, opened 2016-04-20
Add glibc.i686 to specs if its glibc request

nilxam:


osclib/checkrepo.py
```diff
@@ -439,6 +439,12 @@ def check_specs(self, request_id=None, request=None):
         specs = [en.attrib['name'][:-5] for en in root.findall('entry')
                  if en.attrib['name'].endswith('.spec')]
 
+        # special case for glibc.i686, it have not the relevant specfile for glibc.i686
+        # but must be add it to requests list as a dummy request, otherwise the state
+        # has not be check and won't download it's binaries.
+        if 'glibc' in specs:
```
aplanas: Don't you need to check the arch?

nilxam: does it matter here?

aplanas: uhm TBH I do not understand the situation. There are three spec files in glibc (glibc, glibc-utils and glibc-testsuite), and you are adding a new one (glibc.i686). For this last one there will be a request to OBS to get the md5, some _link test and more. If this spec file is not there, there is a chance that the request will be declined.

Did you tested it?

nilxam: yeah, it's a special case, https://build.opensuse.org/package/show/Base:System/glibc.i686 and you can see there is no relevant specfile in glibc what repochecker can recover we've glibc.i686 there, usually it'll be safe if repohcecker did not to check glibc.i686's binaries, BUT when it was together with gcc5, some libraries can not found glibc-32bit as it was generated from glibc.i686.

I'm waiting for glibc.i686 rebuild done in  Base:Ssytem, then I can know will repochecker downloading it's binaries properly, the rest of pre-check in repochecker looks worked.

aplanas: Oh cool, thanks for the explanation.

nilxam: so don't merge this...I guess it will be fail in _check_repo_download as x86_64 is excluded in glibc.i686 anyhow.

osc-check_repo.py
```diff
@@ -51,6 +51,13 @@ def _check_repo_download(self, request):
 
     # XXX TODO - Rewrite the logic here, meanwhile set is to x86_64
     arch = 'x86_64'
+
+    if request.src_package in request.i686_only:
+        # check imported binaries from x86_64 could be safe, but you can not
+        # get rpmlint.log there, and before this step i586 repo should be
+        # are build succeeded. And we have to ignore x86_64 was excluded.
```
aplanas: Can you fix the comment? I do not get the `... and before this step i586 repo should be are build`

PR #490, opened 2016-04-21
Fix ugly workaround

nilxam:


nilxam:
copied wrong version on my previous commit, shame on me... ;/

PR #492, opened 2016-04-28
Add sle reviews

coolo:


openqa-maintenance.py
```diff
@@ -88,6 +90,42 @@ def settings(self, src_prj, dst_prj, packages, req=None):
         return self._settings.copy()
 
 
+class SUSEUpdate(Update):
+    def __init__(self, settings):
+        Update.__init__(self, settings)
```
aplanas: super(Update, self).__init__(settings)

openqa-maintenance.py
```diff
@@ -58,6 +58,8 @@
 
 logger = None
 
+request_name_cache = dict()
```
aplanas: `request_name_cache = {}`

openqa-maintenance.py
aplanas: Or maybe is better to skip entirely `__init__` definition, by default will use the parent contructor

openqa-maintenance.py
```diff
@@ -88,6 +90,42 @@ def settings(self, src_prj, dst_prj, packages, req=None):
...
+            root = ET.parse(osc.core.http_GET(url)).getroot()
+            if root.attrib.get('cicount'):
+                continue
+            request_name_cache[req.reqid] = action.tgt_package
+            break
+
+        return request_name_cache[req.reqid]
+
+    def settings(self, src_prj, dst_prj, packages, req=None):
+        settings = Update.settings(self, src_prj, dst_prj, packages, req)
```
aplanas: `settings = super(SUSEUpdate, self).settings(src_prj, dst_prj, packages, req)`

openqa-maintenance.py
```diff
@@ -88,6 +90,42 @@ def settings(self, src_prj, dst_prj, packages, req=None):
...
+            url = osc.core.makeurl(
+                req.apiurl,
+                ('source', action.src_project, action.src_package, '_link'))
+            root = ET.parse(osc.core.http_GET(url)).getroot()
+            if root.attrib.get('cicount'):
+                continue
+            request_name_cache[req.reqid] = action.tgt_package
+            break
+
+        return request_name_cache[req.reqid]
```
aplanas:
```python
if req.reqid not in request_name_cache:
    request_name_cache[req.reqid] = self._request_name(req)
return request_name_cache[req.reqid]
```

and in `def _request_name(self, req)`

```python
name = 'unknown'
actions = [action for action in req.get_actions('maintenance_release')
           if not action.tgt_package.startswith('patchinfo')]

for action in actions:
    url = osc.core.makeurl(
        req.apiurl,
        ('source', action.src_project, action.src_package, '_link'))
    if root.attrib.get('cicount'):
        continue
    name = action.tgt_package
    break

return name
```

Can be shorted if there is only one instance of such an element

coolo:
updated

aplanas:
not tested but +1 for my side

PR #495, opened 2016-05-03
Use number instead of id for id->number switch issue

nilxam:
eg. https://build.opensuse.org/project/staging_projects/openSUSE:Factory/adi:5

### DIFF
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -32,8 +32,8 @@ class AdiCommand:
             return
         print project, "is ready"
         for req in info['selected_requests']:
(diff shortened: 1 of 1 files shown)


adrianschroeter:
the api should not have changed due to this internal change. Or is this some of your own stuff?

coolo:
that's the API of the dashboard. It returns the AM as JSON

nilxam:
oh, I attached the wrong link, https://build.opensuse.org/project/staging_projects/openSUSE:Factory/adi:5?format=json <- the right one

PR #500, opened 2016-05-09
Use python logging in ToTest manager

lnussel:


aplanas:
You do not need to pass the logger to different objects. You can configure the root logger in the main function, and recover it in each module:

```python
# This is my entry point
import logging

# Configure the logger
logging.basicConfig(filename='totest.log', level=logging.DEBUG)
```
You can use that in this way:
```python
import logging

# Get the root logger
LOG = logging.getLogger()
...
LOG.debug('Something something')
# This also works:
logging.info('Something something')
```

More info: https://docs.python.org/2/howto/logging.html#logging-basic-tutorial

lnussel:
@aplanas how can the last commit affect _invalidate_all() which comes from memoize?

lnussel:
ah, it only gets added once a memoize method is called at least once

aplanas:
@lnussel yep, is the wrong thing about monkey patching. Now I am not sure if I need to fix the tearDown method in the test or use a different approach for your last commit.

PR #502, opened 2016-05-10
Adjust update crawler for 42.2

lnussel:


update_crawler.py
```diff
@@ -32,25 +33,57 @@
...
+        yield l[i:i+n]
 
 class UpdateCrawler(object):
     def __init__(self, from_prj, to_prj):
         self.from_prj = from_prj
         self.to_prj = to_prj
         self.apiurl = osc.conf.config['apiurl']
         self.debug = osc.conf.config['debug']
+        self.filter_lookup = set()
+	self.caching = False
```
coolo: This looks like whitespace foobar - pep will not approve .)

aplanas: yep, tab

PR #503, opened 2016-05-10
Add Leap 42.2 DVD rules

lnussel:


nilxam:
you got some fun with *ugly packagelist loop* already :)

PR #510, opened 2016-05-23
Also update product version in Staging1 TestDVD

nilxam:


coolo:
that is going to break SLE hard IMO

nilxam:
ah, indeed, I'll fix it for openSUSE only.

nilxam:
fixed.

osclib/freeze_command.py
```diff
@@ -140,8 +140,14 @@ def perform(self, prj):
             time.sleep(1)
         self.build_switch_bootstrap_copy('disable')
 
+        # Update the version information found in the Test-DVD package, to match openSUSE-release
+        if self.api.item_exists(prj, "openSUSE-release"):
+            version = self.api.package_version(prj, 'openSUSE-release')
+            for arch in ['x86_64', 'ppc64le']:
```
lnussel: this should use self.cstaging_dvd_archs

nilxam: what is it?

lnussel: for arch in self.api.cstaging_dvd_archs:

PR #514, opened 2016-05-30
Add --split argument to adi command

nilxam:
With --split argument, staging adi command can splits each package to different adi staging project, the argument can work by 'osc staging adi --split' or 'osc staging adi --move --split'. Note that, it will be
ignore --by-develproject argument too.

### DIFF
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -93,7 +93,7 @@ class AdiCommand:
                     # request_id pretended to be index of non_ring_requests
                     non_ring_requests[request_id] = [request_id]
                 else:
-                    if by_dp is True:
+                    if by_dp:
                         devel_project = self.get_devel_project(source_project, source_package)
                         # try target pacakge in Factory
                         # this is a bit against Leap development in case submissions is from Update,
@@ -103,7 +103,7 @@ class AdiCommand:
                         if devel_project is not None:
                             source_project = devel_project
 
-                    if not source_project in non_ring_requests:
+                    if source_project not in non_ring_requests:
                         non_ring_requests[source_project] = []
                     non_ring_requests[source_project].append(request_id)
                 
@@ -144,4 +144,4 @@ class AdiCommand:
         else:
             self.check_adi_projects() 
             if self.api.is_user_member_of(self.api.user, 'factory-staging'):
-                self.create_new_adi((), by_dp, split=split)
+                self.create_new_adi((), by_dp=by_dp, split=split)
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -48,7 +48,7 @@ class AdiCommand:
         else:
             return node.get('project')
 
-    def create_new_adi(self, wanted_requests, by_dp=False):
+    def create_new_adi(self, wanted_requests, by_dp=False, split=False):
         all_requests = self.api.get_open_requests()
 
         non_ring_packages = []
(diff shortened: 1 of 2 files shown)


osclib/adi_command.py
```diff
@@ -89,19 +89,23 @@ def create_new_adi(self, wanted_requests, by_dp=False):
...
-                if not source_project in non_ring_requests:
-                    non_ring_requests[source_project] = []
-                non_ring_requests[source_project].append(request_id)
+                if split:
+                    # request_id pretended to be a source project
+                    non_ring_requests[request_id] = [request_id]
+                else:
+                    if by_dp is True:
+                        devel_project = self.get_devel_project(source_project, source_package)
+                        # try target pacakge in Factory
```
aplanas: package

osclib/adi_command.py
```diff
@@ -89,19 +89,23 @@ def create_new_adi(self, wanted_requests, by_dp=False):
...
-                        devel_project = self.get_devel_project('openSUSE:Factory', target_package)
-                    if devel_project is not None:
-                        source_project = devel_project
-
-                if not source_project in non_ring_requests:
-                    non_ring_requests[source_project] = []
-                non_ring_requests[source_project].append(request_id)
+                if split:
+                    # request_id pretended to be a source project
+                    non_ring_requests[request_id] = [request_id]
```
aplanas: we are using as index `request_id`, but in by_dp we use `source_project`, is `source_project` a request id too? (sorry to be so lazy to check it for myself)

osclib/adi_command.py
```diff
@@ -89,19 +89,23 @@ def create_new_adi(self, wanted_requests, by_dp=False):
...
+                else:
+                    if by_dp is True:
```
aplanas: `if by_dp:`

osclib/adi_command.py
```diff
@@ -89,19 +89,23 @@ def create_new_adi(self, wanted_requests, by_dp=False):
...
+                        devel_project = self.get_devel_project(source_project, source_package)
+                        # try target pacakge in Factory
+                        # this is a bit against Leap development in case submissions is from Update,
+                        # or any other project than Factory
+                        if devel_project is None and self.api.project.startswith('openSUSE:'):
+                            devel_project = self.get_devel_project('openSUSE:Factory', target_package)
+                        if devel_project is not None:
+                            source_project = devel_project
+
+                    if not source_project in non_ring_requests:
```
aplanas: `if source_project not in non_ring_requests:`

osclib/adi_command.py
```diff
@@ -136,8 +140,8 @@ def perform(self, packages, move=False, by_dp=False):
 
             for request, request_project in items:
                 requests.add(request)
-            self.create_new_adi(requests)
+            self.create_new_adi(requests, split=split)
         else:
             self.check_adi_projects() 
             if self.api.is_user_member_of(self.api.user, 'factory-staging'):
-                self.create_new_adi((), by_dp)
+                self.create_new_adi((), by_dp, split=split)
```
aplanas: good point: `self.create_new_adi((), by_dp=by_dp, split=split)`

osclib/adi_command.py
nilxam: some insane old code :)

aplanas: yep, sorry for not catching it early

osclib/adi_command.py
nilxam: no, source_project is a real project path, in ```by_dp``` we needs it to find the devel project if possible, but since --split will ignores --by-develproject, should not have problem in ```by_dp```.

actually the only place are effected by *source_project/request_id* is ```for source_project, requests in non_ring_requests.items():```, but it just using as index, no matter what is the content of ```source_project```.

nilxam: I think I should change the comment out to - use request_id as index

PR #518, opened 2016-06-01
Tons of fixes

lnussel:


osclib/conf.py
```diff
@@ -32,7 +32,7 @@
 DEFAULT = {
     r'openSUSE:(?P<project>[-\w\d:.]+)': {
         'staging': 'openSUSE:%(project)s:Staging',
-        'staging-group': '%(project.lower)s-staging',
+        'staging-group': 'factory-staging',
```
aplanas: This can be a problem for Leap, I think

osclib/stagingapi.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, apiurl, project):
         self.cstaging = conf.config[project]['staging']
         self.cstaging_group = conf.config[project]['staging-group']
         self.cstaging_archs = conf.config[project]['staging-archs'].split(' ')
+        self.cstaging_dvd_archs = conf.config[project]['staging-dvd-archs'].split(' ')
```
aplanas: actually you can skip the `' '` parameter (same above)

osc-staging.py
```diff
@@ -170,19 +172,14 @@ def do_staging(self, subcmd, opts, *args):
...
-                FreezeCommand(api).perform(api.prj_from_letter(prj))
+                FreezeCommand(api).perform(api.prj_from_letter(prj), copy_bootstrap = opts.bootstrap )
         elif cmd == 'frozenage':
             for prj in args[1:]:
                 print "%s last frozen %0.1f days ago" % (api.prj_from_letter(prj), api.days_since_last_freeze(api.prj_from_letter(prj)))
         elif cmd == 'acheck':
             # Is it safe to accept? Meaning: /totest contains what it should and is not dirty
             version_totest = api.get_binary_version(api.project, "openSUSE-release.rpm", repository="totest", arch="x86_64")
-            skip_totest = False
             if not version_totest:
```
nilxam: so never run if openSUSE?

osc-staging.py
```diff
@@ -191,16 +188,12 @@ def do_staging(self, subcmd, opts, *args):
         elif cmd == 'accept':
             # Is it safe to accept? Meaning: /totest contains what it should and is not dirty
             version_totest = api.get_binary_version(api.project, "openSUSE-release.rpm", repository="totest", arch="x86_64")
-            skip_totest = False
-            if not version_totest:
-                # SLE don't have totest repository and openSUSE-release.rpm
-                skip_totest = api.item_exists(api.project, "release-notes-sles")
 
-            if skip_totest or opts.force:
+            if not version_totest or opts.force:
```
nilxam: same here, so never run if openSUSE?

osc-staging.py
```diff
@@ -193,16 +188,12 @@ def do_staging(self, subcmd, opts, *args):
...
```
nilxam: will it being better for Leap, looks Leap have to use force anyway here, since it has no totest manager involved

osclib/conf.py
lnussel: we don't use a separate group for leap, same people as in Factory so far :)

osc-staging.py
lnussel: skip_totest is superflous. The condition so far is if there is no openSUSE-release in the totest repo, it means there is no totest manager involved. So no extra check for sle needed as SLE doesn't have openSUSE-release anyways.

nilxam: yes, that's why I used release-notes-sles to stop continuing if SLE, I agree skip_totest is uperflous here, but now acheck is useless for TW as it had version_totest

lnussel: I can't follow you here.  For TW version_totest will be set so the else branch will be executed, just as before. Or did I miss anything?

lnussel: ah, got it, was looking at wrong if statement :)

PR #519, opened 2016-06-02
Accept cleanup

DimStar77:
This changes 'osc staging accept' to remove any left packages in the staging projects.

it can be skipped by 'osc staging accept --no-cleanup'

It helps us not having to run after all stagings all the time and it intentionally prints the list of what it removed, as often we make use of this to know what we need to link additionally to rings.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -85,6 +85,8 @@ def _full_project_name(self, project):
               help='show version of the plugin')
 @cmdln.option('--no-freeze', dest='no_freeze', action='store_true',
               help='force the select command ignoring the time from the last freeze')
+@cmdln.option('--no-cleanup', dest='no_cleanup', action='store_true',
+              help='do not cleanup remaining packages in staging projects after accept')
 def do_staging(self, subcmd, opts, *args):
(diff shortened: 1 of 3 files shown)


aplanas:
very cool +1

coolo:
Now please fix it so it works for SP2 too - thanks!

DimStar77:
Do you  have details on how it does not work?

DimStar77:
(looking around on IBS - it seems there is no :DVD subproject - this causing the issues for you?)

coolo:
blindly assuming there is a DVD subproject is a problem in itself, yes.

But SP2 leaves more packages behind than what you hardcoded for all. 

I wonder that @aplanas cheered after having implemented a heavy conf.py which makes what the plugin does very specific to the target project instead of hardcoding things in the code that only apply to openSUSE projects.

DimStar77:
Just see that SLE - Staging accept does not even pass the cleanup() action - it has skip_test = True

coolo:
Yeah, we had this problem already before it seems. So there is nothing to "fix" atm, great I was wrong ;)

PR #520, opened 2016-06-02
Displays as Ring0 package if the subpackage was linked to Ring1

nilxam:
If the package was linked to Ring1 from Ring1, staging list command should display it as Ring0 package ie. it's only be able to puts in ring0 staging project. This is work out https://progress.opensuse.org/issues/8266


nilxam:
don't merge, I found a bug...

nilxam:
fixed!

osclib/stagingapi.py
```diff
@@ -131,23 +143,49 @@ def retried_POST(self, url):
...
+                    if list_only and pkg in except_pkgs and prj == except_pkgs[pkg]:
+                        msg = ''
+                    if len(msg):
+                        raise Exception(msg.format(pkg, ret[pkg], prj))
+                if pkg not in ret:
+                    ret[pkg] = prj
+
+                # put the ring1 package to ring0 list if it was linked from ring0 subpacakge
+                if list_only:
+                    if not re.search('0-Bootstrap' , prj):
```
lnussel: prj.endswith()?

osclib/stagingapi.py
```diff
@@ -131,23 +143,49 @@ def retried_POST(self, url):
...
+            for si in ET.parse(root).getroot().findall('sourceinfo'):
+                pkg = si.get('package')
                 # XXX TODO - Test-DVD-x86_64 is hardcoded here
                 if pkg in ret and not pkg.startswith('Test-DVD-'):
                     msg = '{} is defined in two projects ({} and {})'
-                    raise Exception(msg.format(pkg, ret[pkg], prj))
-                ret[pkg] = prj
+                    if list_only and pkg in except_pkgs and prj == except_pkgs[pkg]:
+                        msg = ''
+                    if len(msg):
```
lnussel: maybe reverse the logic in the if to avoid unsetting msg

lnussel:
IIUC the difference between list_only True/False is whether to return the ring a package is actually in or whether to check the link and use the lowest ring, right? I think changing the naming to reflect what the parameter does rather than where it's used would help a lot to understand what's going on :-)

nilxam:
a better naming is welcome :)

I don't get a better idea for the naming, basically it was only useful for list command, in the real case ie. handling package adding/deleting/checking is still rely on ring_packages.

lnussel:
hmm, maybe check_links=True?

PR #521, opened 2016-06-02
list_packages: Return empty list if project does not exist

DimStar77:
@coolo without further info from you, I think this should fix issues you might have on SP2, where there is no :DVD subproject in the staging areas

Care to test and confirm?

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1149,10 +1149,11 @@ class StagingAPI(object):
         url = self.makeurl(['source', project])
         pkglist = []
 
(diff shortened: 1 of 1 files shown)


coolo:
I can't and won't test accept for SP2. But IMO list_packages should error if there is no such project. So I disagree with this workaround.

But that's not the only problem, see the other PR

nilxam:
hmm..now I aware another issue, SLE staging does not update frozenlinks too frequent, they're created linked package in each staging project and keep them there till updating the frozenlinks at some point.

maybe we should change --no-cleanup to default and --cleanup as option?

PR #522, opened 2016-06-05
Cleanup

DimStar77:
This should make the cleanup feature also wok for SLE - the packages to keep are moved to conf.py

I was considering to add the 'default' for cleanup to be a conf.py option as well and, besides --no-cleanup, also add a --cleanup option.

This would allow for example for:
  * openSUSE to specify cleanup by default
  * SLE to not do cleanup by default

and each of them to change the -preconfigured way using --cleanup or --no-cleanup...

but I was not sure if this is worth it or not. Comments on this are welcome


coolo:
looks very reasonable to me - I leave judging the python to others ;)

DimStar77:
Just to re-iterrate (don't want to have confusion here, as it does change how checkins work): the cleanup/no-cleanup option has only been thought of - not implemented.... 

the current form enables --cleanup for both teams

PR #527, opened 2016-06-13
Rework generate pkgs

nilxam:
Just lazy to fix that complicate conflicts in https://github.com/openSUSE/osc-plugin-factory/pull/520


coolo:
LGTM

PR #529, opened 2016-06-14
Avoid crash in SP2

coolo:


lnussel:
or make _fill_package_meta always put a dict, even if a project has no packages

coolo:
I honestly are very puzzled by this crash - I wonder where 'openSUSE:Factory' is referenced in the SP2 context

lnussel:
my fault. cross obs search doesn't work

PR #533, opened 2016-07-08
Devel projects: whitelist system:snappy

DimStar77:
Snappy is approaching

### DIFF
--- a/osc-check_source.py
+++ b/osc-check_source.py
@@ -257,6 +257,7 @@ def _checker_one_request(self, rq, opts):
             self._devel_projects['filesystems:ceph/'] = 'ceph-deploy'
             self._devel_projects['Virtualization:Appliances:Builder/'] = 'python3-kiwi'
             self._devel_projects['isv:ownCloud:desktop/'] = 'owncloud-client'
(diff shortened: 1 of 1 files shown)


osc-check_source.py
```diff
@@ -257,6 +257,7 @@ def _checker_one_request(self, rq, opts):
             self._devel_projects['filesystems:ceph/'] = 'ceph-deploy'
             self._devel_projects['Virtualization:Appliances:Builder/'] = 'python3-kiwi'
             self._devel_projects['isv:ownCloud:desktop/'] = 'owncloud-client'
+            self._devel_projects['system:snappy/'] = 'snappy'
```
aplanas: We have like 10 elements now, maybe time to put it in a JSON document? Or in a dict in the beginning of the file?

DimStar77: Or cleanup again - in the end we only need the list temporarily until at least one package has the devel prj set to anything there.... splitting that out in a differently maintained format sounds like a good idea indeed (but should not block moving forward in the time being)

aplanas:
+1

PR #536, opened 2016-07-15
totest-manager: The information about failed jobs is meant for normal

DimStar77:
operations / usage, and not only for debug / verbose runs.

It's really annoying not to get the list of failed modules unless one specifies --debug or --verbose
Normal program output should just not be killed off

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -169,7 +169,7 @@ class ToTestBase(object):
                 # print json.dumps(job, sort_keys=True, indent=4), jobname
                 failedmodule = self.find_failed_module(job['modules'])
                 url = 'https://openqa.opensuse.org/tests/%s' % job['id']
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -169,7 +169,7 @@ def overall_result(self, snapshot):
                 # print json.dumps(job, sort_keys=True, indent=4), jobname
                 failedmodule = self.find_failed_module(job['modules'])
                 url = 'https://openqa.opensuse.org/tests/%s' % job['id']
-                logger.info("%s %s %s %s"%(jobname, url, failedmodule, job['retry_avbl']))
+                print jobname, url, failedmodule, job['retry_avbl']
```
okurz: Better use Python 3 compatible print()

PR #537, opened 2016-07-18
Use newcand param make sure don't select the processed requests

nilxam:
Scenario: the accepted request from adi staging what was queued for next Factory check-in round, if executing staging select command against a PROJECT , the processed request will picks up by staging tool again. So use 'newcand' param ensure staging-group's review state is new if selecting a PROJECT unless 'mov'e param is show up.

### DIFF
--- a/osclib/select_command.py
+++ b/osclib/select_command.py
@@ -125,7 +125,11 @@ class SelectCommand(object):
             return False
             # FreezeCommand(self.api).perform(self.target_project)
 
-        for request in RequestFinder.find_sr(requests, self.api):
+        # picks new candidate requests only if it's not to move requests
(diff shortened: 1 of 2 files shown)


osclib/select_command.py
```diff
@@ -125,7 +125,11 @@ def perform(self, requests, move=False,
             return False
             # FreezeCommand(self.api).perform(self.target_project)
 
-        for request in RequestFinder.find_sr(requests, self.api):
+        # picks new candidate requests only if it's not to move requests
+        # ie. the review state of staging-project must be new if newcand is True
+        newcand = False if move else True
```
coolo: I think this is 

newcand = not move

PR #544, opened 2016-08-02
Make repochecker give up on Haskell

lnussel:


osclib/checkrepo.py
```diff
@@ -412,6 +412,12 @@ def check_specs(self, request_id=None, request=None):
             self.change_review_state(request_id, 'accepted', message=msg)
             return requests
 
+        if rq.src_project == 'devel:languages:haskell':
+            msg = 'I give up, Haskell is too hard for me'
```
coolo: can't you make this

```
msg = 'let toohard = "haskell " ++ "repochecker"'

peti:
What is the problem with Haskell packages that makes this change necessary? Is there any particular issue that I could maybe help fixing?

coolo:
There are just too many of them - so we can't have them all in one adi or repochecker explodes, but we can't split them either or we miss cycles

peti:
Oh, what a shame. :disappointed:

PR #545, opened 2016-08-04
Schedule test updates

coolo:


openqa-maintenance.py
```diff
@@ -406,9 +484,13 @@ def calculate_repo_hash(self, repos):
     # if that job doesn't contain the proper hash, we trigger a new one
     # and then we know the build
     def trigger_build_for_target(self, prj, u):
+
+        self.update_test_builds[prj] = '20160804-1'
```
coolo: this is just for debugging - don't merge beforel this is removed :)

coolo:
works fine for me now - @lnussel please review especially d03e4f9

lnussel:
I don't get this on-pass mode. It would not do anything if good ==False, right? So should be the same as review_mode=no

coolo:
review_mode=no ignores the result of good - but I want to accept if good == True

PR #548, opened 2016-08-08
Make check_multiple_specfiles return the linked package name instead

nilxam:
Make check_multiple_specfiles return the linked package name rather than
True or False, and if it has not linkinfo node then continue to process
this package.


nilxam:
no bug here, just an improvement :)

PR #557, opened 2016-08-16
follow packaging guides when checking for IDs to prevent manual overr…

jreidinger:
…iding

prevent issues like https://build.suse.de/request/show/119447
I think it do not matter so much which ID is used as long as it is in sync. So we should follow at least our own packaging guide. Order of ids regexp is same as wiki page have it.

### DIFF
--- a/check_tags_in_requests.py
+++ b/check_tags_in_requests.py
@@ -60,18 +60,56 @@ class TagChecker(ReviewBot.ReviewBot):
         if self.factory is None:
             self.factory = "openSUSE:Factory"
         super(TagChecker, self).__init__(*args, **kwargs)
-        needed_tags = [r'bnc#[0-9]+',
+        # try to keep in sync (same sorting ) with
+        # https://en.opensuse.org/openSUSE:Packaging_Patches_guidelines#Current_set_of_abbreviations
+        needed_tags = [r'boo#[0-9]+',
+                       r'pr#[0-9]+',
+                       r'boost#[0-9]+',
+                       r'ceph#[0-9]+',
+                       r'claws#[0-9]+',
                        r'cve-[0-9]{4}-[0-9]+',
+                       r'RT#[0-9]+',
+                       r'deb#[0-9]+',
+                       r'beo#[0-9]+',
                        r'fate#[0-9]+',
-                       r'boo#[0-9]+',
+                       r'fdo#[0-9]+',
+                       r'gnu#[0-9]+',
+                       r'GCC#[0-9]+',
+                       r'gh#[a-zA-Z0-9]+',
+                       r'bgo#[0-9]+',
+                       r'gvz#[0-9]+',
+                       r'bio#[0-9]+',
+                       r'jenk#[0-9]+',
+                       r'kde#[0-9]+',
+                       r'bko#[0-9]+',
+                       r'lp#[0-9]+',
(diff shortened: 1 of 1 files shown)


check_tags_in_requests.py
```diff
@@ -60,18 +60,56 @@ def __init__(self, *args, **kwargs):
...
         super(TagChecker, self).__init__(*args, **kwargs)
-        needed_tags = [r'bnc#[0-9]+',
+        # try to keep in sync (same sorting ) with
+        # https://en.opensuse.org/openSUSE:Packaging_Patches_guidelines#Current_set_of_abbreviations
+        needed_tags = [r'boo#[0-9]+',
+                       r'pr#[0-9]+',
+                       r'boost#[0-9]+',
+                       r'ceph#[0-9]+',
+                       r'claws#[0-9]+',
                        r'cve-[0-9]{4}-[0-9]+',
```
lslezak: Nitpicking: the minimal number of digits for the second part of a CVE identifier is four (seehttps://cve.mitre.org/cve/identifiers/syntaxchange.html#new), it should be actually `cve-[0-9]{4}-[0-9]{4,}`...

check_tags_in_requests.py
```diff
@@ -60,18 +60,56 @@ def __init__(self, *args, **kwargs):
...
+                       r'Ximian#[0-9]+',
+                       r'bxo#[0-9]+',
+                       r'bxc#[0-9]+']
         self.needed_tags_re = [re.compile(tag, re.IGNORECASE) for tag in needed_tags]
         self.review_messages['declined'] = """
 (This is a script running, so report bugs)
 
 We require a ID marked in .changes file to detect later if the changes
-are also merged into openSUSE:Factory. We accept bnc#, cve#, fate#, boo#, bsc# and bgo# atm.
+are also merged into openSUSE:Factory. We accept tags known in
```
lslezak: Maybe `tags listed in` sounds better...

check_tags_in_requests.py
```diff
@@ -60,18 +60,56 @@ def __init__(self, *args, **kwargs):
         if self.factory is None:
             self.factory = "openSUSE:Factory"
         super(TagChecker, self).__init__(*args, **kwargs)
-        needed_tags = [r'bnc#[0-9]+',
+        # try to keep in sync (same sorting ) with
```
lslezak: NP: extra space before `)`

check_tags_in_requests.py
jreidinger: agreed, will fix it

check_tags_in_requests.py
jreidinger: will fix

check_tags_in_requests.py
jreidinger: ok, will modify

nilxam:
actually it's not anymore rely on needed_tags to check the request is valid or not, it's rely on server side issues diff https://github.com/openSUSE/osc-plugin-factory/blob/master/check_tags_in_requests.py#L111

jreidinger:
@nilxam so it is dead code and need to be fixed on server side?

nilxam:
yes, it's dead code, if we still want to use issues diff to ensure the request is valid we have to extend that list in the server side.

lnussel:
so closing

PR #559, opened 2016-08-22
Update skip list 20160822

nilxam:


DimStar77:
this sounds almost like something that would want to be maintained in a file inside 42.2/00Meta - similar to what we do with the known_failures for ttm for example

nilxam:
good idea, don't know will it break leaper though...
*or* I can use another file put on https://build.opensuse.org/package/show/openSUSE:Leap:42.2:Staging/dashboard , will do that later.

PR #562, opened 2016-08-23
Renamed libgloox to gloox.

Mailaender:
I decided to adhere to our current library naming standard right from the beginning and got a decline. :P https://build.opensuse.org/request/show/421161


DimStar77:
doesn't really matter - the package name here is just 'the package that is added in the list of what we have' - not of what we receive.

What is an issue though in the SR you reference is that source and destination package names do not match. You have to properly name the package in the devel prj too

PR #565, opened 2016-08-30
Hard require inkscape due to the pattern change

nilxam:
Inkscape now are Suggests in the pattern rather than Recommends, hard require it in the staging DVD instead of drop the test from openQA.

Already applied it in the packagelists actually.


DimStar77:
:+1:

PR #566, opened 2016-09-04
add Application:Dochazka as devel project

tampakrap:
dochazka is the upcoming application that we are going to use in the Prague
SUSE office as our attendance / payroll system. It comes with a CLI client,
thus it would be really beneficial to add it to factory and upcoming leap


smithfarm:
LGTM - @tampakrap This means `perl-App-Dochazka-Common` will be the first Dochazka package you submit to Factory.

PR #568, opened 2016-09-09
Accept requests that are only stuck in Factory staging

lnussel:


check_source_in_factory.py
```diff
@@ -148,7 +148,14 @@ def _check_requests(self, project, package, rev):
                         return True
                     elif req.state.name == 'review':
                         self.logger.info("request still in review")
-                        return None
+                        for r in req.reviews:
+                            if r.by_project and r.state == 'new' and r.by_project.startswith('openSUSE:Factory:Staging:'):
```
DimStar77: the request will never be 'new' AND pending staging review - then it's still state == review

coolo:
This is *heavily* undermotivated IMO.

coolo:
and this is not breaking factory staging process ?

lnussel:
do you have something in mind? AFAIK the bot is not used in Factory itself, only by maintenance and Leap

coolo:
as long as it's handling this correctly, it might be fine

PR #584, opened 2016-10-19
Add kgraft settings

coolo:


openqa-maintenance.py
```diff
@@ -209,6 +209,54 @@ def settings(self, src_prj, dst_prj, packages, req=None):
 
         return settings
 
+KRAFT_SETTINGS = {
```
aplanas: missing G?

coolo: thanks for spotting - but my code found it when testing :)

PR #586, opened 2016-10-25
Remove dead code

lnussel:


coolo:
if this is dead code, the decline message is wrong

PR #593, opened 2016-11-15
Let Test-DVD is invalid in the beginning of the new round

nilxam:
Make sure Test-DVD won't in green state when new round just started, wipe Test-DVD binary and intended to break kiwi build when accepting the staging.

Next step is let create_test_dvds skips the staging what the build is disabled to avoid create_test_dvds correction the kiwi file.

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -98,6 +98,31 @@ class AcceptCommand(object):
         for package in clean_list:
             print "[cleanup] deleted %s/%s" % (project, package)
             delete_package(self.api.apiurl, project, package, force=True, msg="autocleanup")
+
+        # wipe Test-DVD binaries and breaks kiwi build
+        if project.startswith('openSUSE:'):
+            for package in pkglist:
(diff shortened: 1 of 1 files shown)


nilxam:
merge now, have to test it in the next check-in round :)

osclib/accept_command.py
```diff
@@ -98,6 +98,31 @@ def cleanup(self, project):
...
             print "[cleanup] deleted %s/%s" % (project, package)
             delete_package(self.api.apiurl, project, package, force=True, msg="autocleanup")
+
+        # wipe Test-DVD binaries and breaks kiwi build
+        if project.startswith('openSUSE:'):
+            for package in pkglist:
+                if package.startswith('Test-DVD-'):
+                    # intend to break the kiwi file
+                    arch = package.split('-')[-1]
+                    fakepkgname = 'I-am-breaks-kiwi-build'
```
lnussel: DVD-content-needs-recalculation?

PR #595, opened 2016-11-16
Import http_POST for accept_command

DimStar77:
Since commit 295fcd we newly also require http_POST in order to
put a new file in place after accepting.

Also add quotes around the fakepkgname when injecting it into the kiwi file.


nilxam:
tested

PR #596, opened 2016-11-16
Move testdvd status chack part under the condition protect

nilxam:
that ```|| true``` is prevent set -e kills program


nilxam:
have to deploy it asap

PR #602, opened 2016-11-29
Update version in manager_42 and update_crawler

nilxam:


manager_42.py
```diff
@@ -51,19 +51,23 @@ def __init__(self, from_prj, caching = True):
...
+                'openSUSE:Leap:42.2',
+                'openSUSE:Leap:42.2:NonFree:Update',
+                'openSUSE:Leap:42.2:NonFree',
                 'openSUSE:Leap:42.1:Update',
                 'openSUSE:Leap:42.1',
                 'openSUSE:Leap:42.1:NonFree:Update',
                 'openSUSE:Leap:42.1:NonFree',
                 'openSUSE:Factory',
                 'openSUSE:Factory:NonFree',
-                'openSUSE:Leap:42.2:SLE-workarounds'
```
lnussel: I assume we need to keep that and add a similar project for 42.3

nilxam: we don't have SLE-workarounds in 42.3 now, we can add it later when we created it...

PR #603, opened 2016-11-29
Decline add_role request or incorrect delete request

nilxam:


osc-check_source.py
```diff
@@ -342,6 +342,40 @@ def _checker_one_request(self, rq, opts):
             if self._checker_accept_request(opts, id_, msg, diff=diff):
                 continue
 
+        elif _type == "add_role":
+            # decline add_role request to Factory
+            msg = "It's disallowed to be Factory maintainer, please submit this request against the devel package"
```
DimStar77: I'd probably write this a bit different - will think of something

osc-check_source.py
```diff
@@ -342,6 +342,40 @@ def _checker_one_request(self, rq, opts):
...
+                links = si.findall('linked')
+                if links is None or len(links) == 0:
+                    self._checker_change_review_state(opts, id_, 'accepted',
+                                                  by_group='factory-auto',
+                                                  message="Unchecked request type %s" % _type)
+                else:
+                    linked = links[0]
+                    lprj = linked.get('project')
+                    lpkg = linked.get('package')
+                    msg = "This is an incorrect reuqest, it's a linked package to %s/%s" % (lprj, lpkg)
```
DimStar77: s/reuqest/request/

osc-check_source.py
nilxam: give me the text you're preferred, I'll change it then

PR #604, opened 2016-11-30
Just findall linked tag to instead of search sourceinfo

nilxam:
In this case, sourceinfo is not sub-element but root.


DimStar77:
:+1:

nilxam:
good spot, and updated.

PR #609, opened 2016-12-06
Workaround for OBS caching problem

lnussel:
https://github.com/openSUSE/open-build-service/issues/2301


lnussel:
testsuite can't cope with the extra request. *sigh*

jberry-suse:
1. If this is a regular occurrence then it should be applied to `ReviewBot` instead of just a specific bot.
2. Have we seen this on a regular basis? I would expect all the bot to have trouble.

jberry-suse:
A full year has passed. Re-open if fixed.

PR #611, opened 2016-12-06
Indicate select command progress.

jberry-suse:
Given the select command can take a significant length of time to complete an indication of progress is handy. Uses the same format as zypper during installation.

### DIFF
--- a/osclib/select_command.py
+++ b/osclib/select_command.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 from xml.etree import cElementTree as ET
 
 from osc import oscerr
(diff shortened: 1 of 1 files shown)


jberry-suse:
For example:
```
(1/11) Adding request "444238" to project "openSUSE:Leap:42.3:Staging:D"
(2/11) Adding request "444239" to project "openSUSE:Leap:42.3:Staging:D"
(3/11) Adding request "444241" to project "openSUSE:Leap:42.3:Staging:D"
...
```

DimStar77:
I don't mind it - but I'd prefer if we find out where all the time actually is wasted so we can improve it

osclib/select_command.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import print_function
```
lnussel: why do we need this?

aplanas: Because https://github.com/jberry-suse/osc-plugin-factory/blob/6e0f2165e0b6f25a2fdc7af97d02ed541c5c34d0/osclib/select_command.py#L137

lnussel: didn't know about the end parameter

PR #614, opened 2016-12-13
Enhance ReviewBot based tools to work across build service instances.

jberry-suse:
Additionally, `manager_42.py` is updated in order to generate `lookup.yml` across instances as well.

To use simply set the `--apiurl` (`-A`) option to `htts://api.suse.de` (or configured alias) given configured `osc` access. This would benefit from further testing by those more familiar with the abilities and edge-cases of the affected tools

If one includes the `--debug` option the `apiurl` switches will be indicated. The switches are performed based on the relevant project where applicable. All comments, and context-less `request` API calls use the base URL which is the one passed on command line.

For example,
```bash
# sr#125153 is assumed to be on IBS
./leaper.py --verbose --debug --dry --apiurl ibs id 125153

# sr#445436 is assumed to be on OBS
./leaper.py --verbose --debug --dry id 445436

# will start on IBS and witch to OBS following history
./manager_42.py --debug --dry --apiurl ibs -f SUSE:SLE-12:Update kernel-source
```

It attempts to load the lookup maps for SLE (added fail handling since they are not present yet) and does a very basic check against `SUSE:SLE-12-SP2:Update` if base URL is IBS. This can be expanded in followup requests (perhaps by those a bit more familiar with the expected workflow). Otherwise, I believe it accomplishes the goal, but as stated I can only run the changes in a limited fashion.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -50,8 +50,12 @@ class ReviewBot(object):
     DEFAULT_REVIEW_MESSAGES = { 'accepted' : 'ok', 'declined': 'review failed' }
     REVIEW_CHOICES = ('normal', 'no', 'accept', 'accept-onpass', 'fallback-onfail', 'fallback-always')
 
-    def __init__(self, apiurl = None, dryrun = False, logger = None, user = None, group = None):
-        self.apiurl = apiurl
+    def __init__(self, apiurl = None, dryrun = False, logger = None, user = None, group = None, apiurl_default = None):
+        self._apiurl = apiurl
+        self._apiurl_default = apiurl_default
+        self._apiurl_multi = apiurl_default is not None and apiurl != apiurl_default
+        self.projects = None
+        self.ibs = apiurl.startswith("https://api.suse.de")
         self.dryrun = dryrun
         self.logger = logger
         self.review_user = user
@@ -61,6 +65,26 @@ class ReviewBot(object):
         self._review_mode = 'normal'
         self.fallback_user = None
         self.fallback_group = None
+
+    def apiurl(self, project = None):
+        # Use base URL except when evaluating against multiple API endpoints and
+        # the project context is available. If the project is not  available on
+        # the base endpoint then switch to the default endpoint which is assumed
+        # to contain the project. This could be extended to handle a list of API
+        # endpoints by tracking the projects of each.
+        if self._apiurl_multi:
+            if self.projects is None:
+                self.logger.debug("requesting project list")
+                self.projects = osc.core.meta_get_project_list(self._apiurl)
+
+            if project is not None and project not in self.projects:
+                self.logger.debug("apiurl = default ({}); project = {}".format(self._apiurl_default, project))
+                return self._apiurl_default
+
+            # Only print debug message if in multi mode.
+            self.logger.debug("apiurl = base ({}); project = {}".format(self._apiurl, project))
+
+        return self._apiurl
 
     @property
     def review_mode(self):
@@ -74,7 +98,7 @@ class ReviewBot(object):
 
     def set_request_ids(self, ids):
         for rqid in ids:
-            u = osc.core.makeurl(self.apiurl, [ 'request', rqid ], { 'withhistory' : '1' })
+            u = osc.core.makeurl(self.apiurl(), [ 'request', rqid ], { 'withhistory' : '1' })
             r = osc.core.http_GET(u)
             root = ET.parse(r).getroot()
             req = osc.core.Request()
@@ -130,7 +154,7 @@ class ReviewBot(object):
         if doit == True:
             self.logger.debug("setting %s to %s"%(req.reqid, state))
             if not self.dryrun:
-                osc.core.change_review_state(apiurl = self.apiurl,
+                osc.core.change_review_state(apiurl = self.apiurl(),
                         reqid = req.reqid, newstate = newstate,
                         by_group=self.review_group,
                         by_user=self.review_user, message=msg)
@@ -153,7 +177,7 @@ class ReviewBot(object):
         else:
             raise osc.oscerr.WrongArgs("missing by_*")
 
-        u = osc.core.makeurl(self.apiurl, ['request', req.reqid], query)
+        u = osc.core.makeurl(self.apiurl(), ['request', req.reqid], query)
         if self.dryrun:
             self.logger.info('POST %s' % u)
             return True
@@ -245,7 +269,7 @@ class ReviewBot(object):
             return None
 
     def get_originproject(self, project, package, rev=None):
-        root = ReviewBot._get_sourceinfo(self.apiurl, project, package, rev)
+        root = ReviewBot._get_sourceinfo(self.apiurl(project), project, package, rev)
         if root is None:
             return None
 
@@ -256,7 +280,7 @@ class ReviewBot(object):
         return None
 
     def get_sourceinfo(self, project, package, rev=None):
-        root = ReviewBot._get_sourceinfo(self.apiurl, project, package, rev)
+        root = ReviewBot._get_sourceinfo(self.apiurl(project), project, package, rev)
         if root is None:
             return None
 
@@ -273,7 +297,7 @@ class ReviewBot(object):
     def _get_linktarget(self, src_project, src_package):
 
         query = {}
-        url = osc.core.makeurl(self.apiurl, ('source', src_project, src_package), query=query)
+        url = osc.core.makeurl(self.apiurl(src_project), ('source', src_project, src_package), query=query)
         try:
             root = ET.parse(osc.core.http_GET(url)).getroot()
         except urllib2.HTTPError:
@@ -289,7 +313,7 @@ class ReviewBot(object):
     def can_accept_review(self, request_id):
         """return True if there is a new review for the specified reviewer"""
         states = set()
-        url = osc.core.makeurl(self.apiurl, ('request', str(request_id)))
+        url = osc.core.makeurl(self.apiurl(), ('request', str(request_id)))
         try:
             root = ET.parse(osc.core.http_GET(url)).getroot()
             if self.review_user:
@@ -314,7 +338,7 @@ class ReviewBot(object):
            review = "@by_user='%s'+and+@state='new'"%self.review_user
         else:
            review = "@by_group='%s'+and+@state='new'"%self.review_group
-        url = osc.core.makeurl(self.apiurl, ('search', 'request'), "match=state/@name='review'+and+review[%s]&withhistory=1"%review)
+        url = osc.core.makeurl(self.apiurl(), ('search', 'request'), "match=state/@name='review'+and+review[%s]&withhistory=1"%review)
         root = ET.parse(osc.core.http_GET(url)).getroot()
 
         self.requests = []
@@ -325,7 +349,7 @@ class ReviewBot(object):
             self.requests.append(req)
 
     def set_request_ids_project(self, project, typename):
-        url = osc.core.makeurl(self.apiurl, ('search', 'request'),
+        url = osc.core.makeurl(self.apiurl(project), ('search', 'request'),
             "match=(state/@name='review'+or+state/@name='new')+and+(action/target/@project='%s'+and+action/@type='%s')&withhistory=1"%(project, typename))
         root = ET.parse(osc.core.http_GET(url)).getroot()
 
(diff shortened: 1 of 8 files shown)


lnussel:
Interesting solution. The calls where apiurl() is called without a project can't know the right apiurl though, right?

Also, what about outsourcing the hardcoded ibs/^SUSE: to some config file. Ie to define overrides for specific projects, like e.g.
`openSUSE.org:openSUSE:Factory https://api.opensuse.org`

jberry-suse:
> Interesting solution. The calls where apiurl() is called without a project can't know the right apiurl though, right?

The calls assume that the base URL is correct which from what I can tell works. Typically those calls are related to commenting or direct interaction with an SR. The calls that should change are those that load lookups and search for SRs which have the project context. In fact the searches that then proceed to load an SR also have the project context which is used.

> Also, what about outsourcing the hardcoded ibs/^SUSE: to some config file. Ie to define overrides for specific projects, like e.g.

Sure, I considered this, but figured all the logic surrounding Leap/SLE is very much hardcoded so it seemed moot, but I can do so.

jberry-suse:
Basically, everything should be on the base URL unless a reason to not be and from what I could tell playing around with it and reading the through the code the only places that have a possibility of switching are those with a project context.

lnussel:
Am 15.12.2016 um 20:23 schrieb Jimmy Berry:
> Basically, everything should be on the base URL unless a reason to not
> be and from what I could tell playing around with it and reading the
> through the code the only places that have a possibility of switching
> are those with a project context.

Unfortunately it's the other way around. The algorithm of the
factory source checker iterates over requests in IBS. Then checks for
sources in Factory (can be done either via obs or via openSUSE.org
prefix in ibs), then potentially look for requests in obs (only
possible by talking to obs directly).

jberry-suse:
Looking at `check_source_in_factory.py` all references to `self.apiurl()` include a project even the one to look at requests.

```python
requests = osc.core.get_request_list(self.apiurl(project), project, package, None, ['new', 'review'], 'submit')
```

One of the few places that does not have the context is `ReviewBot.py`:

```python
def set_request_ids(self, ids):
        for rqid in ids:
            u = osc.core.makeurl(self.apiurl(), [ 'request', rqid ], { 'withhistory' : '1' })
```

Which makes sense since that is the starting point when it is merely given a request ID and knowns nothing else other than if the `-A` option was used so it assumes that's the place to look. Given the request ID may (and in many cases does) exist on both instances it really cannot known better and from what I know of usage there aren't any cases when the behavior would be undesirable. I can read through things again, but I believe the few instances where project is not available are all acceptable. Like when the bot goes to make a comment...it only does so on the requests under review which again are on the base url.

Did you run into an issue while trying the code and if so could you give the command line used to reproduce?

jberry-suse:
Of the _49_ references to `self.apiurl()` only _13_ are without project context. Of those almost all are either comment API or load requests without context that can only reasonably be done against base URL. The remainder are in `check_maintenance_incidents.py` and the calling function does have the project context so it could be passed along. I'll try and make out if that is necessary.

jberry-suse:
The 3 questionable calls are in:

- MaintenanceChecker._maintainers()
- MaintenanceChecker.add_devel_project_review()

Both are performing package owner searches and purposefully do not include a project as context. The source or destination projects on the request could be used, but it is not entirely clear what would be desirable from an SLE standpoint. Perhaps in IBS mode both instances should be queried and the resulting owners merged?

jberry-suse:
If the idea is that all packages are to be submitted to factory first then presumably they would have a devel project on OBS and all these queries should target OBS? Otherwise, this simply needs a decision made as all the required context is available.

jberry-suse:
From some test queries, there are owners that makes sense on IBS so potentially this is fine since it will use the base URL for these requests. So upon further consideration this would seem to fit the pattern I originally indicated that all project-less queries make sense to run against base URL.

jberry-suse:
Continuing on the subject of the config file for mapping project to instance. Is it intended to support instances beyond ibs and obs which is quasi supported since one could easily tweak the conditional? If so there are some specifics to work out and decide if really worth the extra complexity since it does not seem we have any usecases that would benefit from this especially given much of the other code already makes a lot of assumptions that would also need to be abstracted.

Assuming it is desired it would seem the `-A` option would need to be abstracted or the config file would have to assume either an alias or absolute URLs to the two instances. Otherwise, the script could end up using three different configs.

For example, the follow config would assume stopping at the first match that starts with prefix and `*` representing the fallback/default.
```yaml
"SUSE:": "https://api.suse.de"
"*": "https://api.opensuse.org"
```

If then the script is called like:

```
$ ./leaper -A ibs id 1234
# osc.conf.get_config(override_apiurl = "ibs")
# request 1234 loaded via ibs
# SUSE: project loaded via https://api.suse.de (may not be the same as ibs or whatever -A option was)
# openSUSE project loaded via https://api.opensuse.org (default from osc config)
```

The config could be altered to use `base` and `default` or whatever terms are desired to reference the two servers and the values then assumed to be default osc config and `-A` option or config options would be needed to define those and perhaps a flag that just says use multi-instance mode. Without some clear usecases it opens up a lot of chances for weirdness.

Unless otherwise indicated I'll leave the config out.

To be clear, as I realize it may not have been, the script currently assumes whatever is loaded as a result of `-A` is the base URL and whatever URL loads as a result of loading the config ignoring the `-A` option is the default URL (which is expected to be a valid OBS config). It is the same result one would get running the script without the `-A` option and once with the `-A` option which seems to be reasonable since it utilizes all the configuration already available through osc and still allows the starting point to be indicated.

The script does check that when an overridden config is used that the URL ends up being `https://api.suse.de` in order to use the IBS/OBS mode. Otherwise it assumes the script is in single instance mode with a custom config which again seems reasonable.

lnussel:
I guess I have to try this.

jberry-suse:
I updated to check against project list for base server as that seems to be the only fully reliable method since the history search can include devel projects which have no consistent prefix.

```diff
diff --git a/ReviewBot.py b/ReviewBot.py
index e0f464d..5befad0 100644
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -53,6 +53,8 @@ class ReviewBot(object):
     def __init__(self, apiurl = None, dryrun = False, logger = None, user = None, group = None, apiurl_default = None):
         self._apiurl = apiurl
         self._apiurl_default = apiurl_default
+        self._apiurl_multi = apiurl_default is not None and apiurl != apiurl_default
+        self.projects = None
         self.ibs = apiurl.startswith("https://api.suse.de")
         self.dryrun = dryrun
         self.logger = logger
@@ -65,16 +67,22 @@ class ReviewBot(object):
         self.fallback_group = None
 
     def apiurl(self, project = None):
-        # Use base URL except when evaluating in IBS mode and the project
-        # context is available. If the project is not developed on IBS then
-        # switch to OBS which is assumed to be the default config URL.
-        if self.ibs:
-            if project is not None and not project.startswith("SUSE:"):
+        # Use base URL except when evaluating against multiple API endpoints and
+        # the project context is available. If the project is not  available on
+        # the base endpoint then switch to the default endpoint which is assumed
+        # to contain the project. This could be extended to handle a list of API
+        # endpoints by tracking the projects of each.
+        if self._apiurl_multi:
+            if self.projects is None:
+                self.logger.debug("requesting project list")
+                self.projects = osc.core.meta_get_project_list(self._apiurl)
+
+            if project is not None and project not in self.projects:
                 self.logger.debug("apiurl = default ({}); project = {}".format(self._apiurl_default, project))
                 return self._apiurl_default
 
-            # Only print debug message if in IBS mode.
-            self.logger.debug("apiurl = base ({})".format(self._apiurl))
+            # Only print debug message if in multi mode.
+            self.logger.debug("apiurl = base ({}); project = {}".format(self._apiurl, project))
 
         return self._apiurl
 
```

I left the `self.ibs` attribute for leaper which makes use of it several times, but made the core `apiurl()` method completely generic and should not care what the two endpoints are.

lnussel:
I tried it with the sudo submission internally but that failed. The project names in ibs and obs are different due to the prefix of course. So I played with it a bit and came up with this:
WDYT? https://github.com/openSUSE/osc-plugin-factory/pull/619

jberry-suse:
Did you include the --factory option to use link instead of directly talking to OBS? I indicated that seems to defeat the point of talking to two instances and doesn't work all the time with devel projects. I can play around with the pull request you came up with, but as you can see from the examples at the top there are a number of different scripts that work with this approach.

jberry-suse:
I commented on the other PR, but from what I can tell the other approach does not work for all the scripts and at minimum would need a bit more work, but still not convinced it is the way to go. I will agree that the requests from link as source seems to be the one remaining issue for this approach.

jberry-suse:
Looks like basically the exact same change you made in #619 is what's needed for this to work.
```python
                    if project.startswith(prefix):
                        apiurl = url
                        project = project[len(prefix):]
```

jberry-suse:
If the only place the link prefix can be encountered is when loading the requests then that should be the only place it needs to be removed (as you did). Presumably, if the rest of the requests start from unprefixed project they should never somehow end up pointing to a prefixed one. If that is not true then removing prefix from project would become a bit more complicated.

jberry-suse:
One could argue this could/should be handled at the osc level as it would improve the cli experience as well.

jberry-suse:
I merged in your changes from #619 and modified a bit. It may make sense to drop part of config map since apiurl part is not used. It's either that or the `ReviewBot.strip_project_namespace()` should be modified to handle changing apiurl and has to be called directly just before anywhere it is possible for project to be namespaced. The locations where possible seem to be only request related (ie when requests are loaded and then subsequent requests are made using the project information from the request). Unfortunately, that would again require all the calling code to always handle this and seems like something that can easily be broken with future additions.

Using the project name as the key seems to be required since IBS requests not originating from the `openSUSE.org:` link and devel project checks end up needing to be handled without any further context.

Since this setup just strips the namespace prefix there could be a potential issue if the exact project name exists on both servers and they are not mirrors of one another. Figured it's time to check.

```bash
$ osc ls / | sort > obs.list
$ osc -A ibs ls / | sort > ibs.list
$ comm -12 obs.list ibs.list | grep -P "^[^h]"
```

```
deleted
Documentation
Documentation:Tools
openSUSE:Backports:SLE-12
openSUSE:Backports:SLE-12-SP1
openSUSE:Backports:SLE-12-SP2
SUSE:SLE-11:GA
SUSE:SLE-11:Update
SUSE:SLE-12:GA
SUSE:SLE-12-SP1:GA
SUSE:SLE-12-SP1:Update
SUSE:SLE-12-SP2:GA
SUSE:SLE-12-SP2:Update
SUSE:SLE-12:Update
```

Those seem fine.

With the additions, the one new case is properly handled (ie an IBS SR whose source project is provided by the `openSUSE.org:` link).

```bash
$ ./check_source_in_factory.py --osc-debug -A ibs --dry --debug id 125419
```

```
DEBUG:check_source_in_factory.py:requesting project list
GET https://api.suse.de/source                                                                                                             
DEBUG:check_source_in_factory.py:apiurl = base (https://api.suse.de); project = None                                                       
GET https://api.suse.de/request/125419?withhistory=1
INFO:check_source_in_factory.py:checking 125419
INFO:check_source_in_factory.py:openSUSE:Factory/sudo@82 -> SUSE:SLE-12-SP3:GA/sudo
DEBUG:check_source_in_factory.py:apiurl = default (https://api.opensuse.org); project = openSUSE:Factory
GET https://api.opensuse.org/source/openSUSE:Factory/sudo?rev=82&view=info
DEBUG:check_source_in_factory.py:checking sudo in openSUSE:Factory
DEBUG:check_source_in_factory.py:apiurl = default (https://api.opensuse.org); project = openSUSE:Factory
GET https://api.opensuse.org/source/openSUSE:Factory/sudo/_meta
DEBUG:check_source_in_factory.py:apiurl = default (https://api.opensuse.org); project = openSUSE:Factory
GET https://api.opensuse.org/source/openSUSE:Factory/sudo?view=info
DEBUG:check_source_in_factory.py:srcmd5 matches
INFO:check_source_in_factory.py:sudo is in openSUSE:Factory
DEBUG:check_source_in_factory.py:apiurl = base (https://api.suse.de); project = None
GET https://api.suse.de/request/125419
INFO:check_source_in_factory.py:can't change state, 125419 does not have the reviewer
INFO:check_source_in_factory.py:125419 accepted: ok
DEBUG:check_source_in_factory.py:125419 review not changed
```

I re-ran a bunch of the commands the two of us had been playing with and everything seems proper. Includes the following.

```bash
$ ./check_source_in_factory.py --osc-debug -A ibs --dry --debug project SUSE:SLE-12-SP3:GA submit
```

jberry-suse:
To be clear those in the common list will default to IBS since they will be found in the project list.

lnussel:
close as per our discussion. Thanks for the suggestion and dicussion anyways.

PR #615, opened 2016-12-15
Disable debuginfo repo

nilxam:


coolo:
I don't think this is clever. At least perl behaves different when it's compiled without debug

nilxam:
@lnussel want it. do you have more details of the risk?

coolo:
@lnussel wanting it is quite a bad reasoning in itself :)

lnussel:
the test in ring2 with disabled debuginfo didn't help much :(

PR #616, opened 2016-12-16
Add SUTs kGraft{0g,1a,21}

fschilling:


coolo:
as you merged: do you also know how to deploy on maintenance VM?

plusky:
hmm is it documented somewhere? - I would cherry-pick this commit on maintenance machine, and I assume that I would have to restart the bot.

PR #622, opened 2016-12-28
Rework ReviewBot.CommandLineInterface to provide class option.

jberry-suse:
Alleviates the need for a lot of duplicate code, some of which is already out-of-sync. Some examples of out-of-sync code are as follows.

- `check_source_in_factory.py`: not include group parameter
- `check_tags_in_requests.py`: missing default user

Either way this is primarily intended as a cleanup. I ran into this when I started porting `osc-check_source.py` to `ReviewBot` and needed to copy/paste a bunch of code for no good reason.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -351,6 +351,7 @@ class ReviewBot(object):
 class CommandLineInterface(cmdln.Cmdln):
     def __init__(self, *args, **kwargs):
         cmdln.Cmdln.__init__(self, args, kwargs)
+        self.clazz = ReviewBot
 
     def get_optparser(self):
         parser = cmdln.Cmdln.get_optparser(self)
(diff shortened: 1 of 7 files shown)


lnussel:
So much better. Thanks! :)
With this the ReviewBot.setup_checker bits could be moved to postoptparse so bots override postoptparse instead. Might be more straight forward I guess.
Would it make sense to pass the class to CommandLineInterface? Save another line in very simple bots.

jberry-suse:
When making the changes I was thinking `postoptparse()` seemed the like the intended place, but imagine that given `setup_checker()` had to be overridden it was simply avoided to add the other method. If changed the `super().postoptparse()` would still need to be called and `self.checker` referenced instead of `bot` returned from `super().setup_checker()`. I am happy to make the change although I am not sure much is gained.

The downside of passing the class to the constructor is that it would have to be removed before passing along to `cmdln.Cmdln.__init__()` which ends up being a tad ugly for what is gained.

PR #623, opened 2016-12-29
Port osc-check_source.py to ReviewBot as check_source.py.

jberry-suse:
In the process of porting many improvements and bug fixes were added. Usage follows the standard `ReviewBot` format and allows for new execution styles, but the following examples replicate previous usage.

```
# Review all requests assigned to factory-auto.
./check_source.py --group factory-auto review

# Review request 13370.
./check_source.py --group factory-auto id 13370

# --project translates to project command.
# No longer automatically includes $project:NonFree.
./check_source.py --group factory-auto project openSUSE:Factory
```

New options available are:

```
--ignore-devel          ignore devel projects for target package
--devel-whitelist=FILE  file containing whitelisted projects (one per line)
--review-team=GROUP     review team group added to requests with > 8 diff
--repo-checker=USER     repo checker user added after accepted review
```

Note that `--ignore-devel` used to be provided as `env[IGNORE_DEVEL_PROJECTS]`.

Some highlights about what was cleaned up:
- cryptic variable names replaced
- replaced custom osc queries with osc.core calls where applicable
- removed the need to load package information for all packages within target project which cuts runtime in half for single review
- removed extraneous code that performed extra steps for no reason
  ```python
  os.path.dirname(os.path.realpath(__file__.replace('.pyc', '.py')))
  os.path.dirname(os.path.realpath(__file__))
  ```
  and `_checker_prepare_dir()` which needlessly `chdir()` twice.
- one logic branch failed to cleanup checkout directory
- new flags provide additional flexibility

The only items I am not certain about are:
- `# Check in the srcmd5 is actually in Factory`
  This would seem to be unnecessary if the bot is reviewing factory since it should be assumed the request is not in factory? I assumed the bot was used for openSUSE point release development which has been replaced by `leaper.py` and as such this would be unneeded. If this is not the case please explain.
- fake_devel_prj submission
  ```python
  act = FakeAction(prj, pkg, tprj, tpkg, rev)
  self.api.submit_to_prj(act, fake_devel_prj, force_enable_build=True)
  ```
  Not sure what this is about.
- skip review functionality, is this needed?

I did not have a `role_add` request to test on, but I tried on `delete` and `submit`. With the new options I would expect this to work for SLE pending #614. Also note this branch includes #622. Once accepted I'll rebase to be sure the merge is clean.


jberry-suse:
I moved the defaults from option definitions into class constructor to be consistent with other ReviewBots.

jberry-suse:
Replaced `os.mkdir(dir)` with `os.makedirs(dir)` to avoid needlessly having to create `co` dir.

lnussel:
The check whether sources are in Factory is an optimization here. So the bot doesn't download the sources again. The purpose of this check in leaper.py is to avoid double work for the review team. However, since this bot may also add the review team the interaction or code sharing with leaper.py would be a topic to investigate indeed.

FakeAction is probably superfluous nowadays since we have adi stagings.

The skip feature is needed to make the bot add review-team and repochecker even if the source check is skipped.

jberry-suse:
Rebased which reduced diff as expected.

On the topic of checking source in factory then, given the task of making `ReviewBot` based code like `leaper.py` work against IBS and that it performs more checks my assumption was that `leaper.py` would be used. If that is the case then perhaps a flag to disable adding review teams should be added which can be used when running against SLE projects so that only the source checks are run while `leaper.py` can be responsible for a more complex analysis and adding review teams. For factory the script will work as it does now.

The expected bots:
- openSUSE:Factory: `check_source.py`
- SUSE: `check_source.py --skip-review-team`, `leaper.py`
- openSUSE:Leap: `leaper.py`

The assumptions being that other than forked changes major development is done in Factory and not Leap while SLE contains both, but in Factory where possible. Is that accurate?

I'll work on adding the skip feature then. Out of curiosity, is the skip feature used manually then?

jberry-suse:
To clarify, it seems prudent to avoid the same set of checks in the two bots if possible. Doing so as an optimization I imagine made sense when the bot was used on point releases in addition to Factory or is there another case in which the bot could see changes already accepted in Factory? If used for SLE obviously this could occur, but it may make sense to simply check if the changes are incoming from Factory as source or is it common to have open requests not from Factory, but containing sources found in Factory? I suppose SLE has same sort of requests from update project like leaper sees? It seems like some thought on the workflow and how the bots are to be utilized could avoid the duplication of code and checks. For instance one bot could add the other bot as a reviewer once it has completed checks. Perhaps `leaper.py` adds `check_source.py` for SLE which avoids the duplication of review team additions and Factory source checks.

lnussel:
Jimmy Berry wrote:
> Rebased which reduced diff as expected.
>
> On the topic of checking source in factory then, given the task of
> making `ReviewBot` based code like `leaper.py` work against IBS and
> that it performs more checks my assumption was that `leaper.py` would
> be used. If that is the case then perhaps a flag to disable adding
> review teams should be added which can be used when running against
> SLE projects so that only the source checks are run while `leaper.py`
> can be responsible for a more complex analysis and adding review
> teams. For factory the script will work as it does now.

Sounds good.

> The expected bots:
> - openSUSE:Factory: `check_source.py`
> - SUSE: `check_source.py --skip-review-team`, `leaper.py`
> - openSUSE:Leap: `leaper.py`
>
> The assumptions being that other than forked changes major development
> is done in Factory and not Leap while SLE contains both, but in
> Factory where possible. Is that accurate?

As soon as check_source is in a sane state I'd like to also enable it
for Leap. The vast majority of packages in Leap of couse is either from
SLE or Factory, a small subset is forks though. For those check_source
does make sense.

> I'll work on adding the skip feature then. Out of curiosity, is
> the skip feature used manually then?

Yes. If you have a good idea to make that more easy to use feel free to.

lnussel:
Jimmy Berry wrote:
> To clarify, it seems prudent to avoid the same set of checks in the
> two bots if possible. Doing so as an optimization I imagine made sense
> when the bot was used on point releases in addition to Factory or is
> there another case in which the bot could see changes already accepted
> in Factory?

maintenace may also want to run check_source.

> If used for SLE obviously this could occur, but it may
> make sense to simply check if the changes are incoming from Factory as
> source or is it common to have open requests not from Factory, but
> containing sources found in Factory?

Yes, for example if the same sources are submitted from the devel
project to Factory and a release in parallel.

> I suppose SLE has same sort of requests from update project like
> leaper sees?

No. SLE uses project links instead.

> It seems like some thought on the workflow and how
> the bots are to be utilized could avoid the duplication of code
> and checks. For instance one bot could add the other bot as a
> reviewer once it has completed checks. Perhaps `leaper.py` adds
> `check_source.py` for SLE which avoids the duplication of review
> team additions and Factory source checks.

Sounds good too.

jberry-suse:
As discussed:
- added `--skip-add-reviews` to `check_source.py`
- added `--check-source-group` to `leaper.py` which is used when `FORK` is detected

```diff
diff --git a/check_source.py b/check_source.py
index 7e42fb0..c17a6f3 100755
--- a/check_source.py
+++ b/check_source.py
@@ -46,6 +46,7 @@ class CheckSource(ReviewBot.ReviewBot):
         self.devel_whitelist = None
         self.review_team = 'opensuse-review-team'
         self.repo_checker = 'factory-repo-checker'
+        self.skip_add_reviews = False
 
     def check_one_request(self, request):
         # Copy original values to revert changes made to them.
@@ -139,12 +140,13 @@ class CheckSource(ReviewBot.ReviewBot):
         if len(checked):
             self.review_messages['accepted'] += "\n\nOutput of check script (non-fatal):\n" + output
 
-        if diff > 8:
-            if self.review_team != '':
-                self.add_review(self.request, by_group=self.review_team, msg='Please review sources')
+        if not self.skip_add_reviews:
+            if diff > 8:
+                if self.review_team is not None:
+                    self.add_review(self.request, by_group=self.review_team, msg='Please review sources')
 
-        if self.repo_checker != '':
-            self.add_review(self.request, by_user=self.repo_checker, msg='Please review build success')
+            if self.repo_checker is not None:
+                self.add_review(self.request, by_user=self.repo_checker, msg='Please review build success')
 
         return True
 
@@ -240,6 +242,7 @@ class CommandLineInterface(ReviewBot.CommandLineInterface):
         parser.add_option('--devel-whitelist', dest='devel_whitelist_file', metavar='FILE', help='file containing whitelisted projects (one per line)')
         parser.add_option('--review-team', dest='review_team', metavar='GROUP', help='review team group added to requests with > 8 diff')
         parser.add_option('--repo-checker', dest='repo_checker', metavar='USER', help='repo checker user added after accepted review')
+        parser.add_option('--skip-add-reviews', dest='skip_add_reviews', action='store_true', default=False, help='skip adding review after completing checks')
 
         return parser
 
@@ -254,6 +257,7 @@ class CommandLineInterface(ReviewBot.CommandLineInterface):
             bot.review_team = self.options.review_team
         if self.options.repo_checker:
             bot.repo_checker = self.options.repo_checker
+        bot.skip_add_reviews = self.options.skip_add_reviews
 
         return bot
 
diff --git a/leaper.py b/leaper.py
index 773f3f8..f81e0a8 100755
--- a/leaper.py
+++ b/leaper.py
@@ -76,6 +76,8 @@ class Leaper(ReviewBot.ReviewBot):
         self.release_manager_group = 'leap-reviewers'
         self.must_approve_version_updates = False
         self.must_approve_maintenance_updates = False
+        self.needs_check_source = False
+        self.check_source_group = None
 
         self.comment_marker_re = re.compile(r'<!-- leaper state=(?P<state>done|seen)(?: result=(?P<result>accepted|declined))? -->')
 
@@ -156,6 +158,7 @@ class Leaper(ReviewBot.ReviewBot):
                 is_fine_if_factory = True
                 not_in_factory_okish = True
                 self.needs_release_manager = True
+                self.needs_check_source = True
                 # fall through to check history and requests
             elif origin.startswith('openSUSE:Leap:42.2'):
                 if self.must_approve_maintenance_updates:
@@ -366,6 +369,19 @@ class Leaper(ReviewBot.ReviewBot):
                     self.review_messages['declined'] += '\nadding opensuse-review-team failed'
                     return False
 
+        if self.needs_check_source and self.check_source_group is not None:
+            add_review = True
+            self.logger.info("%s needs review by %s" % (req.reqid, self.check_source_group))
+            for r in req.reviews:
+                if r.by_group == self.check_source_group:
+                    add_review = False
+                    self.logger.debug("%s already is a reviewer", self.check_source_group)
+                    break
+            if add_review:
+                if self.add_review(req, by_group = self.check_source_group) != True:
+                    self.review_messages['declined'] += '\nadding %s failed' % self.check_source_group
+                    return False
+
         return request_ok
 
     def check_action__default(self, req, a):
@@ -424,6 +440,7 @@ class CommandLineInterface(ReviewBot.CommandLineInterface):
         parser.add_option("--no-comment", dest='comment', action="store_false", default=True, help="don't actually post comments to obs")
         parser.add_option("--manual-version-updates", action="store_true", help="release manager must approve version updates")
         parser.add_option("--manual-maintenance-updates", action="store_true", help="release manager must approve maintenance updates")
+        parser.add_option("--check-source-group", dest="check_source_group", metavar="GROUP", help="group used by check_source.py bot which will be added as a reviewer should leaper checks pass")
 
         return parser
 
@@ -434,6 +451,8 @@ class CommandLineInterface(ReviewBot.CommandLineInterface):
             bot.must_approve_version_updates = True
         if self.options.manual_maintenance_updates:
             bot.must_approve_maintenance_updates = True
+        if self.options.check_source_group:
+            bot.check_source_group = self.options.check_source_group
         bot.do_comments = self.options.comment
 
         return bot
```

jberry-suse:
`leaper.py` can then be run with `--check-source-group factory-auto` to add the `check_source.py` bot with the current configuration.

lnussel:
@nilxam could you take care of deploying and monitoring this to make sure it still does what it should for Factory? :)

nilxam:
besides that try/except for new package submission, I'd prefer logger also print message on screen not just committing review message to the request.

ReviewBot.py
```diff
@@ -286,6 +287,14 @@ def _get_linktarget(self, src_project, src_package):
 
         return (None, None)
 
+    def get_devel_project(self, project, package):
+        m = osc.core.show_package_meta(self.apiurl, project, package)
```
nilxam: I meant,

```
try:
    m = osc.core.show_package_meta(self.apiurl, project, package)
except urllib2.HTTPError:
    return None, None
```

nilxam:
for reference, old check_source will print the message to the screen like below,

```
SUBMIT(449476): Java:Factory/java-1_9_0-openjdk -> openSUSE:Factory/java-1_9_0-openjdk                                                                 │··························································
('NO DEVEL IN', 'rubygem-gli')                                                                                                                         │··························································
('NO DEVEL IN', 'rubygem-rice')                                                                                                                        │··························································
declined Output of check script:                                                                                                                       │··························································
Attention, 1015432.patch is not mentioned in spec files as source or patch.                                                                            │··························································
  Attention, s390-java-opts.patch is not mentioned in spec files as source or patch.                                                                   │··························································
                                                                                                                                                       │··························································
                                                                                                                                                       │··························································
SUBMIT(449478): network/dnsmasq -> openSUSE:Factory/dnsmasq                                                                                            │··························································
accepted Check script succeeded
```

ReviewBot.py
jberry-suse: Fixed:

```diff
diff --git a/ReviewBot.py b/ReviewBot.py
index c48da37..9a7a29f 100644
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -288,12 +288,14 @@ class ReviewBot(object):
         return (None, None)
 
     def get_devel_project(self, project, package):
-        m = osc.core.show_package_meta(self.apiurl, project, package)
-        node = ET.fromstring(''.join(m)).find('devel')
-        if node is None:
-            return None, None
-        else:
-            return node.get('project'), node.get('package', None)
+        try:
+            m = osc.core.show_package_meta(self.apiurl, project, package)
+            node = ET.fromstring(''.join(m)).find('devel')
+            if node is not None:
+                return node.get('project'), node.get('package', None)
+        except urllib2.HTTPError:
+            pass
+        return None, None
 
     def can_accept_review(self, request_id):
         """return True if there is a new review for the specified reviewer"""
```

jberry-suse:
The lack of output is due all information going through the logger. (already provided by ReviewBot). I always ran with `--debug` and many times with `--osc-debug`, but understand it does not seem completely normal to have to do so.

Note the difference:
```bash
./check_source.py --debug --dry --group factory-auto --review-team=None id 447944
```
```
INFO:check_source.py:checking 447944
INFO:check_source.py:X11:Wayland/wlc@34 -> openSUSE:Factory/wlc
INFO:check_source.py:447944 declined: Output of check script:
No changelog. Please use 'osc vc' to update the changes file(s).
DEBUG:check_source.py:447944 review not changed
```

Which seems like everything you want, versus:

```bash
./check_source.py --dry --group factory-auto --review-team=None id 447944
```

Which prints nothing. I am not sure if it makes sense to change the default logger level to include `INFO` (if only for `check_source.py`) as that would seem to suffice for what you want. Otherwise, the output is already there which is why I did not duplicate it.

From `ReviewBot.py`:

```python
        level = None
        if (self.options.debug):
            level = logging.DEBUG
        elif (self.options.verbose):
            level = logging.INFO
```

So using --verbose flag should give you only what you want. Thoughts?

lnussel:
the behavior is consistent with other bots. one needs to run with --verbose or --debug to get something on the console. So this is fine.

nilxam:
Tested on packagelists.

PR #625, opened 2016-12-30
Provide cache for expensive and cache-able staging requests.

jberry-suse:
The two slowest staging API calls are for information that rarely changes. By caching the result the commands typically execute over twice as fast.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -35,6 +35,7 @@ from osclib.list_command import ListCommand
 from osclib.obslock import OBSLock
 from osclib.select_command import SelectCommand
 from osclib.stagingapi import StagingAPI
+from osclib.stagingapi import HTTPCache
 from osclib.unselect_command import UnselectCommand
 from osclib.repair_command import RepairCommand
 
@@ -89,6 +90,8 @@ def _full_project_name(self, project):
               help='do not cleanup remaining packages in staging projects after accept')
 @cmdln.option('--no-bootstrap', dest='bootstrap', action='store_false', default=True,
               help='do not update bootstrap-copy when freezing')
+@cmdln.option('--wipe-cache', dest='wipe_cache', action='store_true', default=False,
+              help='wipe GET request cache before executing')
 def do_staging(self, subcmd, opts, *args):
     """${cmd_name}: Commands to work with staging projects
 
@@ -164,6 +167,9 @@ def do_staging(self, subcmd, opts, *args):
     opts.apiurl = self.get_api_url()
     opts.verbose = False
     Config(opts.project)
+
+    if opts.wipe_cache:
+        HTTPCache.wipe()
 
     with OBSLock(opts.apiurl, opts.project):
         api = StagingAPI(opts.apiurl, opts.project)
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -16,6 +16,7 @@
 
 import json
 import logging
+import os
 import urllib2
 import time
 import re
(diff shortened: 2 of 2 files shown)


jberry-suse:
For example:

```bash
$ osc --debug staging -p openSUSE:Leap:42.3 list
```
```
GET https://api.opensuse.org/source/openSUSE:Leap:42.3:Staging/_attribute/openSUSE:LockedBy
POST https://api.opensuse.org/source/openSUSE:Leap:42.3:Staging/_attribute/openSUSE:LockedBy
GET https://api.opensuse.org/source/openSUSE:Leap:42.3:Staging/_attribute/openSUSE:LockedBy
GET https://api.opensuse.org/search/request?match=state/@name='review'+and+review[@by_group='factory-staging'+and+@state='new']+and+(target[@project='openSUSE:Leap:42.3']+or+target[@project='openSUSE:Leap:42.3:NonFree'])
GET [CACHED] https://api.opensuse.org/source/openSUSE:Leap:42.3:Rings:0-Bootstrap?nofilename=1&view=info
GET [CACHED] https://api.opensuse.org/source/openSUSE:Leap:42.3:Rings:1-MinimalX?nofilename=1&view=info
GET [CACHED] https://api.opensuse.org/source/openSUSE:Leap:42.3:Rings:2-TestDVD?nofilename=1&view=info
GET [CACHED] https://api.opensuse.org/search/package?match=[@project='openSUSE:Factory']
YaST:Head
  sr#447422: yast2-ntp-client               -> 1-MinimalX   (Factory)
devel:languages:lua
  sr#447780: lua                            -> 1-MinimalX   (Factory)
devel:libraries:c_c++
  sr#447162: libunwind                      -> 0-Bootstrap  (~matwey:branches:openSUSE:Leap:42.3)
GET https://api.opensuse.org/source/openSUSE:Leap:42.3:Staging/_attribute/openSUSE:LockedBy
POST https://api.opensuse.org/source/openSUSE:Leap:42.3:Staging/_attribute/openSUSE:LockedBy
```

The only query to run (aside from lock check and aquire) is request search which is desirable.

jberry-suse:
Once the cache framework is accepted I've been playing around with caching strategies for other very slow commands like `select`.

jberry-suse:
To be clear this speeds up `adi` and `select` as well, but I would imagine there are more gains to be had in those commands which will require a careful inspection.

jberry-suse:
I implemented this in a way that presumably could apply to `osc.core.http_request()` and if desired I think it would make sense to submit there.

jberry-suse:
The retried query `https://api.opensuse.org/search/package?match=[@project='openSUSE:Factory']` can timeout and waste quite a lot of time. From testing locally it definitely speeds up the average a lot given the number of times that query will hang.

With a bit more fiddling I've managed to cache and expire _meta requests which drastically speeds up `adi` and `select/unselect` commands. I'll keep playing with it to make sure it works well.

jberry-suse:
Updated this to remove one left over line.

lnussel:
isn't this what we have memoize for?

jberry-suse:
@lnussel memoize is only kept in memory for the current process (unless I am mistaken) which is not terribly useful since the initial requests I targeted are only issued once per process. A disk cache then does not miss every-time like memoize and makes running these commands locally significantly faster.

I am currently utilizing local changes that cache a number of other requests, but until I get around to implementing a collaboration friendly way of expiring them (I have details worked out) it does not make sense to include those additional caches.

jberry-suse:
Interesting, it seems that it does save to cache, I must have confused that with the similarly named pattern. I'll have to investigate using it instead.

lnussel:
obsoleted by #629 IIUC

PR #627, opened 2017-01-05
Add compare_pkglist.py

nilxam:
Comparing package source between two project. https://progress.opensuse.org/issues/14938

### DIFF
--- a/compare_pkglist.py
+++ b/compare_pkglist.py
@@ -1,0 +1,126 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2017 SUSE Linux GmbH
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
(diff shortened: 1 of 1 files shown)


compare_pkglist.py
```diff
@@ -0,0 +1,129 @@
...
+http_GET = osc.core.http_GET
+http_POST = osc.core.http_POST
+http_PUT = osc.core.http_PUT
+
+class CompareList(object):
+    def __init__(self, old_prj, new_prj, verbose):
+        self.new_prj = new_prj
+        self.old_prj = old_prj
+        self.verbose = verbose
+        self.factory = 'openSUSE:Factory'
```
jberry-suse: not: Not used.

compare_pkglist.py
```diff
@@ -0,0 +1,129 @@
...
+            # check if it has diff element
+            diffs = root.findall('files/file/diff')
+            if diffs:
+                return ET.tostring(root)
+        return False
+
+    def crawl(self):
+        """Main method"""
+        # get souce packages from target
+        print 'Gathering the packages list from %s' % self.old_prj
```
jberry-suse: nit: s/packages/package/ (and below)

compare_pkglist.py
```diff
@@ -0,0 +1,129 @@
...
+        old_packages = self.get_source_packages(self.old_prj)
+        print 'Gathering the packages list from %s' % self.new_prj
+        new_packages = self.get_source_packages(self.new_prj)
+
+        for pkg in old_packages:
+            have_diff = False
```
jberry-suse: Any reason to define this as false since it is redefined before use?

compare_pkglist.py
```diff
@@ -0,0 +1,129 @@
...
+            if pkg not in new_packages:
+                logging.debug('%s is not in %s' % (pkg, self.new_prj))
+            else:
+                have_diff = self.check_diff(pkg, self.old_prj, self.new_prj)
+                if have_diff is not False:
+                    print '%s/%s has different source than %s' % (self.new_prj, pkg, self.old_prj)
+                    if self.verbose:
+                        print have_diff
```
jberry-suse: It was not clear when reading this that `have_diff` actually contained the diff or `False`. It seemed it would print out `True/False` until digging deeper. Perhaps just `diff` as variable name?

compare_pkglist.py
```diff
@@ -0,0 +1,129 @@
...
+
+from osc import oscerr
+
+OPENSUSE = 'openSUSE:Leap:42.2'
+SLE = 'SUSE:SLE-12-SP2:GA'
+
+makeurl = osc.core.makeurl
+http_GET = osc.core.http_GET
+http_POST = osc.core.http_POST
+http_PUT = osc.core.http_PUT
```
jberry-suse: nit: Not used.

compare_pkglist.py
```diff
@@ -0,0 +1,129 @@
...
+    uc.crawl()
+
+if __name__ == '__main__':
+    description = 'Compare packages status between two project'
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument('-A', '--apiurl', metavar='URL', help='API URL')
+    parser.add_argument('-d', '--debug', action='store_true',
+                        help='print info useful for debuging')
+    parser.add_argument('-f', '--old', dest='old_prj', metavar='PROJECT',
+                        help='project where to check (default: %s)' % SLE,
```
jberry-suse: not: Perhaps include old/new in descriptions for these as well. Is `-f` and `-t` relating to from/to, otherwise perhaps just `-f/-o`?

PR #628, opened 2017-01-06
Provide a utility to print the list of devel projects for a given project.

jberry-suse:
Simotek in #opensuse-factory was looking for a list of devel projects for Factory. I figured this could be handy and the code already exists in `osc-check_source.py`. Perhaps this is not the right place, but if so here it is.

### DIFF
--- a/devel-project-list.py
+++ b/devel-project-list.py
@@ -1,0 +1,57 @@
+#!/usr/bin/python
+
+import argparse
+import sys
+from xml.etree import cElementTree as ET
+
+import osc.conf
+import osc.core
+from osclib.conf import Config
+from osclib.stagingapi import StagingAPI
+
+
+def devel_projects_get(apiurl, project):
+    """
+    Returns a sorted list of devel projects for a given project.
+
+    Loads all packages for a given project, checks them for a devel link and
+    keeps a list of unique devel projects.
+    """
+    devel_projects = {}
+
+    url = osc.core.makeurl(apiurl, ['search', 'package'], "match=[@project='%s']" % project)
+    root = ET.parse(osc.core.http_GET(url)).getroot()
+    for package in root.findall('package'):
+        devel = package.find('devel')
+        if devel is not None:
+            devel_projects[devel.attrib['project']] = True
+
+    return sorted(devel_projects)
+
+def main(args):
+    osc.conf.get_config(override_apiurl=args.apiurl)
+    osc.conf.config['debug'] = args.debug
+
+    devel_projects = devel_projects_get(osc.conf.config['apiurl'], args.project)
+    if len(devel_projects) == 0:
+        print('no devel projects found')
+    else:
+        out = '\n'.join(devel_projects)
+        print(out)
+
+        if args.write:
+            Config(args.project)
+            api = StagingAPI(osc.conf.config['apiurl'], args.project)
+            api.save_file_content('%s:Staging' % api.project, 'dashboard', 'devel_projects', out)
+
+
+if __name__ == '__main__':
+    description = 'Print out devel projects for a given project (like openSUSE:Factory).'
+    parser = argparse.ArgumentParser(description=description)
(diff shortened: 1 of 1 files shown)


lnussel:
could we maybe run this regularly and upload the result e.g. to the dashboard container? That would allow other scripts like "staging list" to fetch the information. The call sometimes takes 10 minutes and is a constant source of annoyance for me :-)

jberry-suse:
Yeah, this is also one of the calls I cache in #625 which alleviates 99% of time cost. I figured it might be handy to dump somewhere public since it is a requirement of sorts to get a package accepted into a devel project before factory, but I am not aware of an up-to-date list.

jberry-suse:
I went ahead and added a `--write` option to do just that. Assuming this is the desired location and implementation then I'd suggest accepting this, enable automatic running, ensure working properly, and then swap out existing calls to look at this file.

https://build.opensuse.org/package/view_file/openSUSE:Factory:Staging/dashboard/devel_projects?expand=1

```diff
diff --git a/devel-project-list.py b/devel-project-list.py
index 206fb63..cc5c288 100755
--- a/devel-project-list.py
+++ b/devel-project-list.py
@@ -6,6 +6,8 @@ from xml.etree import cElementTree as ET
 
 import osc.conf
 import osc.core
+from osclib.conf import Config
+from osclib.stagingapi import StagingAPI
 
 
 def devel_projects_get(apiurl, project):
@@ -29,12 +31,18 @@ def devel_projects_get(apiurl, project):
 def main(args):
     osc.conf.get_config(override_apiurl=args.apiurl)
     osc.conf.config['debug'] = args.debug
+    Config(args.project)
 
     devel_projects = devel_projects_get(osc.conf.config['apiurl'], args.project)
     if len(devel_projects) == 0:
         print('no devel projects found')
     else:
-        print('\n'.join(devel_projects))
+        out = '\n'.join(devel_projects)
+        print(out)
+
+        if args.write:
+            api = StagingAPI(osc.conf.config['apiurl'], args.project)
+            api.save_file_content('%s:Staging' % api.project, 'dashboard', 'devel_projects', out)
 
 
 if __name__ == '__main__':
@@ -43,6 +51,7 @@ if __name__ == '__main__':
     parser.add_argument('-A', '--apiurl', metavar='URL', help='API URL')
     parser.add_argument('-d', '--debug', action='store_true', help='print info useful for debuging')
     parser.add_argument('-p', '--project', default='openSUSE:Factory', metavar='PROJECT', help='project for which to list devel projects')
+    parser.add_argument('-w', '--write', action='store_true', help='write to dashboard container package')
     args = parser.parse_args()
 
     sys.exit(main(args))
diff --git a/osc-staging.py b/osc-staging.py
index 2ece745..460898e 100644
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -165,6 +165,7 @@ def do_staging(self, subcmd, opts, *args):
     opts.verbose = False
     Config(opts.project)
 
+    print(opts.project)
     with OBSLock(opts.apiurl, opts.project):
         api = StagingAPI(opts.apiurl, opts.project)
 
```

Could also make this part of osc-staging, but not sure that is desired.

jberry-suse:
Presumably this list can be linked to from a relevant wiki page so folks know what projects to target for getting into Factory.

jberry-suse:
What needs to be done to deploy this, presumably alongside some of running bots.

lnussel:
@nilxam or @DimStar77 could you add this to packagelists tmux maybe? For that it would be helpful to rebase on ToolBase.py so we get the common runner function and avoid a shell loop.

jberry-suse:
The one thing that will probably need to be done for this to be more useful is adding other projects to the cronjob. All the main distro projects (like Leap:42.3) and such are also queried and take forever. Additional, I would feel more comfortable increasing the ttl on those caches to a week or more if we had a way to detect package deletions.

nilxam:
well not every main distro projects has the devel project set, like 42.3 :)

PR #629, opened 2017-01-07
Provide cache for expensive and cache-able staging requests.

jberry-suse:
The two slowest staging API calls are for information that rarely changes. By caching the result the commands typically execute over twice as fast. Going further can see improvements of an order of magnitude or more by caching almost all the GET requests.

In contrast to osclib/memoize.py this cache operates at the HTTP request level. This has several advantages:

- Caches the expensive part (ie the HTTP request). There are a number of functions in osc.core and elsewhere that make the same API request, but process the result differently which would require multiple API calls using memoize.
- Handles cases were a loader function uses class attributes as input and output and thus no relevant method parameters or return. An important example is StagingAPI._generate_ring_packages().
- Storage is project aware which allows caches to be deleted when a project is known to have changed.
- Due to project awareness, can utilize OBS /statistics/latest_updated API call to determine which projects need to be expired.

The cache file structure is as follows:

- hostname(apiurl)
  - project
    - sha1(url)
  - sha1(url)

See Cache.PATTERNS for changing the time to live (ttl) or add patterns to be cached.

### DIFF
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -1,0 +1,273 @@
+from __future__ import print_function
+
+import datetime
+import hashlib
+import os
+import osc.core
+import re
+import shutil
+import sys
+import urlparse
+from StringIO import StringIO
+from osc import conf
+from osc.core import urlopen
+from time import time
+
+try:
+    from xml.etree import cElementTree as ET
+except ImportError:
+    import cElementTree as ET
+
+
+def http_request(method, url, headers={}, data=None, file=None):
+    """
+    Wrapper for osc.core.http_request() to provide GET request caching.
+    """
+
+    if method == 'GET':
+        ret = Cache.get(url)
+        if ret:
+            return ret
+    else:
+        # Logically, seems to make more sense after real call, but practically
+        # it should not matter and makes the apitests happy when dealing with
+        # request acceptance which causes a GET to determine target project.
+        Cache.delete(url)
+
+    ret = osc.core._http_request(method, url, headers, data, file)
+
+    if method == 'GET':
+        ret = Cache.put(url, ret)
+
+    return ret
+
+
+class Cache(object):
+    """
+    Provide a cache implementation for osc.core.http_request().
+
+    The cache takes a list of regular expression patterns and time to live (ttl)
+    for API paths. In addition to the ttl the project context is taken into
+    account when available in order to expire all caches related to a project
+    when the remote server indicates a change was made more recently than the
+    local cache reflects. This provides a fairly robust cache that can handle
+    multiple users changing the same projects.
+
+    Cannot safely cache, for lengthy periods, paths that can update without user
+    interaction or that do not trigger the project updated timestamp to change.
+    Such paths include anything related to build status. When a source package
+    is updated the linked packages do not trigger an update of their project. As
+    such sources cannot be reliably cached for too long.
+
+    Any paths without a project context will be cleared when updated using this
+    cache, but obviously not for other contributors.
+    """
+
+    CACHE_DIR = os.path.expanduser('~/.cache/osc-plugin-factory')
+    TTL_LONG = 12 * 60 * 60
+    TTL_SHORT = 5 * 60
+    TTL_DUPLICATE = 3
+    PATTERNS = {
+        # Group members cannot be guaranteed, but change rarely.
+        '/group/[^/?]+$': TTL_SHORT,
+        # Clear target project cache upon request acceptance.
+        '/request/(\d+)\?.*newstate=accepted': TTL_DUPLICATE,
+        "/search/package\?match=\[@project='([^']+)'\]$": TTL_LONG,
+        # Potentially expire the latest_updated since it will be the only way to
+        # tell after an adi staging is removed. For now just cache the calls
+        # that occur in rapid succession.
+        "/search/project/id\?match=starts-with\(@name,'([^']+)\:'\)$": TTL_DUPLICATE,
+        # List of all projects may change, but relevant ones rarely.
+        '/source$': TTL_LONG,
+        # Sources will be expired with project, could be done on package level.
+        '/source/([^/?]+)(?:\?.*)?$': TTL_LONG,
+        # Project will be marked changed when packages are added/removed.
+        '/source/([^/]+)/_meta$': TTL_LONG,
+        '/source/([^/]+)/(?:[^/]+)/(?:_meta|_link)$': TTL_LONG,
+        # Handles clearing local cache on package deletes. Lots of queries like
+        # updating project info, comment, and package additions.
+        '/source/([^/]+)/(?:[^/?]+)(?:\?[^/]+)?$': TTL_LONG,
+        # Presumably users are not interweaving in short windows.
+        '/statistics/latest_updated': TTL_SHORT,
+    }
+
+    last_updated = {}
+
+    @staticmethod
+    def init():
+        Cache.patterns = []
+        for pattern in Cache.PATTERNS:
+            Cache.patterns.append(re.compile(pattern))
+
+        # Replace http_request with wrapper function which needs a stored
+        # version of the original function to call.
+        if not hasattr(osc.core, '_http_request'):
+            osc.core._http_request = osc.core.http_request
+            osc.core.http_request = http_request
+
+    @staticmethod
+    def get(url):
+        match, project = Cache.match(url)
+        if match:
+            path = Cache.path(url, project, include_file=True)
+            ttl = Cache.PATTERNS[match]
+
+            if project:
+                # Given project context check to see if project has been updated
+                # remotely more recently than local cache.
+                apiurl, _ = Cache.spliturl(url)
+                Cache.last_updated_load(apiurl)
+
+                # Use the project last updated timestamp if availabe, otherwise
+                # the oldest record indicates the longest period that can be
+                # guaranteed to have no changes.
+                if project in Cache.last_updated[apiurl]:
+                    unchanged_since = Cache.last_updated[apiurl][project]
+                else:
+                    unchanged_since = Cache.last_updated[apiurl]['__oldest']
+
+                now = datetime.datetime.utcnow()
+                unchanged_since = datetime.datetime.strptime(unchanged_since, '%Y-%m-%dT%H:%M:%SZ')
+                history_span = now - unchanged_since
+
+                # Treat non-existant cache as brand new for the sake of history
+                # span check since it behaves as desired.
+                age = 0
+                directory = Cache.path(url, project)
+                if os.path.exists(directory):
+                    age = time() - os.path.getmtime(directory)
+
+                # If history span is shorter than allowed cache life and the age
+                # of the current cache is older than history span with no
+                # changes the cache cannot be guaranteed. For example:
+                #   ttl = 1 day
+                #   history_span = 0.5 day
+                #   age = 0.75
+                # Cannot be guaranteed.
+                ttl_delta = datetime.timedelta(seconds=ttl)
+                age_delta = datetime.timedelta(seconds=age)
+                if history_span < ttl_delta and age_delta > history_span:
+                    Cache.delete_project(apiurl, project)
+
+            if os.path.exists(path) and time() - os.path.getmtime(path) <= ttl:
+                if conf.config['debug']: print('CACHE_GET', url, file=sys.stderr)
+                return urlopen('file://' + path)
+            else:
+                reason = '(' + ('expired' if os.path.exists(path) else 'does not exist') + ')'
+                if conf.config['debug']: print('CACHE_MISS', url, reason, file=sys.stderr)
+
+        return None
+
+    @staticmethod
+    def put(url, data):
+        match, project = Cache.match(url)
+        if match:
+            path = Cache.path(url, project, include_file=True, makedirs=True)
+
+            # Since urlopen does not return a seekable stream it cannot be reset
+            # after writing to cache. As such a wrapper must be used. This could
+            # be replaced with urlopen('file://...') to be consistent, but until
+            # the need arrises StringIO has less overhead.
+            text = data.read()
+            data = StringIO(text)
+
+            if conf.config['debug']: print('CACHE_PUT', url, project, file=sys.stderr)
+            f = open(path,'w')
+            f.write(text)
+            f.close()
+
+        return data
+
+    @staticmethod
+    def delete(url):
+        match, project = Cache.match(url)
+        if match:
+            path = Cache.path(url, project, include_file=True)
+
+            # Rather then wait for last updated statistics to expire, remove the
+            # project cache if applicable.
+            if project:
+                apiurl, _ = Cache.spliturl(url)
+                if project.isdigit():
+                    # Clear target project cache upon request acceptance.
+                    project = osc.core.get_request(apiurl, project).actions[0].tgt_project
+                Cache.delete_project(apiurl, project)
+
+            if os.path.exists(path):
+                if conf.config['debug']: print('CACHE_DELETE', url, file=sys.stderr)
+                os.remove(path)
+
+        # Also delete version without query. This does not handle other
+        # variations using different query strings. Handy for PUT with ?force=1.
+        o = urlparse.urlsplit(url)
+        if o.query != '':
+            url_plain = urlparse.SplitResult(o.scheme, o.netloc, o.path, '', o.fragment).geturl()
+            Cache.delete(url_plain)
+
+    @staticmethod
+    def delete_project(apiurl, project):
+        path = Cache.path(apiurl, project)
+
+        if os.path.exists(path):
+            if conf.config['debug']: print('CACHE_DELETE_PROJECT', apiurl, project, file=sys.stderr)
+            shutil.rmtree(path)
+
+    @staticmethod
+    def delete_all():
+        if os.path.exists(Cache.CACHE_DIR):
+            shutil.rmtree(Cache.CACHE_DIR)
+
+    @staticmethod
+    def match(url):
+        apiurl, path = Cache.spliturl(url)
+        for pattern in Cache.patterns:
+            match = pattern.match(path)
+            if match:
+                return (pattern.pattern,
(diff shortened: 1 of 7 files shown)


jberry-suse:
See #625 for the first version of this cache style. It suffered from not storing the caches with project context and thus would not gracefully handle other contributors updating projects in between local queries. In addition this approach no longer requires updating the calling code to include `cache` parameter, but instead uses a set of regular expression patterns.

Looks like I'll need to appease the tests, but feel free to play around and see that it works as I will be doing. This caches the same queries I was experimenting with locally on top of the previous pull request. I did not experience any issues while doing staging work, but alas there has not been any new requests since I finished this so time will tell.

jberry-suse:
If this ends up being solid it should be possible to apply it to ReviewBots if desired so for example when they query the full list of packages in Factory it does not need to be loaded again unless Factory has been updated. Or leaper loading the lookup lists and such. Obviously, if speed isn't a concern for the bots since <10 seconds to 1 min per review isn't as painful compared to running commands directly.

jberry-suse:
Removing packages from stagings will be the one tricky part since it will not show up as a change on last updated. Local caches are cleared, but others will not be. Suppose it could be solved by adding some sort of either a) shared cache on OBS, or b) a log of deletes or all changes (or at minimum last changed) which was discussed during call with lnussel.

coolo:
listing the sources of a project should be quick. If it's not, this should be fixed in OBS - and not worked around. OBS can apply much more efficient caching strategies.

jberry-suse:
The notes about listing sources are there to clarify why they are one of the queries not cached. They do not make up the majority nor the expensive queries.

jberry-suse:
I made some optimizations, handled a few more cases, and got the tests passing. The changes made for tests are as follows, may decide a different approach or to skip cache when testing. It seems good that the tests can pass with the cache as that means at least locally it is clearly properly.

```diff
diff --git a/tests/checktags_tests.py b/tests/checktags_tests.py
index 263a0c2..20d1915 100644
--- a/tests/checktags_tests.py
+++ b/tests/checktags_tests.py
@@ -26,7 +26,6 @@ import httpretty
 import osc
 import urlparse
 import sys
+from osclib.cache import Cache
 sys.path.append(".")
 
 from check_tags_in_requests import TagChecker
@@ -42,7 +41,6 @@ class TestTagChecker(unittest.TestCase):
         Initialize the configuration
         """
 
+        Cache.last_updated[APIURL] = {'__oldest': '2016-12-18T11:49:37Z'}
         httpretty.reset()
         httpretty.enable()
 
diff --git a/tests/factory_source_tests.py b/tests/factory_source_tests.py
index 29c8fa5..0ba25d7 100644
--- a/tests/factory_source_tests.py
+++ b/tests/factory_source_tests.py
@@ -21,7 +21,6 @@ import httpretty
 import osc
 import re
 import urlparse
+from osclib.cache import Cache
 
 from check_source_in_factory import FactorySourceChecker
 
@@ -40,7 +39,6 @@ class TestFactorySourceAccept(unittest.TestCase):
         Initialize the configuration
         """
 
+        Cache.last_updated[APIURL] = {'__oldest': '2016-12-18T11:49:37Z'}
         httpretty.reset()
         httpretty.enable()
 
diff --git a/tests/fixtures/statistics/latest_updated b/tests/fixtures/statistics/latest_updated
deleted file mode 100644
index 2b5210f..0000000
--- /dev/null
+++ a/tests/fixtures/statistics/latest_updated
@@ -1,3 +0,0 @@
+<latest_updated>
+  <package name="notreal" project="notreal" updated="2016-12-18T11:49:37Z"/>
+</latest_updated>
diff --git a/tests/obs.py b/tests/obs.py
index 49e765a..b4d3eab 100644
--- a/tests/obs.py
+++ b/tests/obs.py
@@ -23,7 +23,6 @@ import xml.etree.cElementTree as ET
 
 import httpretty
 import osc
+from osclib.cache import Cache
 
 
 APIURL = 'http://localhost'
@@ -108,7 +107,6 @@ class OBS(object):
         if not OBS._self:
             OBS._self = super(OBS, cls).__new__(cls, *args, **kwargs)
 
+        Cache.delete_all()
         httpretty.reset()
         httpretty.enable()
 
@@ -123,9 +121,6 @@ class OBS(object):
         """Instance constructor."""
         self.fixtures = fixtures
 
+        if not hasattr(Cache, '_CACHE_DIR'):
+            Cache._CACHE_DIR = True
+            Cache.CACHE_DIR += '-test'
         httpretty.enable()
 
         oscrc = os.path.join(fixtures, 'oscrc')
```

jberry-suse:
I have been using this during my Leap 42.3 work for a few days without issue. Obviously, package deletion without updating its project can be an issue, since it is not reflected by OBS API, if done by another person so local caches are not cleared. While it would be nice if OBS reflected that I do not imagine it is something likely to change soon if ever. If folks are not opposed keeping a record of last modified times to staging projects by staging tools, perhaps only for deletes since the rest is already tracked, to supplement this seems reasonable.

I posted some output from commands run in debug mode so one can see how much is cached and get an idea of just how much faster the commands run. There is an example of a `freeze` performed before a `select` where you can see the specific staging cache is automatically cleared due to the `freeze`.

https://gist.github.com/jberry-suse/66bc0fd418df3723e9f0e7c0f245eefe

lnussel:
so #625 can be closed?

osclib/cache.py
```diff
@@ -0,0 +1,275 @@
...
+    def init():
+        Cache.patterns = []
+        for pattern in Cache.PATTERNS:
+            Cache.patterns.append(re.compile(pattern))
+
+        # Replace http_request with wrapper function which needs a stored
+        # version of the original function to call.
+        if not hasattr(osc.core, '_http_request'):
+            osc.core._http_request = osc.core.http_request
+            osc.core.http_request = http_request
```
lnussel: you like monkey patching, do you? :)

lnussel:
several tools have a cached_GET function, how does this relate?

osclib/cache.py
jberry-suse: I left #625 as it provides the direct change approach. That is limiting in that methods in `osc.core` cannot be cached and requires a one line change to whatever calls are to be cached. Does the trick.

jberry-suse:
`cached_GET()` is a @memoize based url cache.

```python
    @memoize(add_invalidate=True)
    def _cached_GET(self, url):
        return self.retried_GET(url).read()

    def cached_GET(self, url):
        if self.caching:
            return self._cached_GET(url)
        return self.retried_GET(url).read()
```

As included in original description, `@memoize`/only-URL based approaches like #625 suffer from not being able to expire caches when a project is updated which is a requirement for robust caching of more calls. Just like #625 it also requires the calling to explicitly use it.

osclib/cache.py
jberry-suse: If `osc.core` is ignored (or this implemented there) could use a direct call approach and switch back to caller specifying `ttl` and potentially project as well directly. Beside the afforementioned down-sides this also will require duplicating `ttl` for the same call in different places (ie changing the `ttl` for a particular call may require changes to multiple locations).

lnussel:
let's try it. if it breaks, you have to fix it :)

jberry-suse:
haha, fair enough.

aplanas:
@jberry-suse the memoize implementation from https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/memoize.py can invalidate the internal cache.

As you can see in https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/memoize.py#L43 there is an `add_invalidate` parameter. If set, this will add two methods in the instance: `._invalidate_NAME_DECORATED_METHOD()` and `._invalidate_all()`

You can see some usage examples in https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/stagingapi.py#L558

aplanas:
@jberry-suse also note that memoize.py is process / thread safe via locks. This is convenient for package-list machine, where at the same time a process is doing the reviews, and the user can make other operations in the terminal. Not sure the if the cache is aware of this.

jberry-suse:
@aplanas Right, but then all the calling code would have to figure out all the caches related to a project anytime a call to update it is made and call all the relevant invalidate methods (obviously that isn't practical). Alternatively, `@memoize` could be updated to store things with the project context and invalidate when relevant in the same way that is implemented in this request.

This cache does not lock so that may be something to consider adding. Although, short of deleting a project cache from another process while in the midst of writing to a cache file I would expect there to be little that needs to be locked. Given the use-case exists I can take a look at adding it.

aplanas:
@jberry-suse where to call the invalidation is a matter of how you decide to encapsulate the logic. For example, if the object contains two method: one `get` and one `set`, is clear that the trigger for invalidation of the `get` cache is in the `set` method.

Also with _invalidate_all you can avoid this relation, that is most of the times simple and more correct.

IMHO the current status with multiple caches (if I am not wrong we have three in place), is not good, and the monkey patch of http_XXX makes the debugging more complex.

Also the different TTL for each request based on regex is extremely fragile and prone to bugs. IMHO the easy approach to achieve a similar effect is to abstract the operations that those URL and decorate the method with `@memoize(ttl=xx)` for each action, and not working and `http_request` level. I am sure that some of the operation that those partial URL represents are unchained inside stagingapi.py code, so why to cache the request instead the method that trigger it?

I think that this piece of code needs to be reoriented to a more high level, and abstract more from osc.core in a way that is convenient for you.

But I am super happy that the code is moving again. 👍

jberry-suse:
@aplanas An example of the problem is that adding a package to a project (ie a PUT on $project/$package/...) should invalidate that projects _meta and package list caches which are handled by totally separate methods. Those are also not the only caches to invalidate and may add more things to cache down the road that would need to be cleared.

Rather than

```python
# new package
put(...)

_invalidate_meta()
_invalidate_list()
# etc
```

or

```python
_invalidate_all()
```

which drops many caches that are unchanged. The ones that are desired to be dropped are just those corresponding to that project. I have no doubt that the memorize solution could be modified to handle this with all the same code that was in this PR, but it has the other issues mentioned in summary (like multiple methods that make same underlying call, all of which would have to be needlessly abstracted to provide a function with `@memorize` decorator that just calls the real function.

Anything in `osc.core` like the following has the same problem:

```python
def get_request(apiurl, reqid):
    u = makeurl(apiurl, ['request', reqid], {'withfullhistory': '1'})
    f = http_GET(u)
    root = ET.parse(f).getroot()

    r = Request()
    r.read(root)
    return r
```

would require something like

```python
@memorize(...)
def get_request(apiurl, reqid):
    return osc.core.get_request(apiurl, reqid)
```

So then we end up with a bunch of wrapper functions. It would work and we can do that, but that's what I was trying to avoid. Secondly, a number of similar queries, to same API resource, with different query parameters are made that should share the same `ttl` would have multiple wrapper function definitions and have to write the `ttl` multiple times. As with most things, never a perfect solution. The one I chose avoids the boilerplate wrapper functions and duplicating `ttl`s and instead focuses all caching information inside `cache.py`. If there was a connection between `makeurl()` and the actual request call that could avoid the regex parsing of the  URL that would be great, but short of some serious refactoring or assumptions I do not see that as possible. If one could assume the next `request()` was related to the last `makeurl()` call and that all `request()`s are made with a `makeurl()` preceding them then one could intercept the arguments to `makeurl()` and potentially avoid regex. This does not work for the projects contained within xpath queries which was another reason for not pursuing. So realistically, all `request()` calls would need to have some sort of cache metadata call just before them to avoid parsing or be modified to pass the information directly via additional parameter like #625, but with project context. Any solutions involving direct modification of the calls suffer from the need for wrappers of `osc.core` functions.

jberry-suse:
On the locking front, given that this solution is only used by `stagingapi.py` which already does a project level lock I would expect the possibility for problems to be minimal, but if desired it can investigated although I'd have to sit and think about any weird edgecases where two commands are updating the same project at the same time. Obviously, if we end up using this cache outside of `stagingapi` (for bots and such) locking may be more useful.

aplanas:
@jberry-suse I understand, but my point is that the place to put the cache is not in the `get_request`, but in the piece of code form stagingapi that do the request.

I can imagine that this regex from

https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/cache.py#L74

was triggered here

https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/checkrepo.py#L257

I think that the correct solution is to add the cache in the checkrepo, or where the real call for this URL was made, but not in the http_request level, that completely hide the semantic of the URL

Same here:

https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/cache.py#L81

If you want to get all the sources, cache that in the place where the call is made, or better extract it and add a method in staginapi to do the usual search and memoize it.

For example, I guess that some of the `source` regex can come from:

https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/stagingapi.py#L227

Why not to memoize `get_package_information` instead?

I understand that sometimes the get_request is followed with some processing, making it hard to cache. So do the right thing and refactor the code to separate the part that get the information from the part that process it and cache one of them.

I mean, looking the regex is very hard to see from where it comes, and this makes the maintainability of the code very hard.

jberry-suse:
@aplanas That regex would not be matched since that is for a specific request resource and not the search resource.

```python
def find_request_id(...):
    url = makeurl(self.apiurl, ('search', 'request'), query=query)
```

Produces a path like: `/search/request?stuffs`. Whereas the regex is only for `/request/1234?newstate=accepted`. Since `match = pattern.match(path)` is used instead of `search()` it only matches if it starts with the path (ie the `/search` excludes it from matching as well as the lack of a request id before the query).

To be fair that is my least favorite regex. :) The API call to get updates (`last_updated_load()`) would pickup the change, but this avoids having to make that call since it just expires the local cache. Other users will expire via API result.

Again, the regex are written to be as constrictive as possible to avoid such overlap. Note the specifics in `/source$`. Since `match()` is used it has to start with `/source` and since `$` is used it has to end there so only the literal `/source` will be matched. Whereas `get_package_information()`

```python
url = self.makeurl(['source', project, pkgname], query=query)
```

Would produce `/source/$project/$pkgname?rev=$rev` which would not match.

On the maintainability front, indeed pattern matching will never be as obvious as modifying the calling code directly, but having to remember to cache a new call for similar/same information is a burden as well. Always pros and cons. We can definitely refactor to make explicit, but then we'll need all those wrappers as well. From my perspective, aside from the afore mentioned benefits a properly implemented cache should be transparent and therefore not dealt with by developer. Also note that using `--debug` flag will make it very clear what requests are cached and what exactly is going on.

aplanas:
@jberry-suse actually half of my point: no idea from where it comes, and I was a previous contributor for this code and not a new contributor.

jberry-suse:
@aplanas I am not sure what your last comment means.

To look at it another way, browsers cache at the HTTP level rather than that being implemented by applications (like JavaScript ajax calls). Caching on top of HTTP can also be done, like perhaps memoize decorator to cache python processing of the HTTP result. In the same fashion there are developer tools to see if browser loaded from cache or actually made the call. Assuming cache is correct the developer should not have to care. The `ttl` for browsers comes from heads which are typically implemented with similar patterns on server side.

Or reverse HTTP proxy. Not a new idea, just a few OBS specific parts to make it more robust. It seems straight forward to cache the expensive part rather than what happens after that.

aplanas:
@jberry-suse yes sorry. My last comment try to express that this implementation is hard to maintain, hard to debug and very hard to have in mind. Consider that when I read the code, I will never be sure if this request is cached or not, something critical when I need to understand a problem. I am familiar with this code and I have this perception, so I expect that a newcomer contributor will have a very hard time working on it.

I want to make clear that I understand you argument, but still I think that you are wrong in some points:

* There is a current cache implementation that is interfering with the new one
* The cache needs to be done at higher level, so is easy to understand and to maintain
* Doing the cache at http_request is wrong because:
  * monkey-patching is dangerous and is a nightmare to debug
  * the URLs do not have meaning outside from the place that are used, and there is absolute not guarantee that will not interfere with other request in different code path that you do not want to cache at all.
  * in essence: intrusive code is a hack, and needs to be avoided

I understand that your principal technical argument is: cache invalidation is hard. You are right. But to help to resolve this problem I propose another approach: refactor the code to separate the acquisition of data from the processing of the data, and cache the return value of the method that retrieve the data, in that way is more clear where to invalidate the cache.

jberry-suse:
@aplanas With any cache approach one does not know when reading the code if coming from cache. That is only known at runtime. Even with memoize I do not know the request will come from cache only that it may and only then if I look at the method definition or sub method if called indirectly. Presumably you are looking to have it be clear when reading if it _can_ come from the cache which I agree and have already agree is never as obvious as if stated explicitly. That said developers seem fine dealing with this in browsers and other areas. If the cache returns the correct value why should the developer care? When debugging flip the `--debug` flag on (which you'll likely need to do for other information as well) and all is made clear.

To respond to your points:

- The existing cache is not used extensively since it cannot clear caches in the same way and should not cause issues. I agree having both is redundant in those cases.
- Explicit and implicit is a discussion as old as time. Just as it works in other projects (like browsers) it is manageable here. Already pointed out on a maintenance level, explicit style requires the developer to explicitly use the cache when adding new methods which is a maintenance burden.
- If there was no HTTP request involved there would be no need for a cache, thus it is a logical place for the cache, but certainly not the only one.
  - monkey-patching could be avoided if this were merged into `osc.core`, but given that seems unlikely and would rather test further this allows a compatible approach to be utilized. This is the closest thing to running with a patched `osc.core`.
  - URLs are defined by OBS and by definition have explicit meaning. Certain ones may contain information that cannot be cached, but the meaning does not change depending on what method loads it.
  - If this were OOP code then the class could be extended and simply changed where it is initialized to use the new class. Given that it is not this is the only way to encapsulate the change.

Both approaches have their pros and cons. Some absolute definition of wrong here seems a bit simplistic.

A larger refactoring to separate retrieval from processing does not sound like a bad idea, but until then this was designed to be practical given the code base.

nilxam:
seems we need a configurable CACHE_DIR, otherwise OSError: [Errno 13] Permission denied: '/home/repochecker/.cache/osc-plugin-factory' happened on packagelists.

.cache/ in packagelists actually is linked to /var/cache

PR #630, opened 2017-01-08
Update kGraft host-list

fschilling:
* Add kgraft0h, kgraft0i, kgraft1b and kgraft1c.
* Remove kgraft09 (kernel out of support)


coolo:
as these updates come very frequently - how about you guys invest into fetching this data from remote urls?

PR #631, opened 2017-01-10
Provide ignore/unignore request staging commands.

jberry-suse:
It is not uncommon for a request to be in a pending state which requires action beyond the scope of the staging workflow, but does not make sense to deny the request. For lack of a "postponed" or "pending" state on OBS a sudo-state of "ignore" is provided for the staging workflow. The ignore state will remove a request from the "list" and "adi" commands so as not to be accidentally staged. This avoids the need for keeping a context of what requests should be ignored in one's memory.

It is expected that an ignored request will have comments reflecting what is to be done or one should be added via the -m option of ignore command.

During our weekly call @lnussel seemed to agree that such a facility would be useful.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -31,6 +31,8 @@ from osclib.check_command import CheckCommand
 from osclib.cleanup_rings import CleanupRings
 from osclib.conf import Config
 from osclib.freeze_command import FreezeCommand
+from osclib.ignore_command import IgnoreCommand
+from osclib.unignore_command import UnignoreCommand
 from osclib.list_command import ListCommand
 from osclib.obslock import OBSLock
 from osclib.select_command import SelectCommand
@@ -92,6 +94,7 @@ def _full_project_name(self, project):
               help='do not update bootstrap-copy when freezing')
 @cmdln.option('--wipe-cache', dest='wipe_cache', action='store_true', default=False,
               help='wipe GET request cache before executing')
+@cmdln.option('-m', '--message', help='message used by ignore command')
 def do_staging(self, subcmd, opts, *args):
     """${cmd_name}: Commands to work with staging projects
 
@@ -115,6 +118,10 @@ def do_staging(self, subcmd, opts, *args):
 
     "frozenage" will show when the respective staging project was last frozen
 
+    "ignore" will ignore a request from "list" and "adi" commands until unignored
+
+    "unignore" will remove from ignore list
+
     "list" will pick the requests not in rings
 
     "select" will add requests to the project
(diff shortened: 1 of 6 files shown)


jberry-suse:
I imagine removing from `project/staging_projects/$project` backlog list would be possible and useful as well.

I had implemented a wrapper around `osc staging` that included such an ignore list and have found it rather handy.

jberry-suse:
For testing I went ahead and did so for SR 447422 in openSUSE:Leap:42.3.

lnussel:
IMO good idea. Looks like an ignored request vanishes from the list completely and forever. How can we make sure we don't forget about it and re-evaluate after some time if the maintainer doesn't respond?
Maybe we should call it suspend and put some timeout?

nilxam:
I'd prefer don't ignore it in the list command...just add an additional mark or text for the request is ignored in the list command.

DimStar77:
I agree with @nilxam : hiding them away in the osc staging list overview means we have to think about 'getting them back into the game'; it would be more helpful to have a 'reason why it's still, or again, on the backlog' (stuff we currently put in the comments on the SR, but often enough not checked when doing 'staging list'

aplanas:
I think that is missing a command to list the content of the `ignored` list, and a easy way to wipe out the content

nilxam:
IMO integrate ``ignored`` list into list command should good enough to understand, and select an ``ignored`` request by *select command* will auto-remove the request from the ignored list is better.

jberry-suse:
Presenting the list somewhere definitely makes sense as I considered another command to list them. Annotating the list command output seems like a good plan. On the adi front a similar annotation while continuing to ignore for automatic selection into an adi staging. Automatically removing from ignore list when selected seems like a good addition. I also thought about checking the existing request in list and cleaning them up when another is ignored. For example if a request ended up being denied in the mean time.

As for placing the reason in the ignore list, I see the usefulness and considered, but I figured we are duplicating the information that should be present in comments so others not using staging tools can see it. Could require the message parameter and place a comment on the request and in the ignore file for this purpose, or try and read an annotated comment from request. Thoughts?

jberry-suse:
I tried to take into account the suggestions and reworks it a bit. The `--message` option is now stored in `ignored_requests` file so it can easily be displayed when using `list` or `adi` commands. When an ignored request is `select`ed it should automatically be un-ignored (I need to confirm this works).

Some example output:

```bash
$ osc staging -p openSUSE:Leap:42.3 list
```
```
YaST:Head
  sr#449500: yast2-ntp-client               -> 1-MinimalX   (Factory)
    ignored: depends on augeas-lenses update, see previously sr#449500
network
  sr#449747: bind                           -> 1-MinimalX   (SP1:Update)
```

I am torn on what should be done with `Not in a ring:` output on `list` command. Either duplicate what is in `adi` command, hide from list there, or leave as is and just display them with no annotation there since primary focus is ring packages.

jberry-suse:
See the [referenced `ignored_requests` file](https://build.opensuse.org/package/view_file/openSUSE:Leap:42.3:Staging/dashboard/ignored_requests?expand=1) for an example of the simple yaml storage format.

jberry-suse:
Could rework as `@property` like `ring_packages` if desired.

PR #632, opened 2017-01-11
Fix comments default for openqa-maintenance bot

coolo:
And remove some of the noisy debug output

### DIFF
--- a/openqa-maintenance.py
+++ b/openqa-maintenance.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python
-# Copyright (c) 2015,2016 SUSE LLC
+# Copyright (c) 2015-2017 SUSE LLC
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


openqa-maintenance.py
```diff
@@ -1029,6 +1026,10 @@ def setup_checker(self):
         return bot
 
 if __name__ == "__main__":
+    requests_log = logging.getLogger("requests.packages.urllib3")
+    requests_log.setLevel(logging.WARNING)
+    requests_log.propagate = False
+
```
lnussel: Hmm, not sure this belongs here. Maybe we need to leverage
https://docs.python.org/2/library/logging.config.html#logging-config-api

coolo: urllib3 already leverages it - and this is stackoverflow's recommendation how to shut it up.

lnussel: urllib uses logger but the config file would have to be parsed by the program that has __main__

coolo: Huh? what program would that be? I'm not following

PR #633, opened 2017-01-11
Implement config for manager42

lnussel:
For example for Leap 42.3 it could look like this

### DIFF
--- a/manager_42.py
+++ b/manager_42.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-# Copyright (c) 2015 SUSE Linux GmbH
+# Copyright (c) 2015-2017 SUSE Linux GmbH
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -32,12 +32,11 @@ import urllib2
 import sys
 import time
 import yaml
+from collections import namedtuple
 
 from osclib.memoize import memoize
 
 logger = logging.getLogger()
-
-OPENSUSE = 'openSUSE:Leap:42.3'
 
 makeurl = osc.core.makeurl
 http_GET = osc.core.http_GET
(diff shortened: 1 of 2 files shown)


manager_42.py
```diff
@@ -32,55 +32,47 @@
...
-
-        self.parse_lookup(self.from_prj)
+
+        self.config_defaults = {
+            'project_preference_order' : [],
+            'from_prj' : 'openSUSE:Leap:42.3',
+            'factory' : 'openSUSE:Factory',
+            }
+
+        self.config = self._load_config(configfh)
```
jberry-suse: Presumably the config file changing while a process is running and instantiating multiple `Manager42`/`ReviewBot` instances is just a waste. As such would it make more sense to load the config once as a class variable instead of parsing for each new instance? I did this when I incorporated your config patch into #614 (see https://github.com/openSUSE/osc-plugin-factory/pull/614/commits/1fc8ce3905f20cbf3afe28159a6ae63285d2b1d2).

Given that you had used this same style for a potential `ReviewBot` config, does it make more sense to abstract as a separate config utility that both can use? In the same fashion as `osc.conf`. Although after writing that I remembered `osclib/conf.py` which populates `osc.conf`. Either way I imagine this code might be better to be shared?

And I see `# FIXME: add to ToolBase and rebase Manager42 on that`. Alright. I only saw `biarchtool.py` using `ToolBase` so I didn't use for #628 , should I rebase on that as well? If this is to be used by `ReviewBot` should it be rebased on `ToolBase` or perhaps just call a static method to retrieve config?

lnussel:
Jimmy Berry wrote:
> Presumably the config file changing while a process is running and
> instantiating multiple `Manager42`/`ReviewBot` instances is just a
> waste. As such would it make more sense to load the config once as a
> class variable instead of parsing for each new instance? I did this
> when I incorporated your config patch into #614 (see
> https://github.com/openSUSE/osc-plugin-factory/pull/614/commits/1fc8ce3905f20cbf3afe28159a6ae63285d2b1d2).

The defaults make sense to have in the class indeed. Since the
config is different for each target project it's possible in theory
to have multiple instances with different config within the same
process though.

> Given that you had used this same style for a potential
> `ReviewBot` config, does it make more sense to abstract as a
> separate config utility that both can use? In the same fashion as
> `osc.conf`. Although after writing that I remembered
> `osclib/conf.py` which populates `osc.conf`. Either way I imagine
> this code might be better to be shared?

Absolutely!

> And I see `# FIXME: add to ToolBase and rebase Manager42 on that`.
> Alright. I only saw `biarchtool.py` using `ToolBase` so I didn't
> use for #628 , should I rebase on that as well? If this is to be
> used by `ReviewBot` should it be rebased on `ToolBase` or perhaps
> just call a static method to retrieve config?

IMO a common base class for bots and tools does make sense and helps
to avoid code duplication. ToolBase was my attempt to start with
that. So if it's a good aproach, then yes, tools should use it. If
there's a better way I don't replacing it either.

manager_42.py
```diff
@@ -32,55 +32,47 @@
...
         self.fill_package_meta()
         self.packages = dict()
-        for project in [self.from_prj] + self.project_preference_order:
+        for project in [self.config.from_prj] + self.config.project_preference_order:
             self.packages[project] = self.get_source_packages(project)
 
+    # FIXME: add to ToolBase and rebase Manager42 on that
+    def _load_config(self, handle = None):
+        d = self.__class__.config_defaults
+        y = yaml.safe_load(handle) if handle is not None else {}
```
jberry-suse: The config is still read every-time the class is loaded, but presumably unlike `ReviewBot` this class is not instantiated multiple times so it may not matter. Although if they share the same code they might as well both check that not already loaded.

PR #636, opened 2017-01-12
Correct indentation level of in requests_ignored check.

jberry-suse:
In the previous state this worked for any project except factory, now it works on factory too.

I normally avoid running commands against factory, but given `list` makes no changes it should be fine. Anyway, I seem to have been focused on ensuring the `\n` was added after
```python
result[devel][-1] += ' ({})'.format(source_prj)
```
that I included the check in

```python
if self.api.project != "openSUSE:Factory" and action.find('source') != None:
```

which clearly would not work for factory. Glad to see @DimStar77 already tried to use it though. ;)

### DIFF
--- a/osclib/list_command.py
+++ b/osclib/list_command.py
@@ -87,8 +87,8 @@ class ListCommand:
                     elif source_prj.startswith('home:'):
                         source_prj = '~' + source_prj[len('home:'):]
                     result[devel][-1] += ' ({})'.format(source_prj)
-                    if request_id in requests_ignored:
-                        result[devel][-1] += '\nignored: ' + requests_ignored[request_id]
+                if request_id in requests_ignored:
+                    result[devel][-1] += '\nignored: ' + requests_ignored[request_id]
             else:
                 non_ring_packages.append(target_package)
 


jberry-suse:
With this change I see the ignored SRs in both Leap:42.3 and Factory showing in list.

```
Base:System
  sr#449838: glibc                          -> 0-Bootstrap 
  sr#449922: irqbalance                     -> 2-TestDVD   
KDE:Distro:Factory
  sr#449956: qalculate                      -> 2-TestDVD   
KDE:Extra
  sr#449804: digikam                        -> 2-TestDVD   
    ignored: Waiting for QtAV legal review / checkin to Factory
devel:libraries:c_c++
  sr#448595: zlib                           -> 0-Bootstrap 
    ignored: Put on backlog due to a break
science
  sr#449363: gsl                            -> 1-MinimalX  
server:database
  sr#448102: mariadb                        -> 1-MinimalX  
server:database:postgresql
  sr#444913: postgresql95                   -> 1-MinimalX  
server:mail
  sr#449941: postfix                        -> 1-MinimalX
```

PR #639, opened 2017-01-16
Enhance fallback-onfail review mode

sbahling:
- Add message when setting fallback reviewer indicating the
  ReviewBot failure
- Do NOT set review to "accepted". Nothing has been "accepted" yet;
  just delegated to human review.
- Fixed debug output (state should be newstate)

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -115,21 +115,22 @@ class ReviewBot(object):
 
         by_user = self.fallback_user
         by_group = self.fallback_group
+        msg = self.review_messages.get(state, state)
 
         if state == 'declined':
             if self.review_mode == 'fallback-onfail':
                 self.logger.info("%s needs fallback reviewer"%req.reqid)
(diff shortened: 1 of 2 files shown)


sbahling:
This is for using fallback-onfail with openSUSE:Backports. Not sure about other users of this review mode, and if these changes would break things for others.

DimStar77:
>Do NOT set review to "accepted". Nothing has been "accepted" yet;
just delegated to human review.

The bot has to accept 'his' review - otherwise the human review means nothing.

sbahling:
> The bot has to accept 'his' review - otherwise the human review means nothing.

OK. In my testing I only had one account (mine) to use for all roles (requester, bot, and fallback reviewer) and in that case it appeared as if the fallback review got automatically accepted. I will try testing with separate users.

sbahling:
OK. After testing under more accurate conditions, I don't need this at all - things work as expected. Closing pull request.

PR #643, opened 2017-01-17
check_source_in_factory: Allow checking sources against more than one…

sbahling:
… project.

  --factory option can be specified multiple times to check against
  a list of projects. For example checking against Factory and Leap
  for the Backports project.

  Check will loop over each project and break as soon as a match is
  found. Project search order matches order of --factory options (at least
  in testing).

### DIFF
--- a/check_source_in_factory.py
+++ b/check_source_in_factory.py
@@ -92,13 +92,14 @@ class FactorySourceChecker(ReviewBot.ReviewBot):
     def _package_get_upstream_projects(self, package):
         """ return list of projects where the specified package is supposed to come
         from. Either by lookup table or self.factory """
+        projects = self.factory
         if self.lookup and package in self.lookup:
-            return self.lookup[package]
+            projects = self.lookup[package]
 
-        if type(self.factory) is str:
-            return [self.factory]
+        if isinstance(projects, basestring):
+            projects = [projects]
 
-        return self.factory
+        return projects
 
     def _check_project(self, project, package, rev):
         """check if factory sources contain the package and revision. check head and history"""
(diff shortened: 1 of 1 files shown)


lnussel:
hmm, maybe another case where we should better switch to the leaper bot

check_source_in_factory.py
```diff
@@ -67,28 +67,37 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
...
-    def _package_get_upstream_project(self, package):
-        """ return project where the specified pacakge is supposed to come
+    def _package_get_upstream_projects(self, package):
+        """ return list of projects where the specified package is supposed to come
         from. Either by lookup table or self.factory """
         if self.lookup and package in self.lookup:
             return self.lookup[package]
 
+        if type(self.factory) is str:
+            return [self.factory]
```
lnussel: is there a case where this happens? what about renaming self.factory to self.upstream and require list always?

sbahling: I had that in case there is another consumer of FactorySourceChecker that explicitly sets self.factory. I haven't reviewed all the osc_plugin_factory code. But within this module, it should never be the case if we initialize the variable as a list. Changing the name would further ensure it doesn't happen with existing code ;)

jberry-suse:
Based on the features provided by `check_source_in_factory.py` and usage elsewhere it is not specific to factory and this takes that a step further. We should consider renaming it and refactoring the code the remove references to factory. In #644 it is used in a similar way to check against multiple projects. If not opposed I'll wip up a refactoring PR after this is resolved.

lnussel:
fine with me to make this more generic

check_source_in_factory.py
sbahling: Actually this check is in the wrong place because the code just above it can return a string to the caller expecting a list. The value from the lookup needs to be converted also. I think it's best to keep this check and conversion in the _package_get_upstream_projects method and keep the change hidden from other users of the FactorySourceChecker class (like check_tags_in_requests.py).

sbahling:
So do we want to wait for refactor by @jberry-suse? Or can we take this quick patch now and refactor later? We kind of need this feature urgently to check the latest KDE submissions to Backports which are based on Leap code.

jberry-suse:
I was not indicated we should hold this up.

PR #645, opened 2017-01-18
Leaper tuning for SLE

lnussel:


leaper.py
```diff
@@ -125,35 +127,48 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
...
                 self.logger.info('found package in openSUSE.org:openSUSE:Leap:42.2')
-                return True
 
-            devel_project, devel_package = self.get_devel_project('openSUSE.org:openSUSE:Factory', target_package)
-            if devel_project is not None and devel_package is not None:
-                if self._check_factory(devel_package, src_srcinfo, devel_project) is not False:
+            devel_project, devel_package = self.get_devel_project('openSUSE.org:openSUSE:Factory', package)
+            if devel_project is not None:
+                if devel_package is None:
+                    devel_package = package
```
jberry-suse: This guesses the same package name when handling new packages since there is no devel link? Perhaps toss a comment in there?

leaper.py
```diff
@@ -125,35 +127,48 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
...
+                self.logger.info("expected origin is '%s' (%s)", origin,
+                                 "unchanged" if origin_same else "changed")
+
             # True or None (open request) are acceptable for SLE.
-            if self._check_factory(target_package, src_srcinfo, 'openSUSE.org:openSUSE:Factory') is not False:
-                self.logger.info('found package in openSUSE.org:openSUSE:Factory')
-                return True
+            self.source_in_factory = self._check_factory(package, src_srcinfo, 'openSUSE.org:openSUSE:Factory')
+            if self.source_in_factory is None:
+                self.pending_factory_submission = True
```
jberry-suse: I did not set this since the same style of message would make sense for the rest of the checks so I kept it consistent. The way the messages are built is a bit separated from the rest and perhaps a function with a project as a parameter would be better.

Obviously, this can be done for just Factory if desired.

leaper.py
```diff
@@ -125,35 +127,48 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
...
                     self.logger.info('found package in {}/{}'.format(devel_project, devel_package))
                     return True
             else:
-                self.logger.info('no devel project found for {}/{}'.format('openSUSE.org:openSUSE:Factory', target_package))
+                self.logger.info('no devel project found for {}/{}'.format('openSUSE.org:openSUSE:Factory', package))
 
             self.logger.info('sources not found in Factory, Leap:42.2, or devel project')
 
-            return False
+            return origin_same
```
jberry-suse: During our discussions of workflow I started to wonder what the workflow for ignoring leaper's suggestion would be. I imagine certain requests will intentionally not be on OBS.

leaper.py
```diff
@@ -343,10 +358,11 @@ def check_one_request(self, req):
             return False
 
         request_ok = ReviewBot.ReviewBot.check_one_request(self, req)
-        has_correct_maintainer = self.maintbot.check_one_request(req)
+        if not self.ibs:
+            has_correct_maintainer = self.maintbot.check_one_request(req)
+            self.logger.debug("has_correct_maintainer: %s", has_correct_maintainer)
```
jberry-suse: Yeah, figured that probably made sense. :)

leaper.py
```diff
@@ -91,6 +91,8 @@ def __init__(self, *args, **kwargs):
     def prepare_review(self):
         # SLE workflow is not concerned with origin.
         if self.ibs:
+            self.factory.parse_lookup('SUSE:SLE-12-SP3:GA')
```
jberry-suse: This causes a 404 at the movement. Presumably, this file will exist soon?

leaper.py
```diff
@@ -91,6 +91,8 @@ def __init__(self, *args, **kwargs):
     def prepare_review(self):
         # SLE workflow is not concerned with origin.
```
jberry-suse: Remove this comment since that does not seem to be the case given these changes.

check_source_in_factory.py
```diff
@@ -117,14 +118,16 @@ def _check_project(self, project, package, rev):
...
-        for revision in root.findall('revision'):
+        for revision, i in zip(reversed(root.findall('revision')), count()):
             node = revision.find('srcmd5')
             if node is None:
                 continue
             self.logger.debug("checking %s"%node.text)
             if node.text == rev:
                 self.logger.debug("got it, rev %s"%revision.get('rev'))
                 return True
+            if i == 5:
```
jberry-suse: Perhaps a comment referencing the bug you filed?

leaper.py
lnussel: it gets to that point if a package exists and only specifies a devel project, not a package. Specifying the package is redundant in the meta data if the name is the same. That's what the get_devel_project returns None if the attribute doesn't exist in the meta data.

leaper.py
lnussel: you this is worth refactoring? yes, agree :)

leaper.py
lnussel: I hope so. I just have to get hold of the right person.

leaper.py
lnussel: we'll see how this works in practice. I expect that release managers will complain and the code needs fine tuning. btdt with the guy in the mirror ;)

leaper.py
lnussel: ideally this needs to be reimplemented too. For SLE bugownership in one of the layered projects is used, rather than maintainership in devel projects.

leaper.py
lnussel: fwiw for testing use home:lnussel

leaper.py
jberry-suse: I was not aware the devel link skips the package if the same. If that's the case it probably makes sense to just add that logic to `get_devel_project()`.

Looking at the output that does not seem to be the case:
```xml
<package name="mousepad" project="openSUSE:Factory">
  <title/>
  <description/>
  <devel project="X11:xfce" package="mousepad"/>
</package>
```

Which is what I remembered seeing. Not sure then what case a devel link exists without a package?

PR #649, opened 2017-01-21
staging adi: provide --open-review-by and --loop-when-reviewed options.

jberry-suse:
Together these options help increase the efficiency of adi reviews by opening requests that are waiting on reviews by the specified group so the user can review them. When combined with loop option all pending reviews can be opened and stagings marked ready.

As noted in the comment this does not handle all possible future workflows and may need tweaking as it stands. I'd like to start a discussion about defining the workflows used in the various projects so things like the list of bots that should be considered blocking before prioritizing human review and the release manager related groups that should be opened for reviews can be placed in config files rather than loads of command line arguments that will need to differ for each staging.

For Leap the workflow looks like:
- pending leaper (which may be a while as it waits for Factory acceptance)
- leap-reviewers once leaper has accepted
- auto open requests and loop which will eventually make it through any adi stagings that are blocked by leap-reviewers
- final loop marks requests as ready to accept and deletes adi stagings
- last iteration also stages new requests

This is another on of the things I have been using in a wrapper script for months and I find rather useful. If this approach is too simplistic and would rather wait for a more exhaustive one that may make sense, but hopefully it can be of use in this state.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -103,6 +103,8 @@ def _full_project_name(self, project):
 @cmdln.option('--filter-by', action='append', help='xpath by which to filter requests')
 @cmdln.option('--group-by', action='append', help='xpath by which to group requests')
 @cmdln.option('-i', '--interactive', action='store_true', help='interactively modify selection proposal')
+@cmdln.option('--open-review-by', metavar='GROUP', help='open request (xdg-open) when relevant review matches')
+@cmdln.option('--loop-when-reviewed', action='store_true', help='loop when a request was opened for review')
 def do_staging(self, subcmd, opts, *args):
     """${cmd_name}: Commands to work with staging projects
 
@@ -185,6 +187,7 @@ def do_staging(self, subcmd, opts, *args):
 
     Usage:
         osc staging accept [--force] [LETTER...]
+        osc staging adi [--by-develproject] [--open-review-by GROUP [--loop-when-reviewed]]
         osc staging check [--old] REPO
         osc staging cleanup_rings
         osc staging freeze [--no-boostrap] PROJECT...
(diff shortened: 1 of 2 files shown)


lnussel:
interesting concept to open the requests in the browser. I wonder if it's the right place though, esp since it relies on magic like the alphabetical ordering.

jberry-suse:
I think the solution down the road is some sort of workflow configuration that can handle setting the defaults for a lot of commands based on which project is being acted on. Since this is not replacing default functionality and works for the cases I can think of it does not seem it would have ill effects. I'd be curious to know what other workflows folks use for this sort of thing and if this would be useful to others.

jberry-suse:
Otherwise, I can add an option for which reviews to consider blocking before opening for the desired group. Those are the things I would like to see in a config file.

jberry-suse:
rebased

lnussel:
I'm still undecided about this one. It somehow feels misplaced. Maybe we should extend the staging dashboard code to give a better overview of the pending reviews?

lnussel:
https://github.com/openSUSE/obs_factory

DimStar77:
I think a better dashboard for the adi projects is what we really need (I have a hack board that is updated once a day - it's way too expensive: http://adi.leuenberger.net/dashboard.html )

opening 'all reviews in a browser tab' sounds like nothing I would like to try against openSUSE:Factory

nilxam:
that's what I planned to do in this hackweek, at least a tabbar to switch between normal staging and adi staging will be there.

jberry-suse:
@DimStar77 curious do you normally review them in browser or cli? This code opens them one at a time in browser so it shouldn't open a ton at once.

DimStar77:
'it depends' - when I tend to do 'reviews in a batch' (review team task) I often use "osc review list -i openSUSE:Factory -G opensuse-review-team"

That goes through the list, allows you to show diff, build results and interactively accept/decline/skip them review

jberry-suse:
Seems handy, I'll try that out.

jberry-suse:
Perhaps I should look into adding the blocking group type parameters to the `osc review list` and an option to open if browser if still desired. The `-i` option is definitely very nice.

PR #650, opened 2017-01-24
Don't freeze bash-completion-AGGR

nilxam:
Should not freeze aggregate pacakge in the staging, thus do not freeze bash-completion-AGGR as well. ie. it should not have the build result in the staging project.

### DIFF
--- a/osclib/freeze_command.py
+++ b/osclib/freeze_command.py
@@ -24,6 +24,7 @@ class FreezeCommand(object):
     def __init__(self, api):
         self.api = api
         self.projectlinks = []
+        self.aggrpkgs = ['rpmlint-mini-AGGR', 'bash-completion-AGGR']
 
     def set_links(self):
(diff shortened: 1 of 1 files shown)


lnussel:
so we have a build loop with bash-completion or what is bash-completion-AGGR for? can we break the loop?

nilxam:
bash-completion has been in ring0 but now it was in ring1, in my memory I only recall it is an compromise deal, ie. some stuff runtime requires bash-completion in ring0 however bash-completion changed its build-deps and that change is expensive for ring0 itself.

nilxam:
but bash-completion has build-deps change afterward, should worth to check can we move it back to ring0 instead of aggreegate pkgs.

DimStar77:
I think this is obsolete - I could successfully remove bash-completion-AGGR from ring0 without anything really caring for it

coolo:
there was a bootstrap problem at one point with an util-linux update. This came from SP1 update

PR #653, opened 2017-01-26
ttm: Add --daemon parameter

DimStar77:
When running as a systemd daemon, it is useless to wait for user input
to possible trigger an earlier re-run. Interaction is not possible.

Additionally, there is no stdin assigned to the process when running as
a systemd daemon, which results in raw_input() failing with EOFError

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -16,6 +16,7 @@ import sys
 import urllib2
 import logging
 import signal
+import time
 
 from xml.etree import cElementTree as ET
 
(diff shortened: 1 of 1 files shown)


DimStar77:
This allows ttm to run as a daemon, no longer requiring it to be fired up in a screen/tmux session.

All outputs can be found in the systemd journal

DimStar77:
Updated the pull req - it now auto-detects if stdin is available, so there is no need to pass --daemon when running it as a systemd service...

PR #656, opened 2017-01-27
Added X11:Solus to the Factory devel project whitelist

Mailaender:
To resolve the error at https://build.opensuse.org/request/show/452889 Not sure if this is correct. I guess https://en.opensuse.org/openSUSE:How_to_contribute_to_Factory#How_to_request_a_new_devel_project needs an update.

### DIFF
--- a/check_source.whitelist
+++ b/check_source.whitelist
@@ -13,3 +13,4 @@ server:messaging
 system:snappy
 X11:GNUstep
 FATE
+X11:Solus


DimStar77:
Thanks - looking good (and yes, the wiki needs to be updated since the whitelist was moved out to a file of its own... care to do it please?)

Mailaender:
[Done.](https://en.opensuse.org/index.php?title=openSUSE%3AHow_to_contribute_to_Factory&diff=118795&oldid=118254)

PR #657, opened 2017-01-27
Announcer: put the state of the last version into /var/lib

DimStar77:
I think it's better to store such info in a defined place than in the program binary directory


lnussel:
that would make testing it as normal user harder. what about using XDG_CONFIG_HOME or just change cwd in the startup script?

DimStar77:
just a note (as I think it was missed): I updated the pull req here

PR #658, opened 2017-01-31
Make factory-source checker also accept Leap

lnussel:


DimStar77:
That script no longer runs in this form on the new VM

DimStar77:
The new VM uses a timer-based service, containing:

```
cat source-checker.service 
[Unit]
Description=Validate that sources are in Factory

[Service]
Type=oneshot
ExecStart=/usr/local/bin/check_source_in_factory --verbose review
User=factory-source

[Install]
WantedBy=multi-user.target
```

lnussel:
ok, will modify that then

PR #662, opened 2017-02-01
Remove source devel project check carried over from flawed adi logic

jberry-suse:
Followup on confirmation that previous adi logic was bad and some further testing.

- Remove incorrect devel project lookup copied from adi.
- Move @devel_project to action/target.

### DIFF
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -70,7 +70,7 @@ class RequestSplitter(object):
         target_project = request.find('./action/target').get('project')
         devel = self.devel_project_get(target_project, target_package)
         if devel:
-            request.find('./action/source').set('devel_project', devel)
+            request.find('./action/target').set('devel_project', devel)
 
         ring = self.ring_get(target_package)
         if ring:
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -34,20 +34,19 @@ class RequestSplitter(object):
     def filter_only(self):
         ret = []
         for request in self.requests:
-            target_package = request.find('./action/target').get('package')
-            self.suppliment(request, target_package)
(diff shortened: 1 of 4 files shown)


osclib/request_splitter.py
```diff
@@ -89,13 +89,10 @@ def ring_get(self, target_package):
...
-        # Preserve logic from adi and note that not Leap development friendly.
-        source = request.find('./action/source')
-        devel = self.api.get_devel_project(source.get('project'), source.get('package'))
-        if devel is None and self.api.project.startswith('openSUSE:'):
-            devel = self.api.get_devel_project('openSUSE:Factory', target_project)
-        return devel
+    def devel_project_get(self, target_project):
+        if self.api.project.startswith('openSUSE:'):
+            return self.api.get_devel_project('openSUSE:Factory', target_project)
+        return None
```
lnussel: that is a bit too oversimplified I think :) it stil does make sense to check for devel in the target (leap for example has a handful packages with devel project set). The previous mistake was to check the source. Factory should be used as fallback.

osclib/request_splitter.py
```diff
@@ -89,13 +89,10 @@ def ring_get(self, target_package):
...
         return None
 
-    def devel_project_get(self, request, target_project):
-        # Preserve logic from adi and note that not Leap development friendly.
-        source = request.find('./action/source')
-        devel = self.api.get_devel_project(source.get('project'), source.get('package'))
-        if devel is None and self.api.project.startswith('openSUSE:'):
-            devel = self.api.get_devel_project('openSUSE:Factory', target_project)
-        return devel
+    def devel_project_get(self, target_project):
```
lnussel: is actually target_package but project is needed too, see below

PR #663, opened 2017-02-01
Trigger source service for _product in :zSystems on accept

DimStar77:


DimStar77:
Otherwise the media in :zSystems are not updated with the right version number

nilxam:
LGTM

PR #667, opened 2017-02-06
Return None if has not action/source

nilxam:
eg. DELETE REQUEST


lnussel:
needs to look for action/target anyways. the source package name could be different

nilxam:
the problem here is about has not action/source entry...

nilxam:
and this is inside of devel_project_get()

jberry-suse:
This was fixed in #662 and should be reverted.

jberry-suse:
Looks like it was already reverted when other change was merged (assume @lnussel changed my PR or somesuch).

nilxam:
yeah, we aware of https://github.com/openSUSE/osc-plugin-factory/pull/662 after merged this change.

PR #670, opened 2017-02-08
issue-diff: use bugowner or maintainer as assignee for bug (and small changes)

jberry-suse:
Re-initialized the issue-db repo and granted @coolo write access per discussion.

- update description to include a short reason.
- change default project to SLE-12-SP3.
- use bugowner or maintainer as assignee for bug.

Still not sure on what is desired for description text. @coolo: How wordy do you want it?

### DIFF
--- a/issue-diff.py
+++ b/issue-diff.py
@@ -31,7 +31,7 @@ ISSUE_SUMMARY = u'[{label}]({url}) owned by {owner}: {summary}'
 ISSUE_SUMMARY_PLAIN = u'[{label}]({url})'
 
 
-def bug_create(bugzilla_api, meta, cc, summary, description):
+def bug_create(bugzilla_api, meta, assigned_to, cc, summary, description):
     createinfo = bugzilla_api.build_createbug(
         product=meta[0],
         component=meta[1],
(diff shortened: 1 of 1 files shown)


jberry-suse:
Added link to https://mailman.suse.de/mlarch/SuSE/research/2017/research.2017.02/msg00051.html.

issue-diff.py
```diff
@@ -26,26 +26,49 @@
...
+    url = osc.core.makeurl(apiurl, ('person', userid))
+    root = ET.parse(osc.core.http_GET(url)).getroot()
+    email = root.find('email')
+    return email.text if email is not None else None
+
+def bug_owner(apiurl, package):
+    query = {
+        'binary': package,
+    }
+    url = osc.core.makeurl(apiurl, ('search', 'owner'), query=query)
```
coolo: what apiurl is that? If it's suse.de, this should be fine

jberry-suse: It is, `apiurl_default` is OBS and only used for Factory package list calls. `apiurl` is always suse.de (or to whatever the `-A` option is set)

PR #671, opened 2017-02-08
Delete request must are in_ring package always

nilxam:
Delete reqeut implies a ring type for staging project, according to the staging project design, the delete request can not add to adi staging due to failes to expand the source.

### DIFF
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -42,12 +42,23 @@ class RequestSplitter(object):
     def split(self):
         for request in self.requests:
             self.suppliment(request)
+            delete_request = False
 
             if not self.filter_check(request):
                 continue
 
             target_package = request.find('./action/target').get('package')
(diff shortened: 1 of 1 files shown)


jberry-suse:
It seems like it might make more sense to simulate the package being in a ring fully by assigning it to one in `supplement()` that way it would be filterable/groupable as such with other options.

nilxam:
whether the delete request is ring package or non-ring package, it must to go through normal staging, so we should avoiding show it as pending request for adi staging ie. Not in a ring:... also adi cmd should not process them.

I just provide a quick fix here otherwise I can run staging adi for TW, if you have a good idea in mind to fix it will be welcomed :)

jberry-suse:
Yeah, I understand. I'll go ahead and try making the change in supplement.

nilxam:
https://github.com/openSUSE/osc-plugin-factory/pull/676 instead

PR #672, opened 2017-02-08
maintenance: Do not send notifications for *starting* to test

coolo:


lnussel:
I wonder what happens if neither by_user nor by_group is active

coolo:
I think the API tries all reviews

coolo:
this is the more severe problem :)

ERROR:openqa-maintenance.py:unhandled exception in openQA Bot
ERROR:openqa-maintenance.py:Traceback (most recent call last):
  File "/home/sleopenqa-bot/osc-plugin-factory/openqa-maintenance.py", line 923, in check_one_request
    msg='now testing in openQA')
TypeError: change_review_state() got an unexpected keyword argument 'msg'

PR #674, opened 2017-02-08
request_splitter: delete requests should always be considered in a ring.

jberry-suse:
An alternative approach to #671 which places delete request not otherwise in a ring in a fake ring "delete" since it is only used for filtering/grouping. Could place in an existing ring if desired, but this keeps it out of the adi list.


jberry-suse:
Depends on #676 https://github.com/openSUSE/osc-plugin-factory/pull/676/commits/3de3cae13a9a38589b61477d78f916bf615d16fb

jberry-suse:
I decided to just include in the other PR.

PR #676, opened 2017-02-08
request_splitter: house keeping (and fix for SLE [non-ring] workflow)

jberry-suse:
- list: remove case for delete request without ring.
- non-ring projects consider all requests in rings.
- utilize ring attribute rather then lookup again.
- rather than an empty group name use 00.


jberry-suse:
For those curious I ran a bisect using a script (only works if >= 1 request in queue):
```bash
osc -A ibs staging -p SUSE:SLE-12-SP3:GA list | grep "Not in a ring"
if [ $? -eq 0 ] ; then
  exit 1
fi
```
Which reveal the request_splitter broke SLE workflow.

nilxam:
LGTM

PR #678, opened 2017-02-09
Reload known_failures on each run

lnussel:


DimStar77:
Is there actually a problem in the existing code? ttm-tumbleweed is a long-living service already and it catches changes I add to known_failures whenever they happen; I'd have cursed about this long ago otherwise

lnussel:
you are right. I constructs a new class each time

Issue #677, opened 2017-02-09
check_tags_in_requests: warn about removed references


lnussel:
check_tags_in_requests should warn if a submission removes bug references

PR #679, opened 2017-02-10
check_tags: warn when issue references are removed (plus other fixes)

jberry-suse:
Requested in #677.

- handle failed to load issue diff for existing package.
- correct backwards diff.
- use len() instead of list() since only count is of interest.
- **warn when issue references are removed.**
- grammatical corrections.

Interestingly enough the diff would appear to have been performed backwards. Since only the existence of issues was checked this worked fine, but given the request was about warning when removed it seemed proper to flip rather than look for `added` when meaning `deleted`.

For example, sr#455816 which is clearly adding one bug reference showed it as being `deleted`:

```xml
<sourcediff key="c9b42dc9c3a1f9435342cbe12cad4d3d">
  <old package="python3-requests.6329" project="openSUSE:Leap:42.2:Update" rev="baa02f86911ca86af78b4f0db1b7b882" srcmd5="baa02f86911ca86af78b4f0db1b7b882" />
  <new package="python3-requests" project="openSUSE:Leap:42.3" rev="4" srcmd5="1809a2399024102a6b19ea1ea83e79f2" />
  <files />
  <issues>
    <issue label="boo#912903" name="912903" state="deleted" tracker="bnc" url="https://bugzilla.opensuse.org/show_bug.cgi?id=912903" />
  </issues>
</sourcediff>
```

Whereas after reversing the diff:

```xml
<sourcediff key="148e0ab7e2d155c8933bd5d73d78691d">
  <old package="python3-requests" project="openSUSE:Leap:42.3" rev="4" srcmd5="1809a2399024102a6b19ea1ea83e79f2" />
  <new package="python3-requests.6329" project="openSUSE:Leap:42.2:Update" rev="baa02f86911ca86af78b4f0db1b7b882" srcmd5="baa02f86911ca86af78b4f0db1b7b882" />
  <files />
  <issues>
    <issue label="boo#912903" name="912903" state="added" tracker="bnc" url="https://bugzilla.opensuse.org/show_bug.cgi?id=912903" />
  </issues>
</sourcediff>
```

### DIFF
--- a/check_tags_in_requests.py
+++ b/check_tags_in_requests.py
@@ -78,13 +78,13 @@ by OBS on which this bot relies on.
 
     def checkTagInRequest(self, req, a):
         u = osc.core.makeurl(self.apiurl,
-                             ['source', a.tgt_project, a.tgt_package],
+                             ['source', a.src_project, a.src_package],
                              {'cmd': 'diff',
                               'onlyissues': '1',
                               'view': 'xml',
-                              'opackage': a.src_package,
-                              'oproject': a.src_project,
-                              'orev': a.src_rev})
+                              'opackage': a.tgt_package,
+                              'oproject': a.tgt_project,
+                              'rev': a.src_rev})
         try:
             f = osc.core.http_POST(u)
         except (HTTPError, URLError):
--- a/check_tags_in_requests.py
+++ b/check_tags_in_requests.py
@@ -56,7 +56,7 @@ class TagChecker(ReviewBot.ReviewBot):
         super(TagChecker, self).__init__(*args, **kwargs)
         self.factory = "openSUSE:Factory"
         self.review_messages['declined'] = """
-(This is a script running, so report bugs)
+(This is a script, so report bugs)
 
 The project you submitted to requires a bug tracker ID marked in the
 .changes file. OBS supports several patterns, see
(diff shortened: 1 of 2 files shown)


check_tags_in_requests.py
```diff
@@ -77,32 +77,33 @@ def isNewPackage(self, tgt_project, tgt_package):
...
-        # and it is a new package to target project, then leave it to
-        # human review
-        if is_new:
-            self.logger.info("New package to %s and have not the matched revision in Factory"%a.tgt_project)
+            if self.isNewPackage(a.tgt_project, a.tgt_package):
+                self.review_messages['accepted'] = 'New package'
+                return True
+
+            self.review_messages['accepted'] = 'Warning: failed to load package diff'
             return True
```
lnussel: hmm, this would accept requests that cannot be diffed. I guess None would emulate the old behavior when the bot crashed.

jberry-suse: Yes, so previously when failed to load diff it would only `return True` when `is_new`, which actually meant it would crash entirely on `xml = ET.parse(f)` and fail to mark the review either way. This new code will accept in both cases, but include the warning (which was not present before).

jberry-suse: I re-read what you wrote. Did you want to change this to return `None` is the request is not updated? I imagine it may remain open forever if the diff issues was not transient. In the case of new we seem to agree it needs to remain True.

lnussel: yes, if we return None the review may stay open forever until someone overrides it or fixes the cause. I fear that if we return True right away the bot will approve requests on transient failures. Sometimes OBS is overloaded to create diffs and errors out IIRC. At a later time it may work again.

jberry-suse: Change to `return None`.

Issue #573, opened 2016-09-30
osclib/update_status_comments: put link to staging dashboard


lnussel:
put link to staging dashboard in comment, eg.
https://build.opensuse.org/project/staging_projects/openSUSE:Leap:42.2/adi:84

helps to get an overview if there are many packages

shyukri:
This link triggers 500 error on OBS.

lnussel:
the adi projects are dynamic. just try a number that exists :)

lnussel:
ah, and 42.2 is done. so try Factory or 42.3

PR #683, opened 2017-02-10
stagingapi: update_status_comments() include link to dashboard.

jberry-suse:
Requested in #573.

Fully tested by running:
```python
api.update_status_comments('openSUSE:Leap:42.3:Staging:adi:9', 'select')
```

Which resulted in the correct link in [comment](https://build.opensuse.org/project/show/openSUSE:Leap:42.3:Staging:adi:9).

I debated if the link should be placed around a new piece of text like `(dashboard)` or somesuch, but given the comment is on the project page the only other relevant link related to just the project seems like dashboard. Either way easy to change since url is correct.

The new `extract_staging_short()` from `request_splitter` did the trick and works for both letter and adi. :)

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1199,7 +1199,8 @@ class StagingAPI(object):
 
         meta = self.get_prj_pseudometa(project)
         lines = ['<!--- osc staging %s --->' % command]
-        lines.append('The list of requests tracked in %s has changed:\n' % project)
+        dashboard_url = '{}/project/staging_projects/{}/{}'.format(self.apiurl, self.project, self.extract_staging_short(project))
+        lines.append('The list of requests tracked in [%s](%s) has changed:\n' % (project, dashboard_url))
         for req in meta['requests']:
             author = req.get('autor', None)
             if not author:


jberry-suse:
Could potentially drop the `apiurl` and use a path relative to root (this assumes the server referenced by `apiurl` is not hosted in subdir (which I am not sure if OBS even supports)). The url would just be `/project/staging_projects/...` instead of `http://..../project/staging_projects/...`.

lnussel:
the adi seems gone again. apiurl would be api.o.o, right? We need build.o.o

jberry-suse:
I remember reading the code in `obs_factory`, and in practice it seems to work since the plugin redirect internally for paths that are part of the application and not API. I can change it, but then I imagine it either needs to be hardcoded or derived from the apiurl which is potentially b.o.c specific. Long-story short api.o.c works.

Alternatively the relative path describe in my previous comment.

jberry-suse:
For example: https://api.opensuse.org/project/staging_projects/openSUSE:Leap:42.3/A works just fine.

jberry-suse:
I just confirmed the relative approach works as well so perhaps just use that?
```
[text](/project/staging_projects/openSUSE:Leap:42.3/A)
```

lnussel:
ah, the api link works if logged in. I don't mind either way.

Issue #641, opened 2017-01-17
move commenting feature of leaper to ReviewBot class


lnussel:
leaper posts info messages as comment. That feature would be useful for other bots such as check_source_in_factory too. So the relevant functions need to be moved to the ReviewBot class. Leaper should continue to work of course.

PR #684, opened 2017-02-10
ReviewBot: refactor leaper comment from log functionality.

jberry-suse:
Simple interface:
- comment_handler_add(): Add handler to start recording log messages for comment.
- comment_write(): Write comment from log messages if not similar to previous comment.

Previous implementation had some interesting quirks that I will list in case some were intentional.
- Checked `if self.comment_log:` before processing comment even though that will always be set. I changed to `do_comment` and removed from subsequent functions. Could also move do_comment into ReviewBot, but it seemed more like a guarding flag when turning on the feature rather then something to be part of the API.
- `Leaper.comment_add()` provided a result parameter, but it was never used. It was however calculated, but never passed along.
- Checked state of comment in `find_obs_request_comment()` and again in `add_comment()`.

Notable changes:
- Since the bot that made the comment is still important it now is based on bot class name which means the leaper comments went from `leaper` to `Leaper`. The code does a case-insensitive check to ensure previous comments are still correlated to leaper.
- Log message collection is now done by a handler instead of abusing a filter. Filter is still configurable by either adding a filter to handler or setting log level. Defaults to log level `logging.INFO` as leaper used.
- `comment_write()` automatically drop handler which is then required to be added again to provide existing functionality with one less implementation detail exposed. If there is a case where a comment will be posted and another that includes all the lines from the previous comment this will need to be changed, but that seems highly unlikely.
- The comment marker regular expression supports any state/result instead of fixed list.
- Caller no longer needs to pass request when calling from inside `check_one_request()`
- Caller no longer needs to aggregate log lines unless desired the log handler will be used as default source.

Requested in #641.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -25,6 +25,7 @@ import logging
 from optparse import OptionParser
 import cmdln
 from collections import namedtuple
+from osclib.comments import CommentAPI
 from osclib.memoize import memoize
 import signal
 import datetime
@@ -51,6 +52,8 @@ class ReviewBot(object):
     DEFAULT_REVIEW_MESSAGES = { 'accepted' : 'ok', 'declined': 'review failed' }
     REVIEW_CHOICES = ('normal', 'no', 'accept', 'accept-onpass', 'fallback-onfail', 'fallback-always')
 
+    COMMENT_MARKER_REGEX = re.compile(r'<!-- (?P<bot>[^ ]+) state=(?P<state>[^ ]+)(?: result=(?P<result>[^ ]+))? -->')
+
     # map of default config entries
     config_defaults = {
             # list of tuples (prefix, apiurl, submitrequestprefix)
@@ -73,6 +76,7 @@ class ReviewBot(object):
         self._review_mode = 'normal'
         self.fallback_user = None
         self.fallback_group = None
+        self.comment_api = CommentAPI(self.apiurl)
 
         self.load_config()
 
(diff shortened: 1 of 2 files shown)


ReviewBot.py
```diff
@@ -375,6 +379,63 @@ def set_request_ids_project(self, project, typename):
...
+        self.comment_handler = CommentFromLogHandler(level)
+        self.logger.addHandler(self.comment_handler)
+
+    def comment_handler_remove(self):
+        self.logger.removeHandler(self.comment_handler)
+
+    def comment_find(self, request=None, state=None, result=None):
+        """Return previous comments by current bot and matching criteria."""
+        # Case-insensitive for backwards compatibility.
+        bot = self.__class__.__name__.lower()
```
lnussel: maybe refactor to constructor? it's repeated below but without lower()

jberry-suse: That was intentional since the camelCasing used in class names makes them easier to read if the comment is ever looked at. Extremely minor, but the case here is `lower()` just like the match group to ensure backwards compatibility with old-style `leaper` as noted in original description. If desired I can make the new code print lower case as well.

lnussel: ic. I'll merge but I think there should be a follow up commit to have something like self.botname, in case some bot (or instance of one) needs to override this for whatever reason.

jberry-suse: Created #692.

PR #685, opened 2017-02-10
ReviewBot: use super().check_source_submission() in subclasses.

jberry-suse:
No need to duplicate the log message in multiple places.

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -135,7 +135,7 @@ class Leaper(ReviewBot.ReviewBot):
                 }
 
     def check_source_submission(self, src_project, src_package, src_rev, target_project, target_package):
-        self.logger.info("%s/%s@%s -> %s/%s"%(src_project, src_package, src_rev, target_project, target_package))
+        super(Leaper, self).check_source_submission(src_project, src_package, src_rev, target_project, target_package)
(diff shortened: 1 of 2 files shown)


leaper.py
```diff
@@ -135,7 +135,7 @@ def rdiff_link(self, src_project, src_package, src_rev, target_project, target_p
                 }
 
     def check_source_submission(self, src_project, src_package, src_rev, target_project, target_package):
-        self.logger.info("%s/%s@%s -> %s/%s"%(src_project, src_package, src_rev, target_project, target_package))
+        super(Leaper, self).check_source_submission(src_project, src_package, src_rev, target_project, target_package)
```
lnussel: just curious, is there an elegant way in python to avoid hardcoding the name of the class?

aplanas: In python 3 this parameter is not there anymore, and in python 2 this is the expected way : (

okurz: IMHO in python3 you can just leave out the argument. For python2 compatibility what can work:

```python
super(type(self), self).…
```
-> https://rhettinger.wordpress.com/2011/05/26/super-considered-super/#comment-32

jberry-suse: Yeah, I looked it up a while back since it seemed rather ugly, but came to the same conclusion although type(self) is arguably more flexible.

PR #687, opened 2017-02-11
Do not craft encoded URL with a light heart

coolo:


lnussel:
test suite fails

lnussel:
@jberry-suse could you take a look at this please? The fix is valid, just needs test suite adjustment I suppose

jberry-suse:
@coolo trade for [obs_factory](/openSUSE/obs_factory/pulls/jberry-suse) reviews.

PR #693, opened 2017-02-15
ReviewBot & leaper: provide deduplicate method for leaper comment (and other fixes)

jberry-suse:
- **ReviewBot & leaper: provide deduplicate method for leaper comment.**
- leaper: utilize ReviewBot check_action__default() message.
- ReviewBot: check_action__default() should return None not set ret.
- leaper: drop extra definition of check_action__default().
- ReviewBot: drop duplicate import namedtuple.

This resolves the duplicate line in comment on unhandled request types (like sr#457061).

Before:
```
unhandled request type delete

unhandled request type delete

request needs review by release management
```

After:
```
unhandled request type delete

request needs review by release management
```

I debated making the maintbot check conditional or similar, but seems like this can occur in a variety of ways and even more in the future. If maintbot ends up supporting different types of requests than leaper the message may be appropriate, etc. As such I implemented deduplicate via `list(OrderedDict.fromkeys(self.comment_handler.lines))`.

Additionally, there are two definitions of `check_action__default()` one after the other in `leaper.py`. The second one was added later and is the current behavior so I dropped the older one.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -30,7 +30,6 @@ from osclib.comments import CommentAPI
 from osclib.memoize import memoize
 import signal
 import datetime
-from collections import namedtuple
 
 try:
     from xml.etree import cElementTree as ET
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -25,6 +25,7 @@ import logging
 from optparse import OptionParser
 import cmdln
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


lnussel:
might be also worth rethinking how leaper uses other bots (or pieces of them). Maybe we can find a smarter way.

PR #694, opened 2017-02-16
leaper: notify submit request authors of automatic submission.

jberry-suse:
It is a regular occurrence for submitters to send simultaneous requests to Leap and Factory. Many times the requests may also have followup work before making it into Factory which then results in another set of requests sent to Leap as well. Given that the requests will not be accepted to Leap until they are accepted into Factory all of this seems rather unnecessary and burdensome to all involved. I mentioned to a couple submitters that leaper would automatically create requests for packages sourced from Factory once accepted, but it generally does not seem well known.

Seems like it would be a good idea to highlight this handy feature to reduce workload on everyone involved.

This branch includes #692 so please wait until that is accepted and I will rebase.

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -61,6 +61,7 @@ class Leaper(ReviewBot.ReviewBot):
         self.must_approve_maintenance_updates = False
         self.needs_check_source = False
         self.check_source_group = None
+        self.automatic_submission = False
 
         # project => package list
         self.packages = {}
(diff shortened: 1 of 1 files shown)


jberry-suse:
It occurred to be that adding a direct reference to creator `@creator: ` at the front of message might be handy to ensure they see the comment.

jberry-suse:
Rebased and included `@creator`.

PR #696, opened 2017-02-16
Load the skip pkgs list from OBS instead

nilxam:


fcc_submitter.py
```diff
@@ -153,108 +153,6 @@ def __init__(self, from_prj, to_prj, submit_limit):
...
-                'rubygem-colorator-0',
-                'rubygem-listen-3_0',
-                'rubygem-rouge-1',
-                'rubygem-json-schema-2_2',
-                'computertemp',
-                'libtiger',
-                ]
         self.check_later = [
                 'tulip',
                 'khunphan',
```
lnussel: another hardcoded package list :)

nilxam: hehe, check_later is used nowhere in this tool, just remind myself, but can be add to skip list too since the maintainer should submit them when they think it's ready.

PR #697, opened 2017-02-16
Gcc7 staging also are a valid project to generate packagelist

nilxam:


create_test_dvds.sh
```diff
@@ -158,7 +158,7 @@ function start_creating() {
                     fi
                 fi
 
-                if [[ $prj =~ ^openSUSE.+:[A-Z]$ ]]; then
+                if [[ $prj =~ ^openSUSE.+:[A-Z]$ ]] || [[ $prj =~ ^openSUSE.+:Gcc[0-9]$ ]]; then
```
DimStar77: should we also handle :PIE ?

DimStar77:
LGTM

PR #700, opened 2017-02-17
Add kGraft24 and kGraft0j, remove now unsupported SUTs

fschilling:
* Add kGraft24
* Add kGraft0j
* Remove kgraft-guests which served kernels that are now out of live-
  patching support


plusky:
Merged and deployed.

PR #702, opened 2017-03-01
Select automation: merge and strategies

jberry-suse:
The followup to the `request_splitter` has arrived! This is much closer to what I had envisioned in terms of trying to automate the first select done to requests. I do not expect this to be 100%, but neither is the current manual selection process. I also do not expect that my initial set of strategies will cover all the cases needed by Factory, but I imagine @DimStar77 and @nilxam will have more thoughts there. The strategies should be simple to add and tweak. With the feedback from staging metrics it should be clear if automated changes are having a positive or negative effect.

The primary benefits are as follows:

- removes another human task!
- reduces overall workflow time

The idea being that if the script can be roughly as good as manual staging, but run 24/7 then the build/test process can complete while folks are sleeping or not working rather than that being the first thing done when starting work. Cuts down on the idle time when requests are in backlog and there are available stagings. Also blocking reviews that are accepted at odd times can have the staging automatically cleared and re-staged all while humans are not looking. As a side benefit humans do not have to watch the slow select commands run in mass!

Some examples during the testing period.

- staged a bunch of Leap requests over the weekend which were accepted on Monday, rather than Monday being spend building and testing
- staged `kernel-source` by itself since it matched the `special` strategy

There are a couple items I still want to add to this, but given this has been running since the 18th in this state and the 10th in a more primitive state I am fairly confident that this performs correctly. Any future additions can be made in future requests.

An overview of how I envision this fitting into a more automated process is as follows.

- `accept` automation which I understand @DimStar77 has some prototype scripts
- `freeze` automation which should have similar conditions as accept
- `adi` basically already automated, but could make use of the merge logic here and more complex grouping to merge SRs submitted after initial set is staged
- `select` automation provided by this PR with a few additions

I have been running the last two as a systemd service on Leap 42.3 for the last week and a half or so with good results. The accept being done by those with permission and freeze manually as well.

New flags:

- `--merge` propose merge where applicable and store details to allow future merges
- `--try-strategies` apply strategies and keep any with desireable outcome
- `--strategy` apply a specific strategy

New strategies:

- `none` fallback which just group any remaining packages together
- `requests` list of requests IDs or package names
- `custom` set of filters and grouping options already provided
- `devel` greater than 7 packages from one devel project
- `special` any packages contained by a special package list

See the bottom of `request_splitter.py` to see how simply these strategies can be defined.

Those looking at the Leap stagings may have noticed the `splitter_info` (previously `splitter`) addition to the staging description. This information is set when the initial requests are added to any empty staging and used to merge later requests based on a set of conditions (primarily that the staging is still building and not past 80% complete). The `accept` command has been altered to clear everything in the description except leaving an empty `requests` list to maintain the current logic. Only stagings meeting the conditions and containing this information will be merged when using `--merge` flag. Meaning any manually setup stagings or those that have had the `splitter_info` removed will not be disturbed. Additionally, this process continues to ignore "ignored" requests which should provide a good set of tools to avoid the automation fighting manual changes.

Some additional bits that could presumably be automated are `unselect`ing obsolete requests and replacing superseded requests when the same merge conditions are met.

The primary features need to be added to improve the robustness for Factory (besides additional strategies) are as follows.

- time delayed staging to allow for patterns to develop and be observed
- maximum number of requests matched by a strategy to avoid grouping too many in a single staging

The idea behind time delay is to allow for organic submissions by maintainers without having to worry about the bot schedule making the initial selections. For example if KDE is submitting an update for all applications and only the first couple requests have been made when the bot runs it has no way to see that a full kde update is occurring which should be staged together. If the algorithm requires at least one request matched by a strategy to be over a time threshold (like an hour) then it allows time for the pattern to develop. After waiting an hour there may be several dozen requests at which point that will match a strategy like the `devel` project strategy and they will be staged together. Any requests made after that point will also be merged into that same staging until it no longer meets the criteria.

The maximum number matched is useful for the `none` strategy so that if the backlog is large, but there are no discernible patterns that several hundred requests are not staged together.

Otherwise, in the current state it works quite well for the average load on Leap and presumably SLE. I made fairly granular commits with detailed messages that should make it easier to break down the changes.

The command run by my systemd templated service is:

```bash
osc staging -p "%i" adi --by-develproject && osc staging -p "%i" select --non-interactive --merge --try-strategies
```


jberry-suse:
There is a fair bit of nuance to ensure the that strategy priorities are observed and that requests part of desirable strategy that cannot be staged at the moment are not merged into a less desirable staging (like a none strategy staging). Instead they are internally marked as `postponed`. Additionally, once a strategy is staged the additional constraint of the staging being bootstrapped or not must be observed. Much of this is documented in the commit messages for the particular code or comments. The rough flow for how merging and desirable strategies are followed is contained in `osc-staging.py`.

```python
if opts.merge:
    splitter.merge()
if opts.try_strategies:
    splitter.strategies_try()
if len(requests) > 0:
    splitter.strategy_do('requests', requests=requests)
if opts.strategy:
    splitter.strategy_do(opts.strategy)
elif opts.filter_by or opts.group_by:
    kwargs = {}
    if opts.filter_by:
        kwargs['filters'] = opts.filter_by
    if opts.group_by:
        kwargs['groups'] = opts.group_by
    splitter.strategy_do('custom', **kwargs)
else:
    if opts.merge:
        # Merge any none strategies before final none strategy.
        splitter.merge(strategy_none=True)
    splitter.strategy_do('none')
    splitter.strategy_do_non_bootstrapped('none')
```

A number of small tweaks and improvements were also made that are not necessary for the primary features, but are related and thus included in the request and broken into separate commits.

lnussel:
amazing. I need time to read and understand what you are proposing here :)

PR #704, opened 2017-03-02
osc-staging: provide --cleanup option for unignore.

jberry-suse:
This would make sense to run under some bot credentials to auto-cleanup. I will go ahead and run again Leap 42.3 for the time being. Since the ignored_requests file is version controlled messages and such can always be recovered if some sort of mistake is made.

Here is the output for Factory:
```
Removing 459341 which was superseded 5 days ago                                                                                            
Removing 451216 which was revoked 33 days ago                                                                                              
Removing 448595 which was superseded 45 days ago
Removing 454181 which was superseded 17 days ago
Removing 454246 which was superseded 15 days ago
Removing 459240 which was superseded 5 days ago
Removing 448102 which was revoked 33 days ago
Removing 445239 which was superseded 29 days ago
Removing 445240 which was superseded 28 days ago
Unignoring 9 requests
```

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -93,6 +93,7 @@ def _full_project_name(self, project):
               help='print the plugin version')
 @cmdln.option('--no-freeze', dest='no_freeze', action='store_true',
               help='force the select command ignoring the time from the last freeze')
+@cmdln.option('--cleanup', action='store_true', help='cleanup after completing operation')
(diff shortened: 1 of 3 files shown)


DimStar77:
Feature sounds good

jberry-suse:
Alternatively, I could change it to look for: revoked, accepted, or superseded (final states) with no time delay. Possibly better, but was not sure if declined and not revoked issues would linger. That is likely a separate issue.

PR #714, opened 2017-03-08
osc-staging: provide rebuild command.

jberry-suse:
Often packages will fail to build after becoming stuck or other false negative cases and need to have a rebuild triggered. The process can be tedious if several packages failed in various stagings.

For instance today, there are three Factory stagings that all had stuck packages, but otherwise built fine (E, F, and H). This is another part of the automation flow that I plan to add logic to try and be smart so it can be run automatically, but this is useful as it stands.

I have been keeping a list of stuck and flaky builds that could be used as a basis for the tool. Alternatively, plugging into statistics data should with simple flaky detection pattern should provide fairly decent automation to avoid stuck stagings that just need a package or to rebuilt.

For example the three stuck packages in Factory had the typical:

```
[31113s] qemu-system-x86_64: terminating on signal 15 from pid 21485


Job seems to be stuck here, killed. (after 28800 seconds of inactivity)
```

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -44,6 +44,7 @@ from osclib.stagingapi import StagingAPI
 from osclib.cache import Cache
 from osclib.unselect_command import UnselectCommand
 from osclib.repair_command import RepairCommand
+from osclib.rebuild_command import RebuildCommand
 from osclib.request_splitter import RequestSplitter
 
 OSC_STAGING_VERSION = '0.0.1'
@@ -205,6 +206,8 @@ def do_staging(self, subcmd, opts, *args):
 
     "unlock" will remove the staging lock in case it gets stuck
 
+    "rebuild" will rebuild broken packages in the given stagings or all
+
     Usage:
         osc staging accept [--force] [--no-cleanup] [LETTER...]
         osc staging acheck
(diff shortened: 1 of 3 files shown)


lnussel:
maybe we should name it rebuild-broken? 'rebuild' suggests unconditional rebuild

nilxam:
sorry, I lost the point why not just use ```osc rebuildpac -f```?

if it's about providing new feature to the automation flow, it should stop try to rebuild the package if that package failed to build **few times** ie. it's really broken.

jberry-suse:
@lnussel Could do that, I was trying to be consistent with `osc rebuild` although it would be with `-f`. Probably makes sense to rename.

@nilxam Yes, it is not much different from `osc rebuild -f`. The one helpful difference is that it does subprojects so it includes `:DVD` on letter stagings. As noted I planned to plug it into the build stats and limits on retry attempts, but figured it was useful first step. Since it has the list of broken packages it can perform the necessary logic, but need to read build history and preferably statistics indicating flakiness. If we want to hold off until then like I was originally planning that's fine. I just found it helpful to run `osc staging rebuild E F H` on Factory as I do on Leap regularly.

Issue #713, opened 2017-03-08
remind devel projects of pending submissions or reviews


lnussel:
Devel project maintainers sometimes forget about package submissions to devel projects or submissions that have the devel project/package as reviewer, such as Factory->Leap submissions. So it could be useful to send mail reminders to project maintainers after some period of inactivity.

Maybe the the devel-project-list.py tool could be extended for it. 
Something like

	$ ./devel-project tool list
	Apache
	Apache:Modules
	Apache:Shibboleth
	...
	$ ./devel-project tool requests
	477730 Base:System/efivar home:rwill:Factory/efivar
	...
	$ ./devel-project tool reviews
	459015 mozilla:Factory/kmozillahelper openSUSE:Leap:42.3/kmozillahelper
	...
	$ ./devel-project tool remind --requests --reviews --timeout=3d
	DEBUG: reminding projectmainers wrosenauer, pcerny about 459015 mozilla:Factory/kmozillahelper
	...

PR #718, opened 2017-03-10
Devel project tool: review reminders

jberry-suse:
See #713 for original request. Some additional refactoring has also been done.

- provide --remind and --repeat-age options.
- CommentAPI: provide comment_find() and add_marker() from ReviewBot.
- print age of request and filter via --min-age.
- add reviews subcommand.
- add requests subcommand.
- import http_GET and makeurl.
- provide devel_projects_load().
- check if list changed before saving.
- refactor structure to allow subcommands.
- rename from devel-project-list.

Usage:

```
./devel-project.py list [--write]
./devel-project.py requests [--min-age=AGE] [--repeat-age=AGE] [--remind]
./devel-project.py reviews [--min-age=AGE] [--repeat-age=AGE] [--remind]
```

Instead of separate remind command which would end up essentially running the other two commands with extra processing it seemed to make sense to just provide a `--remind` option. I also added `--repeat-age` option which will allow repeat reminders sent out after the original reminder comment is that many days old.

I refactored the `ReviewBot` `comment_find()` and add marker code into `CommentAPI` so it can be used for this application (from Leaper -> ReviewBot -> CommentAPI :)). I assume the "send mail" reminders makes sense to use comments rather than directly sending email. I ran a number of `leaper` reviews that seem to find the previous comment where appropriate.

This also works for new package submissions.

I also took the liberty of generating review reminders for a couple >20 day year old requests.

- https://build.opensuse.org/request/show/456754
- https://build.opensuse.org/request/show/457685

Presumably this script should run as one of the existing bot users like the `factory-comment-bot `.


jberry-suse:
If one looks at the individual commits you can see a more accurate diff since github seems to summarize the post-move changes as just one new file. :(

jberry-suse:
The bot running `./devel-project-list.py --write` will need to be updated to run `./devel-project.py list --write` as this now reads from that file so even more important to have it kept up-to-date.

lnussel:
it would make sense to not repeat the reminder as comment but rather send an email. If people for whatever reason have misconfigured the notification settings in obs or filter them, a direct mail can help.

lnussel:
(see build-fail-reminder.py)

PR #747, opened 2017-03-14
Leaper adjustments

lnussel:


leaper.py
```diff
@@ -112,6 +112,16 @@ def rdiff_link(self, src_project, src_package, src_rev, target_project, target_p
                 'target_package': target_package,
                 }
 
+    def _check_same_origin(self, origin, project):
+
+        if origin == 'FORK':
+            return True
+
+        if origin.startswith('Devel;'):
+            (dummy, origin, dummy) = origin.split(';')
```
jberry-suse: nit: does it make sense to use `_` instead of `dummy`?

lnussel: I'm obviously not a python native speaker :)

PR #748, opened 2017-03-14
Fix module search path in osc-staging

lnussel:


osc-staging.py
```diff
@@ -27,8 +27,9 @@
 from osc import oscerr
 
 # Expand sys.path to search modules inside the pluging directory
-_plugin_dir = os.path.expanduser('~/.osc-plugins')
-sys.path.append(_plugin_dir)
+# FIXME: osc should do that
+PLUGINDIR = os.path.dirname(os.path.realpath(__file__.replace('.pyc', '.py')))
```
jberry-suse: Is there a reason for `__file__.replace('.pyc', '.py')` as opposed to just `__file__` since the file ends up being chopped off by `dirname()` anyway. Presumably `__file__` always exists and as such will provide a proper result from `realpath()`.

lnussel: no idea, taken from osc-check_repo.py

lnussel:
Note: real fix will be in next osc: https://github.com/openSUSE/osc/pull/279
So we can remove those hacks as soon as that version is widely available.

PR #750, opened 2017-03-15
unselect: cleanup stagings left empty after request removal.

### LINKED ISSUE #735: unselect: clear splitter_info when last request is removed
In addition to `accept` the `unselect` command should remove `splitter_info` when the last request is removed since the staging will then be considered empty and a new strategy may be use. If another strategy is applied the `splitter_info` will be overridden, but if a manual `select` is performed no change will be made which is misleading and may lead to undesired merges.

jberry-suse:
- **unselect: cleanup stagings left empty after request removal.**
- accept: refactor staging cleanup into stagingapi.

Resolves #735 in addition to a missing cleanup case. I have noticed this in occurring in both Factory and Leap and this explains why.

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -69,9 +69,6 @@ class AcceptCommand(object):
                                  message='Accept to %s' % self.api.project)
             self.create_new_links(self.api.project, req['package'], oldspecs)
 
-        # Clear pseudometa since it no longer represents the staging.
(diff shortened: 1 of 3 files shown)


jberry-suse:
Tested on one in Leap, by commenting out the unselect code above affect_projects loop and manually adding B.

Before:
image

After:
image

PR #751, opened 2017-03-15
special strategy: drop python*.

jberry-suse:
Per IRC discussion with @DimStar77 it seems these are only isolated when an issue arises. Is there any other packages you can think of that should be labeled special that I can add while I'm at it?

### DIFF
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -386,8 +386,6 @@ class StrategySpecial(StrategyNone):
         'gcc7',
         'glibc',
         'kernel-source',
-        'python2',
-        'python3',
         'util-linux',
     ]
 


lnussel:
maybe the package lists could be stored separately from the code? Ie either as attribute somewhere or as yet another file in the dashboard?

jberry-suse:
There are a number of options already available and more that will likely be needed. With some of the discussion around Leap specific strategies it seems to make sense to have some project specific configuration it is just a matter of where it is placed.

I am not opposed to a select proposal config in dashboard with a variety of options that can be expanded and enable/disable strategies, etc. Probably still makes sense to have a default list in the code, although perhaps at some stage where it loads config.

PR #758, opened 2017-03-15
Delete requests should be handled by repo checker

DimStar77:
Repo-checker at least has some code to validate if it's safe to delete
a package and gives us the information what else we would break with it.

https://progress.opensuse.org/issues/17746


nilxam:
as verified. LGTM

PR #761, opened 2017-03-16
Prepare repochecker for Leap

lnussel:
Incomplete stil, arch selection needs to be smarter


lnussel:
To avoid hardcoding architectures, an approach similar to abichecker could be used to determine the required repos based on the target:
https://github.com/openSUSE/osc-plugin-factory/blob/master/abichecker/abichecker.py#L908

PR #744, opened 2017-03-14
Add setprio command

lnussel:
raise priority for requests in review

Issue #724, opened 2017-03-10
Reminder: utilize obs_factory staging status aggregation once deployed


jberry-suse:
See openSUSE/obs_factory#69.

jberry-suse:
Pull request accepted and [asked in build-service list how update process works](https://lists.opensuse.org/opensuse-buildservice/2017-03/msg00085.html).

PR #762, opened 2017-03-16
stagingapi: allow project_status() to utilize aggregate backend call.

jberry-suse:
Now that deployed, #724 is completed by this requests. I also refactored the function to allow both the aggregate to be returned direct or just project specific. #744 could then be rebased on this to utilize the improved `stagingapi.project_status()`.

Summary:

- check: utilize project_status().
- adi: utilize extract_staging_short() and project_status().
- **stagingapi: allow project_status() to utilize aggregate backend call.**
- stagingapi: allow staging to be None in project_status() to return all.

For check command it is still more efficient to do a project specific call so it uses the `allow_aggregate` option to force single project query when a project is given.

Summary of testing ([full log](https://gist.github.com/jberry-suse/e12444b4da49c84b00f2712cb9bded5d)):

- Leap `4.5` times faster (72 -> 16 seconds): 74 adi stagings
- Factory `6.5` times faster (236 -> 36 seconds): 250 adi stagings


jberry-suse:
Will likely speed up other commands, like select that ends up checking status repeatedly.

jberry-suse:
Could bump up the length it stores the status query, but since it cannot be determined when it expires I left it at 60 seconds. The one caveat being that memoize appears to count the time from the start of call rather than the end, which means the time it takes for the remote call to return is counted against the ttl. As such if the call is around 60 seconds or adi command is accepting stagings it will have to repeat it. It might be worth caching it longer, but a question of wether it will cause issues elsewhere. From what I understand it should not. Alternatively, adi could be modified to walk the summary list directly rather than calling for each adi.

jberry-suse:
After a bit more thought I think we should stick with just per session.

```diff
diff --git a/osclib/stagingapi.py b/osclib/stagingapi.py
index 687187a..c2e993b 100644
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -816,7 +816,7 @@ class StagingAPI(object):
 
         return log.getvalue()
 
-    @memoize(ttl=60, session=True)
+    @memoize(session=True)
     def project_status(self, staging=None, allow_aggregate=True):
         path = ('project', 'staging_projects', self.project)
         if staging:
```

jberry-suse:
After some more thought and observation I changed the default for `aggregate` from `True` to `False` and explicitly called with `True` in adi code.

jberry-suse:
I went ahead and added:
- stagingapi: refactor _get_staged_requests() to use project_status().

Which is the other most expensive call!! So now `list --supersede` and `select` will drop a ton of queries. Select even more so since it would repeat the same queries as they expired through long selects.

[gist log](https://gist.github.com/jberry-suse/784ae5c259daa92b99486fd24b35a927)

Similar savings as before:

- Leap: 67 -> 15 seconds

jberry-suse:
Test fixture was missing information for `openSUSE:Factory:Staging:U` and thus failed when the refactored code tries to loop over all stagings. Fixed the test fixture data to include an empty staging U.

jberry-suse:
The overall runtime of the 5 commands I run in service dropped from >4 minutes to about 1.5 minutes. :)

jberry-suse:
Works fine against IBS as I would expect since it does not have adi stagings.

PR #776, opened 2017-03-24
Fix wrong call of cstaging

nilxam:


fcrozat:
I can confirm the fix is working

PR #778, opened 2017-03-26
Add legal-auto bot talking to the legaldb API

coolo:


legal-auto.py
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/python
```
okurz: `#!/usr/bin/env python` ?

legal-auto.py
```diff
@@ -0,0 +1,189 @@
...
+                                     reqid = req.reqid, newstate = state,
+                                     by_group=self.review_group,
+                                     by_user=self.review_user, message=self.message)
+        self.delete_from_db(req.reqid)
+
+class CommandLineInterface(ReviewBot.CommandLineInterface):
+
+    def __init__(self, *args, **kwargs):
+        ReviewBot.CommandLineInterface.__init__(self, args, kwargs)
+        self.clazz = LegalAuto
```
okurz: clazz?

legal-auto.py
coolo: no

legal-auto.py
coolo: yes

coolo:
Dare to explain why I would care about pep8 (whatever that is?) for my program?

okurz:
http://lmgtfy.com/?q=pep8

legal-auto.py
```diff
@@ -0,0 +1,189 @@
...
+        if self.dryrun:
+            self.logger.debug("dry setting %s to %s with %s"%(req.reqid, state, self.message))
+            return
+
+        self.logger.debug("setting %s to %s"%(req.reqid, state))
+        osc.core.change_review_state(apiurl = self.apiurl,
+                                     reqid = req.reqid, newstate = state,
+                                     by_group=self.review_group,
+                                     by_user=self.review_user, message=self.message)
+        self.delete_from_db(req.reqid)
```
jberry-suse: This appears to be the only unique line short of dropping functionality from the original. Perhaps call super() and check `self.dryrun` before `delete_from_db()`?

jberry-suse: This appears to be the only unique line short of dropping functionality from the original. Perhaps call super() and check `self.dryrun` before `delete_from_db()`? The message should be settable via the standard `self.review_messages`.

legal-auto.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __init__(self, *args, **kwargs):
+        ReviewBot.ReviewBot.__init__(self, *args, **kwargs)
+
+        self.do_comments = True
+        self.legaldb = None
+        self.commentapi = CommentAPI(self.apiurl)
+        self.apinick = None
+        self.message = None
+        if self.apiurl == 'https://api.suse.de':
```
jberry-suse: nit: `if self.ibs:`

legal-auto.py
coolo: this is also about the message.

legal-auto.py
coolo: yeah, but that is pretty non-symmetric then - as there is no self.obs

legal-auto.py
mimi1vx: pretty bad idea use `env`

PR #782, opened 2017-03-29
stagingapi: invalidate get_open_requests() memoize cache after supersede.

### LINKED ISSUE #780: Superseded request is still show on the list until next run
With refactored supersede logic, `osc staging list --supersede` did superseding request correctly, however the superseded request will still show on the list. Note: next run it disappears.

jberry-suse:
Resolve #780 caused by caching the `get_open_requests()` call which is invoked before supersede and after. As such superseded requests are not reflected in list until the next run.


nilxam:
verified :+1:

PR #785, opened 2017-03-29
Have TTM work through openqa comments

coolo:
Instead of having the known failures in a global file, have it local to the TTM
but allow users to interact with it through openqa comments.

Work flow:
  - openQA tests and finds failures
  - reviewers tag them with (support from automatic carry over)
  - if they are in the ignore list, TTM will not do anything but ignore them
  - if not, TTM will add label:unkown_failure to mark the issue blocking
    a release
  - the release manager adds @ttm ignore as comment for the TTM to
    add the issue to its white list.

Only if all failures are labeled and marked as ignored, a snapshot is released

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -19,6 +19,8 @@ import signal
 import time
 
 from xml.etree import cElementTree as ET
+from pprint import pprint
+from openqa_client.client import OpenQA_Client
 
 import osc
 
@@ -31,6 +33,7 @@ from osclib.conf import Config
 from osclib.stagingapi import StagingAPI
 from osc.core import makeurl
 
+ISSUE_FILE = 'issues_to_ignore'
 
 # QA Results
 QA_INPROGRESS = 1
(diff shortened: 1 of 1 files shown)


coolo:
the openqa api key is put into the packagelists vm in the $HOME of ttm

okurz:
don't want to review the code as you mangled style changes with implementation changes but the idea and workflow LGTM

coolo:
programming python for me always is a style change

coolo:
workflow has been ACKed by Ludwig and Max, so merging and deploying

totest-manager.py
```diff
@@ -162,23 +173,49 @@ def overall_result(self, snapshot):
...
+                    if comment['userName'] == 'ttm' and comment['text'] == 'label:unknown_failure':
+                        labeled = comment['id']
+                    if comment['text'].find('@ttm ignore') >= 0:
+                        to_ignore = True
+                ignored = True
+                for ref in refs:
+                    if not ref in self.issues_to_ignore:
+                        if to_ignore:
+                            self.issues_to_ignore.append(ref)
+                            with open(ISSUE_FILE, 'a') as f:
```
lnussel: this is racy. there are several instances of ttm running. So you need locking or (better) use a separate file per project

totest-manager.py
```diff
@@ -162,23 +173,49 @@ def overall_result(self, snapshot):
...
+                #pprint(comments)
+                labeled = 0
+                to_ignore = False
+                for comment in comments:
+                    for ref in comment['bugrefs']:
+                        refs.add(str(ref))
+                    if comment['userName'] == 'ttm' and comment['text'] == 'label:unknown_failure':
+                        labeled = comment['id']
+                    if comment['text'].find('@ttm ignore') >= 0:
+                        to_ignore = True
```
lnussel: a feature to limit this to certain users is missing

totest-manager.py
```diff
@@ -31,6 +33,7 @@
 from osclib.stagingapi import StagingAPI
 from osc.core import makeurl
 
+ISSUE_FILE = 'issues_to_ignore'
```
lnussel: shouldn't be global. see below.

okurz:
> bad style. reformats and adds new features :(

maybe you will listen to lnussel more than me ;-)

PR #787, opened 2017-03-30
Misc improvements

nilxam:


totest-manager.py
```diff
@@ -194,14 +194,18 @@ def overall_result(self, snapshot):
...
+                    for ref in refs:
+                        if ref not in self.issues_to_ignore:
+                            if to_ignore:
+                                self.issues_to_ignore.append(ref)
+                                with open(ISSUE_FILE, 'a') as f:
+                                    f.write("%s\n" % ref)
+                            else:
+                                ignored = False
+                else:
+                    ignored = False
```
coolo: ignored = len(refs) > 0

was too small of a change? :)

totest-manager.py
```diff
@@ -52,8 +52,8 @@ def __init__(self, project, dryrun=False):
             osc.conf.config['apiurl'], project='openSUSE:%s' % project)
         self.openqa = OpenQA_Client(server='https://openqa.opensuse.org')
         self.issues_to_ignore = []
-        if os.path.isfile(ISSUE_FILE):
-            with open(ISSUE_FILE, 'r') as f:
+        if os.path.isfile("{}_{}".format(self.project, ISSUE_FILE)):
+            with open("{}_{}".format(self.project, ISSUE_FILE), 'r') as f:
```
coolo: then use a variable for this instead of doing this 3 times

totest-manager.py
nilxam: I like this style but fine to change it to a gorgeous one :P

PR #789, opened 2017-04-06
Add devel:openQA:tested to whitelist for auto-submit

okurz:


nilxam:
you really want to use a project name has "testing" included to be a valid devel project? :)

okurz:
yes, because it should be more stable than "devel:openQA" itself. packages are copied to "testing", tested in openQA-in-openQA tests, then submitted

okurz:
also, Dimstar agreed with the name :-)

okurz:
'testing' -> 'tested', as discussed in jangouts call

PR #797, opened 2017-04-11
Hack to skip haskell submissions

lnussel:
It's just too much and needs to be submitted manually

### DIFF
--- a/update_crawler.py
+++ b/update_crawler.py
@@ -37,6 +37,8 @@ import re
 from urllib import quote_plus
 
 from osclib.memoize import memoize
+from osclib.conf import Config
+from osclib.stagingapi import StagingAPI
 
 OPENSUSE = 'openSUSE:Leap:42.3'
 FACTORY = 'openSUSE:Factory'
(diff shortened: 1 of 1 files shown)


update_crawler.py
```diff
@@ -307,6 +310,13 @@ def crawl(self, packages):
         sources = self.get_source_infos(self.from_prj, packages)
         self.update_targets(targets, sources)
 
+    def get_devel_project(self, package):
```
jberry-suse: drop if `StagingAPI` used

update_crawler.py
```diff
@@ -270,8 +270,11 @@ def update_targets(self, targets, sources):
             else:
                 targetinfo = targets[package]
 
-                #if package != 'openssl':
-                #    continue
+                # XXX: make more generic :-)
+                devel_prj, _ = self.get_devel_project(package)
```
jberry-suse: perhaps `StagingAPI.get_devel_project(FACTORY, package)`

lnussel:
I don't understand the travis failure, works locally

jberry-suse:
Failure appears to the same one in all test results #807.

Issue #800, opened 2017-04-11
devel-project: skip reminder comment made by different user that cannot be removed


jberry-suse:
Since I had run the script as `jberry_factory` the `staging-bot` user cannot replace some of my existing comments and crashes. Gracefully skipping removal of these comments and still creating a new reminder would seem preferable.

Alternatively, granting the `staging-bot` user a higher privilege to allow for deletion of such comments might be a good interim solution that would allow for the transition.

PR #801, opened 2017-04-11
devel-project: gracefully skip when removal of previous comment fails.

jberry-suse:
Given that previous comment is used as a basis for determining the time since last reminder if it cannot be removed a new reminder should not be created.

Provides workaround for #800. This code should be irrelevant after all previous reminder comments are either removed or no longer relevant. This also allows for others to run the reminder script without causing each other to crash.

I'll go ahead and delete my previous comments after seeing a full run using this code.


jberry-suse:
Looks like there were only 2 remaining so I deleted them.

PR #806, opened 2017-04-12
request_splitter: provide and utilize an aged attribute on requests.

### LINKED ISSUE #733: select: add time delay to none strategy for both merge and new staging
In order to allow for submissions to not have to be timed in coordination with the schedule on which the select bot is running there needs to be a time delay to allow for patterns to emerge. Simply increasing the interval at which the bot runs will not work since someone may start sending out a series of requests seconds before the bot runs. At that point the bot would not be able to observe the full series for a pattern.

After quite a bit of thought, it seems the delay only needs to be applied to the `none` strategy. The delay (likely 1 hour) will be required before a strategy may be included by a `none` group either for merging or starting a new staging. Any non-none strategies that match will be allowed as soon as they are detected.

As more strategies are added in may become clear that a more complicated algorithm is needed to allow for larger patterns to develop, but this seems adequate for the time being.

The real world delay as seen by the request will be either:

- `1 hour + remaining time until next bot run`
- `< 1 hour if pattern observed + remaining time until next bot run`

I wrote up a rough prototype that added `@age` to requests in `suppliment()` in the form of a boolean indicating if over the delay period. The `none` strategy then simply adds a filter in xpath looking for `True`. The `@postponed` mechanism will likely need to be tweaked to only postpone when a non-none strategy is matched.

jberry-suse:
- **request_splitter: provide and utilize an aged attribute on requests.**
- request_splitter: use capital for boolean strings in supplement().
- request_splitter: s/suppliment/supplement/().
- stagingapi: get_open_requests() provide query_extra option.
- stagingapi: get_open_requests(): replace query string with dictionary.

Fixes #733.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -387,7 +387,7 @@ def do_staging(self, subcmd, opts, *args):
                     print('--move and --from must be used with explicit staging and request list')
                     return
 
-                open_requests = api.get_open_requests()
+                open_requests = api.get_open_requests({'withhistory': 1})
                 if len(open_requests) == 0:
                     print('No open requests to consider')
                     return
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -54,14 +54,14 @@ class RequestSplitter(object):
     def filter_only(self):
         ret = []
         for request in self.requests:
-            self.suppliment(request)
+            self.supplement(request)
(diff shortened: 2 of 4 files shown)


jberry-suse:
Tests pass locally, but have been failing on travisci the last few commits.

PR #812, opened 2017-04-17
Post comment if the issue exists in the issues file

nilxam:


okurz:
https://progress.opensuse.org/projects/openqav3/wiki/#Pinning-comments-as-group-description could be an alternative by using a comment on the group overview page and update the same comment

coolo:
As mentioned in the call: I think these comments create too much noise in the test overview page. Because you will have a result bubble, a job module, an issue and a marker about the comment. But the job of the release manager is finding the label:unknown_failure labels quickly. And the more noise you have, the harder that job is.

See Olli's comment about an alternative how to get the information into the system.

nilxam:
The goal here is how the release manager can be realized the failed job was ignored when seeing at https://openqa.opensuse.org/tests/387452#comments ,  the result bubble might be a problem, but I suspect pinned the ignored issues on the group overview page is a good approach in this case...

nilxam:
https://github.com/openSUSE/osc-plugin-factory/pull/828 instead

Issue #703, opened 2017-03-01
make factory announcer more generic


lnussel:
the factory announcer needs to be reworked to
- support other architectures such as s390
- support other distributions like Leap
- maybe work on the ftp tree instead of DVD

lnussel:
any news here?

jberry-suse:
Will dig in on Monday.

jberry-suse:
Seems like there are a couple of approaches and some assumptions I should verify before proceeding. To start of with, specific to the environment in which the code runs currently. Assumptions:

- has disk access to read DVD iso directly
- does not read directories containing rpms which code supports
- script is triggered as part of the publish process
  - this likely is not duplicated for s390x, but could be for Leap if not currently

Based on the desired features to support and looking over the commonalities between the output information for the three targets it seems the rpm tree is the only reliable source. The repodata file list cannot be used since it is not provided for Factory s390x. Reading from OBS directly could be a possibility, but would likely be fairly slow for this purpose and is not always in-sync with what is published, but that may be a problem regardless.

I have previous written a dashboard that displays package version across the Factory workflow:

- http://tumbleweed.boombatower.com/
- https://github.com/boombatower/opensuse-tumbleweed-dashboard

The dashboard access information both from HTTP download tree and OBS which is why it is clear when the publish to HTTP is out-of-sync.

Given that the output is a diff of the changelog all current and previous changelogs will need to be available to the script. Unless OBS revisions can be used as a source for this information, but given that revisions may be skipped the previous published revision will be required which may be tricky to obtain.

If the HTTP/FTP/rsync tree is used as source then the script will essentially need a full sync of the tree before it can operate. After that, either keep the previous versions of rpms around or cache the information of interest as is done now (via pickle serialization). The current form of serialization store the entire set of meta information for each package for every published snapshot (ie lots of duplicate information).

The tree sync was something I dealt with when trying to make snapshots of Tumbleweed and can really only be solved properly via rsync, but having an entire copy of the tree seems like a lot of excess if the real tree can be accessed.

I had considered and this may still be possible, diffing the file list and downloading the changed/new rpms at which point they could be diffed against the cached meta data if available. Assuming the snapshots are relatively small the about downloaded would not be overwelhming although likely regularly more than the ISO size.

[...]

PR #826, opened 2017-04-19
factory-package-news-web: infer basedir from url_root.

jberry-suse:
Allows for multiple aliases to host different distro/version changes.

Hopefully last code change towards #703.

Was shooting for the fewest changes to leave the basic flow intact. Multiple aliases can be setup which ensure that only the desired paths can be read for changes without hard-coding in the web app.

```conf
WSGIScriptAlias /snapshot-changes/opensuse/tumbleweed .../factory-package-news-web.py
WSGIScriptAlias /snapshot-changes/opensuse/42.3 .../factory-package-news-web.py
```

Similar config changes are pending in my repo on openqa.

### DIFF
--- a/factory-package-news/factory-package-news-web.py
+++ b/factory-package-news/factory-package-news-web.py
@@ -26,16 +26,21 @@ from flask import make_response
 import re
 import os
 import sys
+from urlparse import urlparse
 
 import re
 digits_re = re.compile('^[0-9]+$')
 
-_dir = "/var/lib/openqa/factory/snapshot-changes"
+BASE_DIR = '/var/lib'
 
 app = Flask(__name__)
 
+def get_dir(url):
+    return os.path.join(BASE_DIR, urlparse(url).path.lstrip('/'))
+
 @app.route('/')
 def list():
+    _dir = get_dir(request.url_root)
     fn = os.path.join(_dir, 'current')
     current = None
     if os.path.exists(fn):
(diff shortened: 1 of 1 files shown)


jberry-suse:
This could also be done by extending the routes to include the information, but then likely need acceptable list or some sort of validation. Could check if dir exists, but this works as well.

factory-package-news/factory-package-news-web.py
```diff
@@ -26,16 +26,21 @@
...
 import re
 digits_re = re.compile('^[0-9]+$')
 
-_dir = "/var/lib/openqa/factory/snapshot-changes"
+BASE_DIR = '/var/lib'
 
 app = Flask(__name__)
 
+def get_dir(url):
+    return os.path.join(BASE_DIR, urlparse(url).path.lstrip('/'))
```
lnussel: this needs some sanitation to prevent directory traversal attacks. Put e.g. a regexp to make sure it's the expected path.

lnussel: I guess our config prevents it. nevertheless makes me nervous

jberry-suse: That was the idea with config, as it should only allow for paths in config. I tried some common things like `../` and what not in paths on apache all behaved as expected. The transformation were resolved by apache and the application never even hit just 404.

jberry-suse: For kicks I could check that the resulting path is is within the expected directory or could split it for the two parts and ensure only letters. Seems like this is unnecessary since the paths have to match literal strings in our config.

PR #829, opened 2017-04-19
check_source: skip staging process if only .changes modifications.

jberry-suse:
- check_source: package_source_parse() drop unused apiurl parameter.
- **check_source: skip staging process if only .changes modifications.**

Fixes https://progress.opensuse.org/issues/18648.
Test request https://build.opensuse.org/request/show/489331.

Skips add repo_checker review and accepts factory-staging review if only .changes modifications.


nilxam:
LGTM

PR #831, opened 2017-04-19
unselect: filter superseded when using --cleanup.

### LINKED ISSUE #830: unselect --cleanup need to check if superseded
Based on the discussion in #823 it seems the dashboard api does not differentiate between superseded and revoked when including requests in obsolete list. As such an edge case can happen between `list --supersede` and `unselect --cleanup` where a superseded request is unselected instead of replaced. Either update dashboard api or check condition in `unselect --cleanup`.

jberry-suse:
Fixes #830.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -921,12 +921,13 @@ class StagingAPI(object):
             tobuild += int(repo['tobuild'])
         return final, tobuild
 
-    def project_status_requests(self, request_type):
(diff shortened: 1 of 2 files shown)


jberry-suse:
Confirmed this skips two obsolete, but superseded requests in Factory adi:119.

- 489089
- 489104

Without filter it tries to unselect them. Obviously, if the `list --supersede` command was run first no problem, but the command should be safe by itself and this closes the window for the edge-case.

PR #834, opened 2017-04-19
supersede: separate command, feedback, and force supersede post source project check

### LINKED ISSUE #814: supersede: create command to skip additional checks to complete Leap workflow
After a request is marked as ignored pending human review there is no easy way to continue the process if the ignored request is desired. A supersede command that simply skips the checks that were added and thus forces the request to supersede the other request in question would be handy to ease the process.

jberry-suse:
- cf3ce7435eb6c57bbce01bf12d840cb2ecccb4b4
  stagingapi: bypass ignored list when supersede with target_requests.

- 08a31563c44d7a1b268397b8fdccabe5c5a6061c
  supersede: provide feedback about what requests were superseded.

- e213385ca4b174f1189edc92712e2f744ce2a22d
  supersede: allow list of request IDs or packages.

- acc95dbb8499c90d4d9cb987489a0a6d4afb22e9
  stagingapi: enhance supersede target_pkgs to skip source project check.

  This creates a facility for forcing a request, that was ignored due to
  the source check, to supersede the existing one.

- 54f502da375843ce35a9d09ada00505b8fb595bf
  osc-staging: break out `list --supersede PACKAGES...` as separate command.

  Maintains `list --supersede`, but places the explicit supersede based on
  a list of packages in a separate command.

I confirmed that the following work as expected:

- `list --supersede`
- `supersede REQUEST_ID`
- `supersede PACKAGE_NAME`
- `adi` (with requests to supersede)

The `list` and `supersede` commands now have output indicating which requests were superseded while the `adi` output remains the same and still indicates which requests were superseded.

Something of which I was not aware is that the `list --supersede` command supersedes requests staged in `adi` stagings. This remains the same, but will be more obvious since messages are printed. Just clarifying that there was no change in behavior.

New features:

- print feedback about what was done
- can supersede explicitly by REQUEST_ID in addition to PACKAGE_NAME
- can force a request that was ignored due to source project check to supersede

For Leap, when a request is ignored and after manual review needs to supersede the other request one can do the following.

```
osc staging supersede IGNORE_REQUEST_ID|PACKAGE_NAME
```

Followup on same basic idea behind #753.
Fixes #814.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -263,7 +263,7 @@ def do_staging(self, subcmd, opts, *args):
         osc staging rebuild [--force] [STAGING...]
         osc staging repair [--cleanup] [REQUEST...]
         osc staging setprio [STAGING...]
-        osc staging supersede [PACKAGE...]
+        osc staging supersede [REQUEST...]
     """
     if opts.version:
         self._print_version()
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -47,6 +47,7 @@ from osclib.unselect_command import UnselectCommand
 from osclib.repair_command import RepairCommand
 from osclib.rebuild_command import RebuildCommand
 from osclib.request_splitter import RequestSplitter
+from osclib.supersede_command import SupersedeCommand
 from osclib.prio_command import PrioCommand
 
 OSC_STAGING_VERSION = '0.0.1'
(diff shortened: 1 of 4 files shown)


jberry-suse:
Example output:
```
request 489393 for python-compizconfig superseded 489089 in openSUSE:Factory:Staging:adi:119
request 489395 for compizconfig-settings-manager superseded 489104 in openSUSE:Factory:Staging:adi:119
```

PR #841, opened 2017-04-21
osc-staging: implement colorization using colorama for adi and list, along with some formatting cleanups.

jberry-suse:
Something I was playing around with as it drastically improves the ability to scan/read the output which is relatively busy. Can try and make colorama not required by mocking it and falling back to simple mock when not available.

Try it out and see if it is helpful.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -23,7 +23,10 @@ import tempfile
 import warnings
 import yaml
 
+import colorama
+
 from osc import cmdln
+from osc import conf
 from osc import oscerr
 
 # Expand sys.path to search modules inside the pluging directory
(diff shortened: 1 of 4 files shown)


jberry-suse:
Example, but obviously the color scheme will likely be different based on your terminal setup.

image
image

DimStar77:
Played around with this locally and the coloring looks reasonable here (and it's easy switched off if not wanted)

👍 from me

jberry-suse:
Refactored a bit to add color everywhere in `adi` and some formatting cleanups while in there. @DimStar77 seemed to agree the sr# prefix was more annoying that useful in the context due to being included in copy/paste when double click to select. Also added `staging.color` option to `~/.oscrc` which allows the colors to be disable outright.

jberry-suse:
Key `adi` text changes are:

- `review:` instead of `missing reviews for` (which provides consistent status column
- `ready:` instead of `is ready` in some cases while others had `ready:`

PR #845, opened 2017-04-25
Handle multiple patchinfos

lnussel:


ReviewBot.py
```diff
@@ -226,7 +226,7 @@ def check_one_request(self, req):
     def check_action_maintenance_incident(self, req, a):
         dst_package = a.src_package
         # Ignoring patchinfo package for checking
-        if a.src_package == 'patchinfo':
+        if pkgname == 'patchinfo' or pkgname.startswith('patchinfo.'):
```
coolo: I think you should extract this part - because then you can call it with a.src_package instead of the non existant pkgname

coolo:
See https://github.com/openSUSE/osc-plugin-factory/pull/857 for followup

Issue #815, opened 2017-04-18
rings: tool similar to cleanup_rings, but to make ring changes in stagings


jberry-suse:
Seems like the regular ring changes should be fairly straight forward to automate. Look for un-resolvable packages and see if they there exists a package already in the project (or potentially in open requests) that would provide what is needed and move it to the same ring.

Could be a manual tool like cleanup or automated if either documents changes or easy to tell.

jberry-suse:
The plan:

- `project_status()` to look for un-resolvable packages (perhaps filter to those related to open requests)
- check source project `fileinfo_ext` to see what packages provide its requirements
  - likely makes sense to fallback to primary project and that way if packages are added to rings, followup can add their requirements (and/or check specifically)
  - also possible to fallback to devel project
  - lastly another candidate for `primary.gz` as it would provide overview of what is already provided which short of a new request providing a new thing, but needing to move rings (corner case) should pickup the rest
- use `_builddepinfo` to map the packages back to source packages
- check project for those packages and if exists link them into the highest required ring

If this works well, I think it would be a candidate for adding to `staging-bot`, but would be another nice thing that would be neat to either have `staging-bot` write on diary or somewhere to easily see that it has done something. This should be seen with `check` and such so not a requirement.

jberry-suse:
For SLE updates where builds are disabled looking in factory/devel or `primary.gz` will be the only way.

jberry-suse:
Likely also makes sense to allow `cleanup_rings` to be applied to stagings that way if a request is rejected, but ring modifications were made on its behalf they can easily be undone.

PR #847, opened 2017-04-26
cleanup_rings: runtime requiredby check, whitelist, summary, and substantial refactoring

### LINKED ISSUE #759: cleanup_rings whitelist
we need a list of packages cleanup_rings should not touch.

candidates:
* kernel-syms
* gnome-music

jberry-suse:
- 41e4d5ae2bd09f35889984367bb5ed7c266bb025:
    **cleanup_rings: provide package whitelist for special cases.**

- aca88d19c08c802f94a537979be6342b7ede171c:
    **cleanup_rings: check runtime requiredby if not a build requirement.**
    
    This check only works if the runtime requirement resides in the same ring.
    The fun case being kernel-syms which is required by virtual box (ring-2),
    but must remain in ring-1 to ensure matching build revision with other
    kernel packages.

- ac031bb915fa026b4886a43b7edb2f5169628774:
    stagingapi: provide fileinfo_ext_all() and fileinfo_ext().

- 4d0e98a00c4ce0a5e73569f3261559f3b7188cac:
    **cleanup_rings: print summary of packages to be dropped and commands last.**

- 39a4914e485035bd058bc02331a45b8cedd89da4:
    cleanup_rings: check ring index instead of string comparison.

- ce8818fd3d51e5843d7be5bf0f912c9ac6c8eb12:
    cleanup_rings: utilize splitlines() in check_buildconfig().

- 51c2f6e0e9d5ea0477ac0a805396972ed63a3367:
    cleanup_rings: extract check_image_bdeps() and check_buildconfig().
    
    - change archs loop to use the archs appropriate for the ring instead of
      always using DVD archs
    - drop check_image_bdeps() out of archs loop since DVD should only look
      at DVD archs even if ring 1

- 50166347e1ef3f0dde6474466fc78e91cd48034c:
    stagingapi: provide ring_archs() to return arch relevant to a ring.

- 1198da8eefc597b135ed7c59c8ec3d568d3197fd:
    cleanup_ring: extract repo_state_acceptable().

- 253b6592b4002f002a1f8cb74c0766aad8c5dd49:
    cleanup_rings: indicate the ring being processed.

- 57174711ab6e302693027f512349baa071e45299:
    cleanup_rings: utilize config ring list rather than hard-coded.

[diff out output change](https://gist.github.com/jberry-suse/b6f6962cacf67bc4b02076b009ecd31e) (note some packages reordered, 6 packages ignored due to runtime checks)

Fixes #759.
The runtime check is likely the basis for #815.

It would appear the second `texlive` check can be dropped with this change, but I'll leave for now. The `kernel-syms` could potentially be handled by only moving linked packages together, but since only applies to kernel it is likely not worth it.

These changes also restrict a package from being dropped form ring-0 through 1 and 2 in a single run. Instead only one ring drop is allowed per run which allows for a rebuild to occur and thus the run-time requirements can be checked. I confirmed that `virtualbox` requirement of `kernel-syms` is not show when quering the information for `kernel-syms` since they are in different projects. Obviously, if all the `requiredby` details were checked (ie all packages queried) this could be handled, but that would be very slow. Also note that it handles self references during runtime `requiredby` checks (like `-devel` packages).

Some ideas on getting around this would be to look at `primary.gz` in product repo as this contains everything in one spot, but I'll leave that for another day. Likely this is enough to make this much more reliable.

I spot checked a few of the packages, but feel free to run this locally or look at previously mentioned gist for the new output. Assuming all looks good lets merge and cleanup the rings for real.

### DIFF
--- a/osclib/cleanup_rings.py
+++ b/osclib/cleanup_rings.py
@@ -13,11 +13,9 @@ class CleanupRings(object):
         self.links = {}
 
     def perform(self):
-        self.check_depinfo_ring('{}:0-Bootstrap'.format(self.api.crings),
-                                '{}:1-MinimalX'.format(self.api.crings))
-        self.check_depinfo_ring('{}:1-MinimalX'.format(self.api.crings),
-                                '{}:2-TestDVD'.format(self.api.crings))
-        self.check_depinfo_ring('{}:2-TestDVD'.format(self.api.crings), None)
+        for index, ring in enumerate(self.api.rings):
+            ring_next = self.api.rings[index + 1] if index + 1 < len(self.api.rings) else None
+            self.check_depinfo_ring(ring, ring_next)
 
     def find_inner_ring_links(self, prj):
         query = {
--- a/osclib/cleanup_rings.py
+++ b/osclib/cleanup_rings.py
@@ -133,7 +133,7 @@ class CleanupRings(object):
         for arch in self.api.ring_archs(prj):
             self.fill_pkgdeps(prj, 'standard', arch)
 
-        if prj == '{}:0-Bootstrap'.format(self.api.crings):
+        if self.api.rings.index(prj) == 0:
             self.check_buildconfig(prj)
         else: # Ring 1 or 2.
             # Always look at DVD archs for image, even in ring 1.
--- a/osclib/cleanup_rings.py
+++ b/osclib/cleanup_rings.py
@@ -12,6 +12,11 @@ class CleanupRings(object):
         self.api = api
         self.links = {}
         self.commands = []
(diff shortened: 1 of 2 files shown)


jberry-suse:
I believe the main loop is a fair bit easier to understand given that the major bits and copy/pasted code have been replaced by method calls.

It also checks `i586` (all built archs) for builddeps instead of just `x86_64` (DVD archs) while still only checking DVD for the DVD image.

jberry-suse:
Performed some ring cleanup in Leap 42.3 and seems to be working. Definitely an improvement, but likely room for more.

PR #850, opened 2017-04-26
MaintenanceChecker: include project in owner query to utilize local devel and improvements.

### LINKED ISSUE #835: maintainer check: use devel project setting
when checking for maintainer the bot currently only looks at Factory resp the Factory devel project. Sometimes, e.g in case of kernel-source in Leap the package intentionally has  a different maintainer than in Factory. So the bot needs to check if the target package has a devel project set and use that maintainer instead.

jberry-suse:
- 11dc2e167e0c6d8aa48041c8d1072363cd695640:
    MaintenanceChecker: add memoize to owner_fallback().

- 3d768c3f2ed2d67faf21c35b084ff03f137ad8cb:
    MaintenanceChecker: include project in owner query to utilize local devel.
    
    For Leap the devel project may override the Factory one.

- a1cb4aa772882eb3aca2844ebb4febd43339a313:
    MaintenanceChecker: utilize osc.core.owner().

Tested on https://build.opensuse.org/request/show/490193.

Before:
```
DEBUG:leaper.py:author: tiwai, maintainers: michals,jirislaby,michal-m => need review
GET https://api.opensuse.org/search/owner?binary=kernel-source
DEBUG:leaper.py:Kernel:stable/None already is a reviewer, not adding again
```

After:
```
DEBUG:leaper.py:tiwai is maintainer
```

Confirmed previous behavior remained the same in two other cases:

- https://build.opensuse.org/request/show/490555
- https://build.opensuse.org/request/show/490286

Fixes #835.

### DIFF
--- a/check_maintenance_incidents.py
+++ b/check_maintenance_incidents.py
@@ -50,11 +50,7 @@ class MaintenanceChecker(ReviewBot.ReviewBot):
     # XXX: share with checkrepo
     def _maintainers(self, package):
         """Get the maintainer of the package involved in the package."""
-        query = {
-            'binary': package,
-        }
-        url = osc.core.makeurl(self.apiurl, ('search', 'owner'), query=query)
-        root = ET.parse(osc.core.http_GET(url)).getroot()
+        root = osc.core.owner(self.apiurl, package)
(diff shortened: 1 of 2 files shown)


jberry-suse:
Took a bit to fix the tests as the regex matching seems to behave non-intuitively.

PR #856, opened 2017-04-27
color extended: rebuild, supersede, frozenage, freeze.

jberry-suse:
Adding color to a few more commands.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -24,6 +24,7 @@ import warnings
 import yaml
 
 import colorama
+from colorama import Fore
 
 from osc import cmdln
 from osc import conf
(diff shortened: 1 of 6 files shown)


lnussel:
I like colors. Light grey is a bit hard to read on my ttys with white background though

jberry-suse:
I imagine it will be difficult with more than a couple colors to handle lots of color schemes. I know konsole color schemes override common colors to make them workable and the colors I use are within the common set. Two options then, likely you can override the common colors in your terminal emulator. See below for how it looks in konsole.

image

Alternatively, I added the ability to override the core colors using `.oscrc`. [Relevant keys](https://pypi.python.org/pypi/colorama) being:
```
BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE
```

For example adding the following to the `[general]` section in `.oscrc` overrides white to `RED`.
```
staging.color.white = 31
```

The code passes all values through [`code_to_chars()`](https://github.com/tartley/colorama/blob/master/colorama/ansi.py#L12). See [documentation on specifying colors](https://en.wikipedia.org/wiki/ANSI_escape_code). Hopefully, your terminal emulator can tweak it.

Issue #823, opened 2017-04-18
Consider enabling staging-bot for Factory


jberry-suse:
I have resolved the shortcomings that were on my list before considering for Factory in addition to suggestions from @DimStar77. A few times I had run the other bits of the bot (minus select) which cleanup up obsolete requests, superseded, untracked, and adi work as expected. Undoubtedly more things that can be improved, but it seems like the major points are handled to do an good initial staging job and manage the mundane parts.

Given the bot also has a proper home and thus can be monitored and disabled as needed it seems appropriate to consider its usage for Factory.

From reviewing the `splitter_info` in Factory and feedback clearly the proposals are being tried and used. Are there any other blocking issues?

For reference, some of the recent changes relevant to Factory:

- #816: request_splitter: add super strategy for representing devel groups.
- #810: request_splitter: switch from build percentage to staging age for merge.
- #806: request_splitter: provide and utilize an aged attribute on requests.

The last two allow for a minimum of `55` minutes (less than two 30 min bot cycles) of time before a request will be staged by a none strategy which allows for patterns to be observed. The staging age allows for followup requests to be merged within an 8 hour window after staging was activated. The super groups being a request by dimstar.

If no objections, perhaps just enable it and disable if it becomes an issue. For those who are not aware the `staging-bot-regular@Factory` service on packagelists is the on in question as the `-daily` cleanup service is already enabled. The timer just needs to be enabled/started.

nilxam:
I concern about cleanup "superseded" part, I guess it's related, see https://build.opensuse.org/request/show/486562 and https://build.opensuse.org/request/show/487431 , 486562 was staged and be superseded, before 487431 **superseded it in real* in the staging project, staging-bot removed 486562 from staging project, therefore the others python packages unable to build until select 487431 back to the same staging. this behavior is not good for Factory IMO.

jberry-suse:
I am not sure I understand.

- 486562 staged in E
- 487431 created and supersedes 486562
- bot removes 486562 rather than supersede it?

The bot should just use `list --supersede` like normal to replace 486562 with 487431. The cleanup, `unselect --cleanup` gets the list from the staging dashboard. If that is not updated then perhaps a secondary check before unselecting needs to be performed to ensure it has not been superseded. Although that would seem unlikely since the request would have to be revoked/rejected first anyway to be considered obsolete and then superseded shortly there after in between all this happening.

Perhaps you can clarify what happened for me.

nilxam:
[...]

PR #858, opened 2017-04-28
request_splitter: add staging whitelist config option.

jberry-suse:
For the initial Factory usage the bot will be allowed to stage on a subset of the available stagings.

@DimStar77 suggestion in relation to #823.


DimStar77:
LGTM - The idea is to not give all the power away to a bot but retain some for the human staging master - mainly until we gain enough trust in the bot working in our best interest

PR #859, opened 2017-04-28
lock: provide command to place a multi-command hold on a project (and lock reason).

jberry-suse:
- 630663a1a7557a26461ca8da326011d7bf30e0cf:
    **lock: provide command to place a multi-command hold on a project.**

- 02cd0cf0f8e789e3abd99df0652d9cd34efaaeb9:
    obslock: add reason to indicate what command is being run.

Is backwards compatible so others on older version should still be locked out, but will show the reason as part of username. For example:

```
Lock acquired by [jberry_factory#hold: test] 0:01:55.212672 ago. Try later
```

This provides a reason when seeing a locked out message which will either be the command being run or an indication that a manual hold has been placed on the project.

A normal command lock is still acquired when placing a hold on the project so one will not be able to place a hold if someone else already has a lock. After a hold has been acquired commands may be run like normal and will update the lock to a command level lock. This ensures that one cannot run multiple commands in parallel (just like before) even when a hold is placed on the project.

The documentation update provides a solid example (message being optional):

```
lock -m "checkin round"

list --supersede
adi
accept A B C D E

unlock
```

This new hold can be used to prevent the `staging-bot` (and others) from interrupting.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -157,6 +157,19 @@ def do_staging(self, subcmd, opts, *args):
 
     "list" will list/supersede requests for ring packages or all if no rings.
 
+    "lock" acquire a hold on the project in order to execute multiple commands
+        and prevent others from interrupting. An example:
+
+        lock -m "checkin round"
+
+        list --supersede
+        adi
+        accept A B C D E
+
+        unlock
+
+        Each command will update the lock to keep it up-to-date.
+
     "repair" will attempt to repair the state of a request that has been
         corrupted.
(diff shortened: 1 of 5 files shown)


jberry-suse:
It likely makes sense to have the bot use this while it runs multiple commands.

jberry-suse:
- refactored parsing of `sub_reason` into `_parse()` and `_signature()`.
- implemented `locked` property
- implemented test suite

PR #861, opened 2017-04-29
osc-staging: implement remote config and expose more settings to config.

### LINKED ISSUE #813: request_splitter: provide project level config for options
Given that the most of the options are specific to the project (or at least openSUSE/SUSE) it likely makes sense to read the configuration options from project config file (likely out of dashboard package).

jberry-suse:
- 4dd6eb8df870a50749aaae6e8403d0ab145c8c77:
    request_splitter: expose age thresholds to config.

- ed9488fb35e68f1d7bb8bc12007e474bed8b6bb1:
    cache: include dashboard files.

- a08c55b6028305f4ff64ff06551480a18670d659:
    osc-staging: implement remote config.

Seems to make sense to use the same format as `.oscrc` the only potential downside is storing more complex data structures in config. Note that the config is loaded in two parts since the most basic config information is required to even know where to look for the remote config. As such the early values loaded in `StagingAPI` may not be overridden remotely, but that makes sense since changing the `cstaging` after loading the config from that staging does not make a lot of sense.

Order of precedence:

- code defaults
- remote config `dashboard/config`
- local `.oscrc`

With the cache tweak no remote calls should be required after first run to load config.

Fixes #813.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -338,7 +338,7 @@ def do_staging(self, subcmd, opts, *args):
     opts.project = self._full_project_name(opts.project)
     opts.apiurl = self.get_api_url()
     opts.verbose = False
-    Config(opts.project)
+    config = Config(opts.project)
 
     colorama.init(autoreset=True,
         strip=(opts.no_color or not bool(int(conf.config.get('staging.color', True)))))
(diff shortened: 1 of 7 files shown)


jberry-suse:
I created a [Factory](https://build.opensuse.org/package/view_file/openSUSE:Factory:Staging/dashboard/config?expand=1) and [Leap](https://build.opensuse.org/package/view_file/openSUSE:Leap:42.3:Staging/dashboard/config?expand=1) config in their respective dashboard packages. As such I removed the default for Factory staging whitelist. Can keep local as remote will still win, but this way the line can simply be removed in remote file to remove whitelist and allow all stagings.

jberry-suse:
- implement tests suite.

PR #865, opened 2017-05-02
factory-package-news-web: remove current check from /diff path.

jberry-suse:
The checks seems entirely superfluous.


jberry-suse:
n/m, it always diffs against `current` otherwise it would need to keep track of which snapshots were published to be able to compare against previous.
```python
cmd = [os.path.dirname(os.path.abspath(__file__))+'/factory-package-news.py', \
            'diff', '--dir', _dir, "current", version]
```

PR #866, opened 2017-05-02
factory-package-news-web: do not restrict version to int.

jberry-suse:
Leap version numbers are currently formatted with a leading zero which is stripped when converted to an int.

Another for #703.

### DIFF
--- a/factory-package-news/factory-package-news-web.py
+++ b/factory-package-news/factory-package-news-web.py
@@ -80,10 +80,9 @@ def current():
             return "", 404
         return os.readlink(fn)
 
-@app.route('/diff/<int:version>')
+@app.route('/diff/<version>')
 def diff(version):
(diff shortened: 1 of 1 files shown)


lnussel:
what characters are allowed now?

jberry-suse:
Currently, filtered to `int` this changes to `string`. A directory check is performed and those directories are created by `POST` to `current` which filters on `re.compile('^[0-9]+$')` so presumably still "int like" except that leading zeroes are allowed.

jberry-suse:
It seems this works as expected.

- `/snapshot-changes/opensuse/42.3/diff/abc` results in `invalid version` while
- `/snapshot-changes/opensuse/42.3/diff/0183` spits out changes now that current has been set vs without this change it also says `invalid version` since it ends up look for `183` and not `0183`

PR #872, opened 2017-05-02
unselect: add conditions to obsolete filter used by --cleanup (along with refactor and tests).

### LINKED ISSUE #871: unselect --cleanup: add conditions to make it less aggressive
Per discussion in #823 the `--cleanup` option needs to allow for declines that are expecting changes resulting in a followup request. Unfortunately, there is no `needs changes` state. Once a new request has been created it should supersede the other request via any one of the existing mechanisms in place. The issue is the bot will `unselect` the request after it is `declined` and many times before a new request has been created.

The purpose of the cleanup is to ensure that stagings (especially `adi`) have `obsolete` requests dropped in order to unblock them. Many times resulting in rebuilds so doing so as soon as possible is handy.

It would seem a request in the `revoked` state should always be safe to `unselect`. Requests made by the respective bots for crawling updates do not appear to be `revoked` after `declined` which was the most common case for Leap. As such the following conditions may suffice.

- in state `revoked`
- in state `declined` and by whitelisted users (ie `factory-maintainer` and `leaper`)
- in state `declined` for over `72` hours

The last condition in attempt to allow for a human to create a followup request that will `supersede` the current one. Unfortunately, for requests that are outright `declined` they will sit there for `72` hours. Shorts of looking for specific text in message given when `declined` there simply is not enough information to know for sure. Coming up with a standard hint in message would seem handy as well.

jberry-suse:
- ad9ffc486ee554c24d336c87781e5027be61cb27:
    tests/unselect_tests: add test for cleanup fitler.

- 4e26f2cb27a5b448230e3609252c69b8a979cecd:
    tests/obs: define PROJECT as openSUSE:Factory.

- fb47742431d50d5e03ab74650986e5b6ab6b725d:
    **unselect: add conditions to obsolete filter used by --cleanup.**

- 35fce8136699ae25b1443b86826b0fa020b00a4a:
    stagingapi: project_status_requests(): provide filter function.

- 007ed1a21b98d2344fba9b21b1c8001f636a470b:
    **stagingapi: project_status_requests(): filter requests changed < 5 min ago.**

Tests handle the cases of interest along with the general `< 5 min` case which also applies to `untracked`.

Fixes #871.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -922,7 +922,7 @@ class StagingAPI(object):
                 tobuild += int(repo['tobuild'])
         return final, tobuild
 
-    def project_status_requests(self, request_type, filter_superseded=False):
+    def project_status_requests(self, request_type, filter_function=None):
         key = '{}_requests'.format(request_type)
         requests = []
         for status in self.project_status():
@@ -934,8 +934,10 @@ class StagingAPI(object):
                     # requests whose state has changed in the last 5 minutes.
                     continue
 
-                if not filter_superseded or request['superseded_by_id'] is None:
-                    requests.append(str(request['number']))
+                if filter_function and not filter_function(request, updated_delta):
+                    continue
+
+                requests.append(str(request['number']))
 
         return requests
 
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -16,6 +16,7 @@
 
 from cStringIO import StringIO
 from datetime import datetime
+import dateutil.parser
 import json
 import logging
 import textwrap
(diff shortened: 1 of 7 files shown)


nilxam:
LGTM

jberry-suse:
I changed to `updated_delta.days >= 4`, but could bump higher.

jberry-suse:
Alternatively this may make sense to make it a config variable that can we tweak in the remote config.

jberry-suse:
The wait 5 minutes for dashboard cache is a definite improvement that will resolve staging, `repair --cleanup`, stage again which happens from time to time. The rest seems like an improvement and exposed config options to play around with. Re-open the issue if further conditions to add or change.

nilxam:
then made it remote config and set `7` days for TW, from previous experience, move out several request from staging project leads to a re-bootstrapping, `7` days give us time talking to maintainer to get the fixes and fix the rest fall out; also a big move change will needs longer time than 4 days, eg. patterns split change now is 8 days old in :F.

so I propose:
1) use remote config and set `7` days for TW
2) add a option in the remote config to disable unselect declined request in case we have a big move change in the staging project(or a white list per specific staging project.)

jberry-suse:
Sounds good.

1. I actually changed the default to `7` days so we should be good there.
2. disable can either be done by setting the remote config variable, `unselect-cleanup-days`, to `10000` (or similar). if it seems useful I can add a staging whitelist, but otherwise I believe the current configuration and implementation meets your requirements

nilxam:
oops..I overlooked your another comment and am still thought the default cleanup days is `4` :)

PR #881, opened 2017-05-04
Refactor map_ring_package_to_subject to support corner case

nilxam:
We have a corner case that package A in Ring2 and points to package B in Ring1, map_ring_package_to_subject have to handle that as another non-inner project link case, ie. the sub-package link must be created.

https://github.com/openSUSE/osc-plugin-factory/issues/876

With this change:

```
maxlin_factory@tigercity:~> osc api /source/openSUSE:Factory:Staging:G/libunbound-devel-mini?expand=1
<directory name="libunbound-devel-mini" rev="97a63451d3b9e5fbc28d602e186ddd1b" vrev="4" srcmd5="97a63451d3b9e5fbc28d602e186ddd1b">
  <linkinfo project="openSUSE:Factory:Rings:2-TestDVD" package="unbound" srcmd5="4386c97ad0b6541866be8b714a403f37" lsrcmd5="21343c588907214069c013b4886f7c48" />

Selected to :G

maxlin_factory@tigercity:~> osc api /source/openSUSE:Factory:Staging:G/libunbound-devel-mini?expand=1
<directory name="libunbound-devel-mini" rev="bfdd6c3f78b8dbbae0bb05fb9870aae1" vrev="49" srcmd5="bfdd6c3f78b8dbbae0bb05fb9870aae1">
  <linkinfo project="openSUSE:Factory:Staging:G:DVD" package="unbound" srcmd5="67b8be0d6f387480185d738c73abb0ac" lsrcmd5="67223dc56002f8920fbe53a8d8456fb5" />
  <serviceinfo code="succeeded" lsrcmd5="fc019529629327d5289e25c43cd81d0d" />
```

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -998,16 +998,16 @@ class StagingAPI(object):
         """
         # it's actually a pretty stupid algorithm, but it might become more complex later
 
-        if project.endswith(':DVD'):
-            return project  # not yet
+        # assuming it is in adi staging, workaround for https://progress.opensuse.org/issues/9646
+        if self.is_adi_project(project):
+            return project
 
-        ring_dvd = '{}:2-TestDVD'.format(self.crings)
-        if self.ring_packages.get(pkg) == ring_dvd:
-            if not self.item_exists(project + ":DVD") and self.item_exists(project, pkg):
-                # assuming it is in adi staging, workaround for https://progress.opensuse.org/issues/9646
-                return project
-            else:
-                return project + ":DVD"
+        if self.ring_packages.get(pkg) and self.rings.index(self.ring_packages.get(pkg)) == 2 and not project.endswith(':DVD'):
+            # pkg A in ring1 and points to pkg B in ring2
+            return project + ":DVD"
+        elif self.ring_packages.get(pkg) and self.rings.index(self.ring_packages.get(pkg)) != 2 and project.endswith(':DVD'):
+            # pkg A in ring2 and points to pkg B in ring1
+            return project.replace(':DVD', '')
 
         return project
 
--- a/tests/api_tests.py
+++ b/tests/api_tests.py
@@ -58,6 +58,7 @@ class TestApiCalls(unittest.TestCase):
         # our content in the XML files
         # test content for listonly ie. list command
         ring_packages = {
+            'apparmor': 'openSUSE:Factory:Rings:1-MinimalX',
             'elem-ring-0': 'openSUSE:Factory:Rings:0-Bootstrap',
             'elem-ring-1': 'openSUSE:Factory:Rings:0-Bootstrap',
             'elem-ring-mini': 'openSUSE:Factory:Rings:0-Bootstrap',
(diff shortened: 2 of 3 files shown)


osclib/stagingapi.py
```diff
@@ -998,16 +998,17 @@ def map_ring_package_to_subject(self, project, pkg):
...
+            return project
 
         ring_dvd = '{}:2-TestDVD'.format(self.crings)
-        if self.ring_packages.get(pkg) == ring_dvd:
-            if not self.item_exists(project + ":DVD") and self.item_exists(project, pkg):
-                # assuming it is in adi staging, workaround for https://progress.opensuse.org/issues/9646
-                return project
-            else:
-                return project + ":DVD"
+        if self.ring_packages.get(pkg) == ring_dvd and not project.endswith(':DVD'):
```
jberry-suse: Something I've done elsewhere that I find cleaner than building a string to compare against to determine ring project is the following.

```python
if self.rings.index(self.ring_packages.get(pkg)) == 2:
```

No extra string creation with a pattern that is now hard-coded in another place and the rings line up with array indexes.

An example in [`StagingAPI.ring_archs()` line 1552](https://github.com/nilxam/osc-plugin-factory/blob/c2808dd22cfbe107633954d341dee13d09627cbb/osclib/stagingapi.py#L1552).

osclib/stagingapi.py
```diff
@@ -998,16 +998,17 @@ def map_ring_package_to_subject(self, project, pkg):
         """
         # it's actually a pretty stupid algorithm, but it might become more complex later
 
-        if project.endswith(':DVD'):
-            return project  # not yet
+        # assuming it is in adi staging, workaround for https://progress.opensuse.org/issues/9646
+        if self.is_adi_project(project) and self.item_exists(project, pkg):
```
jberry-suse: Does the `self.item_exists(project, pkg)` check make sense? This code is called during select (`submit_to_prj()`) which means the package likely does not exist in the staging yet. Not to mention the whole point is to avoid `:DVD` when using an `adi` regardless of the state of the `adi`.

I realize it was there before, but not exactly sure how it works properly.

nilxam: not just submit_to_prj(), those code also introduced in rm_from_prj(), the code is mainly to protect a special case like https://progress.opensuse.org/issues/9646 mentioned, if pkg A already staged in adi staging and meanwhile pkg A was added to ring2, while adi accepting it got trouble to delete the package from project as tool thinks it's ring2 package must cleanup package from :DVD, however doesn't have it in adi staging.

osclib/stagingapi.py
nilxam: cool, will update it later.

osclib/stagingapi.py
jberry-suse: Indeed, `rm_from_prj()` is the case to which this PR relates.

The case makes sense, but the condition still seems like it should just be `if self.is_adi_project(project)` since the `:DVD` prefix is never applicable regardless of ring state or package existing in an adi staging.

nilxam: it's just an additional check, fine to remove.

Issue #811, opened 2017-04-14
select: pickup followup requests based on description or for a failing to build package


jberry-suse:
Followup request, those created after issues are seen in a staging, are quite common and would nice to have a more automated method for handling them. Using a special syntax in the description text seems like a nice convenience.

Some possible examples/patterns (perhaps support a few):

- Followup for KDE:Frameworks5 in Staging:E.
- (Stage|Place) in Staging:E
- (Followup )?for(.* in)? Staging:E

These followup requests could be restricted in a few ways to help alleviate abuse and what not.

- Restricted to project/package maintainers so a slightly more trusted user which would prevent random requests from being created and staged all over. Although, likely the request would end up being duplicate and ignored anyway.
- Maintain the automatic 8 hour staging window or if staging is in the state `failed`.

If a request is made for a package failing to build in only one of the stagings, or if not too tricky a dependency of a failing package, it should be placed in that staging.

jberry-suse:
Also worth considering is lifting the `--merge` window for `devel` or `super` so that **all** followup requests are included. I can see this being really handy and potentially annoying. Could end up just getting a trickle of requests at such an interval that the staging never finishes building. Although in many cases that may be the only legitimate way forward. Another possibility would be to just increase it to a much longer period like a week.

jberry-suse:
A simplified version that may be helpful before a more complex algorithm can be implemented would be a slight tweak of the last comment. Where the window is lifted if that staging is not in an `acceptable` state.

jberry-suse:
The followup descriptions could also be used in conjunction with a tool that looks for un-resolvable requests (many times in `adi`s, especially for Leap) that simply need another package submitted. Having a preference order of sources and generate a submit with the reason it was generated in description so it gets staged properly. I find I do this manually already.

jberry-suse:
Also comments on staging projects many times have links or markdown reference to needed requests.

jberry-suse:
Comments can be all over https://build.opensuse.org/request/show/519533

coolo:
For reasons as this we are currently (for quite a while actually) running without automated staging

PR #883, opened 2017-05-04
request_splitter: should merge: allow devel and super if not ready.

jberry-suse:
Simplistic precursor to #811.

If the consensus is that this is generally alright we can accept, but I can see this not always being the best. For things like KDE updates that are long standing stagings with fixes over a week or more this seems very useful, but for troublesome stagings it may just end up adding more clutter.

For example the proposal generated for Factory at the moment:

```yaml
0#devel@Base:System:
  bootstrap_required: true
  group: Base:System
  merge: true
  requests:
    491274: bash
    492375: rpm
    492649: coreutils
    492807: systemd
  staging: B
  strategy:
    name: devel
1#devel@devel:languages:python3:
  bootstrap_required: false
  group: devel:languages:python3
  merge: true
  requests:
    492721: python3-numba
    492722: python3-llvmlite
    492896: python3-fastcluster
    492900: python3-patsy
    492926: python3-statsmodels
    492932: python3-seaborn
  staging: E
  strategy:
    name: devel
```

Otherwise, can wait for more complex approach trying to see if request might fix something in a staging. The side bonus of this approach being that if a fix takes a day or two to come in other requests can be grouped in and accepted together.

Thoughts?

### DIFF
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -177,6 +177,12 @@ class RequestSplitter(object):
         return len(pseudometa['requests']) > 0 and 'splitter_info' in pseudometa
 
     def should_staging_merge(self, status, pseudometa):
+        if (pseudometa['splitter_info']['strategy']['name'] in ('devel', 'super') and
+            status['overall_state'] not in ('acceptable', 'review')):
+            # Simplistic attempt to allow for followup requests to be staged
+            # after age max has been passed while still stopping when ready.
+            return True
+
         if 'activated' not in pseudometa['splitter_info']:
             # No information on the age of the staging.
             return False


DimStar77:
I see less trouble for that for non-bootstrap stagings, but as your example luckily shows, for bootstrap we can sort of end up with re-bootstrapping the staging over and over

jberry-suse:
Could also limit to non-bootstrapped.

jberry-suse:
Another thought, could add an option to staging `splitter_info` that a human simply adds to instruct the bot to keep adding followup requests, or perhaps is defaulted to `True` and a human can disable. May just be cumbersome.

jberry-suse:
Should we make some tweaks or revert this before deployed to packagelists?

DimStar77:
i'd like to see at least the non-bootstrap limitation implemented please

jberry-suse:
Sure, I had not intended for this to be merged. :)

PR #885, opened 2017-05-05
Refactor announcer

lnussel:
- use argparse
- use logging
- allow to specify a config file for options


jberry-suse:
For reference, related to #703.

PR #887, opened 2017-05-09
osc-staging: skip lock for observation commands (and other unlock fix).

### LINKED ISSUE #886: osc staging: implement 'no locking needed' for some commands
There are a number of commands that do not modify the setup of stagings at all and that only are used to read information. Those commands should

a) be able to run even if somebody else has a lock
b) not lock the project, so others can run stagnig plugin without issues

The commands coming to mind are

* acheck
* check
* frozenage

jberry-suse:
- 3d7a7d487b4b9ddf1eac2643c47b422213cc0a70:
    obslock: unlock other user lock if force.

- f4523078b45f4f315eac79f93587516b03b100b1:
    **osc-staging: skip lock for observation commands.**

Fixes #886.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -76,6 +76,12 @@ def _full_project_name(self, project):
     warnings.warn('%s project not recognized.' % project)
     return project
 
+def lock_needed(cmd, opts):
+    return not(
+        cmd in ('acheck', 'check', 'frozenage', 'rebuild', 'unlock') or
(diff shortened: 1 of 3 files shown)


jberry-suse:
`rebuild` is partially questionable since someone could unstage packages that it tries to rebuild, but does not seem worth locking since the end result of that scenario is the package does not need rebuilding since it was removed.

PR #888, opened 2017-05-09
obslock: unlock other user lock if force.

jberry-suse:
Always clear lock by another user if `force` is `True`.


jberry-suse:
Merged in #887.

PR #890, opened 2017-05-09
Add more klickable links

lnussel:


leaper.py
```diff
@@ -221,7 +221,6 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
                     self.needs_release_manager = True
                 if origin == src_project:
                     self.source_in_factory = True
-                    return True
```
jberry-suse: What effect does this have? Does not seem to fall to logic with additional links.

lnussel: it falls through to the is_in_factory branch later

lnussel: but then yeah, actually a noop. just one early return less

leaper.py
```diff
@@ -324,13 +324,17 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
             for p in ('-SP3:GA', '-SP2:Update', '-SP2:GA',
                     '-SP1:Update', '-SP1:GA', ':Update', ':GA'):
                 prj = 'SUSE:SLE-12' + p
-                if self.is_package_in_project(prj, target_package):
-                    self.logger.info('Package is in {}'.format(prj))
+                if self.is_package_in_project(prj, package):
+                    self.logger.info('Package is in [{0}](/package/show/{0}/{1}) ([diff]([{2}]))'.format(prj,
```
jberry-suse: Creates invalid links as seen in https://build.opensuse.org/request/show/506294. Fix in #972. Always good to paste output into an OBS markdown field somewhere to test.

PR #894, opened 2017-05-10
leaper: indicate if submission is already accepted to SLE origin.

### LINKED ISSUE #873: leaper: check for source in SLE when changing origin from SLE
Perhaps always check regardless of origin change, but in the case of submission from Factory or devel project where the source is already in SLE `leaper` should consider this an normal SLE update and not require extra reviews.

Example:
- https://build.opensuse.org/request/show/491864
- https://build.opensuse.org/package/rdiff/SUSE:SLE-12-SP3:GA/perl-Bootloader?linkrev=base&rev=3

```
Base:System/perl-Bootloader@7e85af9fe0ea300a66a2a89a3c129e56 -> openSUSE:Leap:42.3/perl-Bootloader

expected origin is 'SUSE:SLE-12-SP3:GA' (changed)

sr#491865 review by openSUSE:Factory:Staging:H ok

the submitted sources are in or accepted for Factory
```

Automated crawler requests from SP3 will be automatically declined based on the `supersede` logic that I added so really it seems this should just be treated as such.

Thoughts?

jberry-suse:
It seems unlikely that checking for requests would be acceptable for SLE.

Fixes #873.

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -306,6 +306,11 @@ class Leaper(ReviewBot.ReviewBot):
                                 self.logger.info("submission from higher service pack %s:* ok", prj)
                                 return True
 
+                in_sle_origin = self._check_factory(target_package, _src_srcinfo, origin)
+                if in_sle_origin:
+                    self.logger.info('parallel submission, also in {}'.format(origin))
+                    return True
+
                 self.needs_release_manager = True
                 # the release manager needs to review attempts to upgrade to Factory
                 is_fine_if_factory = True


jberry-suse:
If allowed it would be handy to do a request lookup on IBS, but I imagine this is not practical.

lnussel:
ibs is not mapped into obs namespace unfortunately

jberry-suse:
Well, a direct request could be made. More a question of is it alright to expose internal requests before accepted and published to OBS.

Issue #848, opened 2017-04-26
ring: create command for making and recording ring changes and process during accept


jberry-suse:
After discussion this tool would be rather useful in general and likely a precursor to #815.

There are several scenarios to support.

- adding request to ring
  - `select --ring 2 A foo` (enabled instead of disabled)
- changing ring from which a request would normally be placed
  - `select --ring 1 A ring_2_package` (marked as moved)
- adding non-ring request to staging for grouping only
  - `select A foo` (disabled and placed in letter project not DVD like current behavior)
- raising the ring for an existing package without a request
  - `ring A 1 ring_2_package` (move package from ring 2 to 1 in staging A)
  - `ring 1 ring_2_package` (move package from ring 2 to 1 in project rings)
  - `ring 2 ring_1_package` (move package from ring 1 to 2 in project rings)
  - `ring ring_2_package` (remove package from ring 2 in project rings)
- reset ring changes made to a package in a staging
  - `ring --reset A foo` (drops metadata and reverts package to previous ring state)
- reset all changes in a staging
  - `ring --reset A`

Future `select` should utilize the recorded ring changes to override base ring placement. Thus if a package is moved from ring 2 to 1 and a request is then staged in that staging the request will be placed in ring 1.

These commands can be issued from `cleanup_rings` instead of the command pairs that are currently issued. `accept` will then apply the changes made in a staging to the project rings. For example:

- `select --ring 2 A newdep`
- `ring A 1 moving_up_in_the_world` (currently in ring 2)
- `accept A`

Would result in:

- normal `accept` behavior
- `newdep` added to ring 2 in project rings
- `moving_up_in_the_world` moved from ring 2 to ring 1 project rings

Having a command to print out ring changes almost like `check` or a `status` command would seem handy. For now this likely makes sense to just include in `ring` command.

`ring A`

Prints recorded ring changes to staging `A` or indicates if there are none.

jberry-suse:
Touch the project description if removing a package show it shows up as changed in OBS list and clears everyone's caches.

lnussel:
👍

jberry-suse:
#897 will be useful second part to this to ensure packages added to rings in a staging remain enabled when superseded.

nilxam:
BTW we have try to avoid the newdep being add to project rings in case the required package was out of that staging. Ie.

* FOO requires newdep-BAR
* newdep-BAR is selected and recorded
[...]

PR #897, opened 2017-05-10
stagingapi: carry over build state during supersede.

### LINKED ISSUE #896: osc staging supersede: do not change build status to disabled
This is something that has been like this forever, but there is no reason not to address it.

situation:
* a package that newly will be a ring package is added to :LETTER; staging select marks it 'build disabled', which is fine/acceotable
* Staging master went and enabled build for this package (knowing it will be ring relevant)
* the submission is being superseded; at which point staging list --supersede will add the NEW submission to :LETTER, BUT disables build again (with the potential to mask weird errors if for example the old submission was ok, but the new one is not. The staging project willkeep on using the old binaries)

jberry-suse:
Allows for packages being added to rings to remain enabled after being superseded.

Likely need to test on a local OBS, build out a supersede test case (not trivial), or cross-fingers.

DO NOT ACCEPT UNTIL SOMEONE HAS HAD A CHANCE TO SEE THIS WORK.

Fixes #896.

Presumably, it is desirable to keep sub package states which is what I did.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -75,6 +75,8 @@ class StagingAPI(object):
         self._ring_packages_for_links = None
         self._packages_staged = None
         self._package_metas = dict()
+        self._supersede = False
+        self._package_disabled = {}
 
         # If the project support rings, inititialize some variables.
         if self.crings:
@@ -596,6 +598,7 @@ class StagingAPI(object):
             target_requests = []
 
         # get all current pending requests
+        self._supersede = True
         requests = self.get_open_requests()
         requests_ignored = self.get_ignored_requests()
         # check if we can reduce it down by accepting some
@@ -608,6 +611,7 @@ class StagingAPI(object):
             stage_info, code = self.update_superseded_request(rq, target_requests)
             if stage_info:
                 yield (stage_info, code, rq)
+        self._supersede = False
 
     def get_prj_meta(self, project):
         url = make_meta_url('prj', project, self.apiurl)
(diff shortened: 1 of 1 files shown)


lnussel:
so anyone confirmed this working?

jberry-suse:
We could either risk it, create a manual scenario, or wait for me to have time to rework testing setup so this is not a pain to test.

osclib/stagingapi.py
```diff
@@ -760,15 +764,27 @@ def rm_from_prj(self, project, package=None, request_id=None,
 
         self._remove_package_from_prj_pseudometa(project, package)
         subprj = self.map_ring_package_to_subject(project, package)
+        if self._supersede:
+            self.is_package_disabled(subprj, package, store=True)
         delete_package(self.apiurl, subprj, package, force=True, msg=msg)
 
         for sub_prj, sub_pkg in self.get_sub_packages(package):
             sub_prj = self.map_ring_package_to_subject(project, sub_pkg)
+            if self._supersede:
```
nilxam: Actually this should not needed, ring package should always be enabled to build and it's sub-package should do too, carry over build state is only apply to `new package` request.

DimStar77: you mean 'new in ring'? as we might have an existing package that is newly promoted to a ring, paired with a submission introducing new fixes needed too

nilxam: Whatever it's **new package submission to Factory and will be promoted to a ring** or **an existing package update and will be promoted to a ring**, Rings should not have changed until the submission merged to Factory, thus there is `none` from `get_sub_packages`.

jberry-suse: My thought at this point is two-fold:

1. Lets prioritize #1002 (real OBS to test against) so we have some way to walk through all the possible workflows and verify since this is some fairly scary code.
2. It may make sense for me to revisit my prototype `ring` command (#848) and finish it as that would provide supplemental information that would make it crystal clear which packages should be enabled rather than the somewhat wishy-washy alternatives (looking at previous build settings).

nilxam:
I'd like to give it a try in real however staging-bot always win the game :-)
How about we just merge it and see what happened on staging-bot? it should not break things but just about changing build state, we can survived with it in case something goes wrong.

nilxam:
hey @jberry-suse , I tested your code, there is a bug about join(), in this change there is two arguments given to join() hence it fails to work, I think it should mean a tuple of strings ie. ``'/'.join([project, tar_pkg]`` , so I changed all those places to ``'/'.join([project, tar_pkg]`` and tried it again, looks works fine.

jberry-suse:
Thanks for testing. I made the following changes.

```diff
diff --git a/osclib/stagingapi.py b/osclib/stagingapi.py
index 7530d4d..9c2407c 100644
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -782,7 +782,7 @@ class StagingAPI(object):
         meta = ET.fromstring(''.join(meta))
         disabled = len(meta.xpath('build/disable[not(@*)]')) > 0
         if store:
-            self._package_disabled['/'.join(project, package)] = disabled
+            self._package_disabled['/'.join([project, package])] = disabled
         return disabled
 
     def create_package_container(self, project, package, disable_build=False):
@@ -1119,7 +1119,7 @@ class StagingAPI(object):
                 project = self.map_ring_package_to_subject(project, tar_pkg)
 
             if self._supersede:
-                disable_build = self._package_disabled.get('/'.join(project, tar_pkg), disable_build)
+                disable_build = self._package_disabled.get('/'.join([project, tar_pkg]), disable_build)
 
         self.create_package_container(project, tar_pkg,
                                       disable_build=disable_build)
@@ -1147,7 +1147,7 @@ class StagingAPI(object):
             if sub_prj == project:  # skip inner-project links
                 continue
             if self._supersede:
-                disable_build = self._package_disabled.get('/'.join(sub_prj, sub_pkg), False)
+                disable_build = self._package_disabled.get('/'.join([sub_prj, sub_pkg]), False)
             self.create_package_container(sub_prj, sub_pkg, disable_build=disable_build)
 
             root = ET.Element('link', package=tar_pkg, project=project)
```

Issue #764, opened 2017-03-17
don't hardcode architectures in repochecker


lnussel:
Repochecker currently hardcodes i586 and x86_64 as architectures. It needs to look at what the target actually has to determine what to require in the source project. Maybe abichecker could give some hints:
https://github.com/openSUSE/osc-plugin-factory/blob/master/abichecker/abichecker.py#L908

nilxam:
Repochecker not just hardcoded that in one place, which one you're referencing to abichecker? or all?

lnussel:
all places have to be fixed

jberry-suse:
Some code sharing would have been nice rather then loads of copy/paste.

I assume the end goal here is to simply ensure there is a successful build in one of the target architectures of the target project. That seems to be what abichecker is doing.

jberry-suse:
Made some progress. I have the list of architectures being read from main project and utilized in a number of places. The biggest issue seems to be the code previously assumed that a package had to have builds enabled for all considered archs (ie i586 and x86_64) whereas Leap has both enabled, but any project that build against Leap only have x86_64. With some tweaks I can get the check to pass, but need to know what is the desired requirement. Potentially only require that at least one of the considered archs if provided and that the package builds (or is excluded) in all applicable archs included in that repo, but this would not require them for Factory. Otherwise, given the setup of Leap will likely need some sort of hard coded map or somesuch.

PR #904, opened 2017-05-12
repo-checker: refactoring to primarily support multiple archs and the Leap workflow

### LINKED ISSUE #837: repochecker: refactor to support Leap and SLE
Per email request. This issue will track the discussion and subissues related to this task.

First up is #764 to not hardcode architectures.

Part of the work will be understanding what the expected workflow is for Leap and SLE. For instance the first blocker that came up is update requests that have builds disabled are immediately rejected since they are not build against Leap 42.3 (like https://build.opensuse.org/request/show/488094). Does it make sense to add a repo for repochecker in previous Leap version updates repo? Similarly, updates coming from SP3 will have the same problem, but in that case I am not sure if the bridge allows for building without first branching the package (could be done to build entire SP3 against Leap 42.3). Then of course the common Factory submissions which of course also do not build for Leap 42.3. Is the intention to enable such repos? For Factory perhaps the repo checker can look back at the devel project which will likely build against 42.3.

jberry-suse:
The commits provide a fairly good summary of the changes. I have been running the code on a variety of requests against Leap and Factory while developing. Today I cleaned up the code and split into proper commits. As such I will be running some more requests through the checker tomorrow.

Given that `i586` is now checked this does cause new failures in Factory that were not noticed before. Additionally, with the drop of the complex repo logic in `repositories_to_check()` it would be a good idea for any known complex repos to be checked. Let me know if you run any requests and they work properly or have issues. I would suggest we hold off on merging until we are confident this works in a variety of cases in which everyone is comfortable.

This also contains the first bits of `osclib.core` explained by commit and resolves the long standing todo to share the maintainers lookup call. Since I recently enhanced that code to support "local devel" projects for packages like `kernel-source` in Leap it makes sense to ensure repo-checker utilizes those enhancements.

I postponed the rework of the output into the project level comment. I figure if the spammy nature was acceptable before it will be now. Better the limit the number of core changes in one round, but worth revisiting once this settles.

If one is going to thoroughly review I would suggests the commits as they are broken down into small atomic units.

- 34f1ecb6f629de7075e8a0205321e98c4d105660:
    core: extract maintainers_get(), owner_fallback() from MaintenanceChecker.
    
    Allows the code to be properly shared between checkrepo and
    check_maintenance_incidents as a todo suggests. Given that the majority of
    similar cases for code sharing are extension of osc.core it seems to make
    sense to place them in osclib.core.

- 42d01327f47dd0438585a610a58a57bf880b5eb8:
    checkrepo: wait for accepts for all archs before final review accept.

- 087f86e4f8bcd4e9787a14f786da892b3da46e01:
    osc-check_repo: rework primary calls to loop over archs.

- 9c84653a39a38c1449372f55848fdeb22f6f6c27:
    osc-check_repo: protect staging project checks with arch conditions.

- d7570c6c05bdb45652d61c009b7cb7066affabaf:
    osc-check_repo: extract is_buildsuccess() check before _check_repo_group().
    
    _check_repo_group() will become arch specific and is_buildsuccess() already
    checks all applicable archs and does not need to be run multiple times.

- c18b622f18467aa9733aaea202460e1f75fa2900:
    osc-check_repo: mirror all archs.

- fbcab6b68fa0f08679b1e03886965fc3ae01e000:
    checkrepo: convert remaining messages to use target_archs().

- 684ab55ff960fa328130ee756b5c1739fdee991a:
    checkrepo: _whatdependson() and _builddepinfo(): utilize target_archs().

- 48e7d62c23c22980b57a49543f3da55dc4985267:
    checkrepo: is_buildsuccess(): utilize target_archs().

- c77a06d36155e5247b23bb0c535ec5085814447a:
    checkrepo: _toignore(): compress into arch loop.

- 0f5c889fd574267f8d3eacb6b45c0c4096e47e45:
    checkrepo: repositories_to_check(): utilize target_archs().
    
    Drastically simplify the logic while keeping the core the same and
    utilizing archs from project instead of hard-coded.

- 79144beee2392546cd68e249cb09d5bc186422c8:
    checkrepo: utilize the devel project for requests from Factory.
    
    Assuming requests are forwarded to Leap after being accepted into Factory
    the devel project can be reviewed just as when the requests were submitted
    to Factory.

- f09fdd0660fbeab9ed54cf51aeed9328ec3f95b9:
    checkrepo: provide target_archs() method to determine relevant archs.
    
    The default behavior is to return the archs for which the target project
    is built against, but can be used to check any project (like stagings).

- b8973f5e822d5f277f22b870b8cb994cdfc47d5b:
    checkrepo: do not decline for disabled builds.
    
    This behavior is inconsistent in that no builds is ignored, but builds that
    are disabled results in a decline. In order to allow for the request to be
    evaluated in staging it should not be declined so early.

- fcb1c64598a2069d9a91531da51c2fb3cdc07610:
    checkrepo: some spelling and grammar fixes.

Fixes #837 and #764.

### DIFF
--- a/check_maintenance_incidents.py
+++ b/check_maintenance_incidents.py
@@ -36,6 +36,8 @@ import urllib2
 import yaml
 
 from osclib.memoize import memoize
+from osclib.core import owner_fallback
+from osclib.core import maintainers_get
 
 import ReviewBot
 
@@ -47,32 +49,11 @@ class MaintenanceChecker(ReviewBot.ReviewBot):
         ReviewBot.ReviewBot.__init__(self, *args, **kwargs)
         self.review_messages = {}
 
-    @memoize(session=True)
-    def owner_fallback(self, project, package):
-        root = osc.core.owner(self.apiurl, package, project=project)
-        owner = root.find('owner')
-        if not owner or owner.get('project') == project:
-            # Fallback to global (ex Factory) maintainer.
-            root = osc.core.owner(self.apiurl, package)
-        return root
-
-    # XXX: share with checkrepo
-    def _maintainers(self, project, package):
-        """Get the maintainer of the package involved in the package."""
-        root = self.owner_fallback(project, package)
-        maintainers = [p.get('name') for p in root.findall('.//person') if p.get('role') == 'maintainer']
-        if not maintainers:
-            for group in [p.get('name') for p in root.findall('.//group') if p.get('role') == 'maintainer']:
-                url = osc.core.makeurl(self.apiurl, ('group', group))
-                root = ET.parse(osc.core.http_GET(url)).getroot()
-                maintainers = maintainers + [p.get('userid') for p in root.findall('./person/person')]
-        return maintainers
-
     def add_devel_project_review(self, req, package):
         """ add devel project/package as reviewer """
         a = req.actions[0]
         project = a.tgt_releaseproject if a.type == 'maintenance_incident' else req.actions[0].tgt_project
-        root = self.owner_fallback(project, package)
+        root = owner_fallback(self.apiurl, project, package)
 
         package_reviews = set((r.by_project, r.by_package) for r in req.reviews if r.by_project)
         for p in root.findall('./owner'):
@@ -133,7 +114,7 @@ class MaintenanceChecker(ReviewBot.ReviewBot):
                     self.logger.info("{} submitted from {}, no maintainer review needed".format(pkgname, a.src_project))
                     return
 
-        maintainers = set(self._maintainers(project, pkgname))
+        maintainers = set(maintainers_get(self.apiurl, project, pkgname))
         if maintainers:
             known_maintainer = False
             for m in maintainers:
--- a/osc-check_repo.py
+++ b/osc-check_repo.py
@@ -86,7 +86,9 @@ def _check_repo_download(self, request):
             return set()
 
     staging_prefix = '{}:'.format(self.checkrepo.staging.cstaging)
-    if staging_prefix in str(request.group):
+    if (str(request.group).startswith(staging_prefix) and
+        arch in self.checkrepo.target_archs(request.group)
+    ):
         pkglist = self.checkrepo.get_package_list_from_repository(
             request.group, 'standard', arch,
             request.src_package)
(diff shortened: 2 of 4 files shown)


lnussel:
so how do we proceed?

jberry-suse:
Digging back through the local changes I had and working on removing all devel project stuff and sticking to only looking at staging and simplifying the overall process. Based on discussions I'll also likely need to a an arch whitelist per project.

jberry-suse:
After some review, I'm just going to rebuild from scratch as I think the vast majority of the code now duplicates other features, or is irrelevant to staging workflow. Once simplified it should be much easier to confirm and understand. I am considering using `ReviewBot` as well since it seems applicable and has some of the same code.

jberry-suse:
I've compiled notes on what repochecker actually does in detail and what is actually needed. I have a early ReviewDriven based prototype which I'll work on completing.

nilxam:
should we provide some testcase to make sure refactored repochecker pass those testcase?

jberry-suse:
Similar problem to adding any testcases. PITA to mock all of OBS. I believe the way to accomplish that would be to setup real OBS for testing in a similar way that OBS tests itself. That would mean postponing this until that test setup is in place.

jberry-suse:
I ran a complete run against Leap successfully, but need to reinstate the cycles checker. My suggestion going forward is to keep Factory on the old repo checker and run the new one against Leap. Once we're satisfied that it works we can switch out the Factory one.

I am very happy with how this is turning out as the original code did things repeatedly, the same thing in multiple ways, and has a load of checks that are redundant to the staging workflow. I've also discovered a few shortcomings of the old code that should be remedied and implemented the project level comments on failure. The end result should be a massive reduction in queries and quicker filtering of requests that would fail a more expensive check which should result in a rather substantial speed up.

I'll likely provide some rough complexity stats so one can see the massive reduction.

Summary of remaining:

- refactor cycles code to make it usable
- cleanup local changes (half clean and committed)

lnussel:
We'd have to run in on 42.3 then as 15 won't get much requests. Only four weeks left for that. The bot needs to run on packagelists then so anyone in the team can override it in case of problems.

nilxam:
> My suggestion going forward is to keep Factory on the old repo checker and run the new one against Leap. Once we're satisfied that it works we can switch out the Factory one.

Fine with me. :+1:

PR #909, opened 2017-05-15
unslect: --cleanup handle superseded request that is never replaced.

jberry-suse:
- [492287](https://build.opensuse.org/request/show/492287)
  - superseded by [494593](https://build.opensuse.org/request/show/494593), but declined

Such requests will never be unselected since the request is `superseded`. This still allows for requests to be reopened or the supersede chain continued while eventually cleaning it up if left to languish.


jberry-suse:
The example is only `4` days old this this logic will not yet apply with default `7` day grace period.

PR #912, opened 2017-05-17
Fix wrong parsing from root

nilxam:
It's a regression from https://github.com/openSUSE/osc-plugin-factory/commit/8db3461a00784e999cb3218c70d16a10b2e70f38

``show_project_sourceinfo`` is returned a string of `sourceinfolist` element. ie.

```
<sourceinfolist>
  <sourceinfo XXX >
    <linked XXX/>
  </sourceinfo>
</sourceinfolist>
```

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -216,7 +216,7 @@ class CheckSource(ReviewBot.ReviewBot):
             return None
 
         # Decline the delete request against linked package.
-        links = root.findall('linked')
+        links = root.findall('sourceinfo/linked')
         if links is None or len(links) == 0:
             if not self.skip_add_reviews and self.repo_checker is not None:
                 self.add_review(self.request, by_user=self.repo_checker, msg='Is this delete request safe?')


check_source.py
```diff
@@ -215,8 +215,10 @@ def check_action_delete(self, request, action):
         except urllib2.HTTPError:
             return None
 
-        # Decline the delete request against linked package.
-        links = root.findall('linked')
+        for si in root.findall('sourceinfo'):
+            # Decline the delete request against linked package.
+            links = si.findall('linked')
```
jberry-suse: Would `links = si.findall('sourceinfo/linked')`work? Based on your xml example it seems like it would, but the previous code placed everything inside of the loop.

982bd86c6ba1b724f4266419aee54370a3c5f24a

nilxam: yeah, that's my coding style.

jberry-suse:
Is the referenced commit (982bd86c6ba1b724f4266419aee54370a3c5f24a) message a case that needs to be handled?
```
Just findall linked tag to instead of search sourceinfo
In this case, sourceinfo is not sub-element but root.
```

nilxam:
no, why?

nilxam:
the regression is from https://github.com/openSUSE/osc-plugin-factory/commit/8db3461a00784e999cb3218c70d16a10b2e70f38 , the old way is https://github.com/openSUSE/osc-plugin-factory/commit/8db3461a00784e999cb3218c70d16a10b2e70f38#diff-5167bd7d0287ecc261c2203e8d9dc068L366 which was `sourceinfo` element indeed, change to use ReviewBot is is called https://github.com/openSUSE/osc-plugin-factory/commit/8db3461a00784e999cb3218c70d16a10b2e70f38#diff-0a4a97910d5aa0961bacac3e93ea3002R191 and that one is `sourceinfolist` element. why referenced to https://github.com/openSUSE/osc-plugin-factory/commit/982bd86c6ba1b724f4266419aee54370a3c5f24a ? I couldn't understand!

jberry-suse:
The reason I referenced is that it changed this xpath logic in a _very_ similar manor to this change presumably for a reason. Looking at the history the previous commit (6d3324ee8184ca5e7666eabd6c5de07a83995e6d) adds that code which I guess never worked and the referenced commit (982bd86c6ba1b724f4266419aee54370a3c5f24a) corrects it. The message made me think that it was trying to resolve some corner case instead of a general correction and I wanted to make sure we didn't cause a different regression.

nilxam:
yes, https://github.com/openSUSE/osc-plugin-factory/commit/6d3324ee8184ca5e7666eabd6c5de07a83995e6d does not worked as we expected wrt delete request, the followup https://github.com/openSUSE/osc-plugin-factory/commit/982bd86c6ba1b724f4266419aee54370a3c5f24a corrects it and worked, https://github.com/openSUSE/osc-plugin-factory/commit/8db3461a00784e999cb3218c70d16a10b2e70f38 broke it afterward. and no, there is no any special corner case.

jberry-suse:
Yep, made sense once I saw the previous commit.

:+1:

check_source.py
```diff
@@ -216,7 +216,8 @@ def check_action_delete(self, request, action):
             return None
 
         # Decline the delete request against linked package.
-        links = root.findall('linked')
+        links = root.findall('sourceinfo/linked')
+
```
jberry-suse: Perhaps drop blank line since no longer a block of code.

PR #916, opened 2017-05-19
openSUSE Leap 42.1 is EOL (poo#19272)

okurz:


DimStar77:
dup of #914  ?

okurz:
yep, sorry. I was too late ;-) -> https://progress.opensuse.org/issues/19272

PR #929, opened 2017-05-30
check_source: accept delete repository requests.

### LINKED ISSUE #917: gnome-review-bot no longer operational
Seems the gnome-review-bot (small brother of factory-auto) is no longer functional.

It uses the same sources, just slightly different command line parameters:

> /usr/local/bin/check_source --ignore-devel --skip-add-reviews review
```
ERROR:check_source:argument of type 'NoneType' is not iterable
  File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 553, in runner
    workfunc()
  File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 521, in work
    self.checker.check_requests()
  File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 127, in check_requests
    good = self.check_one_request(req)
  File "/usr/local/bin/check_source", line 40, in check_one_request
    return super(CheckSource, self).check_one_request(request)
  File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 221, in check_one_request
    ret = func(req, a)
  File "/usr/local/bin/check_source", line 213, in check_action_delete
    result = osc.core.show_project_sourceinfo(self.apiurl, action.tgt_project, True, (action.tgt_package))
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 3780, in show_project_sourceinfo
    query.extend(['package=%s' % quote_plus(p) for p in packages])
  File "/usr/lib64/python2.7/urllib.py", line 1303, in quote_plus
    if ' ' in s:
TypeError: argument of type 'NoneType' is not iterable
```

jberry-suse:
Fixes #917.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -209,6 +209,16 @@ class CheckSource(ReviewBot.ReviewBot):
         return False
 
     def check_action_delete(self, request, action):
+        if action.tgt_repository is not None:
+            if action.tgt_project.startswith('openSUSE:'):
+                self.review_messages['declined'] = 'The repositories in the openSUSE:* namespace ' \
+                    'are managed by the Release Managers. For suggesting changes, send a mail ' \
+                    'to opensuse-releaseteam@opensuse.org with an explanation of why the change ' \
+                    'makes sense.'
+                return False
+            else:
+                self.review_messages['accepted'] = 'unhandled: removing repository'
+                return True
         try:
             result = osc.core.show_project_sourceinfo(self.apiurl, action.tgt_project, True, (action.tgt_package))
             root = ET.fromstring(result)


DimStar77:
Hmm.. this might be the right thing for a devel prj (like GNOME:Factory) running the bot... but would we want to accept a delete request of openSUSE:Factory/standard?

jberry-suse:
Not likely, I was thinking in the context of this bot being used by leaper or similar, but Factory uses directly.

For `add_role()` the bot has:
```
# Decline add_role request (assumed the bot acting on requests to Factory or similar).
```

Should we continue this course or just add a check for target project being Factory and behave differently? Alternatively always decline.

DimStar77:
my immediate reaction is a decline - but then, it's my own fault for having started to use the same bot on a devel prj (just with --no-devel option) and I think some other devel prjs would likely want to start using it too...

I propose a 'accept by bot' for * except projects in openSUSE:* namepsace

jberry-suse:
Seems reasonable, updated to do so.

check_source.py
```diff
@@ -209,6 +209,13 @@ def check_action_add_role(self, request, action):
         return False
 
     def check_action_delete(self, request, action):
+        if action.tgt_repository is not None:
+            if action.tgt_project.startswith('openSUSE:'):
+                self.review_messages['declined'] = 'removing repository'
```
DimStar77: """The repositories in the openSUSE:* namespace are managed by the Release Managers. For suggestions on changes, please send a mail to opensuse-releaseteam@opensuse.org with an explanation why you think that change makes sense""" ?

After all this is a message we give to the user as an explanation why his request is declined. "removing repository" as a decline message sounds like it would still be done.

jberry-suse: Tweaked slightly, but replaced old message.

DimStar77:
👍  - LGTM

PR #931, opened 2017-05-31
stagingapi: rework supersede logic to handle different request types.

### LINKED ISSUE #808: Handle requests of different types that supersede each other
https://build.opensuse.org/request/show/487630 not sure how it does happened

jberry-suse:
- e91e12ab68ef782364aec79ced106f00f574df7f:
    **stagingapi: rework supersede logic to handle different request types.**

- 39799f0235cbda9e96b1e5d7ac3a83a5c1e48e05:
    stagingapi: include new request ID in supersede review message.

Fixes #808.

Since the supersede logic is rather complicated to handle with the given test setup there are no tests. As such this cannot be fully tested without finding one of each case in the wild.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -537,7 +537,7 @@ class StagingAPI(object):
             # Remove the old request
             self.rm_from_prj(stage_info['prj'],
                              request_id=stage_info['rq_id'],
-                             msg='Replaced by newer request',
+                             msg='Replaced by sr#{}'.format(request_id),
                              review='declined')
             # Add the new one that should be replacing it
             self.rq_to_prj(request_id, stage_info['prj'])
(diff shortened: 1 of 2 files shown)


jberry-suse:
Successfully superseded a normal submit pair for Leap so the code would seem to have to previous functionality in tact.

jberry-suse:
The gist:

Always supersede if old is `superseded`/`revoked`

- `delete -> delete`: ignored
- `delete <-> submit`: newer is declined and requests user to revoke the other if desired
- `submit -> submit`: different source project same as before

DimStar77:
delete -> delete: the new request could be declined with a message that a delete request is already pending

jberry-suse:
@DimStar77: corrected my comment. the code declines.

DimStar77:
👍

PR #936, opened 2017-06-08
Extract patchinfo check into a function

mimi1vx:


mimi1vx:
fixes https://github.com/openSUSE/osc-plugin-factory/pull/857

PR #940, opened 2017-06-09
Split data and export *_TEST_ISSUES variable

mimi1vx:


mimi1vx:
@coolo please review or any comments?

openqa-maintenance.py
```diff
@@ -23,7 +23,7 @@
 import sys
 from datetime import date
 import md5
-import json
+import simplejson as json
```
coolo: Better make that an exception handled - as we for ElementTree in other places - simplejson has its name for a reason (I assume)

openqa-maintenance.py
```diff
@@ -79,6 +79,12 @@
 #
 #        return [p for p in packages if p.name in in_channel]
 
+with open("./osc-plugin-factory/data/kgraft.json", 'r') as f:
+    KGRAFT_SETTINGS = json.load(f)
```
coolo: please do not hardcode the checkout path. Better access a relative path __file__ - https://stackoverflow.com/a/595332

openqa-maintenance.py
mimi1vx: few lines after is ```from simplejson``` but try/expect make sense

openqa-maintenance.py
```diff
@@ -19,14 +19,25 @@
...
-import json
+
+try:
+    import simplejson as json
+except ImportError:
+    import json
+
+try:
+    from simplejson import JSONDecodeError
+except ImportError:
```
coolo: shouldn't we prefer json? CC @aplanas

aplanas: Yes please

mimi1vx: why ? json in only older version of simplejson included in stdlib

mimi1vx: and err, this is buggy we really need simplejson because json doens't support JSONDecodeError  before python 3.5

coolo: ok, then go with simplejson only and avoid the try/except

aplanas: And why do you need jsondecodeerror? The current use in 485 is kind of weird. IMHO do not justify a new dependency

mimi1vx: @aplanas not new dependency. Is from first code drop :) Btw. python-simplejson is part of SLE12 and also is faster than stdlib version + more compactible with py3 implemenation :)

openqa-maintenance.py
```diff
@@ -673,29 +498,41 @@ def calculate_repo_hash(self, repos):
...
+    @staticmethod
+    def calculate_incidents(incidents):
+        """
+        get incidet numbers from SUSE:Maintenance:Test project
+        returns dict with openQA var name : string with numbers
+        """
+        l_incidents={}
+        for kind, prj in incidents.items():
+            url = "SUSE:Maintenance:Test:" + prj
+            iid = ','.join([x.split('.')[1] for x in osc.core.meta_get_packagelist("https://api.suse.de", url)])
```
plusky: Does the url have to be hardcoded?

coolo: there is no need to hardcode the api url - at some point we might want to sync SLE and openSUSE maintenance ways

coolo: and please split this line - e.g. first get the packagelist and then split it

openqa-maintenance.py
```diff
@@ -722,7 +559,11 @@ def trigger_build_for_target(self, prj, u):
 
         buildnr = "%s-%d" % (today, buildnr + 1)
 
-        for s in u['settings']:
+        for s in data['settings']:
+            if 'incidents' in data.keys():
+                incidents = self.calculate_incidents(data['incidents'])
```
coolo: if calculate_incidents is a static method you don't need self. But I suggest not to make it static so you have the apiurl handy

openqa-maintenance.py
```diff
@@ -673,29 +498,41 @@ def calculate_repo_hash(self, repos):
             m.update(cs.text)
         return m.hexdigest()
 
-    def jobs_for_target(self, u):
-        s = u['settings'][0]
+    @staticmethod
+    def calculate_incidents(incidents):
+        """
+        get incidet numbers from SUSE:Maintenance:Test project
```
coolo: typo

openqa-maintenance.py
mimi1vx: calculate_incidents isnt static method now :)  and you need self :)

coolo:
then happy deployment. Pro-Tip: if you change the repohash calculation in production you can force a new snapshot.

But we need to change the repohash calculation anyway soonish to be based on the list of issues tracked. But not before openqa tests were changed

PR #948, opened 2017-06-15
stagingapi: submit_to_prj(): create sub-packages for adi requests.

### LINKED ISSUE #946: adi command should try to create sub-package in case multiple-specfile
adi command should to create sub-package in case multiple-specfile, otherwise sub-package potentially build failed after merged into Factory as it is never built in adi staging.

jberry-suse:
Rather non-trivial to test given testing setup. If anyone has a case to try it on that would be great.

Fixes #946.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -807,8 +807,10 @@ class StagingAPI(object):
 
         for sub_prj, sub_pkg in self.get_sub_packages(package):
             sub_prj = self.map_ring_package_to_subject(project, sub_pkg)
-            if sub_prj != subprj:  # if different to the main package's prj
-                delete_package(self.apiurl, sub_prj, sub_pkg, force=True, msg=msg)
+            # Skip inner-project links for letter stagings as they should
+            # already be in place, but for adi packages remove them.
+            if not self.is_adi_project(project) and sub_prj == project: continue
+            delete_package(self.apiurl, sub_prj, sub_pkg, force=True, msg=msg)
 
         self.set_review(request_id, project, state=review, msg=msg)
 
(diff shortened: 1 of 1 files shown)


osclib/stagingapi.py
```diff
@@ -807,8 +807,10 @@ def rm_from_prj(self, project, package=None, request_id=None,
 
         for sub_prj, sub_pkg in self.get_sub_packages(package):
             sub_prj = self.map_ring_package_to_subject(project, sub_pkg)
-            if sub_prj != subprj:  # if different to the main package's prj
-                delete_package(self.apiurl, sub_prj, sub_pkg, force=True, msg=msg)
+            # Skip inner-project links for letter stagings as they should
+            # already be in place, but for adi packages remove them.
+            if not self.is_adi_project(project) and sub_prj == project: continue
```
nilxam: I think it should be ``sub_prj == subprj`` due to ``project`` does not go through ``map_ring_package_to_subject()``

nilxam:
I'll test it on https://build.opensuse.org/project/show/openSUSE:Factory:Staging:adi:8

osclib/stagingapi.py
jberry-suse: I would expect `map_ring_package_to_subject()` to have no effect on an adi project given:
```python
if self.is_adi_project(project):
    return project
```
As such it should not matter, but seems better stylistically to check project.

nilxam:
I did a small test, this does not work, because https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/stagingapi.py#L1073 returned the empty list as the request went to adi means it doesn't exist in ring, we might need to check ``.spec`` from souce in this case.

jberry-suse:
Can that method just be adapted to look at the target package in the main project and fallback to source project for new packages?

jberry-suse:
This can likely be fixed by making a call against the source project using the same call the old repo checker used.

jberry-suse:
Closing in favor of #1110.

PR #950, opened 2017-06-15
request_splitter: fallback to source project when devel is not present.

### LINKED ISSUE #947: staging-bot should group request to adi staging per the devel project  on TW
The origin default logic of TW adi staging is grouping request from the same devel project to an adi staging in one go, however we see staging-bot does not did that as our expected.

eg. https://build.opensuse.org/project/show/openSUSE:Factory:Staging:adi:15

<pre>
Staging Bot staging-bot wrote 2 days ago

Requests (dashboard):

    request#503055 for package python3-hl7apy submitted by @DocB
    request#503058 for package ghc-sibe submitted by @psimons
</pre>

or we should adjust the parameter for staging-bot?

jberry-suse:
Fixes #947.

See issue for details.

### DIFF
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -105,10 +105,13 @@ class RequestSplitter(object):
             delta = datetime.utcnow() - created
             request.set('aged', str(delta.total_seconds() >= self.request_age_threshold))
 
+        request_type = request.find('./action').get('type')
         target = request.find('./action/target')
         target_project = target.get('project')
         target_package = target.get('package')
         devel = self.devel_project_get(target_project, target_package)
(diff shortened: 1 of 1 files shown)


jberry-suse:
Confirmed this works on package in Factory currently.

```
fallback: ('python-pycups', 'devel:languages:python')
```

nilxam:
LGTM

jberry-suse:
It occurred to me that this changes the list behavior for SLE to make it repetitive which is why Factory does not display source projects separately from group. A side benefit is that the devel related strategies can be used for SLE. Incoming PR to include SLE in hide_source exclusion.

PR #952, opened 2017-06-15
Detect duplicate binaries within a project

jberry-suse:
Fixes https://progress.opensuse.org/issues/17996.

`osc staging list` works to see output.

What is the desired outcome of the command? Doubtful there can be any automated resolution as it will likely require deleting packages. Simply print a list of duplicates?

Seems like we want to integrate with repo checker potentially and a one off command potentially similar to `devel-projects.py` that could generate a map/list in dashboard container to be picked up by repo checker and refreshed nightly. Could enhance to only rescan packages that have been updated.

### DIFF
--- a/issue-diff.py
+++ b/issue-diff.py
@@ -21,6 +21,7 @@ import osc.conf
 import osc.core
 
 from osclib.cache import Cache
+from osclib.core import package_list
 
 # Issue summary can contain unicode characters and therefore a string containing
 # either summary or one in which ISSUE_SUMMARY is then placed must be unicode.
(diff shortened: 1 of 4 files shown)


jberry-suse:
This avoids the regex parsing mentioned by re-using the methods created during my `cleanup_rings` enhancement.

jberry-suse:
@lnussel is the devel-projects style list dump and tool to check requests or current project for dups and print them?

lnussel:
a plain list of duplicated packages is fine, yes.

jberry-suse:
Debated if I should create another individual script like `devel-project.py`, but needs `StagingAPI` and such which are all provided through `osc-staging.py`. Should probably sit down and either iron out `ToolBase` or decide how to structure such tools that to be run once a day automatically rather than as part of regular commands.

Output will be located in [dashboard/duplicate_binaries](https://build.opensuse.org/package/view_file/openSUSE:Leap:42.3:Staging/dashboard/duplicate_binaries?expand=1). YAML structure: `arch -> binary -> packages (that provide it)`. I am running locally, but it will likely take a while.

I dropped the secondary call which provides parsed file names since I do not believe that is even necessary here. The package name is already known since it is one of the loops and literally want to see if any duplicate `rpm`s are provided within an architecture. As such the whole file name seems fine.

jberry-suse:
Since this was changed to include noarch the list no contains packages that provide a noarch and $arch of same name.

PR #961, opened 2017-06-20
Bot

mimi1vx:


openqa-maintenance.py
```diff
@@ -505,9 +505,7 @@ def calculate_incidents(self, incidents):
         """
         l_incidents=[]
         for kind, prj in incidents.items():
-            # TODO: unfortuanetly this works correctly only for IBS ( OBS using
-            # incident_NR )
-            l_incidents.append((kind + '_TEST_ISSUES', ','.join([x.split('.')[1] for x in osc.core.meta_get_packagelist(self.apiurl, prj)])))
+            l_incidents.append((kind + '_TEST_ISSUES', ','.join([x.replace('_', '.').split('.')[1] for x in osc.core.meta_get_packagelist(self.apiurl, prj)])))
```
coolo: please split this statement so I don't have to scroll the diff to see it

openqa-maintenance.py
```diff
@@ -505,9 +505,10 @@ def calculate_incidents(self, incidents):
         """
         l_incidents=[]
         for kind, prj in incidents.items():
-            # TODO: unfortuanetly this works correctly only for IBS ( OBS using
-            # incident_NR )
-            l_incidents.append((kind + '_TEST_ISSUES', ','.join([x.split('.')[1] for x in osc.core.meta_get_packagelist(self.apiurl, prj)])))
+            l_incidents.append(
+                ( kind + '_TEST_ISSUES', ','.join(
+                    [x.replace('_', '.').split('.')[1] for x in osc.core.meta_get_packagelist(self.apiurl, prj)]
+                )))
```
coolo: this wasn't exactly the meaning of split I was refering too - some local variables to make this more readable. But next round...

PR #963, opened 2017-06-22
Avoid start KGraft jobs for updates without correct settings

mimi1vx:


openqa-maintenance.py
```diff
@@ -474,6 +474,11 @@ def check_action_maintenance_release(self, req, a):
             settings = update.settings(a.src_project, a.tgt_project, packages, req)
             settings['INCIDENT_PATCH'] = patch_id
             if settings is not None:
+                # if is old style kgraft check if hash correctly set all options
```
jberry-suse: **if** is old style kgraft check if has**h** correctly set all options

`is old style kgraft check if all options correctly set`

PR #964, opened 2017-06-23
Complete rewrite of repo_checker based on ReviewBot with a large number of improvements

### LINKED ISSUE #482: factory-repo-checker: add comment if check is delayed
the factory-repo-checker currently only changes the review if it couldn't approve the request. There are no notifications for that kind of events. factory-repo-checker should therefore add an obs comment if it couldn't approve a request for let's say 24 hours.

### LINKED ISSUE #764: don't hardcode architectures in repochecker

### LINKED ISSUE #837: repochecker: refactor to support Leap and SLE

jberry-suse:
Well, it's finally here! The `ReviewBot` based `repo_checker`!

The primary focus was to rework to utilize the staging workflow. Anything already done by the staging worklfow was then thrown out to remove duplication and complexity. Additionally, refactoring was done where it made sense to keep things clean and pretty. As usual any custom code that could be replaced by `osc.core` or `osclib` methods was done. The end result is a significantly simplified, actually follow-able code, and faster tool.

Summary of what was lost and gained.

**gained**

- A mix of source project and staging project check were performed so some requests may be accepted based on source project results while others based on staging project. This means that not all packages in a staging are
reviewed during an install check as a goup. The new code always looks at staging group together.
- Removed inconsistent `tgt_package` vs `src_package` usage for check which could have unwanted side-effects.
- Ignores requests until staged which allows for focusing on requests that have the highest probability for success or more meaningful failures. This also means there is no review spam on the request for early failures.
- Consistent directory structure and shared download method for both staged and product rpms. Previous these were two completely different code paths the staged rpms requiring a lot of extra code.
- Dropped rpmling.log check duplicated by legal review process which removes the extra call to download `rpmlint.log` and related calls that led up to the call.
- Proper shell escaping for robustness and security.
- Handling of mirroring errors. Previosly the repo_checker could operate on stale rpms if the mirroring failed.
- Dropped the need for lastsuccess which should elevate the need for openSUSE/open-build-service#3114.
- Staging project commits to make failures visible and provide notifications to maintainers.
- Multiple arch support and whitelist configuration option. (`repo_checker-arch-whitelist` which can be in remote config!)
- Does not require builds enabled in source project which allow for SLE updates to Leap.
- MUCH simpler codebase.

**lost**

- Sub-packages not built in stagings are no longer checked. This applies to ring packages without all subs in rings in addition to non-ring packages with subpackages. The non-ring packages will be resolved by fixing #946 which is better than looking in the source project anyway.

**basic complexity analysis**

```bash
# BEFORE
grep -w for osc-check_repo.py osclib/checkrepo.py | wc -l # = 118
grep -w for osclib/cycle.py | wc -l # = 51
# enough said

# AFTER
grep -w for repo_checker.py | wc -l # = 18
grep -w for osclib/cycle.py | wc -l # = 37

# before  169
# after    55
# removed 114 (0.674556213018)
```

I may do some speed comparisons and compare the amount of queries. I expect both to be significant improvements.

**example project comment (cropped)**

The `cycle(s)` text links the the OBS repo page showing cycles with clickable links to packages. This avoids the need to make the emails extremely cluttered while still having quick clicking access to packages.

The `install check` text links to the dashboard container product `installcheck` file in case relevant.

repo_checker-comment

**deployment**

As noted in previous PR (#904) I suggest this be deployed in parallel with the old `repo_checker` to allow for testing against Leap before using in Factory and SLE as well. I already have a few improvements I will likely make to this after confirming this version works more broadly. Any thoughts on accept the PR and utilize a static copy for Factory similar to the announcer setup when the changes for Leap were accepted. Otherwise, this can be utilize directly from my branch.

**appendix**

Resolves three issues.

Fixes #837.
Fixes #764.
Fixes #482.

See commit log for additional details and insight.


jberry-suse:
See [gist](https://gist.github.com/jberry-suse/3a683755fbb8bd8f2d2b8abacfe03aff) for example of output (no failures).

jberry-suse:
Looking for opinions/agreement on my deployment plan. The [existing code should restrict to project and user](https://github.com/openSUSE/osc-plugin-factory/blob/866a28adada9f814383c92937ed5815c9d50af5c/osclib/checkrepo.py#L241), but [`ReviewBot` only has by project or by user](https://github.com/openSUSE/osc-plugin-factory/blob/866a28adada9f814383c92937ed5815c9d50af5c/ReviewBot.py#L356). Additionally, it requires a request type for project which does not seem to work with `any` or similar. Either change `ReviewBot` or run the bot multiple times for each type of request and the main and non-free projects. Alternatively, just create a new user which may be best anyway since we could drop the `factory` portion from the name (perhaps just `repo-checker`).

jberry-suse:
Secondly, someone with access will need to setup Leap:42.3 to add the agreed upon user to each review in OBS config the way it is configured for Factory.

jberry-suse:
May also be useful to run them alongside each other in Factory when making the transition which for which separate users would also be useful.

lnussel:
sorry I couldn't look at this yet

jberry-suse:
Did some work to set things up on packagelists:

- froze off `/var/lib/osc-plugins/osc-check_repo.py` by making new clone in `/home/repochecker/src/osc-plugin-factory-froze-for-repo-checker` and re-symlinking to that
- created `repo-checker` user on OBS
- created `_opensuse.org-repo-checker` on packagelists and logged into osc
- created pre-accept clone of this branch in previous user home
- created systemd service and timer: `repo-checker@.service` and `repo-checker@.timer`
- enabled timer and started a run, currently configured with debug and dry run, just to see that it it builds up mirrors and works properly

Will need to add `repo-checker` to openSUSE:Leap:42.3 required reviews before it will start updating reviews. Otherwise, once I removed `--dry` it will post comments on projects and requests when issues discovered, but still not do anything to reviews.

jberry-suse:
Fixed tests. My suggestion would be to accept this PR if no objections and make any followup changes in new PRs. Once satisfied with capabilities we can switch over Factory to use it, but for now keep it on the frozen checkout.

jberry-suse:
Currently, running cleanly against all non-failing stagings in Leap. I started a run against Factory to see what it turns up.

DimStar77:
FYI: i reverted the relinking of the plugins on pkglist VM: create-test-dvd failed completely for the entire night (with permission denied, apparmor blocks)

also, it seems like the new check repo did not see cycles happening with the unbound submission (but reports a lot of cycles as comments on :B where there are no really new ones)

jberry-suse:
As discussed in IRC, the cycles check missing is the old bot still, but likely occurs for the same reason the new one added a comment to :B. Seems like OBS `_buildepinfo` gets stale.

jberry-suse:
Per discussion with @lnussel I pushed the following commit:

- 2abb0f117979cdf2aff16e2d3d0ebe188d13113d:
    repo_checker: remove maintainer check for delete requests.
    
    This appears to be an artifact from before factory-auto/leaper require
    maintainer review for requests not created by one.

@DimStar77 @nilxam : can you confirm the reasoning here?

jberry-suse:
After resolving the pre-existing issue with `repo-checker.pl` one can see the pretty new comment on [adi:16](https://build.opensuse.org/project/show/openSUSE:Leap:42.3:Staging:adi:16) for a bogus package I created to test the new `repo_checker` in production.

nilxam:
> Per discussion with @lnussel I pushed the following commit:
>
> 2abb0f1:
> repo_checker: remove maintainer check for delete requests.
>
> This appears to be an artifact from before factory-auto/leaper require
maintainer review for requests not created by one.

> @DimStar77 @nilxam : can you confirm the reasoning here?

I think it's reasonable for delete request, *and* I'd suggest we should add a markdown CC on the request, the maintainer should be noticed at least.

jberry-suse:
@nilxam Reasonable that the maintainer has to sign off on delete? Sure. That the maintainer has to create the delete? For example the series of deletes you and I made in `Leap:42.3:Staging:D` would fail this rule and all be declined. `Leaper` will add a maintainer review as I thought `factory-auto` does in Factory right?

nilxam:
> @nilxam Reasonable that the maintainer has to sign off on delete? Sure. That the maintainer has to create the delete? For example the series of deletes you and I made in Leap:42.3:Staging:D would fail 

IMHO I don't think the DR must be created by one of maintainer.

> this rule and all be declined. Leaper will add a maintainer review as I thought factory-auto does in Factory right?

Nope, factory-auto doesn't add maintainer review eg. https://build.opensuse.org/request/show/506447 , factory-auto just add repochecker review then repochecker can pick up this delete request, that's all. release manager check what repochecker say about the delete request and evaluate it. Is leaper adds a maintainer review to the request?

And what I suggested here just that repochecker *can* add a comment remaining maintainer this delete request exists.

jberry-suse:
Misunderstood what you mean in your previous response. So it seems we are in agreement. :) (the referenced requests seems like the wrong one)

Perhaps `factory-auto` should be updated to add the review just like it does for submits and be consistent with leaper rather than keep this as a `repo-checker`. I would expected `repo-checker` to still be used for whatdependson check and installcheck with delete.

jberry-suse:
`repo-checker` detected a legitimate problem in `Leap:42.3:Staging:adi:44` to which submitters started discussion on notification email to resolve!! Big improvement over nearly invisible error output.

After one request was unselected (after being revoked) the comment was updated since the removed package was one of the offending ones. :)

```
can't install trytond_account_invoice_stock-4.2.0-1.1.noarch:
  package trytond_account_invoice_stock-4.2.0-1.1.noarch requires trytond, but none of the providers can be installed
  package trytond-4.2.4-1.3.noarch requires python3-relatorio, but none of the providers can be installed
  nothing provides python3-pycha needed by python3-relatorio-0.6.4-1.1.noarch
found conflict of python-PyWebDAV-0.9.8-3.2.noarch with python3-PyWebDAV3-GNUHealth-0.10.1-2.1.noarch:
  - /usr/bin/davserver
```

jberry-suse:
rebase to resolve conflicts post #952.

jberry-suse:
Ran into a unique corrupt request case that was not handled by standard cleanup. Handled in `repo_checker` and fixed request.

```
osc staging repair 500626
```

Change deployed to packagelists.

jberry-suse:
For the record a [rough comparison of the old and new checkers](https://gist.github.com/jberry-suse/3c6a313ed18545cffe06518929baa125). Both checkers had the same run performed twice to allow for mirroring and caching to more accurately reflect running in a loop. The new checker was limited to `x86_64` to be fair and adi stagings were used since the checks are more similar for those. Also note that the old checker makes a series of individual download requests for staged packages along with a variety of other queries. The test set was 8 requests from 8 different adi stagings. The time would be virtually identical if all requests from each staging were included for the new checker as it just has to update reviews. The old checker would require a series of `lastsuccess` and package build result walking which will take a significant time. The new checker performs an installcheck on all packages in letter stages which takes a bit of time, but is beneficial and relatively small compared to the old checker additional api calls. Also note the old checker gains a 30 second head start by using cached `_meta` API calls for request list while the new checker has to wait 30 seconds for dashboard JSON API to return (only performed once per run).

Anyway, with that aside:

new:
```
$ time ./repo_checker.py --dry --debug --osc-debug id 509398 509207 509750 508134 501668 508137 509123 502896
real    4m11.848s
user    3m11.275s
sys     0m6.911s
```

old:
```
$ time osc -d check --verbose --dry 509398 509207 509750 508134 501668 508137 509123 502896
real    14m55.320s
user    4m0.497s
sys     0m12.536s
```

73% time reduction. Based on observation of the two on packagelists I would bet the real world reduction is even larger, especially considering the new checker will ignore requests that fail to build in stagings and does not keep looping over unstaged requests that may not be satisfied until staged.

To top it off Dominique ran into a case where the old checker was stuck on cycle in the devel project that did not exist in the staging so the new checker was happy. :)

PR #965, opened 2017-06-23
Second part of data split for openqa-maintenance bot

mimi1vx:


openqa-maintenance.py
```diff
@@ -240,127 +242,15 @@ def settings(self, src_prj, dst_prj, packages, req=None):
...
-                'VERSION': '42.3',
-                'FLAVOR': 'Maintenance',
-                'ARCH': 'x86_64',
-                'ISO': 'openSUSE-Leap-42.3-DVD-x86_64.iso',
-            }),
-    ],
-}
+PROJECT_OPENQA_SETTINGS = {}
+
+for i,j in _BASE_POS.items():
```
coolo: what is BASE_POS and why the _?

mimi1vx: dictionary filled by incidents.json and '_ ' marks it as internal, but .. You are right , in reality isn't needed

openqa-maintenance.py
```diff
@@ -240,127 +240,16 @@ def settings(self, src_prj, dst_prj, packages, req=None):
...
-                'ARCH': 'x86_64',
-                'ISO': 'openSUSE-Leap-42.3-DVD-x86_64.iso',
-            }),
-    ],
-}
+PROJECT_OPENQA_SETTINGS = {}
+
+with open(opa.join(data_path, "data/incidents.json"), 'r') as f:
+    for i,j in json.load(f).items():
+        if i.split(":")[0] == 'SUSE':
```
jberry-suse: `i.startswith('SUSE:')`

openqa-maintenance.py
```diff
@@ -240,127 +240,16 @@ def settings(self, src_prj, dst_prj, packages, req=None):
...
+            PROJECT_OPENQA_SETTINGS[i] = [SUSEUpdate(k) for k in j]
+        elif i.split(":")[0] == 'openSUSE':
```
jberry-suse: `i.startswith('openSUSE:')`

coolo:
Note that you omited the : now - I don't think it matters though - it's unlikely we will have SUSEopen soonish :)

PR #967, opened 2017-06-23
Test DVDs: use xdelta3 instead of xdelta

DimStar77:


DimStar77:
@lnussel that means you also need xdelta3 in ring1 in Leap 42.3 - without this diff, it's openQA failing because xdelta3 would not be installable (xdelta is scheduled to be removed)

jberry-suse:
Added xdelta3 to ring-1.

PR #968, opened 2017-06-26
Use frozenlinks for submission from the specific project

nilxam:
Create frozenlinks in case had matched source project in expand-allpkgs-from-same-srcprj config, the frozenlinks contains all packages from the specific source project.

Tested it in real: https://build.opensuse.org/project/show/openSUSE:Factory:Staging:adi:181

https://progress.opensuse.org/issues/19830

### DIFF
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -5,6 +5,7 @@ from colorama import Fore
 from osc import oscerr
 from osc.core import delete_project
 from osc.core import show_package_meta
+from osc import conf
 
 from osclib.select_command import SelectCommand
 from osclib.supersede_command import SupersedeCommand
@@ -15,6 +16,7 @@ from xml.etree import cElementTree as ET
 class AdiCommand:
     def __init__(self, api):
         self.api = api
+        self.config = conf.config[self.api.project]
 
     def check_adi_project(self, project):
         query_project = self.api.extract_staging_short(project)
@@ -58,6 +60,7 @@ class AdiCommand:
             self.check_adi_project(p)
 
     def create_new_adi(self, wanted_requests, by_dp=False, split=False):
+        source_projects_expand = self.config.get('source_projects_expand', '').split()
         requests = self.api.get_open_requests()
         splitter = RequestSplitter(self.api, requests, in_ring=False)
         splitter.filter_add('./action[@type="submit"]')
@@ -96,7 +99,9 @@ class AdiCommand:
                 # Only create staging projec the first time a non superseded
                 # request is processed from a particular group.
                 if name is None:
-                    name = self.api.create_adi_project(None)
+                    use_frozenlinks = group in source_projects_expand and not split
+                    name = self.api.create_adi_project(None,
+                            use_frozenlinks, group)
 
                 if not self.api.rq_to_prj(request_id, name):
                     return False
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -41,7 +41,9 @@ from osc.core import http_GET
 from osc.core import http_POST
 from osc.core import http_PUT
 from osc.core import rebuild
+from osc.core import search
 from osc.core import show_project_meta
+from osc.core import show_project_sourceinfo
 from osc.core import streamfile
 
 from osclib.cache import Cache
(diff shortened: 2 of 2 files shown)


adrianschroeter:
Can't you use the api call for creating frozen links meanwhile?

`POST /source/$PROJECT/_project?cmd=freezelink`

should do.

nilxam:
@adrianschroeter If I understand it correct it does create(or update to current) per the project link, and it can not put customize specific packages within frozenlinks, right?

osclib/adi_command.py
```diff
@@ -58,6 +60,10 @@ def check_adi_projects(self):
             self.check_adi_project(p)
 
     def create_new_adi(self, wanted_requests, by_dp=False, split=False):
+        needed_expand_allpkgs = self.config.get('expand-allpkgs-from-same-srcprj')
```
jberry-suse: I am never a fan of mixing abbreviations, but perhaps we can at least drop "same" since it does not really apply. It's a whitelist of source projects from which to expand all packages.

Perhaps even, `expand-from-source-projects` or `source-projects-expand` or `...-to-expand`.

Similar rule could be applied to variable names as they are rather different.

Better still:

```python
source_projects_expand = self.config.get('source_projects_expand', '').split()
```

Then the rest of the code can just be dropped.

osclib/stagingapi.py
```diff
@@ -1570,7 +1570,42 @@ def _candidate_adi_project(self):
             adi_index = i + 2
         return self.adi_prj_from_number(adi_index)
 
-    def create_adi_project(self, name):
+    def update_adi_frozenlinks(self, name, src_prj):
+        xpath = "(@project='%s' and devel/@project="\
+                "'%s')" % (self.project, src_prj)
+        query = {'match': xpath}
+        try:
+            url = self.makeurl(['search', 'package'], query=query)
```
jberry-suse: `osc.core.search()` should be able to handle this.

```python
search(apiurl, {'package': xpath})['package']
```

osclib/stagingapi.py
```diff
@@ -1570,7 +1570,42 @@ def _candidate_adi_project(self):
...
+        except urllib2.HTTPError, e:
+            raise e
+
+        # all packages had matched devel project defined
+        pkglist = [p.attrib['name'] for p in collection.findall('package')]
+
+        flink = ET.Element('frozenlinks')
+        fl_prj = ET.SubElement(flink, 'frozenlink', {'project': self.project})
+
+        url = self.makeurl(('source', self.project), {'view': 'info', 'nofilename': '1'})
```
jberry-suse: `osc.core.show_project_sourceinfo()`

osclib/stagingapi.py
```diff
@@ -1581,27 +1616,49 @@ def create_adi_project(self, name):
...
-            <path project="{1}" repository="standard"/>
-            <arch>x86_64</arch>
-          </repository>
-        </project>""".format(name, self.project, self.extract_adi_number(name))
+        if use_frozenlinks:
+            meta = """
+            <project name="{0}">
+              <title></title>
+              <description></description>
+              <link project="{1}"/>
```
jberry-suse: If this is the only line different surely complete duplication of the string can be avoided. Either parse as XML and add element, or add a string placeholder.
```
'<description></description>{}'.format("<link project="{1}"/>\n")
```
etc

osclib/adi_command.py
```diff
@@ -96,7 +102,11 @@ def create_new_adi(self, wanted_requests, by_dp=False, split=False):
                 # Only create staging projec the first time a non superseded
                 # request is processed from a particular group.
                 if name is None:
-                    name = self.api.create_adi_project(None)
+                    use_frozenlinks = False
```
jberry-suse: `use_frozenlinks = group in expand_allpkgs` and drop extra two lines

osclib/stagingapi.py
```diff
@@ -1570,7 +1572,40 @@ def _candidate_adi_project(self):
...
 
-    def create_adi_project(self, name):
+    def update_adi_frozenlinks(self, name, src_prj):
+        search = {
+            'package': "@project='%s' and devel/@project='%s'" % (self.project, src_prj),
+        }
+        try:
+            collection = osc.core.search(self.project, **search)['package']
+        except urllib2.HTTPError, e:
+            raise e
```
jberry-suse: Any reason to catch the exception if only raised when caught, otherwise drop try/except?

osclib/stagingapi.py
```diff
@@ -1570,7 +1572,40 @@ def _candidate_adi_project(self):
...
+                if linked.get('package') in pkglist:
+                    ET.SubElement(fl_prj, 'package', {'name': pkg, 'srcmd5': si.get('lsrcmd5'), 'vrev': si.get('vrev')})
+
+        # commit frozenlinks
+        url = self.makeurl(['source', name, '_project', '_frozenlinks'], {'meta': '1'})
+        l = ET.tostring(flink)
+        try:
+            http_PUT(url, data=l)
+        except urllib2.HTTPError, e:
+            raise e
```
jberry-suse: Same as above. Any reason to catch the exception if only raised when caught, otherwise drop try/except?

PR #969, opened 2017-06-26
Update testdvd template

nilxam:
Mainly to added new requirement for testing textmode and docker, and adjust the
template per patterns change.


DimStar77:
LGTM - do we want/need the Leap equivalent too? @lnussel

lnussel:
would be fine with me if Max does it :) Needs quite some ring tuning I guess. At this point we first need to clean up the current mess I'd say

nilxam:
I'll wait 42.3 settled then add textmode test to it, yes, adjust ring and add temporary testdvd template is required. Let's merge this first.

PR #988, opened 2017-07-06
AcceptCommand: update version of all products found in the project

DimStar77:
in opeSUSE:Factory, for example, there is also openSUS-Kubic.product and the
versions are supposed to stay in sync.

We newly find all product definitions inside a project and update all of them,
with the same limiation as before: only if version was a 8 digit version it will
be updated to the current date stamp

Additionally, this is a preparation for the new product builder, as _product is no
longer hardcoded, but found.

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -191,21 +191,20 @@ class AcceptCommand(object):
         # XXX TODO - This method have `factory` in the name.  Can be
         # missleading.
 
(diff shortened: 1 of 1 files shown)


DimStar77:
Even though my main motivation for this was the new product builder, I actually squashed https://github.com/openSUSE/osc-plugin-factory/issues/937 as a side effect too

PR #993, opened 2017-07-07
Speed up duplicate binaries check

lnussel:
We can leverage an api call that returns all binaries.

### DIFF
--- a/osclib/check_duplicate_binaries_command.py
+++ b/osclib/check_duplicate_binaries_command.py
@@ -1,42 +1,64 @@
 from __future__ import print_function
-from osc.core import get_binarylist
+from osc.core import http_GET
 from osclib.core import package_list
 from osclib.core import target_archs
+from xml.etree import cElementTree as ET
+from pprint import pprint
 import re
+import urllib2
 import yaml
 
 
 class CheckDuplicateBinariesCommand(object):
     def __init__(self, api):
         self.api = api
+        # some packages create packages with the same name but
+        # different architecture than built for.
+        self.ignore_extra_archs = {
+            'i586': {
+                'glibc.i686': ('i686',)
+            },
+            'x86_64': {
+                'syslinux': ('s390x', 'ppc64le',)
+            }
+        }
 
     def perform(self, save=False):
         duplicates = {}
         for arch in sorted(target_archs(self.api.apiurl, self.api.project), reverse=True):
-            print('arch {}'.format(arch))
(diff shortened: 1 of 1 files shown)


osclib/check_duplicate_binaries_command.py
```diff
@@ -13,30 +16,41 @@ def __init__(self, api):
...
+		    parch = match.group(4)
+		    if parch in ('src', 'nosrc'):
+			continue
 
-                    name = match.group(1)
-                    if name in binaries:
-                        print('DUPLICATE', package, binaries[name], name)
+		    # XXX: we assume that proper export filters for
+		    # the architecures exist. Examples are glibc and
+		    # syslinux.
```
jberry-suse: Not sure what this comment is trying to convey.

In general a `noarch` does duplicate a `x86_64` and the same should be true of other arch combinations. As such restricting binary check to `parch` seems wrong. `glibc` being a special case since it is allowed to have both `i586` and `i686`. As such it seems to make sense to provide a whitelist or other method for filtering such cases out.

osclib/check_duplicate_binaries_command.py
```diff
@@ -13,30 +16,41 @@ def __init__(self, api):
...
 
-                        if name not in duplicates[arch]:
-                            # Only add the first package found once.
-                            duplicates[arch][name] = [binaries[name]]
+                    name = match.group(1)
+                    if name in binaries[parch]:
+			duplicates.setdefault(arch, {})
+                        duplicates[arch].setdefault(name, set()).add(package)
+                        duplicates[arch][name].add(binaries[parch][name])
+                        #print('DUPLICATE binary {} built by {}'.format(name, ', '.join(duplicates[arch][name])))
```
jberry-suse: left over

osclib/check_duplicate_binaries_command.py
```diff
@@ -13,30 +16,41 @@ def __init__(self, api):
     def perform(self, save=False):
         duplicates = {}
         for arch in sorted(target_archs(self.api.apiurl, self.api.project), reverse=True):
-            print('arch {}'.format(arch))
+	    url = self.api.makeurl(['build', self.api.project, 'standard', arch], { 'view': 'binaryversions' })
+	    try:
+		data = http_GET(url)
+	    except urllib2.HTTPError:
+		return
```
jberry-suse: Any reason to bother catching exception since this will just hide the error and stop immediately? Would seem preferable to just let the exception free.

PR #994, opened 2017-07-07
Ignore failures to delete adi's

lnussel:
If someone branched the ADI we are not able to delete it. Just ignore that.

### DIFF
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -1,4 +1,5 @@
 import json
+import urllib2
 
 from colorama import Fore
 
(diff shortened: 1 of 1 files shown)


okurz:
yep, sorry ;-)

lnussel:
no idea. the tool would keep complaining about the adi on next run though. So the adi won't linger unnoticed at least

DimStar77:
wihtout force it would linger until the last user removed the reference to the adi from his project - that can be a looooong time in my experience

PR #995, opened 2017-07-07
improve biarch tool

lnussel:


biarchtool.py
```diff
@@ -46,28 +46,70 @@ def __init__(self, project):
...
+
+    def _has_baselibs_in_files(self, files):
+        for n in files.findall("./entry[@name='baselibs.conf']"):
+            return True
+        return False
+
+    def _is_link(self, files):
+        for n in files.findall("./entry[@name='_link']"):
+            return True
+        return False
```
jberry-suse: Perhaps:
```
return len(files.findall("./entry[@name='_link']")) > 0
```
or
```
return files.find("./entry[@name='_link']") is not None
```
to avoid `findall()`.

biarchtool.py
```diff
@@ -46,28 +46,70 @@ def __init__(self, project):
...
+                    'skelcd-openSUSE',
+                    'plasma5-workspace',
+                    ]),
+                }
+
+
+    def _has_baselibs_in_files(self, files):
+        for n in files.findall("./entry[@name='baselibs.conf']"):
+            return True
+        return False
```
jberry-suse: Same as below.

biarchtool.py
```diff
@@ -46,28 +46,70 @@ def __init__(self, project):
...
+        if self._has_baselibs_in_files(files):
             logger.debug('%s has baselibs', package)
             ret = True
-            return True
+        elif self._is_link(files):
+            files = ET.fromstring(self.cached_GET(self.makeurl(['source', self.project, package], {'expand':'1'})))
+            if self._has_baselibs_in_files(files):
+                logger.warn('%s is linked to a baselibs package', package)
         self._has_baselibs[package] = ret
+        #logger.debug('%s has no baselibs', package)
```
jberry-suse: left over?

biarchtool.py
```diff
@@ -46,28 +46,70 @@ def __init__(self, project):
...
 
         ret = False
         files = ET.fromstring(self.cached_GET(self.makeurl(['source', self.project, package])))
-        for n in files.findall("./entry[@name='baselibs.conf']"):
+        if self._has_baselibs_in_files(files):
             logger.debug('%s has baselibs', package)
             ret = True
-            return True
+        elif self._is_link(files):
+            files = ET.fromstring(self.cached_GET(self.makeurl(['source', self.project, package], {'expand':'1'})))
```
jberry-suse: Perhaps replace both of these calls with `osc.core.meta_get_filelist()` returns list of files names. So the above xpath checks could be replaces with `'_link' in files`.

biarchtool.py
```diff
@@ -46,28 +46,70 @@ def __init__(self, project):
...
+            return True
+        if package in self.whitelist[self.arch]:
+            logger.debug('%s is whitelisted', package)
+            return True
         r = self.has_baselibs(package)
-        if not r and package in self.rdeps:
+        if r:
+            return r
+        if package in self.rdeps:
+            #logger.debug('checking rdeps for %s (%d)', package, len(self.rdeps[package]))
```
jberry-suse: left over?

biarchtool.py
lnussel: that would not use caching though

jberry-suse: Using the explicit `cached_GET()` no. Too many different ways.

biarchtool.py
```diff
@@ -46,28 +46,64 @@ def __init__(self, project):
...
+                'i586': set([
+                    'belle-sip',
+                    'release-notes-openSUSE',
+                    'openSUSE-EULAs', # translate-toolkit
+                    'skelcd-openSUSE',
+                    'plasma5-workspace',
+                    ]),
+                }
+
+    def get_filelist(self, project, package, expand = False):
```
jberry-suse: nit: parameter defaults typically have no space (same during usage calls below)

PR #997, opened 2017-07-10
comments: avoid attempt to delete non-existent parent comments.

### LINKED ISSUE #911: accept: tries and fails to delete already deleted parent comments
This happened now twice in a row on accepting stagings - so somehow, somebody broke something

```
Server returned an error: HTTP Error 404: Not Found
Couldn't find Comment with 'id'=231339
```

This happens at the very end after accepting a group - and as the group is then empty and nothing left to be done, can't just be reproduced for a test again...

I'll try to remember running osc -d next tim around, hopefully being able to gather more info

jberry-suse:
```
# Parent comments that have been removed are still returned
# when children exist and are authored by _nobody_. Such
# should not be deleted remotely, but only marked internally.
```

Fixes #911.

### DIFF
--- a/osclib/comments.py
+++ b/osclib/comments.py
@@ -160,10 +160,14 @@ class CommentAPI(object):
             if comment['parent']:
                 parents.append(comment['parent'])
 
-        for id_ in comments.keys():
-            if id_ not in parents:
-                self.delete(id_)
-                del comments[id_]
+        for comment in comments.values():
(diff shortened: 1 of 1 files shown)


jberry-suse:
I confirmed this with a test project where I created a comment, replied, and then deleted the original parent comment. When attempting to remove via `delete_from()` the 404 error occurs, with the new code no problem.

PR #998, opened 2017-07-12
dist/spec: initial specfile including relevant sub-packages.

### LINKED ISSUE #989: Create OBS package with subpackages for various deployed bits vs staging tools
I have considered this for a while, but it was requested in email conversation as well so it seems worth at least creating an issue. I do not imagine this will be too difficult accomplish and seems better than running out of git checkout for those not actively developing the tools.

This may also be a reason to provide some sort of releases with decent release notes that alert folks to the new features. Also should make it easier to keep deployments in-sync rather than ad-hock updates.

jberry-suse:
Definitely some areas that should be improved, but those can be done in coordination with `Makefile` updates once a base spec file exists.

Having an updated package was something I have considered for a while and was requested when communicating with others. There are a wide range of tools present in this repo that are not applicable to everyone and as such I have created a number of subpackages. I used the name `openSUSE-release-tools` as that seems more applicable than the current repository name and others like `SUSE-...` given other package names and such. Eventually it might make sense to move this repository, but for now I'll stick with the package.

For SLE some python packages are missing (python-cmdln, python-httpretty, python-mock), but are available on IBS in NON_Public project. I would prefer not to have to build this package twice (once on OBS and once on IBS), but short of those dependencies becoming available or branched on OBS in some fashion there are not too many options.

Otherwise, there are a number of odd/inconsistent behaviors performed by Makefiles that should be cleaned up, but this spec packages what currently exists in appropriate sub-packages.

Seems to make sense to submit to `openSUSE:Tools` project on OBS and have this package automatically updated once a week. I can imagine a variety of methods for auto-updating and other packages already have various methods in place. If others know the preferred method please let me know.

Once this all makes it through and is built in proper repos a next step would be to re-deploy existing services on packagelists and IBS like leaper using the package. This should make moving and cleaning things up a lot more controlled.

As a side-note the `osc` package is ironically using its deprecated plugin directory. The plugin packages I looked at use a conditional for the directory that should pickup the proper location so I'll submit a request to update the osc package to provide both for a grace-period. The `Makefile` in this repo uses the new location unconditional so I package it that way, but have to include the main directory. Packagelists is also using the deprecated directory.

Lastly, I am sure that the `make` process should be updated to provide the compiled python files and it might also make sense to package `osclib` as a proper python modules so other tools could be spawned in separate repositories while still benefiting from that code.

See the [package on OBS](https://build.opensuse.org/package/show/home:jberry:release-tools/openSUSE-release-tools).

Fixes #989.


jberry-suse:
Also, need to come up with a consistent `.service` file style that works for both OBS and IBS and related users for those services for `osc` credentials. I have some thoughts on that I need to formalize.

jberry-suse:
Created https://build.opensuse.org/request/show/509599 to `Provide non-deprecated plugin directory and macro for plugins.` and updated my package, but I'll save updating the spec here until it is accepted.

jberry-suse:
Rebased post #964 and provided travis-ci integration for build spec file.

jberry-suse:
The spec build seems to fail due to Tumbleweed mirror issues (thread on Factory), the build works locally.

jberry-suse:
As suspected, hard-coding a working mirror resolves the problem.

PR #1004, opened 2017-07-14
travis: add deployment to OBS package on master branch commit.

jberry-suse:
Should be updated to point to proper home once accepted. Could also use a different user, not sure if there is a standardized one for this type of deployment.

Might also be worth submitting a proper OBS deployment plugin for Travis.


jberry-suse:
I want to get the various pieces moving, but feel free to comment regarding followups.

PR #1013, opened 2017-07-18
check_source: request review team to review delete requests.

### LINKED ISSUE #983: factory-auto: add maintainer review for delete requests
Per #964 the `repo-checker` will no longer require deletes originate from a maintainer, but the maintainer should still sign-off. As such it seems best to have `factory-auto` add a review for the maintainer just like it does for submits.

jberry-suse:
Since the new repo_checker does not require maintainer submission it makes sense to have some human review of delete requests.

Fixes #983.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -221,6 +221,8 @@ class CheckSource(ReviewBot.ReviewBot):
         # Decline the delete request against linked package.
         links = root.findall('sourceinfo/linked')
         if links is None or len(links) == 0:
+            if self.review_team is not None:
+                self.add_review(self.request, by_group=self.review_team, msg='Please review delete request')
             if not self.skip_add_reviews and self.repo_checker is not None:
                 self.add_review(self.request, by_user=self.repo_checker, msg='Is this delete request safe?')
             return True


DimStar77:
Makes no sense to me - the review team is responsible for packaging guidelines - a delete request can't possibly violate them

jberry-suse:
Was looking for a group that could be responsible for human checking delete requests make sense. I would imagine many are related to packaging changes, renames or splits which the team reviews, or un-maintained packages which just need a sign-off. You indicated the maintainer review is bad for the latter. Otherwise, could assign to `factory-maintainers` similar to `leap-reviewers`.

Beyond that seems stuck with maintainer review to ensure some human signs off on it.

DimStar77:
Seems like having the 'package' set as reviewer is the best course then - if the prj maintainer reacts, fine - if he doesn,t and the devel prj is setup correctly, factory-maiintainers is 'inherited maintainer' from the group and can accept the review instead

jberry-suse:
Alright, so back to the original logic I proposed in the issue. It would seem to make sense to re-use `MaintenanceChecker` logic for doing so.

PR #1014, opened 2017-07-18
check_source: utilize project config to allow for remote entries.

jberry-suse:
- 0ec6d70981096de7c0459d75cbfbeac1f54ef7b5:
    check_source: utilize project config to allow for remote entries.
    
    Rather than updating a local whitelist file for temporary devel project
    whitelist entries it makes more sense to allow for remote configuration.

- 03c46f18d77b926aa4a6a6bc215a3c5950cb9cee:
    ReviewBot: provide staging_api() from repo_checker.

Seems like this is a bit more practical than having PRs to add temporary entries. Instead just add them to remote config.

```
devel-whitelist = project1 some:other:project project3000
```

If accepted, wiki should be updated to indicate a ping in IRC unless we want to accept SRs to dashboard container. :)


jberry-suse:
I went ahead and checked the existing projects in whitelist:

```
devel:languages:D
hardware:sdr
isv:ownCloud:desktop
network:cluster:ohpc
network:mail:zarafa
Novell:NTS
network:messaging:matrix
security:tls
server:messaging
server:php:extensions:php7
system:snappy
X11:Solus
devel:openQA:tested
```

Of those the following are still not devel projects:

```
isv:ownCloud:desktop (last updated 1 year ago, newer subprojects)
network:mail:zarafa (no longer exists)
Novell:NTS (last updated 3 years ago)
network:messaging:matrix (last updated 9 days)
server:messaging (last updated 7 days ago)
system:snappy (last updated 2 months ago)
X11:Solus (last updated 24 days ago)
```

I placed all except `network:mail:zarafa` in the remote config so they will continue to be whitelisted post this change. Feel free to further prune the list.

DimStar77:
works for me (actually we had this in the past) - make sure to update the doc at https://en.opensuse.org/openSUSE:How_to_contribute_to_Factory#How_to_request_a_new_devel_project then though, so we can point contributors at the right way to get a new devel prj on the whitelist

jberry-suse:
@DimStar77 as in a file in dashboard container? Was it intentionally changed or side-effect of some bot change?

jberry-suse:
Also where exactly? The oldest white-list I find is https://github.com/openSUSE/osc-plugin-factory/commit/89b122da294c9f41b2139c27fece7d2b16b838e1#diff-8f33b35196c4caa105bb6675c869a2a4R219 (6 years ago) and it appears to be introducing the concept and not switching from some other source.

DimStar77:
> Also where exactly? 

Actually, I was wrong - this was embedded into the .py file directly - and later split at least in an own file... so it's just logical to move it into the Staging/dashboard container now

jberry-suse:
Yeah, I split it when I rewrote bot based on ReviewBot. Ok, just making sure there was not a reason for wanted it local.

DimStar77:
> works for me (actually we had this in the past) - make sure to update the doc at https://en.opensuse.org/openSUSE:How_to_contribute_to_Factory#How_to_request_a_new_devel_project then though, so we can point contributors at the right way to get a new devel prj on the whitelist

That part is missing; we currently don't tell our users how 'to do it right'

jberry-suse:
Indeed. Made the relevant changes https://en.opensuse.org/index.php?title=openSUSE%3AHow_to_contribute_to_Factory&type=revision&diff=120672&oldid=119010. Although packagelists has not yet been updated.

PR #1017, opened 2017-07-19
check_source: change repo_checker default to new OBS username.

jberry-suse:
To distinguish from the old factory-repo-checker and use a generic name applicable to Leap as well.

Seems like the best plan to use the same user for identical bots to remove any possible confusion by being consistent. There are no other references, outside of tests, in this repo, and I would not expect them elsewhere.


jberry-suse:
This PR should likely not be accepted until ready to make the switch.

PR #1025, opened 2017-07-27
WIP: Major rework of the openqa bot 

coolo:
I'm working on this for a week and more - it's no longer looking at requests, but at incident projects and test updates and then only aggregates into the request if it opened.

This way the release managers get to know about problems in incident *before* they open the request into testing.

But I didn't find a good way to make this work for opensuse too - and the kgraft testing needs work too.

### DIFF
--- a/openqa-maintenance.py
+++ b/openqa-maintenance.py
@@ -910,6 +910,7 @@ class OpenQABot(ReviewBot.ReviewBot):
         incident_project = str(job['project'])
         comment_info = self.find_obs_request_comment(project_name=incident_project)
         comment_id = comment_info.get('id', None)
+        comment_build = str(comment_info.get('revision', ''))
 
         openqa_posts = []
         for prod in API_MAP.keys():
@@ -918,12 +919,12 @@ class OpenQABot(ReviewBot.ReviewBot):
         for s in openqa_posts:
             jobs = self.incident_openqa_jobs(s)
             # take the project comment as marker for not posting jobs
-            if not len(jobs) and comment_info.get('revision', '') != job['openqa_build']:
+            if not len(jobs) and comment_build != str(job['openqa_build']):
                 if self.dryrun:
                     print 'WOULD POST', json.dumps(s, sort_keys=True)
                 else:
                     ret = self.openqa.openqa_request('POST', 'isos', data=s, retries=1)
-                openqa_jobs += self.incident_openqa_jobs(s)
+                    openqa_jobs += self.incident_openqa_jobs(s)
             else:
                 print s, 'got', len(jobs)
                 openqa_jobs += jobs
--- a/openqa-maintenance.py
+++ b/openqa-maintenance.py
@@ -436,7 +436,6 @@ class OpenQABot(ReviewBot.ReviewBot):
         started = []
         # then check progress on running incidents
         for req in self.requests:
-            # just patch apiurl in to avoid having to pass it around
             jobs = self.request_get_openqa_jobs(req, incident=True, test_repo=True)
             ret = self.calculate_qa_status(jobs)
             if ret != QA_UNKNOWN:
@@ -472,6 +471,9 @@ class OpenQABot(ReviewBot.ReviewBot):
             self.logger.warn("not handling %s" % a.tgt_project)
             return None
 
+        # TODO - this needs to be moved
+        return None
+
         packages = []
         # patchinfo collects the binaries and is build for an
         # unpredictable architecture so we need iterate over all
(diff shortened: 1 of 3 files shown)


coolo:
Hmm, the api is a bit above my head right now. Perhaps coffee helps

openqa-maintenance.py
```diff
@@ -91,73 +96,191 @@
 with open(opa.join(data_path, "data/repos.json"), 'r') as f:
     TARGET_REPO_SETTINGS = json.load(f)
 
+with open(opa.join(data_path, "data/apimap.json"), 'r') as f:
+    API_MAP = json.load(f)
+
+MINIMALS = dict()
```
mimi1vx: this should be 'set'

openqa-maintenance.py
```diff
@@ -91,73 +96,191 @@
...
+        root = ET.fromstring(repomd.text)
+
+        cs = root.find(
+            './/{http://linux.duke.edu/metadata/repo}data[@type="updateinfo"]/{http://linux.duke.edu/metadata/repo}location')
+        url = repo + cs.attrib['href']
+
+        # python 3 brings gzip.decompress, but with python 2 we need to store to
+        # temporary file to uncompress
+        repomd = requests.get(url).content
+        tfile = NamedTemporaryFile()
```
mimi1vx: whole python3 magic :

```
def decompress(data):
    """Decompress a gzip compressed string in one shot.
    Return the decompressed string.
    """
    with GzipFile(fileobj=io.BytesIO(data)) as f:
        return f.read()
```

jberry-suse:
hmm, assumed this was going to be changed to non-WIP in title once ready for final review and perhaps commits like "one more" changed or squashed. Rather messy to merge this as is, that's why git-rebase exists. :(

jberry-suse:
Also looks like the one comment I made regarding using the comment code that we now have rather than adding/maintaining yet another copy was also left unchanged...

openqa-maintenance.py
```diff
@@ -927,9 +969,9 @@ def test(self):
                 result = 'declined'
                 state = 'done'
             comment = "<!-- openqa state=%s result=%s revision=%s -->\n" % (state, result, job['base'].get('openqa_build'))
-            comment += "\n" + msg
+            comment += "\nCC @coolo\n" + msg
 
-            #print comment
+            # print comment
```
jberry-suse: huh?

openqa-maintenance.py
```diff
@@ -915,7 +957,7 @@ def test(self):
                     openqa_jobs += jobs
             if not openqa_done or len(openqa_jobs) == 0:
                 continue
-            #print openqa_jobs
+            # print openqa_jobs
```
jberry-suse: ??

openqa-maintenance.py
```diff
@@ -879,34 +920,35 @@ def check_product(self, job, product_prefix):
...
         for inc in requests.get('https://maintenance.suse.de/api/incident/active/').json():
-            if not inc in ['a4871', 'a5146', 'a2129', '5219', '5217', '5230']: continue
-            #if not inc.startswith('52'): continue
+            if not inc in ['a4871', 'a5205', 'a2129', 'a5219', '5258', '5223']:
+                continue
+            if not inc.startswith('52'):
+                continue
             print inc
-            #continue
+            # continue
```
jberry-suse: ??

openqa-maintenance.py
```diff
@@ -269,10 +288,22 @@ def kernel_target(req):
...
+        if not minimal:
+            return []
+
+        settings = settings.copy()
+        settings['FLAVOR'] += '-Minimal'
+        return [settings]
+
+    def add_kernel_settings(self, prj, settings):
+        settings = settings.copy()
+        # not right now
```
jberry-suse: indeed, that's what `return` means.

openqa-maintenance.py
```diff
@@ -792,15 +841,119 @@ def check_one_request(self, req):
...
+        if comment_info.get('comment', '') == comment:
+            self.logger.debug("%s comment did not change", incident_project)
+            return
+
+        self.logger.debug("adding comment to %s, state %s", incident_project, state)
+        #self.logger.debug("message: %s", msg)
+        if not self.dryrun:
+            if comment_id is not None:
+                self.commentapi.delete(comment_id)
+            self.commentapi.add_comment(project_name=str(incident_project), comment=str(comment))
```
jberry-suse: Virtually all of this is generalized in `ReviewBot`....

mimi1vx:
@jberry-suse, in reality, this code was used in production for more than a month :( so it mainly synncs with reality

jberry-suse:
I did not comment on that and do not doubt it. Just asking that we raise the quality bar, just a tad. If I didn't rebase my garbage commits this repo would have 10k useless and unintelligible commits from me.

jberry-suse:
I do the same myself, but that does not prevent me from going back and reworking all my commits to actually make sense and have useful messages instead of two undescriptive words.

coolo:
just realized Ondrej merged :(

This was never meant to be merged as is - this was supposed to be reviewed and tested and improved.

Issue #1012, opened 2017-07-18
repo_checker: resolve installcheck/fileconflict issues in Factory rings


jberry-suse:
[Relevant issues](https://gist.githubusercontent.com/jberry-suse/99329e902ecae6a03a903fa8623c0ce3/raw/23bbb36d1b3f37dd1ea50cee49d958f00d36bfe9/repo_checker.openSUSE:Factory:Staging:L)

Meta issue to track bugs and submit requests.

jberry-suse:
[llvm4](https://build.opensuse.org/request/show/508876) and [llvm3_8](https://build.opensuse.org/request/show/508877)

jberry-suse:
[dummy-release](https://build.opensuse.org/request/show/511064) conflict with `openSUSE-Kubic-release`

jberry-suse:
[installation-images](https://build.opensuse.org/request/show/511067) all subpackages

jberry-suse:
Looks like primarily:
```
found conflict of libopenssl-1_0_0-devel-32bit-1.0.2l-2.2.x86_64 with libopenssl-1_1_0-devel-32bit-1.1.0f-1.3.x86_64:
  - /usr/lib/libcrypto.so
  - /usr/lib/libssl.so
```
and
```
found conflict of openSUSE-Kubic-release-20170704-2.1.x86_64 with openSUSE-release-20170704-3.1.x86_64:
  - /etc/motd
  - /etc/os-release
```

installation-images resolves most of them.

@DimStar77 What's the best way to solve the `*-release` ones? I already fix dummy-release, but they should all probably just provide a "release" virtual package and conflict with other providers. Do you have a preference?

jberry-suse:
[openssl-1_1_0](https://build.opensuse.org/request/show/511068) conflict with [openssl-1_0_0](https://build.opensuse.org/request/show/511069)

jberry-suse:
@DimStar77 Also given that openssl-1_1_0 fails to build for i586 in Factory and devel I suppose this update will not translate to a fix since no new rpms will be produced. Given the number of problems indicated perhaps a special fix will be needed (binary wipe of -devel or ugly temporary whitelist for repo checker).

DimStar77:
> @DimStar77 What's the best way to solve the *-release ones? I already fix dummy-release, but they should all probably just provide a "release" virtual package and conflict with other providers. Do you have a preference?

[...]

PR #1033, opened 2017-07-28
repo_checker.pl: switch from multi-line regex to single line and full (and cleanups)

jberry-suse:
- 8658aa56dc7f4259242db2e622d1a041c3fcc0d4:
    repo_checker.pl: rework installcheck parsing like fileconflicts to resolve bugs.

- a9fdf43411395e77913a20290dcb78242636b800:
    repo_checker.pl: s/INSTALL/CONFLICTS/ for file pointer for more clarity.

- 9f4241a828aa783e5300ebffd523157dfabe27ba:
    **repo_checker.pl: switch from multi-line regex to single line and full.**
    
    The can't install regex was actually matching across multiple lines and
    skipping various entries in the install check.

- ee787adfabd01f6b6596310fb58655fc24b18de7:
    Rename repo-checker.pl to repo_checker.pl to be consistent.

- 7790817641b60a58a3215f77b8949f6fc9ff0b1d:
    repo-checker.pl: drop old debug lines.

- 906182f5666ed456a45e114fd77c2c585367d660:
    repo-checker.pl: replace [^-]* with [^-]+ as it is more accurate.

This was actually hiding additional `installcheck` failures and thus made #1012 look more reasonable than it actually was. Between this and the corruption problem I can see how these failures made it into Factory....the tool kinda worked and could fail often.

I made the two parsing loops follow the same pattern so they are easier to deal with since they were needlessly different.


jberry-suse:
rebased post merge

jberry-suse:
Solves another serious bug that hides problems.

Issue #944, opened 2017-06-13
risk: ideas for risk assessment tool


jberry-suse:
@lnussel and I have bounced around the idea of reworking the leaper bot based on the current usage as the code is rather ungainly.

From performing a large number of `leap-reviewers` reviews it seems clear there are classes of updates that should not require human review (like SLE updates) from the perspective of is the update desirable for Leap. Instead only relevant for technical/legal or maintainer sign off. Otherwise, it simply becomes a nuisance to weed through loads of requests that simply should be accepted.

The leaper bot could be reworked to generate a risk and/or badness index (related #943) and only require human review for packages above a certain threshold. Seems more a nuisance to require `leap-reviewer` reviews of KDE LTS updates which are clearly alright from a leap perspective.

The primary areas for concern of origin changes, delete requests. Origin change include major version updates for packages inherited from previous Leap version. Packages that are tracking Factory or a special Leap devel are unobjectionable from the `leap-reviewers` perspective.

Generally, the sources for Leap packages are from places that have already had human review and the only decision to make from the Leap perspective is do we want to support this or is this compatible with the rest of Leap. Combined with a way to shut-off update crawler requests #942 should make for a much smoother daily operation.

Likely, it makes sense to start developing these indexes and outputting additional information that went into the index and if manual review confirms these indexes then the bot can be switched to utilize the index for deciding when to require the review.

lnussel:
Now that we are getting closer to the release it may be desirable to do more and stricter reviews actually.  At some point even a KDE lts update may no longer be desiable. The question then just is at which level to decline.

jberry-suse:
This was intended for next Leap cycle as I do not expect to get to this soon, but a good note to plan for dialing back setting for next cycle.

jberry-suse:
The more I've run into the issue leaper really needs to be able to check for parallel requests to SLE especially since that is the encouraged workflow now.

jberry-suse:
Broke off origin-manager issue, refocused to risk assessment tool, and removed from origin manager milestone.

coolo:
No longer relevant with closed leap gap

PR #1037, opened 2017-08-01
leaper: transition from SLE-12-SP3 and Leap:42.3 to SLE-15 and Leap:15.0.

### LINKED ISSUE #1036: adjust leaper for SLE15
please adjust leaper for SUSE:SLE-15:GA, replacing the SP3 code path

jberry-suse:
This shows just how much something like #944 (cleanup) is needed.

In order to maintain existing logic, but disable until relevant again I placed a `False and` in front of disabled conditions along with a comment about when it should be re-enabled.

Fixes #1036.

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -78,21 +78,13 @@ class Leaper(ReviewBot.ReviewBot):
         # update lookup information on every run
 
         if self.ibs:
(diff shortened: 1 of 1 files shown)


jberry-suse:
Any checks that produce remote calls I dropped the extraneous future service pack references, whereas simple string checks I left after updating.

PR #1040, opened 2017-08-02
dist/package/spec: utilzie %osc_plugin_dir provided by osc post sr#512849.

jberry-suse:
Presumably this will fail travis until new `osc` is in Tumbleweed. Now that the package is situated it may make sense to rework build test to build out of `openSUSE:Tools`.


jberry-suse:
For reference https://build.opensuse.org/request/show/512849.

jberry-suse:
Works fine when built against new `obs`, see https://build.opensuse.org/package/show/home:jberry:branches:openSUSE:Tools/openSUSE-release-tools. So we'll just wait until in TW.

Issue #1029, opened 2017-07-28
repo_checker: project_only check for full build completed


jberry-suse:
The previous script just ran with whatever state the repo was in, as does #1028, but it likely makes sense to either wait for build to complete, or perhaps run against the snapshot repo if that is simple enough. That way false-positives due to incomplete rebuild are not included.

This is somewhat tricky in that the check either needs to be run repeatedly to find the right time when everything has settled or needs to be triggered along with other snapshot related bits.

jberry-suse:
This may also end up being best applied to request reviews as well, but given it would slow things down is likely worth the potential for temporary false-negatives.

PR #1042, opened 2017-08-03
repo_checker: provide --post-comments option on project_only subcommand (and cleanups).

### LINKED ISSUE #1030: repo_checker: project_only parse output and comment on related packages
As I noted in #981 and #1028 it is likely possible and beneficial to parse the output and relate problems back to the package responsible for creating the binaries using `/build/openSUSE:Factory/standard/x86_64?view=binaryversions`. As an aside, the `check_duplicate_binaries` tool could be reworked to do something similar to notify maintainers.

The one area that is unclear is about which packages to notify. I can envision a common package being uninstallable and causes a number of other packages to also be uninstallable. At that point a comment would be placed on all effected packages. That does not seem incorrect in the sense that maintainers likely want to know their package is no longer installable, but depending on how many levels removed they are from the source may not be able to do anything about it. As such it may make more sense to either only comment about source packages at the root of a tree or those and a level up. I would also expect these problems to be less common and likely warrant attention when they occur. It is certainly simpler to just post on all effected. It may be one of those try it out and see if it becomes noisy.

Where to place the comment is also in question. It seems straight forward to place it on the package in question, meaning `openSUSE:Factory/packagefoo`, but depending on how notifications work it may make more sense to place the comment on the source package in the devel project. Of course the problem may be fixed there depending on what it was and could cause confusion. If the notifications do not trickle as desired the maintainers could be mentioned explicitly as the `./devel-project.py` reviews do. Another benefit of placing on the target project would be easy to find if folks are looking to confirm not something specific to their set of repos and that the problem has been reported.

jberry-suse:
- f42cc4bcc9f599b1059127e7ce310fe926e3eef6:                                                                                                
    repo_checker: truncate long messages to avoid crashing OBS.                                                                            
                                                                                                                                           
- 722764a1f528e0cb9bd1cc9a29086c7f04656c32:
    **repo_checker: provide --post-comments option on project_only subcommand.**

- ce35a686ddafa6dd8943ef808fa5e7ee1e0c4f93:
    repo_checker: provide optional parsing of install check output and mapping to package.

- d87ee4c38ea59ed86f87774113b9d5bcf060374f:
    repo_checker: package_whitelist() does not need to support group splitting to find product.

- 924db6024b3587a177ce860a27b9d5d46886d9f0:
    osclib/core: provide package_binary_list() to efficiently obtain binary to package map.

- 0452fb266bcf1548b63f53535120062234419992:
    osclib/core: add package to BinaryParsed tuple.

- 16c5c92bcc2eddb29add606e930255f79659206c:
    osclib/core: provide BINARY_REGEX named group pattern and utilize.

- cfbd479ed74db16906d66de26f4bf2a151870cc8:
    ReviewBot: comment_write(): support package.

Deployment as service should be blocked by #1029 even though old `repo_checker` dumped to file regardless this is noisy and makes false positives more annoying.

Invocation is simple:

```
./repo_checker.py --verbose project_only --post-comments openSUSE:Factory
```

I went ahead and did a manual run in order to get the cleanup started especially with SLE/Leap 15 branching from Factory. The script produced `111` comments on devel packages containing the aggregated problems.

To be clear this approach:

- parses the `installcheck` & `file conflict` output
- maps the binary names back to packages
- groups all messages related to the same package
- sorts the output and posts as comment on package in devel project in order to trigger notifications

I debated posting bugs to bugzilla, but that would be a fair bit more work especially considering the clean comments api we have for avoiding useless updates, but ensuring a single comment. I did envision shared bugs between two packages that have file conflicts which would be neat, but this just duplicates those messages to both packages.

A hash representing the binaries and number of issues is generated to avoid updating comment for version updates or release increments while still properly picking up new or changes to existing issues.

Depending on how this works out and opinions this approach can be broadened for the rest of repo_checker or at minimum include for duplicate binary checks.

After the recent fixes that were hiding all `-32bit` imported binaries in the `x86_64` space added to the already exposed issues there are quite a few even after all my fixes.

See [dashboard/repo_checker](https://build.opensuse.org/package/view_file/openSUSE:Factory:Staging/dashboard/repo_checker) for an updated report.

The end result are package comments like the following on [devel:languages:ruby:extensions/rubygem-celluloid](https://build.opensuse.org/package/show/devel:languages:ruby:extensions/rubygem-celluloid).

![image](https://user-images.githubusercontent.com/22943929/28904145-83aae4d6-77ce-11e7-82ec-856f8167a4d8.png)

One caveat being that there is no simple way of removing these comments since the OBS comment API does not provide a way of querying all comments by a user so each devel project would have to be scanned. Alternatively, on accept comments for all effected packages could be removed. I would expect these comments to be relatively rare once the new `repo_checker` is enforcing cleanliness on new requests so this is more of a stop-gap for getting maintainers to act for the initial cleanup. It seems simple enough for maintainers to remove them once fixes are accepted.

Fixes #1030.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -446,15 +446,16 @@ class ReviewBot(object):
 
         comments = self.comment_api.get_comments(**kwargs)
         comment, _ = self.comment_api.comment_find(comments, self.bot_name, info)
+        debug_key = '/'.join(kwargs.values())
         if (comment is not None and
             ((identical and comment['comment'] == message) or
              (not identical and comment['comment'].count('\n') == message.count('\n')))
         ):
             # Assume same state/result and number of lines in message is duplicate.
-            self.logger.debug('previous comment too similar to bother commenting again')
+            self.logger.debug('previous comment on {} too similar'.format(debug_key))
             return
 
-        self.logger.debug('adding comment to {}: {}'.format('/'.join(kwargs.values()), message))
+        self.logger.debug('adding comment to {}: {}'.format(debug_key, message))
 
         if not self.dryrun:
             if comment is None:
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -426,11 +426,13 @@ class ReviewBot(object):
     def comment_handler_lines_deduplicate(self):
         self.comment_handler.lines = list(OrderedDict.fromkeys(self.comment_handler.lines))
 
-    def comment_write(self, state='done', result=None, project=None, request=None,
-                      message=None, identical=False):
+    def comment_write(self, state='done', result=None, project=None, package=None,
+                      request=None, message=None, identical=False):
         """Write comment from log messages if not similar to previous comment."""
         if project:
             kwargs = {'project_name': project}
+            if package:
+                kwargs['package_name'] = package
         else:
             if request is None:
                 request = self.request
@@ -452,7 +454,7 @@ class ReviewBot(object):
             self.logger.debug('previous comment too similar to bother commenting again')
             return
 
-        self.logger.debug('adding comment to {}: {}'.format(kwargs.itervalues().next(), message))
+        self.logger.debug('adding comment to {}: {}'.format('/'.join(kwargs.values()), message))
 
         if not self.dryrun:
             if comment is None:
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -12,6 +12,8 @@ from osc.core import owner
 from osc.core import show_project_meta
 from osclib.memoize import memoize
 
+BINARY_REGEX = r'(?:.*::)?(?P<filename>(?P<name>.*?)-(?P<version>[^-]+)-(?P<release>[^-]+)\.(?P<arch>[^-\.]+))'
+RPM_REGEX = BINARY_REGEX + '\.rpm'
 BinaryParsed = namedtuple('BinaryParsed', ('filename', 'name', 'arch'))
 
 
@@ -91,18 +93,18 @@ def request_staged(request):
 def binary_list(apiurl, project, repository, arch, package=None):
     parsed = []
     for binary in get_binarylist(apiurl, project, repository, arch, package):
-        result = re.match(r'(.*)-([^-]*)-([^-]*)\.([^-\.]+)\.rpm', binary)
+        result = re.match(RPM_REGEX, binary)
         if not result:
             continue
 
-        name = result.group(1)
+        name = result.group('name')
         if name.endswith('-debuginfo') or name.endswith('-debuginfo-32bit'):
             continue
         if name.endswith('-debugsource'):
             continue
-        if result.group(4) == 'src':
+        if result.group('arch') == 'src':
             continue
 
-        parsed.append(BinaryParsed(binary, name, result.group(4)))
+        parsed.append(BinaryParsed(result.group('filename'), name, result.group('arch')))
 
     return parsed
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -108,3 +108,28 @@ def binary_list(apiurl, project, repository, arch, package=None):
         parsed.append(BinaryParsed(package, result.group('filename'), name, result.group('arch')))
 
     return parsed
+
+def package_binary_list(apiurl, project, repository, arch, package=None):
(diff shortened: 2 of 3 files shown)


jberry-suse:
Could also use the new `package_binary_list()` for duplicate check and the ignore list of deletes to avoid individual package queries. The check command being relevant to #993.

jberry-suse:
hehe, this already worked...I got an email notifying me on one of my packages about an issue of which I was not aware.

jberry-suse:
As seen in this example the comments are resulting in fixes!

https://build.opensuse.org/package/show/devel:languages:python/pymol

jberry-suse:
Re-ran today for stats `(package, len(sections))` [112 packages]:

```
('rubygem-rhc', 6)
('rubygem-celluloid', 6)
('perl-Selenium-Remote-Driver', 1)
('nss_ldap', 1)
('dice', 2)
('deepin-music-player', 2)
('hdjmod', 1)
('zerobranestudio', 2)
('pcfclock', 1)
('rubygem-r10k', 6)
('openmpi', 1)
('epour', 2)
('ghc-dice', 2)
('ghc-postgresql-simple-migration', 2)
('espeak-gui', 4)
('installation-images', 10)
('vegastrike-speech', 1)
('kernel-syms', 1)
('yast2-docker', 1)
('sysdig', 1)
('ghc-protobuf-simple', 2)
('libfabric', 1)
('virtaal', 4)
('python3-geoip2', 2)
('enchant', 1)
('php5-ZendFramework', 2)
('gns3-server', 2)
('tkman', 2)
('vhba-kmp', 1)
('rubygem-puppet_forge', 6)
('opentoonz', 2)
('octave-forge-ocs', 2)
('gns3', 2)
('ghc-convert-annotation', 2)
('prosody', 2)
('dkms', 1)
('openssl-1_1_0', 1)
('xapps', 1)
('rubygem-bundler', 8)
('libqt5-qt3d', 4)
('boost', 2)
('ginac', 2)
('xtables-addons', 1)
('rubygem-gpgme', 6)
('rubygem-sdoc', 6)
('hdf5', 3)
('rubygem-logging', 6)
('libqt5-qtwebengine', 2)
('vegastrike-music', 1)
('rubygem-ra10ke', 4)
('python-web.py', 1)
('buzztrax', 1)
('docker-bench-security', 1)
('bbswitch', 1)
('deepin-tool-kit', 14)
('rubygem-winrm-2_1', 4)
('nagvis', 4)
('ghc-inline-r', 2)
('icingaweb2-module-director', 2)
('virtualbox', 2)
('jhbuild', 2)
('openssl-1_0_0', 1)
('ruby2.4', 2)
('ha-cluster-bootstrap', 2)
('Mesa', 1)
('geis', 1)
('kaccounts-integration', 1)
('kiwi', 2)
('python-carrot', 2)
('paperwork', 2)
('rubygem-winrm-fs', 6)
('python-kiwi', 1)
('yast2-drbd', 1)
('perl-CPAN-Meta-YAML', 2)
('patterns-caasp', 1)
('rubygem-winrm-elevated', 6)
('libstatgrab', 1)
('kopano', 58)
('hgview', 4)
('python-launchpadlib', 2)
('hawk2', 2)
('rubygem-rake', 2)
('kopano-webapp', 2)
('rubygem-faraday_middleware-0_9', 4)
('rubygem-pry_debug', 6)
('rubygem-jekyll', 4)
('python-plaso', 2)
('rubygem-nokogiri-1_6', 6)
('libpinyin', 1)
('arpack-ng', 2)
('openmpi2', 1)
('docker-compose', 1)
('rubygem-studio_api', 6)
('scim-bridge', 2)
('libvoikko', 1)
('scalapack', 2)
('octave-forge-secs1d', 2)
('kdebase3-SuSE', 4)
('fwupdate', 3)
('protobuf', 2)
('nss-pam-ldapd', 1)
('gobby', 4)
('fusecompress', 2)
('purpose', 2)
('python-sqlalchemy-migrate', 2)
('vegastrike-extra', 1)
('ktexteditor', 2)
('bind', 1)
('pymol', 1)
('tulip', 16)
('dcmtk', 6)
('octave-forge-secs3d', 2)
```

PR #1051, opened 2017-08-11
ReviewBot: add_review(): raise exception on failure and avoid duplicate reviews

### LINKED ISSUE #1050: ReviewBot: add_review() failure should cause crash and be able to pickup later
Eg. https://build.opensuse.org/request/show/514577 and https://build.opensuse.org/request/show/511736

review team group doesn't added on it, thus it does not pass human review actually.

due to https://github.com/openSUSE/osc-plugin-factory/blob/master/check_source.py#L121

jberry-suse:
Also update calls based on default behavior change.

Fixes #1050 unless changes to diff algorithm are also desired.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -206,20 +206,12 @@ class ReviewBot(object):
         u = osc.core.makeurl(self.apiurl, ['request', req.reqid], query)
         if self.dryrun:
             self.logger.info('POST %s' % u)
-            return True
+            return
 
-        try:
-            r = osc.core.http_POST(u, data=msg)
-        except urllib2.HTTPError, e:
-            self.logger.error(e)
-            return False
-
+        r = osc.core.http_POST(u, data=msg)
         code = ET.parse(r).getroot().attrib['code']
         if code != 'ok':
-            self.logger.error("invalid return code %s"%code)
-            return False
-
-        return True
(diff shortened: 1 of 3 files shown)


jberry-suse:
After some thought it does not seem to make sense to ever add a review twice regardless of the state. The only time one might want that behavior, like with `repo-checker` when accepted and then a staging is modified, it would not happen anyway since `factory-auto` would not rerun. The only time it is currently used is for fallback reviewers, but again that is a human review that makes little since to add twice if it was accepted already and declined will work automatically. See further explanations in code comment and commit messages.

I left the parameter, but starting to think even that is unnecessary.

- 078ec6869706d227ca54a4332ab45a19705d9138:
    ReviewBot: add_review(): provide allow_duplicate parameter and default of False.
    
    Also update calls based on default behavior change. As described in the
    comment above add_review() there seems to be no reason to ever add a
    duplicate review.
    
    The check in leaper would skip unless state == declined. In such a case
    the review would automatically be reopened along with the request so that
    logic is unnecessary.

- ef93709e16b25cfb2fb26dd7afc6b6bf71e429ce:
    ReviewBot: add_review(): raise exception instead of returning boolean.
    
    This ensures that calling bots crash rather than continue which can cause
    non-atomic state changes that miss reviews when OBS crashes or is
    unavailable.

PR #1076, opened 2017-08-18
ReviewBot: get_devel_project(): correct error handling to only allow 404.

jberry-suse:
- 164b04903cfc56828b08e0fbb6041d3cb15c6e67:
    ReviewBot: add_review(): remove left-over print() line.

- d565f46123bc7208f034bfc8c2388a0eb4c79081:
    ReviewBot: get_devel_project(): correct error handling to only allow 404.
    
    The original intent was clearly to allow 404 (package does not exist for
    new packages), but crash on anything else. Instead it consumes all
    exceptions and does nothing different even with the e.code condition.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -208,7 +208,6 @@ class ReviewBot(object):
             raise osc.oscerr.WrongArgs("missing by_*")
 
         for r in req.reviews:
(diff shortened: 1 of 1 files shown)


nilxam:
LGTM

PR #1077, opened 2017-08-21
stagingapi: Ensure the links is valid before processing

nilxam:
Make sure the links of package in Rings must are valid before processing, otherwise someone had package linked to the Ring package in somewhere else than Rings will breaks select/delete/rm_to_prj, ie. _link will be created. Therefore, only return the candidate if it has proper linked from another ring and have different package name.

Example of failure:
```
mlin@rana:~> osc api -X POST `/source/openSUSE:Factory:Rings:2-TestDVD/telepathy-logger-qt5?cmd=showlinked
<collection>
    <package project="home:Vogtinator:branches:openSUSE:Factory:Rings:2-TestDVD" name="telepathy-logger-qt5"/>
</collection>
```

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1072,17 +1072,17 @@ class StagingAPI(object):
 
         return project
 
-    def get_sub_packages(self, pkg, project=None):
+    def get_sub_packages(self, package, project=None):
         """
         Returns a list of packages that need to be linked into rings
         too. A package is actually a tuple of project and package name
         """
         ret = []
         if not project:
-            project = self.ring_packages.get(pkg)
+            project = self.ring_packages.get(package)
         if not project:
             return ret
-        url = self.makeurl(['source', project, pkg],
+        url = self.makeurl(['source', project, package],
                            {'cmd': 'showlinked'})
 
         # showlinked is a POST for rather bizzare reasons
@@ -1090,7 +1090,9 @@ class StagingAPI(object):
         root = ET.parse(f).getroot()
 
         for pkg in root.findall('package'):
-            ret.append((pkg.get('project'), pkg.get('name')))
+            # ensure sub-package is valid
+            if pkg.get('project') in self.rings and pkg.get('name') != package:
+                ret.append((pkg.get('project'), pkg.get('name')))
 
         return ret
 


osclib/stagingapi.py
```diff
@@ -1072,25 +1072,27 @@ def map_ring_package_to_subject(self, project, pkg):
...
                            {'cmd': 'showlinked'})
 
         # showlinked is a POST for rather bizzare reasons
         f = http_POST(url)
         root = ET.parse(f).getroot()
 
         for pkg in root.findall('package'):
-            ret.append((pkg.get('project'), pkg.get('name')))
+            # ensure sub-package is valid
+            if pkg.get('project') in self.rings and pkg.get('name') != package:
```
jberry-suse: Is `pkg.get('project') in self.rings` a problem for `adi` trying to bring in subpackages?

nilxam: Should not matter with `adi`, `get_sub_packages` is checking the sub-package in ring and adi just be ignored.

jberry-suse: but we had plans to include subpackages in adi.... #948 of which this would get in the way.

nilxam: do you planning to change `get_sub_package` to cope with adi? because `get_sub_package` is mainly for ring package, if you've plan to do that then yes this can be a problem against adi, but if do so, there is another problem, `showlinked` call does not make sense for adi package, the **layout**(all specs are identical to the new request; all sub-package's link are created in the devel project) is not easy to guaranteed. For https://github.com/openSUSE/osc-plugin-factory/pull/948 , created main package in adi and then create sub-packages for rest of specfile is the proper way IMO.

jberry-suse: I am not sure what you mean. #948 was created to solve #946 created by you which asked for adi requests to support creating subpackages. So yes, we do have plans. Either way the mere fact that this would break such a feature indicates that the check is likely not correct since being in a ring is not the only "valid" case.

nilxam: We're misunderstand each other?! Let me ask several question(please don't considering #948 in this change):
1) do you agree `get_sub_packages` is mainly for ring package only? if you don't think so, please point me where is the real case(not returned a empty list https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/stagingapi.py#L1084) this function is used for adi in the code.
2) do you agree all `valid` sub-package link in Rings must from/to another ring? ie. ring1->ring1, ring1->ring2, etc. is valid; linking to `home:` or whatever else is invalid.

jberry-suse: 1. sure, but we want to make it work for `adi` so why make work for ourselves by needlessly introducing more code that will have to be changed?
2. I agree that is a subset of what is valid. In factory right now there are packages with linked subpackages that are not in rings. `adi` code will need to look at the packages in target project to get a list. This same code will need to work in that case. A valid link should be to the something that starts with the target project.

My point was not that this change by itself will cause issues, but why write code that will cause us issues down the road.

nilxam: It's a function mainly for ring package only, verifying the link is from/to another ring makes sense in this function IMO, I don't know why you bring `adi` in here, if there will make it work for `adi` in the future, then that is the future thing.

jberry-suse: Much of the problems with this code base is shooting future developers in the foot. The method name is `get_sub_packages()` and getting sub packages applies to both adi and non-adi packages. The fact that it is currently not used on adi packages doesn't mean we should intentionally make it only work on non-adi packages. If it was a load of extra work I would agree, deal with it in the future, but this is simply short-sighted and due to not properly defining the problem. The problem is validating that the links are not to external projects not that they are links to ring projects.

Most of the features I added requiring changing massive amounts of code that did similar things, rather than actually solving the problem the method implied it was solving. This is the definition of technical debt.

nilxam: The description of `get_sub_packages()` is **Returns a list of packages that need to be linked into rings too.**

I will not continue the discussion here any further. You can revert it, change it to what you've expected, or whatsoever. Just note that, the issue here is - for ring sub-package, the links are not to external projects and they are links to ring projects.

jberry-suse: Indeed, it just means I get to pay the technical debt like the rest of the things of this nature.

PR #1079, opened 2017-08-22
stagingapi: Do not disable build if the build state is enabled manually

nilxam:
Do not disable build of package if the build state is enabled manually, we need to keep that state in case the request is superseding a new package and that package is enabled to build manually.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -827,6 +827,21 @@ class StagingAPI(object):
         :param disable_build: should the package be created with build
                               flag disabled
         """
+        if disable_build:
+            # keep build state in case builds is not disabled during the
+            # reqeust is superseding a new package
+            try:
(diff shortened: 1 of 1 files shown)


osclib/stagingapi.py
```diff
@@ -827,6 +827,21 @@ def create_package_container(self, project, package, disable_build=False):
         :param disable_build: should the package be created with build
                               flag disabled
         """
+        if disable_build:
+            # keep build state in case builds is not disabled during the
+            # reqeust is superseding a new package
```
jberry-suse: I believe you are trying to say: `keep enabled build state for a superseding request providing a new package`

nilxam:
I didn't noticed https://github.com/openSUSE/osc-plugin-factory/pull/897 , let's move to https://github.com/openSUSE/osc-plugin-factory/pull/897 then.

PR #1080, opened 2017-08-22
Freeze Cmd: Implement freezing adi staging project

nilxam:
Since we have introduced the source project expand function in adi staging project, a new function to freeze a adi staging project is needed as well.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -395,7 +395,7 @@ def do_staging(self, subcmd, opts, *args):
             CheckDuplicateBinariesCommand(api).perform(opts.save)
         elif cmd == 'freeze':
             for prj in args[1:]:
-                prj = api.prj_from_letter(prj)
+                prj = api.prj_from_short(prj)
                 print(Fore.YELLOW + prj)
                 FreezeCommand(api).perform(prj, copy_bootstrap=opts.bootstrap)
         elif cmd == 'frozenage':
--- a/osclib/freeze_command.py
+++ b/osclib/freeze_command.py
@@ -118,6 +118,15 @@ class FreezeCommand(object):
 
     def perform(self, prj, copy_bootstrap=True):
         self.prj = prj
+
+        if self.api.is_adi_project(prj):
+            src_prj = self.api.find_devel_project_from_adi_frozenlinks(self.prj)
+            if src_prj is None:
+                raise Exception("{} does not have a valid frozenlinks".format(self.prj))
+            else:
+                self.api.update_adi_frozenlinks(self.prj, src_prj)
+            return
+
         self.set_links()
 
         self.freeze_prjlinks()
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -370,6 +370,17 @@ class StagingAPI(object):
         projects = [p for p in self.get_staging_projects() if self.is_adi_project(p)]
         return sorted(projects, key=lambda project: self.extract_adi_number(project))
 
+    def find_devel_project_from_adi_frozenlinks(self, prj):
+        try:
+            url = self.makeurl(['source', prj, '_project', 'frozenlinks'], {'meta': '1'})
+            root = ET.parse(http_GET(url)).getroot()
+        except urllib2.HTTPError, e:
+            if e.code == 404:
+                return None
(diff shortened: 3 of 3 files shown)


osc-staging.py
```diff
@@ -395,9 +395,14 @@ def do_staging(self, subcmd, opts, *args):
             CheckDuplicateBinariesCommand(api).perform(opts.save)
         elif cmd == 'freeze':
             for prj in args[1:]:
-                prj = api.prj_from_letter(prj)
+                is_adi = False
+                if 'adi:' in prj:
+                    prj = api.adi_prj_from_adi_name(prj)
+                    is_adi = True
+                else:
+                    prj = api.prj_from_letter(prj)
```
jberry-suse: `api.prj_from_short()` should handle both cases and provide the full project name.

osc-staging.py
```diff
@@ -395,9 +395,14 @@ def do_staging(self, subcmd, opts, *args):
...
                 print(Fore.YELLOW + prj)
-                FreezeCommand(api).perform(prj, copy_bootstrap=opts.bootstrap)
+                FreezeCommand(api).perform(prj, copy_bootstrap=opts.bootstrap, freeze_adi=is_adi)
```
jberry-suse: `freeze_adi` seems odd, as `prj` already indicates if `adi` or not and passing to `freeze` command which would imply one wants to freeze `prj`. I would drop the parameter and just check if `adi` inside freeze since the implementation of freeze differs based on project.

osclib/freeze_command.py
```diff
@@ -116,8 +116,18 @@ def verify_bootstrap_copy_codes(self, codes):
                     return False
         return True
 
-    def perform(self, prj, copy_bootstrap=True):
+    def perform(self, prj, copy_bootstrap=True, freeze_adi=False):
         self.prj = prj
+
+        if freeze_adi:
```
jberry-suse: `if self.api.is_adi_project()`

osclib/freeze_command.py
```diff
@@ -116,8 +116,18 @@ def verify_bootstrap_copy_codes(self, codes):
                     return False
         return True
 
-    def perform(self, prj, copy_bootstrap=True):
+    def perform(self, prj, copy_bootstrap=True, freeze_adi=False):
```
jberry-suse: Drop `freeze_adi`.

osclib/freeze_command.py
```diff
@@ -116,8 +116,18 @@ def verify_bootstrap_copy_codes(self, codes):
...
         self.prj = prj
+
+        if freeze_adi:
+            src_prj = self.api.find_devel_project_from_adi_frozenlinks(self.prj)
+            if src_prj is None:
+                raise Exception("{} does not have a valid frozenlinks".format(self.prj))
+            else:
+                self.api.update_adi_frozenlinks(self.prj, src_prj)
+            # do not continue
```
jberry-suse: `return` = "do not continue". A comment should indicate what the code does not. In this case two different implementations of freeze and adi implementation is done so do not perform non-adi freeze.

osclib/stagingapi.py
```diff
@@ -370,6 +370,17 @@ def get_adi_projects(self):
...
 
+    def find_devel_project_from_adi_frozenlinks(self, prj):
+        try:
+            url = self.makeurl(['source', prj, '_project', 'frozenlinks'], {'meta': '1'})
+            root = ET.parse(http_GET(url)).getroot()
+        except urllib2.HTTPError, e:
+            if e.code == 404:
+                return None
+        packages = root.findall('./frozenlink/package')
+        # check the first one package's devel project is good enough
```
jberry-suse: `s/check //`
`s/one //`

osclib/stagingapi.py
```diff
@@ -1197,6 +1208,11 @@ def adi_prj_from_number(self, number):
             return number
         return '{}:adi:{}'.format(self.cstaging, number)
 
+    def adi_prj_from_adi_name(self, name):
```
jberry-suse: Use `prj_from_short()` instead. `short` means short-hand names like: `A` or `adi:17` which is what you want.

PR #1082, opened 2017-08-22
stagingapi: avoid making an empty package-diff comment.

### LINKED ISSUE #1081: select: Avoid adding comments about no-ops
Sometimes a select command can result in no operation being done, e.g. by moving a request that has already been moved or selecting a request that does not exist. In those cases a useless comment gets added to the staging:

```
<!-- osc-staging type=package-diff -->

Requests: 0 added, 0 removed; using select command
```

jberry-suse:
In the say way that I had to rework the supersede logic to actually keep track of what changes were made to print messages indicating the changes the select and unselect commands should likely know if the command they were asked to evaluate ended up doing nothing rather than just calling `update_status_or_deactivate()` always. Short of that ends up not being an issue except for `update_status_comments()` so just check for empty `package-diff` comments to avoid making them.

Fixes #1081.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1481,6 +1481,9 @@ class StagingAPI(object):
                     req['prefix'] = 'removed '
                     requests.append(req)
 
+            if not len(requests):
+                # Nothing changed so no sense posting comment.
+                return
         else:
             dashboard_url = '{}/project/staging_projects/{}/{}'.format(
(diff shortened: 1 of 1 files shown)


jberry-suse:
Confirmed in the wild attempting to restage a request in same adi staging.

```
osc staging select adi:168 512096
```

Without patch a blank comment is posted.

PR #1092, opened 2017-08-23
repo_checker: use <pre> instead of ``` to avoid html entity escaping.

jberry-suse:
Other `<pre>` switch in #1089, but I'll just leave it there as a lot to break out.


jberry-suse:
Properly formats things like `>=` in comments.

PR #1093, opened 2017-08-23
repo_checker: rework to handle multiple stagings per group and proper repo layering.

### LINKED ISSUE #1044: repo_checker: rename sub-packages of existing package can cause false file conflicts
See https://build.opensuse.org/request/show/510161 for an example where both checkers see:

```
found conflict of python3-zope-proxy-4.1.1-6.2.x86_64 with python3-zope.proxy-4.2.1-10.2.x86_64
```

Note the rename from `-proxy` to `.proxy`. This should be a matter of using the binaryversions query to determine the binaries produced in target project for all packages within a staging project rather than looking at the binaries produced in the staging and ignoring the versions from the target project.

Both repo checkers use the same workflow style here, only the old one one would get the list from the devel project using lots of API calls whereas the new one uses the binary list produced by `bs_mirrorfull` from the staging project.

Since this is a pre-existing bug I would not expect this to prevent deployment. I am curious however why I have not seen reports about this before as clearly this request has been pending the old repo checker review for over 20 days (ie stuck on this). I can't imagine this is the first case where something like this is manually skipped. Would be good to know of any other cases Factory maintainers are familiar with.

jberry-suse:
- 4280b8934af4b5113772d37dae05691a6d4c12f2:
    cycle: cycles() rename group parameter to staging to better reflect usage.

- beef4a7a4d5ebd83395646d78ed24050e5eb8b4e:
    repo_checker: rework to handle multiple stagings per group (ie :DVD).

- a6a4eb9870c053af7281d8b1b10d14144cdff775:
    core: add memoize(session=True) to package_binary_list().

- 572868558a98b6e34e41e6127da4a6ab55e4d0db:
    stagingapi: provide staging_walk() to cleaning handle :DVD subproject.

- 601dc6755aecd53f243fe1a3fa1b583063125f97:
    repo_checker: replace separate ignore methods with new algorithm.

- 3c3d4a735bada71e8b2ff1150421e5b5aae3afd1:
    repo_checker.pl: support multiple directories to check.

- f8e3cd9972d9276a67104a0a159a9571f0691c49:
    repo_checker: move result_comment() headers into respective check methods.
    
    This will be necessary for proper context in future commits and seems to
    make more sense.

- 5d4b62a9575bc2f85e7b516507b8cbabdf4f04f7:
    repo_checker: only bother to parse if --post-comments.

Fixes #1044.


jberry-suse:
Tweaked to avoid trying `:DVD` subproject on SUSE:* by setting `staging-dvd-archs` to empty list in acced3a66c63921b8e3ad2db11605ae2bc3890e4 and checking in `staging_walk()` cfc21469fd50521b920b01931bb247790a116d46.

PR #1098, opened 2017-08-24
accept: call accept_other_new() for all projects instead of just openSUSE.

lnussel:
TODO: make configurable


jberry-suse:
Perhaps use a prefix in commit message to make it clear to what aspect of the repo this applies.

```
accept: call accept_other_new() for all projects instead of just openSUSE.
```

PR #1100, opened 2017-08-25
Leaper fixes

lnussel:


jberry-suse:
Again, it would be nice to start prefixing commits with relevant tool given the larger separate tools that live in this repo.

`leaper:`

lnussel:
mea culpa

jberry-suse:
Thanks for the fancy commit messages.

This will end up posting an empty comment since messages about adding reviews are created after `comment_write()` and that is the only thing it ends up doing:
```
<!-- Leaper state=done result=accepted -->


```

I'll go ahead and make a followup PR to check if comments extracted from logger ends up being empty.

PR #1101, opened 2017-08-25
Ttm

DimStar77:


totest-manager.py
```diff
@@ -313,10 +313,10 @@ def maxsize_for_package(self, package):
         if re.match(r'.*-dvd5-.*', package):
             return 4700372992  # a DVD needs to match
 
-        if re.match(r'.*-image-livecd-x11.*', package):
+        if re.match(r'.*livecd-openSUSE:livecd-x11.*', package):
```
jberry-suse: Based on the final name below: `livecd-openSUSE:livecd-x11` it would seem the first `.*` is necessary and the second set since this is the package name, not binary.

totest-manager.py
```diff
@@ -313,10 +313,10 @@ def maxsize_for_package(self, package):
...
             return 681574400  # not a full CD
 
-        if re.match(r'.*-image-livecd.*', package):
+        if re.match(r'.*livecd-openSUSE:livecd-.*', package):
```
jberry-suse: Same comment about `.*` as above only second set makes sense here.

totest-manager.py
```diff
@@ -384,7 +384,7 @@ def is_snapshottable(self):
 
                 for arch in ['i586', 'x86_64']:
                     for product in self.livecd_products:
-                        if not self.package_ok('openSUSE:%s:Live' % self.project, product, 'standard', arch):
+                        if not self.package_ok('openSUSE:%s:Live' % self.project, product, 'images', arch):
```
jberry-suse: This would appear to not work for the multibuild packages:

```
osc api '/build/openSUSE:Factory/_result?package=livecd-openSUSE:livecd-kde&repository=images&arch=x86_64'
Server returned an error: HTTP Error 404: unknown package 'livecd-openSUSE livecd-kde'
unknown package 'livecd-openSUSE:livecd-kde'
```

totest-manager.py
jberry-suse: Could leave I suppose since this applies to other products where this may make sense even though I do not see any using it currently.

totest-manager.py
jberry-suse: Package does not exist in `Factory`, but does in `Factory:Live`.

PR #1103, opened 2017-08-27
osclib/stagingapi: create_adi_project(): force rpmlint strict.

### LINKED ISSUE #1102: repo-checker no longer blocks invalid licenses
The old repo-checker inspected rpmlint.log and did not let packages pass with an invalid license.

The main issue here is that rpmlint-Factory-strict is not enabled across OBS - as it is too annoying in devel projects and even worse in 3rd party repos... so the invalid licenses are flagged as warnings there only

In adi stagings, this is also not enabled (as it builds against openSUSE:Factory/standard - just like any other 3rd party prj would) - 'proper' staigngs are fine, as they build against openSUSE:Factory:Staging/standard, where the strict filter is injected unconditionally

We need to either find a setup that works for adi to get Factory-strict enabled or have repo-checker kick this out again, as in the past.

jberry-suse:
Fixes #1102.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1671,6 +1671,10 @@ class StagingAPI(object):
         # put twice because on first put, the API adds useless maintainer
         http_PUT(url, data=meta)
 
+        # Force strict rpmlint like :Staging project does.
+        url = make_meta_url('prjconf', name, self.apiurl)
+        http_PUT(url, data='Support: rpmlint-Factory-strict')
+
         if use_frozenlinks:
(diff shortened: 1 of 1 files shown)


jberry-suse:
Likely, needs configuration for what package name to use, especially now that SLE is using adi stagings. We can apply this patch to packagelists to resolve until that bit is cleared up.

If nothing else could make a config option for it, but kinda leads me to believe the `Staging:adi` passthru project would be a better approach.

jberry-suse:
Close in favor of #1109.

PR #1106, opened 2017-08-28
totest-manager: follow another live media setup change (no multibuild)

DimStar77:


DimStar77:
multibuild was a nice idea, but is nasty, as we needed to check build status in PKG:FLAVOR and for the actual release, it was on PKG; that would be more refactorying than I am willing to do for now

nilxam:
since it's hot-patched on packagelists-VM, let's merge it now.

PR #1107, opened 2017-08-28
SUSE: wants to build test dvds on 'local' workers

coolo:


coolo:
I removed the disabling of images - I no longer need it

PR #1108, opened 2017-08-28
osclib/conf: do not disable remote config for SLE-15.

### LINKED ISSUE #1104: staging config should come from attribute
For SLE there are so many staging managers, it's no longer funny. And with the no-cleanup-list being so important for the health, it's quite important that they all share the config.

As such: please support some attribute in $PRJ:Staging that the config is taking from.

jberry-suse:
Fixes #1104.

### DIFF
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -72,7 +72,6 @@ DEFAULT = {
         'openqa': None,
         'lock': 'SUSE:%(project)s:Staging',
         'lock-ns': 'SUSE',
-        'remote-config': False,
     },
     r'SUSE:(?P<project>.*$)': {
         'staging': 'SUSE:%(project)s:Staging',


coolo:
I will look into having the no-cleanup white list in there then

jberry-suse:
@coolo Can you elaborate on what that is? The original issue indicated staging config which is what `conf.py` is and the only reference to "cleanup whitelist" is in `unselect` which I imagine is not what you are referring, but is already possible to set in remote config as we did for Leap 42.3.

coolo:
the main config I want to sync is cstaging_nocleanup

jberry-suse:
Ok, then easy. Just add `nocleanup-packages = ...` the same way it would appear in `.oscrc` to `SUSE:SLE-15:GA:Staging/dashboard/config` as it looks like @lnussel already added a config a few days ago.

PR #1109, opened 2017-08-28
osclib/stagingapi: create_adi_project(): use :Staging passthru project.

### LINKED ISSUE #1102: repo-checker no longer blocks invalid licenses

jberry-suse:
This allows for project configuration to be injected.

Fixes #1102.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1661,10 +1661,12 @@ class StagingAPI(object):
             <enable/>
           </debuginfo>
           {4}
+            <path project="{5}" repository="standard"/>
             <path project="{1}" repository="standard"/>
             <arch>x86_64</arch>
           </repository>
-        </project>""".format(name, self.project, self.extract_adi_number(name), linkproject, repository)
+        </project>""".format(name, self.project, self.extract_adi_number(name), linkproject, repository,
+                             self.cstaging)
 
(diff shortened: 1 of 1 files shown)


jberry-suse:
Update to point at just `:Staging`.

jberry-suse:
Updated to include both `:Staging` and `openSUSE:Factory` in path.

DimStar77:
Generally should be fine - the commit message does not really represent what is changed though anymore imho

jberry-suse:
I updated it, not sure when you looked:

```
osclib/stagingapi: create_adi_project(): add :Staging passthru project.
This allows for project configuration to be injected.
```

That seem clear or not?

DimStar77:
yes - that seems fine... must have had a stale screen here

PR #1110, opened 2017-08-29
stagingapi: refactor get_sub_packages() to support adi package

nilxam:
The new logic is introduced, the description is below:
* If adi package: return sub-package list according to the specfiles of main package.
* If ring package: return sub-package list by showlinked call and the sub-package must be in another ring.
* The others: return empty list, no reason to return anything or even checking it in the devel project. Otherwise selecting a non-ring package to a letter staging probably would created unneeded sub-package.

With this, https://github.com/openSUSE/osc-plugin-factory/pull/948 should work.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -260,7 +260,7 @@ class StagingAPI(object):
 
         return package_info
 
-    def get_filelist_for_package(self, pkgname, project, extension=None):
+    def get_filelist_for_package(self, pkgname, project, expand=None, extension=None):
         """
         Get a list of files inside a package container
         :param package: the base packagename to be linked to
@@ -272,6 +272,8 @@ class StagingAPI(object):
         query = {
             'extension': extension
         }
+        if expand:
+            query['expand'] = expand
 
         if extension:
             url = self.makeurl(['source', project, pkgname], query=query)
@@ -836,7 +838,7 @@ class StagingAPI(object):
             self.is_package_disabled(subprj, package, store=True)
         delete_package(self.apiurl, subprj, package, force=True, msg=msg)
 
-        for sub_prj, sub_pkg in self.get_sub_packages(package):
+        for sub_prj, sub_pkg in self.get_sub_packages(package, project):
             sub_prj = self.map_ring_package_to_subject(project, sub_pkg)
             if self._supersede:
                 self.is_package_disabled(sub_prj, sub_pkg, store=True)
@@ -1111,27 +1113,42 @@ class StagingAPI(object):
 
         return project
 
-    def get_sub_packages(self, package, project=None):
+    def get_sub_packages(self, package, project):
         """
-        Returns a list of packages that need to be linked into rings
-        too. A package is actually a tuple of project and package name
+        Returns a list of packages that need to be linked to main package.
+        For adi package, check specfiles according to the main package.
         """
         ret = []
-        if not project:
+
+        # Started the logic. Note that, return empty tuple in case selecting
+        # non-ring package to a letter staging.
+        if self.is_adi_project(project):
+            if not self.item_exists(project, package):
+                return ret
+            # For adi package, do not trust the layout in the devel project, we
+            # must to guarantee the sub-pacakges are created according to the
+            # specfiles of main package. Therefore, main package must be
+            # created before through get_sub_packages().
+            filelist = self.get_filelist_for_package(pkgname=package, project=project, expand='1', extension='spec')
+            mainspec = "{}{}".format(package, '.spec')
+            if mainspec in filelist:
+                filelist.remove(mainspec)
+            for spec in filelist:
+                ret.append((project, spec[:-5]))
+        elif self.ring_packages.get(package):
             project = self.ring_packages.get(package)
-        if not project:
-            return ret
-        url = self.makeurl(['source', project, package],
-                           {'cmd': 'showlinked'})
 
-        # showlinked is a POST for rather bizzare reasons
-        f = http_POST(url)
-        root = ET.parse(f).getroot()
+            url = self.makeurl(['source', project, package],
+                               {'cmd': 'showlinked'})
 
-        for pkg in root.findall('package'):
-            # ensure sub-package is valid
-            if pkg.get('project') in self.rings and pkg.get('name') != package:
-                ret.append((pkg.get('project'), pkg.get('name')))
+            # showlinked is a POST for rather bizzare reasons
+            f = http_POST(url)
+            root = ET.parse(f).getroot()
+
+            for pkg in root.findall('package'):
+                # ensure sub-package is valid in rings
+                if pkg.get('project') in self.rings and pkg.get('name') != package:
+                    ret.append((pkg.get('project'), pkg.get('name')))
 
         return ret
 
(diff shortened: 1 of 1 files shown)


osclib/stagingapi.py
```diff
@@ -1088,27 +1088,43 @@ def map_ring_package_to_subject(self, project, pkg):
 
         return project
 
-    def get_sub_packages(self, package, project=None):
+    def get_sub_packages(self, package, project):
         """
-        Returns a list of packages that need to be linked into rings
-        too. A package is actually a tuple of project and package name
+        Returns a list of packages that need to be linked into main package.
```
jberry-suse: nit: `s/linked into/linked to/`

osclib/stagingapi.py
```diff
@@ -1088,27 +1088,43 @@ def map_ring_package_to_subject(self, project, pkg):
...
+        For adi package, check specfiles according to the main package.
         """
         ret = []
-        if not project:
+
+        # Started the logic. Note that, return empty tuple in case selecting
+        # non-ring package to a letter staging.
+        if self.is_adi_project(project):
+            # The additional check ensure the package is exist
```
jberry-suse: nit: `item_exists(project, package)` is prety clear that checking if the package exists in project without comment.

osclib/stagingapi.py
```diff
@@ -1088,27 +1088,43 @@ def map_ring_package_to_subject(self, project, pkg):
...
+            if not self.item_exists(project, package):
+                return ret
+            # For adi package, do not trust the layout in the devel project, we
+            # must to guarantee the sub-pacakges are created according to the
+            # specfiles of main package. Therefore, main package must be
+            # created before through get_sub_packages().
```
jberry-suse: Is this check necessary as ring packages follow the same workflow? In fact on delete I imagine this would prevent proper cleanup.


```python
         delete_package(self.apiurl, subprj, package, force=True, msg=msg)
  
 -        for sub_prj, sub_pkg in self.get_sub_packages(package):
 +        for sub_prj, sub_pkg in self.get_sub_packages(package, project):
```

Meaning the package in `item_exists()` would be deleted just before `get_sub_packages()` is called. Since the source project is all that is needed and possible to be destroyed, perhaps just for deletes that should be checked. On submits it should create the project first (same as for rings) before linking to it. If the creation fails then the rest of the comments should probably crap just as they do for rings.

jberry-suse:
Alternatively, perhaps the removal logic should be changed to remove sub-packages first which more properly reflects the dependencies between them. Remove the end nodes before the root node.

nilxam:
Good catch! I didn't aware it will causes problem in removal. Fixed. Also I copied change from PR#948 as it was closed, we still requires it for https://github.com/openSUSE/osc-plugin-factory/issues/946 , let's get everything ready here.

osclib/stagingapi.py
jberry-suse: Still not sure this check is good as it might just hide failure to properly stage a package, but re-ordering delete resolved that part.

osclib/stagingapi.py
```diff
@@ -821,12 +821,15 @@ def rm_from_prj(self, project, package=None, request_id=None,
...
         subprj = self.map_ring_package_to_subject(project, package)
-        delete_package(self.apiurl, subprj, package, force=True, msg=msg)
 
-        for sub_prj, sub_pkg in self.get_sub_packages(package):
+        for sub_prj, sub_pkg in self.get_sub_packages(package, project):
             sub_prj = self.map_ring_package_to_subject(project, sub_pkg)
-            if sub_prj != subprj:  # if different to the main package's prj
-                delete_package(self.apiurl, sub_prj, sub_pkg, force=True, msg=msg)
+            # Skip inner-project links for letter staging
+            if not self.is_adi_project(project) and sub_prj == subprj: continue
```
jberry-suse: As I changed in #948 it would seem better to use `sub_prj == project` especially given that is the form use for the same check in other methods. Given no difference might as well be consistent.

nilxam:
I can't find an adi request has multiple specfiles what pending in staging today, thus I didn't pushing this, I forget to change the subject to [DO NOT MERGE] ;/
I'll find a one tomorrow or create a fake one for testing. but tomorrow is a moving day for me, if you see a multiple specfiles adi request then please just give it a try.

osclib/stagingapi.py
nilxam: hmm.. I think use `sub_prj == subprj` is correct here, `subprj` had through 'map_ring_package_to_subject` but `project` has not. Think about a scenario: FOO package in ring2 and FOO-bar in ring1.

nilxam:
okay, I tested it, and I see a bug related to expand the source, fixed now. PR is updated.

osclib/stagingapi.py
jberry-suse: The other locations with this same loop and logic have it the other way. Otherwise, those should likely be changed, but not reason for them to be different.

nilxam: alright, I changed it to https://github.com/openSUSE/osc-plugin-factory/pull/1110/commits/2b5e6a99da9b5a259033737785680d82a49dc4f6 , the same logic in the other way, also used the same variable name.

nilxam:
Fixed conflicts...

DimStar77:
I had to revert back to 3c1a264 on my machine, as osc staging accept failed - the reviews are not being cleared before the accept happened, resulting in 'permission denied - submission is not in state new'

<status code="post_request_no_permission">
  <summary>Request is in review state. You may use the force parameter to ignore this.</summary>
</status>

After reverting, osc staging accept was working again
@DimStar77

nilxam:
does it happened on adi request?

DimStar77:
Saw it on letter staging accepts - first package already failed, tried a 2nd run, again failing on the first package to accept

nilxam:
Do you have the failed request id? the only place I can guessing is https://github.com/openSUSE/osc-plugin-factory/blob/master/osclib/stagingapi.py#L851

DimStar77:
On Sat, 2017-09-16 at 02:35 -0700, Max Lin wrote:
> Do you have the failed request id? the only place I can guessing is h
> ttps://github.com/openSUSE/osc-plugin-
> factory/blob/master/osclib/stagingapi.py#L851
> 

The first one that failed was virtualbox (don't have the number) and
the 2nd attempt it failed on 522987; both looked normal and I could
accept their review manually and let them be checkd in as
accept_other_new in the end.
> 
>   
>   
> 
> 
> 
> 
> 
> {"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c5
> 5493e4bb","name":"GitHub"},"entity":{"external_key":"github/openSUSE/
> osc-plugin-factory","title":"openSUSE/osc-plugin-
> factory","subtitle":"GitHub repository","main_image_url":"https://clo
> ud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-
> 95fc-
> 7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent
> .com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-
> b52498112777.png","action":{"name":"Open in GitHub","url":"https://gi
> thub.com/openSUSE/osc-plugin-
> factory"}},"updates":{"snippets":[{"icon":"PERSON","message":"@nilxam
>  in #1110: Do you have the failed request id? the only place I can
> guessing is https://github.com/openSUSE/osc-plugin-factory/blob/maste
> r/osclib/stagingapi.py#L851"}],"action":{"name":"View Pull
> Request","url":"https://github.com/openSUSE/osc-plugin-factory/pull/1
> 110#issuecomment-329957748"}}}

nilxam:
I found it! The issue would happened on a ring2 package, eg. virtualbox, https://github.com/openSUSE/osc-plugin-factory/pull/1140 should fix this issue.

PR #1112, opened 2017-08-29
totest-manager: Adapt zSystems to new productbuilder

Soulofdestiny:


okurz:
nice! LGTM

DimStar77:
Well, we know it won't work yet - but it's a step in the right direction (ttm will now be able to detect when the build is readay to be 'released' (not published!) - but the release code won't work, as it does not know the 000product container to be the relevant part of the _multibuild enabled build

totest-manager.py
```diff
@@ -566,7 +566,7 @@ class ToTestFactoryzSystems(ToTestBase):
     main_products = ['_product:openSUSE-dvd5-dvd-s390x',
```
DimStar77: Actually, when we switch to the new prod builder, the DVD ones also are produced by it - there can;t be a mix between both builders.

Please adjust those as well.

Soulofdestiny:
I must have missed the notification mail for that
updated now...

DimStar77:
ok - changes are good.. I just know that we need more changes (more structural) in ttm, as the release process we currently have does not work on multibuild packages...

DimStar77:
After more discussion with the OBS Team, releasing 000product:FOO is now possible - which would mean this PR is completed (there are likely small issues appearing in the version checks for openSUSE-release, but we will need to approach them one by one; nothing on the OBS side should make it overly complex now)

okurz:
The main problem we wanted to adress is a new snapshot of openSUSE Tumbleweed on zSystems which is most likely more blocked by build errors that no one cares about and not an incomplete release process but I could be wrong.

DimStar77:
@okurz The build fails are 'fatal' - but we also know that the old TTM would never have done a release, since :zSystems was switched to use the new prod builder (000product) and ttm would only have touched _product; of course, - build fails is something people have to care for and submit fixes

PR #1115, opened 2017-08-30
repo_checker: rename package-whitelist to binary-whitelist to be more accurate.

### LINKED ISSUE #1096: repo_checker: remote config whitelist: s/package/binary/
The remote config variables are prefix `repo_checker-package-whitelist`, but really they are a binary whitelist.

jberry-suse:
I duplicated Factory config so it will work both ways. Once this has been confirmed deployed we can drop the old form, but best to do this sooner rather than later.

Fixes #1096.

### DIFF
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -170,7 +170,7 @@ class RepoChecker(ReviewBot.ReviewBot):
             # Only bother if staging can match arch, but layered first.
             directories.insert(0, self.mirror(project, arch))
 
-            whitelist = self.package_whitelist(project, arch)
+            whitelist = self.binary_whitelist(project, arch)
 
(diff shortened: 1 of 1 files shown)


jberry-suse:
@DimStar77 just to make sure, does this seem like better name to you? Based on confusion elsewhere I figured it was good to differentiate between the package container on OBS that produces the binaries which are the things repo_checker actually inspects. Meaning subpackages and such will not be whitelisted just because the parent container is whitelisted.

DimStar77:
@jberry-suse yes, I think calling this here binaries when it inspects binaries is better. using 'package; when we talk about about package containers inside a project, and binaries when we reference rpms

PR #1116, opened 2017-08-31
Moved packages not to cleanup into remote config

coolo:
@jberry-suse correct?

### DIFF
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -64,7 +64,6 @@ DEFAULT = {
         'staging-group': 'sle-staging-managers',  # '%(project.lower)s-staging',
         'staging-archs': 'i586 x86_64',
         'staging-dvd-archs': '',
-        'nocleanup-packages': '000product.local kernel-firmware',
         'rings': 'SUSE:%(project)s:Rings',
         'nonfree': None,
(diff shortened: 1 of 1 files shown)


coolo:
File "/suse/coolo/.osc-plugins/osclib/stagingapi.py", line 71, in __init__
    self.cstaging_nocleanup = conf.config[project]['nocleanup-packages'].split()
KeyError: 'nocleanup-packages'

Doesn't work - this needs to be lazy loaded

jberry-suse:
A yeah, I remember this caveat, since the dashboard loading functions are in stagingapi, it has to be loaded first, but obviously those config variables are loaded then. Can't remember everything. :( I had two thoughts if these were ever to be loaded from remote config:

1: lazy load/reload once remote config loaded
2: move dashboard content loading functions outside of `stagingapi` (at this point to `osclib/core`)

The first option felt a bit weird as somewhat have to pick and choose which are allowed since things like `cstaging` should not be allowed to change as the end result would seem rather odd, but perhaps not really an issue. The second option still relies on the `cstaging` variable in order to know where to look for dashboard.

At this point probably want to just lazy load it. I'll create a PR.

PR #1117, opened 2017-08-31
osclib/stagingapi: lazy-load nocleanup-packages to allow for placement in remote config.

jberry-suse:
This seems like the best solution since any other default values make little sense to change in remote config.

Replaces #1116.


jberry-suse:
For accurate trail (related to #1104).

PR #1126, opened 2017-09-01
dist/ci: rework spec.sh as obs-build-target using `osc build` and provide easy local run, Leap:42.3, and SLE-12-SP3 builds.

jberry-suse:
- cbadb74069906611546bdcb03614519d4193b1c3:
    travis: add distribution builds for Leap 42.3 and SLE-12-SP3.

- 19f6d4af604a594713f7bb876239ab72aab26de5:
    dist/ci: rework spec.sh as obs-build-target using `osc build`.
    
    Instead of performing a slim rpm build and check directly inside the
    container a proper build is performed using `osc build`. The target
    package is checked out the _service file pointed at the local copy
    and then `osc service disabledrun` before building.

- 11b0f473c6e2e62a8818c08a4a74d8a1d94441c1:
    dist/ci: extract common osc setup as osc-init and provide in /usr/bin.

- 18f61d57d00b792621001f337191019fd9204792:
    dist/ci/Dockerfile: include obs-deploy in /usr/bin.
    
    Paving the way for this image to be built outside of this repository.

- 6c7c5a19a1d4705004ced06c0d982d2968ebeded:
    dist/ci: rename deploy.obs.sh to obs-deploy.

- 844325e31a170fcce84c960986985761621bb402:
    dist/ci/deploy: only commit if changes available to commit.

- 903bb0921fba82da3dacd5a2dd544edf5643d73e:
    dist/ci/deploy: only add line if missing from _servicedata.
    
    Avoids duplication when multiple PRs are merged quickly after one another.

- 65ce835ffce899da24456288895c274767683ae2:
    dist/ci/Dockerfile: switch to base tumbleweed image.
    
    The official base image is now updated regularly so no need to use
    custom rebuilding image.


jberry-suse:
Unfortunately, this will not work without an obs user that can be used during merge requests which is very unfortunate. Really just looking to do an anonymous `osc checkout` and local `osc build`.

jberry-suse:
Magic! The script now falls back to `/public` API if no pass so it will automatically switch to public during merges and build verification while using proper credentials during deploy.

jberry-suse:
Well, close. The checkout succeeds, but the `/source/openSUSE:Tools/openSUSE-release-tools?cmd=getprojectservices` call returns a `404` which seems doubly wrong since: a) its a permission issue, and b) why is it restricted. Lastly seems interesting that when doing a `localrun` the OBS service is still queried to parse the _service file sitting on my machine?

jberry-suse:
Looks like I can patch `osc` to not be silly. Will do later.

jberry-suse:
Created openSUSE/osc#342 documented the tool deficiencies faced.

jberry-suse:
Rebased and added a throwaway account since OBS rejected easy API relaxing.

jberry-suse:
It is now trivial to test spec changes the same way they are run on travis, but locally.

PR #1128, opened 2017-09-03
totest-manager: Prevent old tests in same build blocking release with "latest=1"

okurz:


DimStar77:
Should be ok - even though it is merely a workaround against reposted jobs (which should not happen to start with) - but no reason to block that, should it happen

PR #1131, opened 2017-09-07
leaper: fix iterating through list

lnussel:


mimi1vx:
+1 , only isn't through list:)

PR #1132, opened 2017-09-07
Generate minimals with set compression

mimi1vx:
Also fixes problem with unnecessary spaces in list of packages


coolo:
read a python book? :)

mimi1vx:
@coolo Python Essential Reference, 9 years ago :) but in this time was set and comprehension hot news ( set comprehension comes with 2.7 ).

The main problem fixed by this commit is missing rstrip on line:) Unfortuanetly some records in this file have whitespaces on end


Btw: this construction in Python is directly taken from Haskell

PR #1136, opened 2017-09-12
Implements virtually accept the delete request

nilxam:
https://progress.opensuse.org/issues/20420

Workflow:
* selecting a request to a staging, the pseudometa is changed to 
{id: 100, package: FOO, author: user, type: submit}
{id: 101, package: BAR, author: user, type: delete}

* during accepting a delete request:
If 'type' is exists in pseudometa and delreq-monitor was set in conf.py, the delete request will through *virtual accept* process, the binary will be removed on standard repo. If 'type' is not exists, go through the old route make sure it will not break the current stuff.

* in the later check-in round:
If it has found the virtual accepted request's package doesn't have binary on any build repo in the main project, the accept command accepts delreq-monitor review and continue accepts this request to the main project via accept_other_new().

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -3,9 +3,9 @@ import urllib2
 import warnings
 from xml.etree import cElementTree as ET
 
-from osc.core import change_request_state
+from osc.core import change_request_state, show_package_meta, wipebinaries
 from osc.core import http_GET, http_PUT, http_DELETE, http_POST
-from osc.core import delete_package
+from osc.core import delete_package, search, set_devel_project
 from datetime import date
 
 
@@ -14,7 +14,40 @@ class AcceptCommand(object):
         self.api = api
 
     def find_new_requests(self, project):
-        query = "match=state/@name='new'+and+(action/target/@project='{}'+and+action/@type='submit')".format(project)
+        query = "match=state/@name='new'+and+action/target/@project='{}'".format(project)
+        url = self.api.makeurl(['search', 'request'], query)
+
+        f = http_GET(url)
+        root = ET.parse(f).getroot()
+
+        rqs = []
+        for rq in root.findall('request'):
+            pkgs = []
+            act_type = None
+            actions = rq.findall('action')
+            for action in actions:
+                act_type = action.get('type')
+                targets = action.findall('target')
+                for t in targets:
+                    pkgs.append(str(t.get('package')))
+
+            rqs.append({'id': int(rq.get('id')), 'packages': pkgs, 'type': act_type})
+        return rqs
+
+    def virtual_accept_request_has_no_binary(self, project, package):
+        filelist = self.api.get_filelist_for_package(pkgname=package, project=self.api.project, expand='1', extension='spec')
+        pkgs = self.api.extract_specfile_short(filelist)
+
+        for pkg in pkgs:
+            query = {'view': 'binarylist', 'package': pkg, 'multibuild': '1'}
+            pkg_binarylist = ET.parse(http_GET(self.api.makeurl(['build', project, '_result'], query=query))).getroot()
+            for binary in pkg_binarylist.findall('./result/binarylist/binary'):
+                return False
+
+        return True
+
+    def find_virtually_accepted_requests(self, project):
+        query = "match=state/@name='review'+and+(action/target/@project='{}'+and+action/@type='delete')+and+(review/@state='new'+and+review/@by_group='{}')".format(project, self.api.delreq_review)
         url = self.api.makeurl(['search', 'request'], query)
 
         f = http_GET(url)
@@ -51,6 +84,24 @@ class AcceptCommand(object):
         content = ET.tostring(root)
         http_PUT(url + '?comment=accept+command+update', data=content)
 
+    def virtually_accept_delete(self, request_id, package):
+        self.api.add_review(request_id, by_group=self.api.delreq_review, msg='Request accepted. Cleanup in progress - DO NOT REVOKE!')
+
+        filelist = self.api.get_filelist_for_package(pkgname=package, project=self.api.project, expand='1', extension='spec')
+        pkgs = self.api.extract_specfile_short(filelist)
+
+        # Disable build and wipes the binary to the package and the sub-package
+        for pkg in pkgs:
+            meta = show_package_meta(self.api.apiurl, self.api.project, pkg)
+            meta = ''.join(meta)
+            # Update package meta to disable build
+            self.api.create_package_container(self.api.project, pkg, meta=meta, disable_build=True)
+            wipebinaries(self.api.apiurl, self.api.project, package=pkg, repo=self.api.main_repo)
+
+            # Remove package from Rings
+            if self.api.ring_packages.get(pkg):
+                delete_package(self.api.apiurl, self.api.ring_packages.get(pkg), pkg, force=True, msg="Cleanup package in Rings")
+
     def perform(self, project, force=False):
         """Accept the staging project for review and submit to Factory /
         openSUSE 13.2 ...
@@ -73,16 +124,22 @@ class AcceptCommand(object):
             self.api.rm_from_prj(project, request_id=req['id'], msg='ready to accept')
             packages.append(req['package'])
             msg = 'Accepting staging review for {}'.format(req['package'])
-            print(msg)
 
             oldspecs = self.api.get_filelist_for_package(pkgname=req['package'],
                                                          project=self.api.project,
                                                          extension='spec')
-            change_request_state(self.api.apiurl,
-                                 str(req['id']),
-                                 'accepted',
-                                 message='Accept to %s' % self.api.project)
-            self.create_new_links(self.api.project, req['package'], oldspecs)
+            if 'type' in req and req['type'] == 'delete' and self.api.delreq_review:
+                msg += ' and started handling of virtual accept process'
+                print(msg)
+                # Virtually accept the delete request
+                self.virtually_accept_delete(req['id'], req['package'])
+            else:
+                print(msg)
+                change_request_state(self.api.apiurl,
+                                     str(req['id']),
+                                     'accepted',
+                                     message='Accept to %s' % self.api.project)
+                self.create_new_links(self.api.project, req['package'], oldspecs)
 
         self.api.accept_status_comment(project, packages)
         self.api.staging_deactivate(project)
@@ -128,6 +185,16 @@ class AcceptCommand(object):
 
     def accept_other_new(self):
         changed = False
+
+        if self.api.delreq_review:
+            rqlist = self.find_virtually_accepted_requests(self.api.project)
+            for req in rqlist:
+                if self.virtual_accept_request_has_no_binary(self.api.project, req['packages'][0]):
+                    # Accepting delreq-review review
+                    self.api.do_change_review_state(req['id'], 'accepted',
+                                                    by_group=self.api.delreq_review,
+                                                    message='Virtually accepted delete {}'.format(req['packages'][0]))
+
         rqlist = self.find_new_requests(self.api.project)
         if self.api.cnonfree:
             rqlist += self.find_new_requests(self.api.cnonfree)
@@ -135,12 +202,23 @@ class AcceptCommand(object):
         for req in rqlist:
             oldspecs = self.api.get_filelist_for_package(pkgname=req['packages'][0], project=self.api.project, extension='spec')
             print 'Accepting request %d: %s' % (req['id'], ','.join(req['packages']))
+            if req['type'] == 'delete':
+                # Remove devel project/package tag before accepting the request
+                self.remove_obsoleted_develtag(self.api.project, req['packages'][0])
             change_request_state(self.api.apiurl, str(req['id']), 'accepted', message='Accept to %s' % self.api.project)
             # Check if all .spec files of the package we just accepted has a package container to build
             self.create_new_links(self.api.project, req['packages'][0], oldspecs)
             changed = True
 
         return changed
+
+    def remove_obsoleted_develtag(self, project, package):
+        xpath = {
+            'package': "@project='%s' and devel/@project=@project and devel/@package='%s'" % (project, package),
+        }
+        collection = search(self.api.apiurl, **xpath)['package']
+        for pkg in collection.findall('package'):
+            set_devel_project(self.api.apiurl, project, pkg.attrib['name'], devprj=None)
 
     def create_new_links(self, project, pkgname, oldspeclist):
         filelist = self.api.get_filelist_for_package(pkgname=pkgname, project=project, extension='spec')
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -44,6 +44,8 @@ DEFAULT = {
         'openqa': 'https://openqa.opensuse.org',
         'lock': 'openSUSE:%(project)s:Staging',
         'lock-ns': 'openSUSE',
+        'delreq-review': 'factory-maintainers',
+        'main-repo': 'standard',
     },
     r'openSUSE:(?P<project>Leap:[\d.]+)': {
         'staging': 'openSUSE:%(project)s:Staging',
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -75,6 +75,8 @@ class StagingAPI(object):
         self.cproduct = conf.config[project]['product']
         self.copenqa = conf.config[project]['openqa']
         self.user = conf.get_apiurl_usr(apiurl)
+        self.delreq_review = conf.config[project]['delreq-review']
+        self.main_repo = conf.config[project]['main-repo']
         self._ring_packages = None
         self._ring_packages_for_links = None
         self._packages_staged = None
@@ -259,6 +261,11 @@ class StagingAPI(object):
         package_info['package'] = linkinfo.attrib['package']
 
         return package_info
+
+    def extract_specfile_short(self, filelist):
+        packages = [spec[:-5] for spec in filelist if re.search('\.spec$', spec)]
+
+        return packages
 
     def get_filelist_for_package(self, pkgname, project, expand=None, extension=None):
         """
@@ -748,7 +755,7 @@ class StagingAPI(object):
     def clear_prj_pseudometa(self, project):
         self.set_prj_pseudometa(project, {})
 
-    def _add_rq_to_prj_pseudometa(self, project, request_id, package):
+    def _add_rq_to_prj_pseudometa(self, project, request_id, package, act_type=None):
         """
         Records request as part of the project within metadata
         :param project: project to record into
@@ -761,13 +768,14 @@ class StagingAPI(object):
         for request in data['requests']:
             if request['package'] == package:
                 # Only update if needed (to save calls to get_request)
-                if request['id'] != request_id or not request.get('author'):
+                if request['id'] != request_id or not request.get('author') or not request.get('type'):
                     request['id'] = request_id
+                    request['type'] = act_type
                     request['author'] = get_request(self.apiurl, str(request_id)).get_creator()
                 append = False
         if append:
             author = get_request(self.apiurl, str(request_id)).get_creator()
-            data['requests'].append({'id': request_id, 'package': package, 'author': author})
+            data['requests'].append({'id': request_id, 'package': package, 'author': author, 'type': act_type})
         self.set_prj_pseudometa(project, data)
 
     def set_splitter_info_in_prj_pseudometa(self, project, group, strategy_info):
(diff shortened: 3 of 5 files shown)


nilxam:
TODO:

* no need delreq-monitor anymore, replace it to factory-maintainers group.
* verify sub-package's binary also empty.

nilxam:
TODO list is clean up, please review!

osclib/accept_command.py
```diff
@@ -32,6 +32,39 @@ def find_new_requests(self, project):
             rqs.append({'id': int(rq.get('id')), 'packages': pkgs})
         return rqs
 
+    def virtual_accept_request_has_no_binary(self, project, package):
+        pkgs = []
+        filelist = self.api.get_filelist_for_package(pkgname=package, project=self.api.project, expand='1', extension='spec')
+        for spec in filelist:
+            pkgs.append(spec[:-5])
```
jberry-suse: Since this same sort of logic is done in #1110, perhaps extracting as another method in StagingAPI and using for both would make sense (ie a method to return list of subpackages based on spec files in main package). Might as well merge the other and refactor in this request along with all the additional usages.

osclib/accept_command.py
```diff
@@ -32,6 +32,39 @@ def find_new_requests(self, project):
...
+
+        for pkg in pkgs:
+            query = {'view': 'binarylist', 'package': pkg, 'multibuild': '1'}
+            pkg_binarylist = ET.parse(http_GET(self.api.makeurl(['build', project, '_result'], query=query))).getroot()
```
jberry-suse: Without running the queries to double check this seems like the same information provided by `osclib.core.package_binary_list()`. From the comment and my memory multibuild results are returned there as well. Perhaps try it and just and check if that returns an non-empty list. Otherwise the `binary_list()` method is the one used by duplicate binary check.

The main difference is the details of the information returned, but you only seem to care if any binaries or not so I would expect them to be sufficient.

osclib/accept_command.py
```diff
@@ -51,6 +84,23 @@ def reset_rebuild_data(self, project):
         content = ET.tostring(root)
         http_PUT(url + '?comment=accept+command+update', data=content)
 
+    def virtual_accept_to_delete(self, request_id, package):
+        self.api.add_review(request_id, by_group=self.api.delreq_monitor)
+
+        pkgs = []
+        filelist = self.api.get_filelist_for_package(pkgname=package, project=self.api.project, expand='1', extension='spec')
+        for spec in filelist:
+            pkgs.append(spec[:-5])
```
jberry-suse: Another need for having a method to return list of subpackages based on spec files as noted above.

osclib/accept_command.py
```diff
@@ -51,6 +84,23 @@ def reset_rebuild_data(self, project):
...
+
+        # Disable build and wipes the binary to the package and the sub-package
+        for pkg in pkgs:
+            meta = show_package_meta(self.api.apiurl, self.api.project, pkg)
+            meta = ''.join(meta)
+            # Update package meta with disabling build
```
jberry-suse: nit: `s/with disabling/to disable/`

osclib/accept_command.py
```diff
@@ -51,6 +84,23 @@ def reset_rebuild_data(self, project):
...
+            self.api.create_package_container(self.api.project, pkg, meta=meta, disable_build=True)
+            url = self.api.makeurl(['build', self.api.project], {'cmd': 'wipe', 'repository': self.api.main_repo, 'package': pkg})
```
jberry-suse: `osc.core.wipebinaries()`

osclib/accept_command.py
```diff
@@ -78,11 +128,15 @@ def perform(self, project, force=False):
             oldspecs = self.api.get_filelist_for_package(pkgname=req['package'],
                                                          project=self.api.project,
                                                          extension='spec')
-            change_request_state(self.api.apiurl,
-                                 str(req['id']),
-                                 'accepted',
-                                 message='Accept to %s' % self.api.project)
-            self.create_new_links(self.api.project, req['package'], oldspecs)
+            if 'type' in req and req['type'] == 'delete' and self.api.delreq_monitor:
+                # Virtual accept to the delete request
```
jberry-suse: nit: `s/Virtual/Virtually/`, `s/to //`

osclib/accept_command.py
```diff
@@ -78,11 +128,15 @@ def perform(self, project, force=False):
...
+                self.virtual_accept_to_delete(req['id'], req['package'])
```
jberry-suse: nit: just as above the `to` is a bit odd. Perhaps `virtual_accept_delete` or `virtually_accept_delete`.

osclib/accept_command.py
```diff
@@ -140,8 +194,32 @@ def accept_other_new(self):
             self.create_new_links(self.api.project, req['packages'][0], oldspecs)
```
jberry-suse: `oldspecs` here is another example of the need for that method, also note it does not use `expand` is that correct in this case?

osclib/accept_command.py
```diff
@@ -140,8 +194,32 @@ def accept_other_new(self):
...
 
+        if self.api.delreq_monitor:
+            rqlist = self.find_virtual_accepted_requests(self.api.project)
+            for req in rqlist:
+                if self.virtual_accept_request_has_no_binary(self.api.project, req['packages'][0]):
+                    print 'Accepting request %d: %s' % (req['id'], ','.join(req['packages']))
+                    # Accepting delreq-monitor review
+                    self.api.do_change_review_state(req['id'], 'accepted',
+                                                    by_group=self.api.delreq_monitor,
+                                                    message='Virtual accepted to delete {}'.format(req['packages'][0]))
```
jberry-suse: nit: same wording change as above. `s/Virtual/Virtually/`, `s/to //`

Any reason to list the packages in the review state change as I would expect that to be clear given that is placed on the request? The rest of the text would seem sufficient.

osclib/accept_command.py
```diff
@@ -140,8 +194,32 @@ def accept_other_new(self):
...
+                    # Delete sub-package links before accepting the request
+                    self.delete_obsoleted_links(self.api.project, req['packages'][0], str(req['id']))
+                    change_request_state(self.api.apiurl, str(req['id']), 'accepted', message='Accept to %s' % self.api.project)
+                    changed = True
+
         return changed
 
+    def delete_obsoleted_links(self, project, pkgname, request_id):
+        filelist = self.api.get_filelist_for_package(pkgname=pkgname, project=project, expand='1', extension='spec')
```
jberry-suse: Another case for that method and the `spec[:-5]`. Definitely should be extracted as a stand-alone method.

osclib/accept_command.py
```diff
@@ -140,8 +194,32 @@ def accept_other_new(self):
...
+                                                    by_group=self.api.delreq_monitor,
+                                                    message='Virtual accepted to delete {}'.format(req['packages'][0]))
+                    # Delete sub-package links before accepting the request
+                    self.delete_obsoleted_links(self.api.project, req['packages'][0], str(req['id']))
+                    change_request_state(self.api.apiurl, str(req['id']), 'accepted', message='Accept to %s' % self.api.project)
+                    changed = True
+
         return changed
 
+    def delete_obsoleted_links(self, project, pkgname, request_id):
```
jberry-suse: Curious, so previously linked subpackages were not removed?

osclib/conf.py
```diff
@@ -44,6 +44,8 @@
         'openqa': 'https://openqa.opensuse.org',
         'lock': 'openSUSE:%(project)s:Staging',
         'lock-ns': 'openSUSE',
+        'delreq-monitor': 'factory-maintainers',
+        'main-repo': 'standard',
```
jberry-suse: This should likely be used in `repo_checker` as well since for lack of having a proper variable I kept the hard-coded `standard`. I'll update that once this is merged.

osclib/conf.py
```diff
@@ -44,6 +44,8 @@
         'openqa': 'https://openqa.opensuse.org',
         'lock': 'openSUSE:%(project)s:Staging',
         'lock-ns': 'openSUSE',
+        'delreq-monitor': 'factory-maintainers',
```
jberry-suse: I am not clear what is the purpose behind the word `monitor` in the key. Seems more like a `virtual delete request queue` or `backlog` or `pending list`. I understand the list will be monitored to determine when it is appropriate to actually accept the delete, but given this controls the `virtual delete` request feature it seems like that would be more appropriate to be in the key name.

Perhaps `virtual-delete-review` or somesuch as that provides a much better idea of what this is for at a glance.

osclib/stagingapi.py
```diff
@@ -1254,7 +1263,7 @@ def list_requests_in_prj(self, project):
 
         return list
 
-    def add_review(self, request_id, by_project=None, by_group=None, msg=None):
+    def add_review(self, request_id, by_project=None, by_group=None, by_user=None, msg=None):
```
jberry-suse: I assume this is no longer necessary since using `factory-maintainers` group?

osclib/conf.py
```diff
@@ -58,6 +60,8 @@
...
```
jberry-suse: Would it make sense to use a different user/group like you originally did to avoid the continued mixing of `factory` in `non-factory` projects?

osclib/accept_command.py
```diff
@@ -32,6 +32,39 @@ def find_new_requests(self, project):
...
+
+        for pkg in pkgs:
+            query = {'view': 'binarylist', 'package': pkg, 'multibuild': '1'}
+            pkg_binarylist = ET.parse(http_GET(self.api.makeurl(['build', project, '_result'], query=query))).getroot()
+            for binary in pkg_binarylist.findall('./result/binarylist/binary'):
+                return False
+
+        return True
+
+    def find_virtual_accepted_requests(self, project):
```
jberry-suse: Rather than duplicate all the dom walking code from `find_new_requests()` it would seem to make more sense to rename that method to somethign like `find_requests()` and add a parameter to indicate looking for virtual delete requests.

osclib/accept_command.py
nilxam: 1) Nope, it will be in *broken* state and be removed manually
2) The main reason, if the package has set *devel* package, accepting delete request won't work, eg. BAR is FOO's sub-package and it has set FOO is its devel package, accepting the delete request against FOO will fail due to `Package is used by following packages as devel package: BAR`.

nilxam: discussed with Dominique, this should not need, what we need is removing *devel* tag in meta in case it was the sub-package of FOO, and FOO is a delete request.

osclib/conf.py
DimStar77: the idea here was that 'the one that will have to accept it into oS:F in the end must be in factory-maintainers' anyway; and the 'last step', removing the review of the 'delreq-monitor', would happen directly in the ```osc staging accept``` call; so definitively by a user in the factory-maintainers group.

If we take a 2nd group, it _must_ have all the member of factory-maintainers, plus potentially more (which did not seem to realistic)

jberry-suse: So I guess this is a symptom of the larger problem that that group is also used for Leap already. As such it makes sense to use in this context, but should be considered for changing for Leap in general at some point.

DimStar77: indeed - so fo far the group allowed to write into the product directly was small and the same set of users is used across TW and Leap (to back each other up);

jberry-suse: So perhaps `masters-of-the-obs` :)

DimStar77: That sounds more like the obs admins; happy not to be part of that (most of the times)

osclib/accept_command.py
nilxam: it's correct.

osclib/accept_command.py
nilxam: hmm... I don't want to hardcoding arch, the binary must be removed in any arch.

nilxam:
updated, still need to test it in real.

DimStar77:
Fai

```
DELETE https://api.opensuse.org/source/openSUSE:Factory:Staging:D:DVD/ecj-bootstrap?comment=ready+to+accept&force=1
GET https://api.opensuse.org/request/527833?withfullhistory=1
GET https://api.opensuse.org/request/527833?withfullhistory=1
GET https://api.opensuse.org/request/527833?withfullhistory=1
POST https://api.opensuse.org/request/527833?newstate=accepted&cmd=changereviewstate&by_project=openSUSE%3AFactory%3AStaging%3AD
Accepting staging review for ecj-bootstrap
CACHE_MISS https://api.opensuse.org/source/openSUSE:Factory/ecj-bootstrap?extension=spec (does not exist)
GET https://api.opensuse.org/source/openSUSE:Factory/ecj-bootstrap?extension=spec
CACHE_PUT https://api.opensuse.org/source/openSUSE:Factory/ecj-bootstrap?extension=spec openSUSE:Factory
GET https://api.opensuse.org/source/openSUSE:Factory:Staging/_attribute/openSUSE:LockedBy
POST https://api.opensuse.org/source/openSUSE:Factory:Staging/_attribute/openSUSE:LockedBy
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python2.7/site-packages/osc/babysitter.py", line 61, in run
    return prg.main(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 501, in onecmd
  File "/usr/lib/python2.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
  File "/home/dimstar_suse/.osc-plugins/osc-staging.py", line 433, in do_staging
    if cmd.perform(api.prj_from_letter(prj), opts.force):
  File "/home/dimstar_suse/.osc-plugins/osclib/accept_command.py", line 130, in perform
    self.virtually_accept_delete(req['id'], req['package'])
  File "/home/dimstar_suse/.osc-plugins/osclib/accept_command.py", line 88, in virtually_accept_delete
    self.api.add_review(request_id, by_group=self.api.delreq_reivew)
AttributeError: 'StagingAPI' object has no attribute 'delreq_reivew'
```

osclib/accept_command.py
```diff
@@ -51,6 +84,20 @@ def reset_rebuild_data(self, project):
         content = ET.tostring(root)
         http_PUT(url + '?comment=accept+command+update', data=content)
 
+    def virtually_accept_delete(self, request_id, package):
+        self.api.add_review(request_id, by_group=self.api.delreq_reivew)
```
DimStar77: typo.. this should be self.api.delreq_review

nilxam: sigh...

osclib/accept_command.py
```diff
@@ -128,20 +179,42 @@ def cleanup(self, project):
...
             change_request_state(self.api.apiurl, str(req['id']), 'accepted', message='Accept to %s' % self.api.project)
             # Check if all .spec files of the package we just accepted has a package container to build
             self.create_new_links(self.api.project, req['packages'][0], oldspecs)
             changed = True
 
         return changed
 
+    def remove_obsoleted_develtag(self, project, package):
+        xpath = {
+            'package': "@project='%s' and devel/@project='%s' and devel/@package='%s'" % (project, project, package),
```
jberry-suse: This removes the sub-packages which have their devel project set to the target project right? Not sure if you like, but one tweak that I checked works is:

```
@project='%s' and devel/@project=@project and devel/@package='%s'
```

to avoid adding project twice.

osclib/accept_command.py
```diff
@@ -14,7 +14,40 @@ def __init__(self, api):
         self.api = api
 
     def find_new_requests(self, project):
-        query = "match=state/@name='new'+and+(action/target/@project='{}'+and+action/@type='submit')".format(project)
+        query = "match=state/@name='new'+and+action/target/@project='{}'".format(project)
```
jberry-suse: This will now include things like `change_devel` requests. Are you sure the rest of the code can handle them properly or should it explicitly looks for submit or delete requests?

I had to restricted explicitly for SLE in #1121 which is the equivalent of `change_devel` in Factory so it be good to make sure we do not introduce another case.

osclib/accept_command.py
```diff
@@ -128,20 +179,42 @@ def cleanup(self, project):
...
 
         return changed
 
+    def remove_obsoleted_develtag(self, project, package):
+        xpath = {
+            'package': "@project='%s' and devel/@project='%s' and devel/@package='%s'" % (project, project, package),
+        }
+        collection = search(self.api.apiurl, **xpath)['package']
+        pkglist = [p.attrib['name'] for p in collection.findall('package')]
+        for pkg in pkglist:
```
jberry-suse: Any reason to build a list and then iterate over it rather than just looping over the original data structure?

```python
for pkg in collection.findall('package'):
    set_devel_project(self.api.apiurl, project, pkg.attrib['name'], devprj=None)
```

osclib/accept_command.py
```diff
@@ -78,11 +125,15 @@ def perform(self, project, force=False):
             oldspecs = self.api.get_filelist_for_package(pkgname=req['package'],
                                                          project=self.api.project,
                                                          extension='spec')
-            change_request_state(self.api.apiurl,
-                                 str(req['id']),
-                                 'accepted',
-                                 message='Accept to %s' % self.api.project)
-            self.create_new_links(self.api.project, req['package'], oldspecs)
+            if 'type' in req and req['type'] == 'delete' and self.api.delreq_review:
+                # Virtually accept to the delete request
```
jberry-suse: nit: `s/to //`

Although really it now says also exactly the same thing the method name does.

osclib/accept_command.py
DimStar77: No, it only removes the devel attribute from the packages - it does not remove the package (but as long as the devel attribute is set, accepting the delete request fails; with or without the virtual delete

osclib/accept_command.py
nilxam: any kind of request must be reviewed by human and bot, if it's state is `new` then means all reviews are passed, in the past, it was merged manually(either change_devel request to Factory or add_role to SLE), this change just auto-merge those kind of request while executing accept_other_new.

osclib/accept_command.py
nilxam: just different coding style not you like...

osclib/accept_command.py
nilxam: It only removes the devel tag from the package if it has set devel pacakge in the same project.

DimStar77:
Last accept round was done with this PR merged on my local branch

The package ```python-argparse``` was handled by the delete process as defined:

* the package has been build disabled in openSUSE:Factory
* The binaries have been removed from openSUSE:Factory/standard (but kept in /totest and /snapshot)
* The actual delete request recevied a new review for 'factory-maintainers'; no more in the staging backlog

Once the current snapshot (0928) is being published, the binaries should all be gone and the 'accepting of the delete request' should be completed (to be seen)

osclib/accept_command.py
jberry-suse: I understand they were reviewed, a question of can the code handle the structure of those different types of requests or does it make assumptions as the code I had to restrict did.

Is this needed for virtual delete requests? I do not remember this change in original PR and seems like a different topic entirely.

osclib/accept_command.py
jberry-suse: Indeed, nothing incorrect about it, just curious as typically looping twice is worse than looping once.

osclib/accept_command.py
jberry-suse: Yes, that's what I meant. Obviously, deleting a devel project would be a bit crazy.

osclib/accept_command.py
jberry-suse: Well, loops an extra time and allocates an entire extra array. Again time, but I see no upside.

DimStar77:
> Assuming the drop of request type filter in find_new_requests() does not cause issues with change_devel requests. It would seem to make sense to restrict to the types that the code actually wants to handle.

Imho, accept_other_new should really handle all 'new' requests; any incoming request on openSUSE:Factory receives a 'review' at least on factory-staging and a couple bots.

at least factory-staging is a 'human review' for non-submit and non-delete requests, as those are not grouped into staging projects (and staging list / staging select ignores them)

nilxam:
Added an additional change to cope with 'virtual accept' in Rings. The package is deleted in theory thus should happened in Rings for real.

<pre>
if self.api.ring_packages.get(pkg):
    delete_package(self.api.apiurl, self.api.ring_packages.get(pkg), pkg, force=True, msg="Cleanup package in Rings")
</pre>

PR #1139, opened 2017-09-15
Disable ppc64le for now in Tumbleweed stagings

DimStar77:


jberry-suse:
Looks like tests need to be updated. :)

DimStar77:
meh - it was too easy :) this is not a setting that happens to be read from config, right?

jberry-suse:
From remote config? no, but could lazy load like the cleanup list.

PR #1140, opened 2017-09-16
rm_from_prj: always use the origin project path to set_review()

nilxam:
The variable project might got changed through map_ring_package_to_subject(), according to release process, we have to given the origin project path to set_review() always.

The case would like - a ring2 pacakge's review was a letter staging project and it can not be accepted as the `project` variable is changed to `$PROJECT:DVD` , accepts `:DVD` review is wrong. Regression from https://github.com/openSUSE/osc-plugin-factory/commit/831c73a3bc2cfca45bbb69a10244ad295329ea40

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -832,6 +832,7 @@ class StagingAPI(object):
         if not package or not request_id:
             return
 
+        orig_project = project
         self._remove_package_from_prj_pseudometa(project, package)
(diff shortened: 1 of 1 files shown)


nilxam:
merged now avoiding breaks next check-in round, also can be verify in the next check-in round(although seems the last snapshot have product issue on openqa...)

CC @DimStar77

PR #1143, opened 2017-09-19
osclib/stagingapi: accept_status_comment(): avoid making comments for empty projects.

### LINKED ISSUE #1142: adi: fails to delete project and repeatedly triggers empty accepted comments
Two parts:

1) the failure to delete project seems like OBS issue, but need to debug
2) the empty comments should be avoided in general as they can occur when a request is revoked, unselected and adi left empty

An example, is `adi:40` in Factory:

image

jberry-suse:
Fixes #1142. The comments part anyway.


jberry-suse:
Hot-patched packagelists to stop the email spam.

PR #1145, opened 2017-09-20
osclib/adi: include force=True when invoking delete_project().

### LINKED ISSUE #1144: adi: *force* project deletion to handle branched references
See #1142. Such requests could automatically be issued and is probably the only reasonable automated approach since it still allows for human sign-off.

jberry-suse:
If users build against an adi project OBS will refuse to delete, but in
that case breaking other projects is fine as the branches should only be
used to work on fixes seen in adi stagings.

See #1142 for an example, details, and discussion.

Fixes #1144.

### DIFF
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -52,7 +52,7 @@ class AdiCommand:
                 packages.append(req['package'])
             self.api.accept_status_comment(project, packages)
             try:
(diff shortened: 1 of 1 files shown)


jberry-suse:
Tested locally and `adi:40` (#1142) was successfully deleted.

DimStar77:
👍

PR #1147, opened 2017-09-26
Package list generator TNG

coolo:
For whatever reason it was impossible to create a PR from my fork :(


coolo:
@lnussel - it's a bit underdocumented at this point, but it works very reliable for SLE 15 at this point

pkglistgen.py
```diff
@@ -0,0 +1,622 @@
...
+        archs = ('*',) + ARCHITECTURES
+        for arch in archs:
+            x = self.toxml(arch)
+            print(ET.tostring(x, pretty_print=True))
+
+
+class PkgListGen(ToolBase.ToolBase):
+
+    def __init__(self, repostr):
+        ToolBase.ToolBase.__init__(self)
```
mimi1vx: ToolBase.ToolBase is new style class so `super(PkgListGen, self).__init__()` is better

PR #1148, opened 2017-09-26
Rework openqa-maintenance bot

mimi1vx:


oqamaint/ReviewBot.py
```diff
@@ -0,0 +1 @@
+../ReviewBot.py
```
jberry-suse: Is this considered good practice in python world? In other languages one would use from parent location.

It seems relative imports to parent directory is also a thing in python.

oqamaint/cli.py
```diff
@@ -0,0 +1,126 @@
...
+import simplejson as json
+
+import ReviewBot
+from openqabot import OpenQABot
+from suse import SUSEUpdate
+from opensuse import openSUSEUpdate
+from logger import ColorFormatter
+
+from openqa_client.client import OpenQA_Client
+import osc
```
jberry-suse: Perhaps alphabetize or something consistent?

oqamaint/logger.py
```diff
@@ -0,0 +1,44 @@
...
+
+RESET_SEQ = "\033[0m"
+COLOR_SEQ = "\033[1;{}m"
+
+COLORS = {
+    'WARNING': YELLOW,
+    'INFO': GREEN,
+    'DEBUG': BLUE,
+    'CRITICAL': RED,
+    'ERROR': RED}
```
jberry-suse: Colors are already handled using colorama, please grep for plenty of usage examples, and perhaps investigate using setup that allows for color overrides in oscrc (see `osc-staging.py`).

oqamaint/cli.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if self.options.force:
+            bot.force = True
+
+        bot.do_comments = self.options.comment
+
+        if not self.options.openqa:
+            raise osc.oscerr.WrongArgs("missing openqa url")
+
+        bot.openqa = OpenQA_Client(server=self.options.openqa)
```
jberry-suse: The normal style is to initialize things like this in the bot constructor and avoid overriding `postoptparse` unless absolutely necessary. Initializing those in constructing using the arguments passed in (like the url) still allows for flexible usage, but does not inconvenience folks trying to use the class from another class as many of the bots do.

Everything in here is either copied from `ReviewBot.setup_checker` or could be placed elsewhere to match style and provide better flexibility. See `leaper.py` for a decent example. This code was previous refactored to avoid the duplication present here as all the bots had it. The standard interface is to have the constructor only require the parameters defined by `ReviewBot` and optional ones can be provided for people using class directly, but cli classes just set them afterwards to avoid duplicating the construction which has all the same stuff.

oqamaint/opensuse.py
```diff
@@ -0,0 +1,39 @@
...
+
+class openSUSEUpdate(Update):
+
+    @staticmethod
+    def repo_prefix():
+        return 'http://download.opensuse.org/repositories'
+
+    @staticmethod
+    def maintenance_project():
+        return 'openSUSE:Maintenance'
```
jberry-suse: Any reason not to use static properties instead which can obviously have a getter defined if more complex logic is needed as done in `stagingapi.py`?

oqamaint/osclib
```diff
@@ -0,0 +1 @@
+../osclib
```
jberry-suse: Again, this seems less than ideal. My understanding is that since you are defining the subdir as a package you can use the following.

```python
from ..osclib import stagingapi
```

oqamaint/openqabot.py
```diff
@@ -0,0 +1,588 @@
...
+except ImportError:
+    import json
+
+QA_UNKNOWN = 0
+QA_INPROGRESS = 1
+QA_FAILED = 2
+QA_PASSED = 3
+
+Package = namedtuple('Package', ('name', 'version', 'release'))
+pkgname_re = re.compile(r'(?P<name>.+)-(?P<version>[^-]+)-(?P<release>[^-]+)\.(?P<arch>[^.]+)\.rpm')
```
jberry-suse: Nearly identical to:

https://github.com/openSUSE/osc-plugin-factory/blob/4011159df44ce05b3c74a53e29ffe007d1b27a26/osclib/core.py#L15

`osclib/core.py` is intended for general `osc/core.py` related helpers. Stuff that is generalize to that level ought to live there. Many of the tools have duplicated lots and lots of methods which is why I started collecting them there (and adding new ones). This follows the same method signature as `osc/core.py` rather than `StagingAPI` requiring a lot more to initialize.

oqamaint/openqabot.py
```diff
@@ -0,0 +1,588 @@
...
+comment_marker_re = re.compile(
+    r'<!-- openqa state=(?P<state>done|seen)(?: result=(?P<result>accepted|declined|none))?(?: revision=(?P<revision>\d+))? -->')
```
jberry-suse: As noted on @coolo's submission, I did a lot of work on comment code in `ReviewBot` and `osclib/comments.py` lets use that instead of duplicating it in openqa bot. See `leaper.py` or `repo_checker.py` for fairly extensive use of the comment stuff. Also note the `ReviewBot` also has the ability to automatically collect comment content from a logger like leaper uses, but all the replacing old comment code should remove the need for this regex and related code.

oqamaint/openqabot.py
```diff
@@ -0,0 +1,588 @@
...
+        self.project_settings = project_settings
+        self.api_map = api_map
+        super(OpenQABot, self).__init__(*args, **kwargs)
+
+        self.force = False
+        self.openqa = None
+        self.commentapi = CommentAPI(self.apiurl)
+        self.update_test_builds = {}
+        self.pending_target_repos = set()
+        self.openqa_jobs = {}
```
jberry-suse: Per comments on CLI portion this constructor should be refactored.

oqamaint/openqabot.py
```diff
@@ -0,0 +1,588 @@
...
+            self.update_test_builds[prj] = buildnr
+            jobs = self.jobs_for_target(u, build=buildnr)
+            self.openqa_jobs[prj] = jobs
+            if self.calculate_qa_status(jobs) == QA_INPROGRESS:
+                self.pending_target_repos.add(prj)
+
+    # reimplemention from baseclass
+    def check_requests(self):
+
+        if self.apiurl.endswith('.suse.de'):
```
jberry-suse: There already is:

https://github.com/openSUSE/osc-plugin-factory/blob/4011159df44ce05b3c74a53e29ffe007d1b27a26/ReviewBot.py#L72

oqamaint/openqabot.py
```diff
@@ -0,0 +1,588 @@
...
+        if not j:
+            return QA_UNKNOWN
+        if in_progress:
+            return QA_INPROGRESS
+        if has_failed:
+            return QA_FAILED
+
+        return QA_PASSED
+
+    def add_comment(self, msg, state, request_id=None, result=None):
```
jberry-suse: See `ReviewBot.write_comment()` and related note on comment regex at the top of this file. Should be able to drop all of this.

oqamaint/openqabot.py
```diff
@@ -0,0 +1,588 @@
...
+
+        except Exception:
+            import traceback
+            self.logger.error("unhandled exception in openQA Bot")
+            self.logger.error(traceback.format_exc())
+            ret = None
+
+        return ret
+
+    def find_obs_request_comment(self, request_id=None, project_name=None):
```
jberry-suse: Should also be able to drop.

oqamaint/ReviewBot.py
mimi1vx: no relative imports work only inside of the package. So not usable in this case.

oqamaint/cli.py
mimi1vx: make sense, now is only copy paste from original `openqa-maintenance.py`

oqamaint/cli.py
mimi1vx: hmm, l good idea:) only one small thing `ReviewBot.post_optparse` is buggy

oqamaint/logger.py
mimi1vx: no, colorama isn't in SLE ( and bot is running on a very old instance of SLE ) and we need only a few well-defined ANSI colors for better orientation in logging output

oqamaint/openqabot.py
mimi1vx: nearly ... not same

oqamaint/openqabot.py
mimi1vx: No, unfortunately, most important part of this regex is the **revision** on which depends the whole functionality of bot

oqamaint/openqabot.py
mimi1vx: nice , thx

oqamaint/osclib
mimi1vx: as above, relative imports works only inside package..

oqamaint/openqabot.py
mimi1vx: btw. usage of `cmdln` is the worst thing in our tooling

oqamaint/logger.py
jberry-suse: It is in SLE 12, which was the entire reason I used it over other libraries. Perhaps an old version it is not, but that seems more like a case for updating the machine than rewriting libraries.

oqamaint/ReviewBot.py
jberry-suse: Perhaps you could spend the effort to setup `osclib` as a python module and fix spec so it could be loaded properly as the current layout of repo is a bit cumbersome and most things seem to be in root directory to avoid this issue.

oqamaint/cli.py
jberry-suse: If there are issues with `ReviewBot`, lets fix them instead of copy/paste and changing the code.

oqamaint/openqabot.py
jberry-suse: That's what refactoring is about, really need to copies of 99% identical regex where the prefix is different? The prefix `(?:.*::)?` being optional I would expect this to still work. This repo already has far too much copy/paste instead of refactoring and re-use...

oqamaint/openqabot.py
jberry-suse: The code supports a dict of key values. "revision=<version>" is a key and value...how will it not work?

oqamaint/openqabot.py
jberry-suse: I agree, hence the new code I've added does not use it.

oqamaint/openqabot.py
jberry-suse: Note that `repo_checker` even generates a hash which is essentially the same as the revision used here.

https://github.com/openSUSE/osc-plugin-factory/blob/3bbeacd75f0a99b725a92c2689c91efa621657df/repo_checker.py#L74-L86

jberry-suse: All that is needed is `ReviewBot.comment_write()` to provide optional additional info and pass it along to `CommentAPI.comment_write()`.

Currently, it just builds to dict from the only two keys that bots used, but that's a restriction present in `ReviewBot` only. Considering how much duplicate code this eliminates it's worth the tiny refactoring.

https://github.com/openSUSE/osc-plugin-factory/blob/3bbeacd75f0a99b725a92c2689c91efa621657df/ReviewBot.py#L456

oqamaint/openqabot.py
jberry-suse: Even worse `osc` ships with a copy of `cmdln` nested inside itself. The `osc` tied parts would like be more cumbersome to implement without it given they are built atop `osc`, but anything outside of that surely can.

oqamaint/openqabot.py
jberry-suse: Even if `ReviewBot` was not refactored the `CommentAPI` methods should be used instead of re-inventing them as well, but really I've seen improvements from `ReviewBot` copied here and the code short of the `revision` does the exact same thing and trivial to implement. Anytime one sees that it's a case for refactoring not copying.

oqamaint/logger.py
mimi1vx: unfortuanetly bot runs on LTSS versions.. and modules arent supported for ltss

oqamaint/openqabot.py
mimi1vx: revision in this case is totallu different  thing .. and calculating has of binaries in osclib is prettty stupid .. OBS repodata.xml has revision , simply read it

oqamaint/openqabot.py
mimi1vx: old copy ( and upstream is dead 3 years )

oqamaint/openqabot.py
mimi1vx: """Write comment from log messages if not similar to previous comment.""" ... comments from logger ? really ? No way

jberry-suse: In the time you spent to write this comment you could have read the method and seen that it provides everything you need. I am not sure why you are so extremely resistant. This code-base is far from ideal, but I've refactored a lot to make this possible. The comment does not cover all the aspects of the function.

`leaper.py` was built that way so this function started with that as base functionality.

oqamaint/openqabot.py
jberry-suse: What are you talking about. I embedded the code related to hash it has nothing to do with the one OBS provides. The only feature you want is a single extra key=value pair in addition to state and result. The `CommentAPI` functions were enhanced to provide just that. The `ReviewBot` code simply needs to be changed from only providing `state` and `result` to either just an `info` dict or adding an `additional_info` parameter.

That is the only thing missing. In fact the other methods used to insert the marker and find existing comments can be used directly from `CommentAPI` just fine. The `ReviewBot` version is just the highest level wrapper that does everything and unfortunately limits this one, teany, tiny feature for no reason other than it was never needed before.

Why you are assuming so much instead of just reading when I spent a fair amount of time reviewing this to try and improve it is utterly beyond me. I made the effort to embed the relevant code bits and you clearly didn't even skim ready them (all 10 lines of them).

jberry-suse: Tada! https://github.com/openSUSE/osc-plugin-factory/pull/1154/commits/b886949e836efd45367bfa748dec4530fc90bc14 I am still sweating from the effort.

oqamaint/openqabot.py
jberry-suse: Added some long-form documentation: https://github.com/openSUSE/osc-plugin-factory/pull/1154/commits/76c4b499e317aa7f24d199e9cbf3871b4d42c12f

oqamaint/openqabot.py
mimi1vx: I'll plan to rewrite it to comment api. "revision" used in QAM requests/incidents is different than hash you use for differentiating comments. Our revision is highest repository revision from channels. If it changes we retrigger needed job. And also it can be exactly read in another internal tooling.

oqamaint/cli.py
mimi1vx: @jberry-suse how about move initialization of logger to setup_checker method? + current ReviewBot brokes without `--debug` or `--verbose` parameters:)

oqamaint/openqabot.py
mimi1vx: still "logging" purpose isn't to hold state ...

jberry-suse: a) indeed, it was shortcut make in `leaper`, not by me.
b) unrelated to the topic at hand since you can pass `message` parameter.

oqamaint/openqabot.py
jberry-suse: The revision in the comment to tell if the already posted results are for same thing as last time? It really does not matter, from comment API it is simply used to match a previous comment. So as long as you can specify `revision` key and the value as whatever you want and have it match it behaves as you need. I never implied the one in `repo-checker` was the same, merely that is is an arbitrary hash used for comment matching...

oqamaint/cli.py
jberry-suse: Sounds good to me. Not sure what all the smiles are about. I did not write the majority of this, just advocating for not making a mess.

oqamaint/ReviewBot.py
mimi1vx: Perhaps. .. osclib maybe , How about ReviewBot.py ?

jberry-suse: Probably wishful thinking since apparently it only works on modules. Not sure why there is not a better solution to this. Works fine in other languages.

.travis.yml
```diff
@@ -70,7 +71,7 @@ matrix:
         # Needs python prefix to use the correct interpretor.
         - python ./obs_clone.py --cache --debug --apiurl-target local
       script:
-        - nosetests --with-coverage --cover-package=. --cover-inclusive
+        - nosetests --with-coverage --cover-package=. --cover-inclusive --exclude-dir=./oqamaint
```
jberry-suse: What's the reason to not include the `./oqamaint` directory? Previously, only `osclib` was included and gave a rather false sense of coverage.

mimi1vx: links between packages.

jberry-suse:
I see the comment handling has not been refactored. Not sure what the plan is going forward.

mimi1vx:
@jberry-suse planned but lowest priority of my tasks  + any changes in comments generate Big bunch emails to overloaded people:(

jberry-suse:
Sure, this has been sitting unmerged for quite some time so wasn't sure what the plan is/was.

lnussel:
@mimi1vx is this now meant to be the final version? The subject still says WIP?
I assume this code is used in production already anways, right?

coolo:
Let's get this in and improve it in master

jberry-suse:
image

Hehe, ok...I'll save myself the effort of reviewing next time.

jberry-suse:
I mentioned commit typos and we couldn't even fix that. I'll ignore next time.

image

oqamaint/ReviewBot.py
coolo: I don't know what they were required for - the bot works fine without it.

PR #1149, opened 2017-09-26
adi_command: do not checking adi staging status if there is no selected_requests

nilxam:
Wihtout selected_requests then that adi staging project is an empty project, adi command should clean it up rather than go through the logic to check the project status. Since we have introduced a kind of frozenlinks adi staging, the project status might be build broken or so but it doesn't have any request
staged in that staging actually, those kind of projects must be deleted.

### DIFF
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -22,27 +22,28 @@ class AdiCommand:
     def check_adi_project(self, project):
         query_project = self.api.extract_staging_short(project)
         info = self.api.project_status(project, True)
-        if len(info['building_repositories']):
-            print query_project, Fore.MAGENTA + 'building'
-            return
-        if len(info['untracked_requests']):
-            print query_project, Fore.YELLOW + 'untracked:', ', '.join(['{}[{}]'.format(
-                Fore.CYAN + req['package'] + Fore.RESET, req['number']) for req in info['untracked_requests']])
-            return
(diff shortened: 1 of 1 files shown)


osclib/adi_command.py
```diff
@@ -22,27 +22,28 @@ def __init__(self, api):
...
-            print query_project, Fore.RED + 'broken:', ', '.join([
-                Fore.CYAN + p['package'] + Fore.RESET for p in info['broken_packages']])
-            return
-        for review in info['missing_reviews']:
-            print query_project, Fore.WHITE + 'review:', '{} for {}[{}]'.format(
-                Fore.YELLOW + review['by'] + Fore.RESET,
-                Fore.CYAN + review['package'] + Fore.RESET,
-                review['request'])
-            return
+        if len(info['selected_requests']):
```
jberry-suse: Seems like a good case for stopping instead of conditioning all the logic and is the style being used already.

```python
if not len(info['selected_requests']):
    return
```

Does the dashboard api properly handle the case when psuedometa is broken/missing and still set `selected_requests` to be an empty list? Since that seems to be your desired case to fix it'd be good to be sure, but just double checking.

jberry-suse: Spoke too soon, you want this to continue on and be cleaned up so this suggestion is not useful, but good to know the api handling.

PR #1151, opened 2017-09-27
Check user when fixing duplicated submit requests

jordimassaguerpla:
Otherwise, you get an error that you don't have permissions

fix issue#1150

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

### DIFF
--- a/osc-check_dups.py
+++ b/osc-check_dups.py
@@ -8,7 +8,7 @@
 import osc.conf
 
 def _checker_check_dups(self, project, opts):
-    user = osc.conf.get_configParser().get(opts.apiurl, "user")
+    user = conf.config['api_host_options'][apiurl]['user']
     url = makeurl(opts.apiurl, ['request'], "states=new,review&project=%s&user=%s&view=collection" % (project, user))
     f = http_GET(url)
     root = ET.parse(f).getroot()
--- a/osc-check_dups.py
+++ b/osc-check_dups.py
@@ -5,8 +5,11 @@
 # Copy this script to ~/.osc-plugins/ or /var/lib/osc-plugins .
 # Then try to run 'osc checker --help' to see the usage.
 
+import osc.conf
+
 def _checker_check_dups(self, project, opts):
-    url = makeurl(opts.apiurl, ['request'], "states=new,review&project=%s&view=collection" % project)
+    user = osc.conf.get_configParser().get(opts.apiurl, "user")
+    url = makeurl(opts.apiurl, ['request'], "states=new,review&project=%s&user=%s&view=collection" % (project, user))
     f = http_GET(url)
     root = ET.parse(f).getroot()
     rqs = {}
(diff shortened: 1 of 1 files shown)


DimStar77:
This is the code used by a TW bot to decline duplicate submissions - The bot-user does not ever create any submission - but it superseded the old one with the new one (which it should be allowed as target maintainer)

You seem to have a different usecase/expectation here

jordimassaguerpla:
@DimStar77 : We are using this script for checking duplicates in the CASP 2.0 project. However, we (containers core team) are not the only ones submitting requests, but also sle team is. This script checks both, our requests and theirs, but of course have no permission to modify theirs.

DimStar77:
as said: this change would break the function of the TW bot checking for duplicates; Is the script executed by a maintainer of the CASP2.0 project? Imho, a maintainer of $TARGET should be able to supersede different requests on the way to $TARGET

nilxam:
or add a parameter `user` for user given from command line instead of from config file, if `user` param has assigned then add `user=%s` to api call?

osc-check_dups.py
```diff
@@ -5,8 +5,11 @@
 # Copy this script to ~/.osc-plugins/ or /var/lib/osc-plugins .
 # Then try to run 'osc checker --help' to see the usage.
 
+import osc.conf
+
 def _checker_check_dups(self, project, opts):
-    url = makeurl(opts.apiurl, ['request'], "states=new,review&project=%s&view=collection" % project)
+    user = osc.conf.get_configParser().get(opts.apiurl, "user")
```
jberry-suse: Rather than reparsing the config it makes sense to just pull the already parsed value out the same way the lock code does.

https://github.com/openSUSE/osc-plugin-factory/blob/4011159df44ce05b3c74a53e29ffe007d1b27a26/osclib/obslock.py#L38

jordimassaguerpla:
what about it now?

osc-check_dups.py
```diff
@@ -38,7 +46,8 @@ def _checker_check_dups(self, project, opts):
                 print("DUPS found:", id, oldid)
             rqs[type + package] = [id, source]
 
-
+@osc.cmdln.option('--by-user',action='store_true',
+               dest='by_user')
```
jberry-suse: `dest='by_user'` is default just drop and code should still work.

jberry-suse:
Also github uses `[Ff]ixes #issue_id` and will highlight in PR description if it parses it as actionable. Not sure if it handles other forms in commit message, but both contain `issue#issue_id` which is not supported as far as I am aware. See [documentation](https://help.github.com/articles/closing-issues-using-keywords/)

lnussel:
@jordimassaguerpla ?

jberry-suse:
@jordimassaguerpla: please re-open with fixed commits. thanks.

PR #1152, opened 2017-09-27
Include CLEANUP => 1 during temp directory creation in perl scripts.

jberry-suse:
As noted by @DimStar77 the packagelists machine filled up `/tmp` and these two instances seem to at least leave directories behind and are thus worth fixing. It is also possible that the perl scripts crashing will leave behind the tmp directories.

I ran `check_source.py` locally and verified the fix works correctly before and after. Currently running `repo_checker.py`

### DIFF
--- a/check_source.pl
+++ b/check_source.pl
@@ -234,7 +234,7 @@ if (-d "$old") {
 }
 
 my $odir = getcwd;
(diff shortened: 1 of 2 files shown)


DimStar77:
👍  - LGTM

jberry-suse:
`repo-checker` succeeded.

Issue #730, opened 2017-03-11
Provide overview/log of  staging related actions


jberry-suse:
Having an idea of what actions were taken by humans vs automated tools will likely be handy in understanding what is going on. This information could be either directly stored by the staging related tools or sourced from the OBS database through `obs_factory`.

Rough example:

- bot select 96 KDE:Applications to C
- DimStar unselect kate due to upstream license issue
- bot triggered 4 rebuilds in C
- lnussel accept C

The strategy information would likely be handy to display in the dashboard as well.

PR #1159, opened 2017-10-04
Initial metrics ingestion, processing, write to influxdb, and grafana dashboards.

jberry-suse:
A lot of room for improvement and additional metrics that can be extracted. Including non-final state requests would allow for analyzing the current staging state instead of only historical state. Additionally, the current state can be used to present an activity log.

Handling incremental updates is non-trivial given the deltas are evaluated and stored in sum state. A few possible approaches, but likely not worth the hassle given the relatively short processing time and infrequent desire to update data (daily at minimum).

One of the biggest things missing is walking the history of the dashboard container to know the state of config options and psuedo-ignore list so those requests can be separated out from backlog.

For simplicity the non-final requests are not included at all which means that for `openSUSE:Factory` the totals for backlog and such will not reflect currently open request.

Additionally, I have it set to infinitely cache request results until I decide how to handle only looking for new requests. Removing the cache directory (`~/.cache/osc-plugin-factory-metrics`) or the last few pages will handle looking for new ones. Perhaps the easiest tweak would be sorting the search results on last update for requests to make it easy to append newly changed requests.

The overall structure is to loop over all request creating measurement points, many of which are deltas (like add one to this bucket, or minus one from another). After all requests are process the points are walked to evaluate the deltas and re-create the state at that time.

Due to a number of issues with the data returned from OBS the Factory graphs will show negative values at a couple locations due to completely incorrect dates. The final numbers zero out correctly since the incorrect dates are still within dataset start and end dates. One caveat is that Grafana seems to not always grab the last record and thus shows a trailing one, but the data generated correctly ends at zero. The following are the OBS related issues:

- openSUSE/open-build-service#3857
- openSUSE/open-build-service#3858
- openSUSE/open-build-service#3897
- openSUSE/open-build-service#3898

I also have an experimental version of #730 being generated by Grafana. With a few tweaks and including open requests (at least for the log) I expect this to provide a rather nice activity log.

Lastly, a test could be added in a several ways, but this it seems like overkill to add to in-depth of a test for this. The simplest approach would be a test against a finished project using live data and confirm some points of interest and total records created and such. Alternatively, a hand-crafted set of requests could be returned using the current mock setup...or a full OBS setup and release requests setup to be queried. The live approach seems better since the data should not change, provides a large and unique enough data-set to be relevant, and does not requiring adding large XML dumps to this repository.

Finally, ingesting the annotation events from an external data source would definitely be preferable if one exists. Additionally, ingesting upstream release dates of interest or annotating version changes of certain packages signifying major updates may also be interesting.

Since [open-build-service team did not indicate interest in hosting the tools](https://lists.opensuse.org/opensuse-buildservice/2017-09/msg00103.html) I filed an [IT ticket](https://infra.nue.suse.com/SelfService/Display.html?id=92842) to have a publicly facing machine so the metrics can be available to anyone. May also encourage others to expand what is extracted.


jberry-suse:
Some interesting trends were certainly notable. See a snapshot of graphs generated using everything in this PR.

https://imgur.com/a/kKxnb

jberry-suse:
If anyone would plans to review and would like more time please comment, otherwise I'll merge this tomorrow.

jberry-suse:
rebased post other merges

PR #1160, opened 2017-10-04
leaper: ignore 404

lnussel:
Some projects may not exist yet, such as SLE :Update.
Don't bail out because of that.

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -89,9 +89,15 @@ class Leaper(ReviewBot.ReviewBot):
     def get_source_packages(self, project, expand=False):
         """Return the list of packages in a project."""
         query = {'expand': 1} if expand else {}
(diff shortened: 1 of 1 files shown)


leaper.py
```diff
@@ -89,9 +89,15 @@ def prepare_review(self):
...
-        packages = [i.get('name') for i in root.findall('entry')]
+        try:
+            root = ET.parse(osc.core.http_GET(osc.core.makeurl(self.apiurl,['source', project],
+                                     query=query))).getroot()
+            packages = [i.get('name') for i in root.findall('entry')]
+        except urllib2.HTTPError, e:
+            # in case the project doesn't exist yet (like sle update)
+            if e.code == 404:
+                packages = []
+                pass
```
jberry-suse: The [pass](https://docs.python.org/2/tutorial/controlflow.html#pass-statements) keyword does nothing. I had to fix a different occurrence of this pattern in #1076. Use that style.

PR #1161, opened 2017-10-04
check_source: allow usage by SLE to primarily add repo-checker review

### LINKED ISSUE #1015: repo_checker: workflow for Leap 15 / SLE 15
Presumably we are looking to have the new `repo_checker` for 15 and not far behind SLE 15. I imagine it makes sense to resolve any issues in Leap 15 first and forward such changes to SLE 15.

The question of the workflow comes to mind. For Factory `check_source.py` is responsible for adding `factory-repo-checker` user to requests after they pass its checks. For Leap 15 we could do a few different approaches:

- add as required review to project added to all requests (presumably requests are declined by other bots anyway)
- run `check_source` bot on Leap; could either run through `leaper` or by itself. could also skip actual review bits for requests not coming from already checked sources like `SLE-workarounds`, but still add `repo-checker`
- modify `leaper` to add the review in the same manor as `check_source` (ie once it passes its checks [or at least not initially declined])

Given the new `repo_checker` will wait for requests to be staged and build successfully adding the review to project may be simplest as it is a required review just as legal review, but obviously other reviews may decline before `repo_checker` runs.

The one caveat being that `check_source` will skip the entire staging process for requests that only include modifications to `.changes` files. I would imagine that many times those requests are then forwarded on to Leap and should likely have the same skip applied. As such it seems to make sense to perhaps run `check_source` after reviewing it for any needed Leap related modifications. I would imagine `leaper` could remain a total separate review as it is more concerned with the Leap workflow the request contents.

With all that in mind, the second choice, running `check_source` with some tweaks for Leap, seems like a good path forward. Thoughts?

jberry-suse:
- a825af82a0f9f04d2a82766556ebf8c4da6cec9a:
    check_source: lookup staging group using StagingAPI rather than flag.

- c081de602564c16b96e19de459621f3808b08a07:
    check_source: interpret "None" as python None for review-team and repo-checker.
    
    Allows for disabling adding those reviews from the command line.

- 1e9c389617b0b9f9dee77ef36cb27879652fd8de:
    check_source: drop inaccurate "with > 8 diff" from --review-team option.

- 73df8871f14812dabb78a6dac675e9fc7169b70c:
    check_source: drop unnecessary dest from add_option().

Fixes #1015.

For SLE `review-team` can be disabled and devel check disabled. Leap works with just devel check disabled.

```
./check_source.py -A ibs --debug --dry --ignore-devel --review-team None id 142766
[I] checking 142766
[I] openSUSE.org:openSUSE:Factory/console-setup@9979d4d0d4103aca0a1067e221fd614e -> SUSE:SLE-15:GA/console-setup
[I] POST https://api.suse.de/request/142766?cmd=addreview&by_user=repo-checker
[I] can't change state, 142766 does not have the reviewer
[I] 142766 accepted: Check script succeeded
[D] 142766 review not changed
```

For Factory, keeps functioning as it used to.

```
./check_source.py --debug --dry id 522873
[I] checking 522873
[I] network/atftp@34 -> openSUSE:Factory/atftp
[D] skipped adding duplicate review for opensuse-review-team
[D] skipped adding duplicate review for repo-checker
[I] can't change state, 522873 does not have the reviewer
[I] 522873 accepted: Check script succeeded
[D] 522873 review not changed
```

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -30,7 +30,6 @@ class CheckSource(ReviewBot.ReviewBot):
         self.ignore_devel = False
         self.review_team = 'opensuse-review-team'
         self.repo_checker = 'repo-checker'
-        self.staging_group = 'factory-staging'
         self.skip_add_reviews = False
 
     def check_source_submission(self, source_project, source_package, source_revision, target_project, target_package):
@@ -113,14 +112,24 @@ class CheckSource(ReviewBot.ReviewBot):
 
             if self.only_changes():
                 self.logger.debug('only .changes modifications')
-                if not self.dryrun:
+                staging_group = self.staging_group(target_project)
+                if staging_group and not self.dryrun:
                     osc.core.change_review_state(self.apiurl, str(self.request.reqid), 'accepted',
-                        by_group=self.staging_group,
+                        by_group=staging_group,
                         message='skipping the staging process since only .changes modifications')
             elif self.repo_checker is not None:
                 self.add_review(self.request, by_user=self.repo_checker, msg='Please review build success')
 
         return True
+
+    def staging_group(self, project):
+        try:
+            return self.staging_api(project).cstaging_group
+        except urllib2.HTTPError, e:
+            if e.code != 404:
+                raise e
+
+        return None
 
     def is_devel_project(self, source_project, target_project):
         # Load project config and allow for remote entries.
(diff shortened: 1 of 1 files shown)


jberry-suse:
@lnussel With these changes `check_source` bot can be run against Leap/SLE 15 and double checked that new `repo_checker` still works fine against SLE 15. Someone will need to review project level issues and setup whitelist similar to Factory.

PR #1175, opened 2017-10-09
pkglistgen: initial work for openSUSE

lnussel:
- allow to specify architecures
- need to specify default support status
- use correct api (still dirty)

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -46,6 +46,7 @@ from osclib.memoize import CACHEDIR
 logger = logging.getLogger()
 
 ARCHITECTURES = ('x86_64', 'ppc64le', 's390x', 'aarch64')
+DEFAULT_REPOS = ("openSUSE:Factory/standard")
 
 class Group(object):
 
@@ -319,12 +320,9 @@ class Group(object):
 
 class PkgListGen(ToolBase.ToolBase):
 
-    def __init__(self, repostr):
+    def __init__(self):
         ToolBase.ToolBase.__init__(self)
-        self.repos = []
-        for repo in repostr.split(','):
-            project, reponame = repo.split('/')
-            self.repos.append({'project': project, 'repo': reponame})
+        self.repos = DEFAULT_REPOS
         # package -> supportatus
         self.packages = dict()
         self.default_support_status = 'l3'
@@ -454,8 +452,7 @@ class PkgListGen(ToolBase.ToolBase):
         self.lockjobs[arch] = []
         solvables = set()
         for prp in self.repos:
-            project = prp['project']
-            reponame = prp['repo']
+            project, reponame = prp.split('/')
             repo = pool.add_repo(project)
             s = os.path.join(CACHEDIR, 'repo-{}-{}-{}.solv'.format(project, reponame, arch))
             r = repo.add_solv(s)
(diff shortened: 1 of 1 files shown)


lnussel:
@coolo when merging this your scripts need to specify --default-support-status=l3

lnussel:
ah, and -A ibs

pkglistgen.py
```diff
@@ -46,23 +46,22 @@
 logger = logging.getLogger()
 
 ARCHITECTURES = ('x86_64', 'ppc64le', 's390x', 'aarch64')
```
coolo: not used anymore as default?

jberry-suse: seems like on line 539

lnussel: exactly

generate-groups.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
```
jberry-suse: Could all the scripts be placed in `scripts` directory as root is getting a tad large? I tried this concept in https://github.com/openSUSE/osc-plugin-factory/pull/1177/files#diff-b67911656ef5d18c4ae36cb6741b7965R25, preferably prefix with `osrt-` for basic namespacing.

I can merge that request today, but the `Makefile` change to install them will mean the `spec` needs to be updated to own the files. I plan to make a packagelist subpackage which should likely own them.

lnussel: script or scripts? Do you want to merge the cited request first so I can rebase on it?

jberry-suse: `script`, yeah I just did.

jberry-suse: Not sure if we want to go with multiple directories, but since I was installing them in path I went with `osrt-$tool-...`, but as long as we have them it git it's a step forward.

lnussel:
moved

lnussel:
Jimmy, maybe you can help with the install part here. generate-groups.sh should go to pkgdatadir I think. Only *-leap15.sh are meant to be called. Needs some more shell script hackery to actually work both from git as well as installed. Or we just install all script to pkgdatadir.

jberry-suse:
@lnussel My general approach is to install anything to be executed from service in `/usr/bin` with the remaining code like `osclib` all loaded out of main directory. Preferably the tools would all load things from proper python modules that were installed as such, but not there yet.

Additionally, I've been trying to stay away from making things `openSUSE` specific so I use full project names to allow for `SUSE:...`. Most of the services call code directly, but the `update_crawler` has wrapper scripts like this service. You can see I named them with full project suffix: `script/osrt-leaper-crawler-openSUSE:Leap:15.0`. Everything in that directory is then installed in `/usr/bin` and the tools it calls `osrt-update_crawler` (link to `update_crawler.py`) are as well.

Currently, the old-style service file needlessly includes the environment file used by announcer since the only variable it uses is project which is a parameter to the templated service file. It also runs as `factory-repo-checker`, but given this service is rather separate from the others should probably have its own user and thus subpackage that creates the user. Given I packaged up most of the other services there should be plenty of examples to go off.

I would suggest only placing the wrapper scripts to be called directly in `script/` dir and name them based to the service they are wrapping (in this case `pkglistgen.py`. Perhaps `osrt-pkglistgen-openSUSE:Leap:15.0`. Could also drop `osrt-` prefix and do a transformation in `makefile`, but seems better to just include.

If you add `pklistgen` to `pkgdata_BINS` in `Makefile` it will be symlinked into `/usr/bin` and thus can be easily called from wrapper scripts. As for development short of overriding links in bin or `~/bin` not sure what other sort of generalized method can be employed.

Should be able to copy `repo-checker` subpackage and style used in service files to provide a subpackage for all this. If you're not up for all that I suppose you could move the files back and I can package them and move them, but would be nice if we started creating this this way.

jberry-suse:
To clarify, I have been grouping services by subpackage and the user to which they related since they all work together to provide a larger "service". Like the various `staging-bot` services or parts to making the `leaper` workflow work (like `manager_42` and `leaper` review).

jberry-suse:
You can see the plethora of services in [`systemd`](https://github.com/openSUSE/osc-plugin-factory/tree/master/systemd) that I've adapted from `packagelists`.

Issue #1006, opened 2017-07-14
Create standardized service files for use by IBS and OBS


jberry-suse:
Given there are two variables for most services a single template variable for a service file is likely not enough. As such pointing to an environment variable file/config file might make more sense. It would be nice if the user could be a parameter based on that file, but guessing that it would not be possible to load environment file in that manor. Alternatively, a wrapper script could be provided to start the desired service using the config file as the appropriate user.

As a different approach the users themselves could also be standardized and created by the package so then they simply need to have their credentials setup for the appropriate OBS instance and user. I would imagine the bots/scripts that run as different users on OBS should also be run separately on IBS so this may simplify things.

jberry-suse:
Performed an inventory of the services on `packagelists` machine.

## services

- announcer
- build-fail-reminder
- create-test-dvds
- factory-auto-submit
- facotry-dvl-prj-list
- factoryauto [gnome-review-bot] -> check_source
- leaperbot
- maintbot
- manager_42
- openqa-comments
- osc-check_dups [likely merged into repo_checker]
- repo-checker
- source-checker -> check_source_in_factory
- staging-bot-daily
- staging-bot-regular
- staging-bot-reminder
- suppkg_rebuild
- trigger-rebuilds
- tmm-arm, tmm-leap, tmm-s390, tmm-tumbleweed -> totest-manager
- update-crawler (for leap)

A slight lack of consistency in naming and the template argument being partial or full project name or built into service file. I would suggest the full project name be used for all templated services to allow for easy usage by IBS for `SUSE:` projects as well.

As for naming and additional configuration, it would seem to make sense to prefix services with an abbreviation of `openSUSE-release-tools` so `osrt-` as the prefix, otherwise the full name. A common config location like `/etc/openSUSE-release-tools` with sub-directories for services requiring them and the project name can be included for templated service files requiring different config per project. This will merge all the random config paths into one location with a standard path setup.

I may also rename some of the services that are a bit misleading or obfuscated to make their role more clear. As an example, I plan to do the official `repo_checker` deployment via the package from OBS and will submit the PRs for the service files before then. Something along the lines of `openSUSE-release-tools-repo-checker@openSUSE:Factory` or prefixed if preferred as `osrt-repo-checker@openSUSE:Factory`.

## users

```
$ grep -hoPR 'User=\K.*' systemd/packagelists/ | sort | uniq
```

- factorymaintainer -> factory-maintainer
- factory-source
[...]

PR #1176, opened 2017-10-09
dist: provide staging-bot subpackage for related services.

jberry-suse:
Another towards #1006.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ SUBDIRS = factory-package-news abichecker
 
 include Makefile.common
 
-pkgdata_BINS=repo_checker
+pkgdata_BINS=devel-project repo_checker
 pkgdata_SCRIPTS=$(wildcard *.py *.pl *.sh)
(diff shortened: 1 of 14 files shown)


jberry-suse:
Added `dist: provide staging-bot-devel-list service (changed from leaper).` which is a change from running as `leaper` on `packagelists`, but makes sense since other `devel-project` services are run as `staging-bot`.

jberry-suse:
`dist: include suppkg_rebuild in staging-bot subpackage.` not repo-checker related and `staging-bot` already handles other types of rebuilds.

jberry-suse:
`dist: include supersede service in staging-bot subpackage.` Again not `repo-checker` related and given that `staging-bot` is doing everything staging related and reviewbots have separate users this seems to make the most sense. Also called the service `supersede` since that seem more clear given we have things like check for duplicate binary as a service too. I imagine most of this is duplicated by the revamped supersede logic in `StagingAPI`, but primarily focused on getting everything on `packagelists` into git.

PR #1177, opened 2017-10-10
dist: provide leaper services and related configuration.

jberry-suse:
Another towards #1006.

I do not seem to have access to `maintenance` machine to grab SLE configs, but those should be added. The `scripts` were adapted from `packagelists` in order to provide a complete setup, but seem like they should really be part of the same sort of config used by `manager_42`.


jberry-suse:
rebased

PR #1189, opened 2017-10-11
osclib: provide ConfigCommand and expose via osc-staging (utilize in repo_checker)

### LINKED ISSUE #1122: osclib/conf: provide mechanism for staging specific setting overrides
Per [mailing list discussion about how to handle ring packages whose install-time requirements are not in rings](https://lists.opensuse.org/opensuse-releaseteam/2017-08/msg00013.html) @DimStar77 requested to leave things as is and provide a way to keep a staging specific whitelist that is cleared on accept. This removes the need to whitelist problems, caused by stagings that have not been rebased, globally and instead can be restricted to individual stagings.

This feels like it could be done by extending remote config to provide staging specific overrides, but may end up being `repo-checker` specific. The ring command (#848) will work in similar manor, storing staging specific information that is cleared on accept, but the intention was to use psuedometa data. The two concepts are not that different, but if remote config either the entire global whitelist from the target project will need to be included along with the staging specific entries or a method for appending to the global whitelist will be required.

It also may make sense to introduce a command line tool for setting the staging specific config to provide a simple workflow.

minimal requirements:

- staging specific `repo-checker-binary-whitelist*` settings
- clear staging specific values on accept of that staging

jberry-suse:
- 1e39042569e5b05cd64a4dc3419920453202175c:                                                                                                
    repo_checker: filter empty whitelist entries.

- f0ff9cb542ded92de4279cffa2b31e8b063df235:
    repo_checker: utilize staging configuration for whitelist additions.

- d2b50c0a1bb8e5edfd764f283830fdb0b407d459:
    osclib: provide ConfigCommand and expose via osc-staging.

- 336a0e54addf4f8122e1e338c4134acf3c47c63b:
    osclib/stagingapi: get_staging_projects(): provide include_dvd option.

I played around with various approaches for modifying `osclib/conf.py` to be able to apply staging config, but given that for the `repo_checker` use-case the values are to be appended and I cannot foresee trivial uses of this elsewhere I left it up to utilizing code to use the values. Currently, the config can only be loaded with project level overrides since a project is always being operated on, but stagings are mixed in many commands and there does not appear to be a clear way to implement automatically nor a real reason for it. Once we come across a second desirable case hopefully this will be more obvious. The the usage in `repo_checker` is quite trivial.

Fixes #1122.

### DIFF
--- a/openqa-comments.py
+++ b/openqa-comments.py
@@ -231,6 +231,5 @@ if __name__ == '__main__':
     if args.staging:
         openQA.report(api.prj_from_letter(args.staging))
     else:
-        for staging in api.get_staging_projects():
-            if not staging.endswith(':DVD'):
-                openQA.report(staging)
+        for staging in api.get_staging_projects(include_dvd=False):
+            openQA.report(staging)
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -454,7 +454,7 @@ def do_staging(self, subcmd, opts, *args):
         elif cmd == 'select':
             # Include list of all stagings in short-hand and by full name.
             existing_stagings = api.get_staging_projects_short(None)
-            existing_stagings += [p for p in api.get_staging_projects() if not p.endswith(':DVD')]
+            existing_stagings += api.get_staging_projects(include_dvd=False)
             stagings = []
             requests = []
             for arg in args[1:]:
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -38,6 +38,7 @@ from osclib.check_command import CheckCommand
 from osclib.check_duplicate_binaries_command import CheckDuplicateBinariesCommand
 from osclib.cleanup_rings import CleanupRings
 from osclib.conf import Config
+from osclib.config_command import ConfigCommand
 from osclib.freeze_command import FreezeCommand
 from osclib.ignore_command import IgnoreCommand
 from osclib.unignore_command import UnignoreCommand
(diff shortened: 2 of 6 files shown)


jberry-suse:
@DimStar77 this provides the workaround you were looking for with a fairly nice interface to see all such additions. I would expect your normal usage to be:

```
osc staging config --append C D E adi:17 repo_checker-binary-whitelist leaky-dep-to-avoid-rebase
```

jberry-suse:
An overview of all such changes being easy to view using:

```
osc staging config
```

DimStar77:
@jberry-suse That command line interface looks ok; I can certainly live with that.

and as it's a part of the config, it should be rather simple to extend 'osc staging freeze and osc staging adi to reset their special config (freeze usually means the leaks are gone, and adi, well, would reset when a prj is deleted (ready to accept)

jberry-suse:
Was there something you had in mind? This provides every sort of read/write/clear/append combination that would seem useful. Far more than minimal read/write or go manually editor psudometa.

jberry-suse:
The problem can still be resolved by making the install requirements policy consistent with build requirements policy.

jberry-suse:
Added the freeze bit to clear staging config.

DimStar77:
👍 

The most common issue seen lately are i586 issues for ring2 - and I'm still not willing to add i586 to ring2 (:DVD) builds - unless somebody sponsors another 100 dedicated workers for Stagings

But with the interface to add ignores for a staging that are wiped on rebase, we at least get accurate info of what is there for which stagings.

nilxam:
LGTM

PR #1190, opened 2017-10-12
check_source: move all relevant flags to config to work properly without flags.

jberry-suse:
- c05a26d154818b2725a1dc1e1d07d89df478a2a0:
    check_source: move all relevant flags to config to work properly without flags.

- 52af5b3d75a1840132b65c616e473def136f5840:
    check_source: print message if checking devel project.
    
    Currently, no indication unless ends in decline. More important when moving
    option to config rather than flag.

This approach seems preferable to flags, just as I already moved devel whitelist to be able to load from remote config. It means users do not need to know the right flag combination to run the tool, rather it just figures it out. This also means we can run same service with same options and same user on OBS which again seems preferable.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -36,7 +36,7 @@ class CheckSource(ReviewBot.ReviewBot):
         super(CheckSource, self).check_source_submission(source_project, source_package, source_revision, target_project, target_package)
 
         if not self.ignore_devel:
-            # Check if target package exists and has devel project.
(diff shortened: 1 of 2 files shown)


nilxam:
:+1:

DimStar77:
Hmm.. this broke the GNOME Review bot, which was started using:

```/usr/local/bin/check_source --ignore-devel --skip-add-reviews review```

whereas ```--ignore-devel``` is no longer valid

PR #1194, opened 2017-10-12
check_source: decline the delete request if duplicated

nilxam:
Decline the delete request in case there is pending delete/submit request to the same package, if there is another submit request against the same package that we should prefer the submit one which should be passed staging already, what it means just queued for accepting, like request go through adi or virtual accept delete request.

Fixes https://github.com/openSUSE/osc-plugin-factory/issues/1193

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -223,6 +223,16 @@ class CheckSource(ReviewBot.ReviewBot):
         except urllib2.HTTPError:
             return None
 
+        # Decline the delete request if there is another delete/submit request against the same package
+        query = "match=state/@name='new'+and+(action/target/@project='{}'+and+action/target/@package='{}')"\
+                "+and+(action/@type='delete'+or+action/@type='submit')".format(action.tgt_project, action.tgt_package)
+        url = osc.core.makeurl(self.apiurl, ['search', 'request'], query)
+        matches = ET.parse(osc.core.http_GET(url)).getroot()
+        if int(matches.attrib['matches']) > 1:
+            ids = [rq.attrib['id'] for rq in matches.findall('request')]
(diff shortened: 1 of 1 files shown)


jberry-suse:
The supersede logic in `StagingAPI` already does this for staged requests where a second request either delete or submit for same package comes in. Should this be made specific to virtual deletes since that seems to be the issue? The other code handle multi-origin checks and such as well.

For example [sr#533384](https://build.opensuse.org/request/show/533384) was declined by this code and should handle the existing submit. If the submit is no longer staged (like adi in "ready to accept") then it will allow a second submit or delete.

The overall point being that we expect to use this bot for Leap as well and will decline requests with a much less complex comparison. Perhaps simply restrict to state "new".

nilxam:
Ah, you're right. I forgot that `stagingapi` logic would handle those `review` cases, yes, this should restrict to `new` state requests only.

check_source.py
```diff
@@ -223,6 +223,15 @@ def check_action_delete(self, request, action):
         except urllib2.HTTPError:
             return None
 
+        # Decline the delete request if there is another delete/submit request against the same package
+        query = "match=state/@name='new'+and+(action/target/@project='{}'+and+action/target/@package='{}')"\
+                "+and+(action/@type='delete'+or+action/@type='submit')".format(action.tgt_project, action.tgt_package)
+        url = osc.core.makeurl(self.apiurl, ['search', 'request'], query)
+        matches = ET.parse(osc.core.http_GET(url)).getroot()
+        if int(matches.attrib['matches']) > 1:
+            self.review_messages['declined'] = "There is pending request to %s/%s in process." % (action.tgt_project, action.tgt_package)
```
jberry-suse: `There is [a] pending...`

I would include the request ID in the message so people can find it easily.

PR #1195, opened 2017-10-12
request_splitter: implement autoselect: False

lnussel:
allow to exempt stagings from automatic selects bases on strategies
by putting autoselect: False in pseudometa

### DIFF
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -211,6 +211,7 @@ class RequestSplitter(object):
 
     def is_staging_considerable(self, project, pseudometa):
         return (len(pseudometa['requests']) == 0 and
+                pseudometa.get('autoselect', True) and
                 self.api.prj_frozen_enough(project))
 
     def stagings_load(self, stagings):


nilxam:
you can configure it by `splitter-whitelist`  in dashboard/config, or it does not work?

jberry-suse:
Unless intended to be a temporary marking the psuedometa is wiped everytime a staging is emptied or accepted. The current approach it to remove `splitter_info` if no longer want followups. I that's the desire to disable follow-ups without removing strategy info could add a `merge: False` option as that would be more applicable.

Otherwise, the other code that clears psuedometa would need to be changed to make an exception for this. If permanent I would suggest the config like Factory uses.

lnussel:
I was just wondering how to deal with Staging:X in Leap. It's not meant to get any requests.

DimStar77:
For TW we have this in 'config':

> splitter-whitelist = B C D E F G H I J

We want to have manual control over A (at least one ring0 for us) and the project K, L and M (and the two new N and O are also not to be touched by the bot)

lnussel:
ok, I'll try that one then. thanks!

PR #1196, opened 2017-10-13
pkglistgen: shell scripts to actually commit

lnussel:


lnussel:
I'm not sure which user to use yet, maybe we need a dedicated one, maybe not.

lnussel:
hmm, why does github not update the pull request when I force push?

lnussel:
@dirkmueller this is the change you need for ports. there's also some docu here

PR #1184, opened 2017-10-10
ttm: update for Leap 15.0

lnussel:

PR #1198, opened 2017-10-16
Fix undefined variables 'release'

dirkmueller:
The totest-manager crashed on Factory:ARM when it tried to release
a build.


jberry-suse:
From #1184.

PR #1201, opened 2017-10-16
Flake8 / pep8 formatting fixes

dirkmueller:


dirkmueller:
@jberry-suse see #1202

jberry-suse:
@dirkmueller Does this PR need to be updated to enable checks for issues that is fixes?

jberry-suse:
Also is the max line length, presumably what these are to fix, configurable? I find 80 to just be small, 100 is much more reasonable.

ReviewBot.py
```diff
@@ -209,11 +209,11 @@ def add_review(self, req, by_group=None, by_user=None, by_project=None, by_packa
...
-                r.by_project == by_project and
-                r.by_package == by_package and
-                r.by_user == by_user and
-                # Only duplicate when allow_duplicate and state != new.
-                (not allow_duplicate or r.state == 'new')):
+                    r.by_project == by_project and
+                    r.by_package == by_package and
+                    r.by_user == by_user and
+                    # Only duplicate when allow_duplicate and state != new.
+                    (not allow_duplicate or r.state == 'new')):
```
jberry-suse: Hmm, this class of indent is useful since everything lines up vertically to scan rather than being offset from the first entry.

metrics.py
```diff
@@ -20,7 +20,7 @@
 
 # Duplicate Leap config to handle 13.2 without issue.
 osclib.conf.DEFAULT[r'openSUSE:(?P<project>[\d.]+)'] = \
-osclib.conf.DEFAULT[r'openSUSE:(?P<project>Leap:[\d.]+)']
+    osclib.conf.DEFAULT[r'openSUSE:(?P<project>Leap:[\d.]+)']
```
jberry-suse: Same idea here and below.

jberry-suse:
I suppose fighting the tool, hmm.

metrics.py
```diff
@@ -19,8 +19,9 @@
 from osclib.stagingapi import StagingAPI
 
 # Duplicate Leap config to handle 13.2 without issue.
-osclib.conf.DEFAULT[r'openSUSE:(?P<project>[\d.]+)'] = \
-osclib.conf.DEFAULT[r'openSUSE:(?P<project>Leap:[\d.]+)']
+osclib.conf.DEFAULT[
+    r'openSUSE:(?P<project>[\d.]+)'] = osclib.conf.DEFAULT[
+    r'openSUSE:(?P<project>Leap:[\d.]+)']
```
jberry-suse: That's certainly interesting. :)

PR #1202, opened 2017-10-17
Enable travis / flake8 checks

dirkmueller:
This at least should be able to find syntax errors and the like before it gets merged. most of the scripts and almost all checks are blacklisted right now to get a green result without too much code churn. this can be cleaned up in followups.


dirkmueller:
#1198 should be fixing the travis failure (to be rechecked and turning green once the other one merged)

.travis.yml
```diff
@@ -40,3 +46,7 @@ deploy:
   on:
     branch: master
     condition: $TEST_SUITE = distribution
+
+notifications:
+  on_success: change
+  on_failure: always
```
jberry-suse: Are these not the defaults?

flake8.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
jberry-suse: This file should be placed in `dist/ci/`

flake8.sh
```diff
@@ -0,0 +1,6 @@
...
+# There are 3 levels: F (FATAL), E (ERRORS) and W(WARN)
+# Everything is blacklisted here only due to excessive hits in existing code
+# base. The ignores should be removed in the severity level (e.g. F first, then E, then W)
+find -type f -name "*.py" -print | grep -v abichecker | grep -v openqa | \
```
jberry-suse: Ideally, this file should allow a git reference to be passed (which travis can provide) and use the git diff to only check the files that were changed. I suppose if the check runs fast and we end up cleaning everything this approach also works, but seems like it avoid you having to fix all the issue. If the pip8 has the ability to run on diffs then we can fix things whenever the line is changed, which would be the real preference.

jberry-suse: Ideally, this file should allow a git reference to be passed (which travis can provide) and use the git diff to only check the files that were changed. I suppose if the check runs fast and we end up cleaning everything this approach also works, but seems like it avoid you having to fix all the issue. **If the pip8 has the ability to run on diffs then we can fix things whenever the line is changed, which would be the real preference.**

flake8.sh
jberry-suse: This file should be placed in `dist/ci/`

.travis.yml
jberry-suse: Are these not the defaults?

osc-cycle.py
```diff
@@ -15,6 +15,9 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
 import osc.core
+from xml.etree import cElementTree as ET
+
+from osc.core import get_dependson
```
jberry-suse: Should this not be after line 17?

.travis.yml
dirkmueller: Could be. Is it had to make them explicit?

flake8.sh
dirkmueller: I see where this is going, but I would prefer to run it locally on a list of files. As a background, i have that integrated in my editor and it is difficult to even save a file of it doesn't pass.

None of those checks are difficult or time-consuming to fix and it just takes about a second to run over all of what is in git.

dirkmueller: In one of the projects that I am involved in we're doing that increments code style fixing and it causes every pull request to become bloated with the change plus any style fixes necessary to pass testing and it makes every review harder because of both of it mangled in one place, *and* it is frustrating for the submitter because of the need to care about a bunch of unrelated stuff.

jberry-suse: The script could easily work in both forms, but we can keep it this way. Still not sure I like the level of reformatting in the original PR.

.travis.yml
jberry-suse: No, I just usually leave defaults alone as it keeps things smaller and easier to focus on what is actually being done.

.travis.yml
```diff
@@ -8,6 +8,12 @@ env:
 
 matrix:
   include:
+    - env: TEST_SUITE=flake8
```
jberry-suse: super nit: perhaps keep with sorting based on `TEST_SUITE`.

Issue #1204, opened 2017-10-18
factory-auto is dead


DimStar77:
```
Oct 18 11:29:11 packagelists check_source[29648]: [I] checking 534622
Oct 18 11:29:12 packagelists check_source[29648]: [E] 'CheckSource' object has no attribute 'repo_checker'
Oct 18 11:29:12 packagelists check_source[29648]: Traceback (most recent call last):
Oct 18 11:29:12 packagelists check_source[29648]:   File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 646, in runner
Oct 18 11:29:12 packagelists check_source[29648]:     workfunc()
Oct 18 11:29:12 packagelists check_source[29648]:   File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 614, in work
Oct 18 11:29:12 packagelists check_source[29648]:     self.checker.check_requests()
Oct 18 11:29:12 packagelists check_source[29648]:   File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 144, in check_requests
Oct 18 11:29:12 packagelists check_source[29648]:     good = self.check_one_request(req)
Oct 18 11:29:12 packagelists check_source[29648]:   File "/home/repochecker/src/osc-plugin-factory/ReviewBot.py", line 257, in check_one_request
Oct 18 11:29:12 packagelists check_source[29648]:     ret = func(req, a)
Oct 18 11:29:12 packagelists check_source[29648]:   File "/usr/local/bin/check_source", line 248, in check_action_delete
Oct 18 11:29:12 packagelists check_source[29648]:     if not self.skip_add_reviews and self.repo_checker is not None:
Oct 18 11:29:12 packagelists check_source[29648]: AttributeError: 'CheckSource' object has no attribute 'repo_checker'
```

DimStar77:
Fallout of https://github.com/openSUSE/osc-plugin-factory/pull/1190

nilxam:
@jberry-suse 

I've temporary added a workaround, http://susepaste.org/37899259

those config does not load in check_action_delete() but check_source_submission() only therefore the bot crashes in check_action_delete(), maybe you have a better solution here, so I leave it to you :)

PR #1207, opened 2017-10-19
Provide local OBS setup, base data creation, test harness, and initial set of tests.

### LINKED ISSUE #1002: Bring up real OBS against which tests may be run
In order to avoid the over complexity of writing mocks for everything when adding tests it would seem to make far more sense to just run the tests against the real thing especially given the level of complexity.

Several approaches:

- OBS vagrant (won't work on travis-ci since vm inside vm not allowed)
- docker image: last checked no official ones
- copy OBS travis-ci setup

In order to use the last option locally either run tests against the standard vagrant development setup or build a docker based local equivalent. If a docker based version is created it might be just a effective to use on travis-ci.

Just to see if it is possible my first approach would be to copy all relevant travis-ci setup from OBS and try to run an osc command against the instance in a travis-ci worker. Assuming that works will need to look at creating a separate test harness.

Similar to the tests introduced in #998 I would expect these to run in separate job from existing tests. Potentially having the existing tests phased out.

jberry-suse:
- bda410b5aa0c4fd1c99726d4ea60465c5e9c7459:
    travis: provide local OBS setup and clone base data.

- 40f8ed27288f3ca877ad2709dc27269a080bde81:
    tests/devel_project: provide basic test against local OBS.

- d3cf5fddf13faf45756411cd79d39f03d1bfcc6b:
    tests/check_source: provide basic test against local OBS.

- 98963a8ebd9c25ac33355f9d47a50973b76d6d21:
    tests/OBSLocal: provide based class for testing against local OBS.

- d87cd051a3be95fa687958aca68e1e4b2e2ebd06:
    obs_clone: clone projects and dependencies between OBS instances.

- 438350ba4bd3f60ed0001396568b1a1a9b983b59:
    devel-project: list: only remove self if in devel project list.

Summary of what this accomplishes:

- bring up local OBS from source using new `docker-compose` setup from `OBS`
- clone all relevant projects, groups, users, and packages to provide a proper facsimile of `openSUSE:Factory` for realistic testing
  - the clone is fairly fancy since it recursively traverses entities so one can say clone `openSUSE:Factory:Staging/dashboard` and you get just about everything you need minus `opensuse-review-team` group
- provide base test case:
  - handles all `osc` wrangling and allows access to run as any cloned user
  - useful asserts for `ReviewBot` style tests
  - error handling and plentiful output for debugging
- initial stab at travis integration (needs resolving)

To test locally, either bring up your own OBS instance, or use `./dist/ci/docker-compose-obs` to do so. Then run tests. This bit does not work on travis due to pip not installing `osc` binary in path and no throw-away account. I am considering moving the script into the python base-class of sorts which could also be a solution to `osc` in path.

I am very pleased with this setup as it is MUCH MUCH simpler to get enough data to write a real test and does not require rewriting OBS in mocks to do so. I included a test that would have caught #1206. Also should be easy to add one that would have caught #1204.

I did not setup `obs_factory` as Ruby was giving me some cryptic gem errors, but in theory it should not be very hard to add it.

I would expect that we can add basic tests for all ReviewBots on top of this rather rapidly and once `obs_factory` is present even more.

Suffers from same throw-away OBS account issue as #1126. I created an account for this since OBS team not interested in a proper solution, but login fails even after password recovery.

Fixes #1002.

The side effect is that if one adds the following to `~/.oscrc`

```
[http://0.0.0.0:3000]
user=Admin
pass=opensuse
aliases=local
```

That one can run queries against local instance and thus develop, change data, and inspect without doing so on production. Gives a lot more insight into cryptic OBS crashes and such since one can see all the output. Hopefully, others are excited by this and help fill out the tests once this is merged.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.open-build-service/
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
-exclude = abichecker, openqa, openqa-maintenance.py
+exclude = .open-build-service/, abichecker, openqa, openqa-maintenance.py
 max-line-length = 100
 ignore = E501,F401,E302,E228,E128,E251,E201,E202,E203,E305,F841,E265,E261,E266,E712,E401,E126,E502,E222,E241,E711,E226,E125,E123,W293,W391,E731,E101,E227,E713,E225,E124,E221,E127,E701,E714,W503,E129,E303,E741,E722
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .coverage
 .coveralls.yml
 .docker-tmp/
+.open-build-service/
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,16 +49,26 @@ matrix:
         - flake8
     - env: TEST_SUITE=nosetests
       sudo: required
+      services:
+        - docker
       language: python
       python: 2.7
       before_install:
         # provides xmllint used by test_bootstrap_copy (tests.freeze_tests.TestFreeze)
         - sudo apt-get install libxml2-utils
       install:
-        # needed to install osc from git in requirements.txt
-        - pip install pycurl urlgrabber
+        # urlgrabber needed to install osc from git in requirements.txt
+        # m2crypto for osc to be runable as used in docker-compose-obs
+        - pip install pycurl urlgrabber m2crypto
         - pip install -r requirements.txt
         - pip install python-coveralls
+      before_script:
+        # travis-ci/travis-ci#7008: stop services to make room for OBS setup
+        - sudo service mysql stop
+        - sudo service memcached stop
+        - ./dist/ci/docker-compose-obs
+        # Needs python prefix to use the correct interpretor.
+        - python ./obs_clone.py --debug --apiurl-target local
       script:
         - nosetests --with-coverage --cover-package=. --cover-inclusive
       after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,4 +92,4 @@ cache:
   timeout: 604800
 before_cache:
   - sudo rm -rf .docker-tmp/build-root
-  - sudo chmod o+rw .docker-tmp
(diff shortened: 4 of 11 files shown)


jberry-suse:
Just checkout the two new test files to see how trivial it is to write some real tests.

Much improvement can be had by just invoking all bots to ensure they do no crash.

DimStar77:
you are crazy! And I mean this in a positive way :)

jberry-suse:
Rebased on #1126 which includes throwaway account. Detected the `pip` version of `osc` and create alias.

Remaining issue is OBS `rake docker:build` creating invalid yaml file on travis for some reason. Additionally, I should update the `check_source` test include a change before the sr creation.

jberry-suse:
Due to ubuntu's `echo`... :(

jberry-suse:
After quite a battle...

Rebased and squashed my WIP commits, then made more. :) I got everything running on travis!!!! I had to disable my `check_source` tests since they use `osc vc` which requires `build` package which is cumbersome to install on the ubuntu travis host. After consideration, `repo-checker.pl` and similar will have other deps that need to be installed and we really don't care about `openSUSE` so forcing those packages to be maintained for Ubuntu (even if they already exist) seems like a bad idea.

Also have to workaround several travis/ubuntu specific issues that would be resolved by running on `openSUSE`. As such my thinking is as follows:

- merge test running into distribution jobs (after)
  - install built rpms (gives us all required deps to run)
  - run tests
  - run additional tests to see that installed locations and setup works proper

This removes the need to deal with the ubuntu base image entirely and allows us to test the installed packaged files.

Alternatively, add `openSUSE:Tools` ubuntu repo on travis and install build package and cross-fingers about new deps in the future and leave workarounds. This also does not allow us to test installed files.

jberry-suse:
Dropped a couple hundred lines off, and `check_source` test pending #1221. Cleanup up changes and extracted local docker fixes to another branch for submission later.

jberry-suse:
`Coverage increased (+0.9%) to 31.208%`!! Half of `devel-project.py`.

PR #1213, opened 2017-10-20
travis: include all files instead of just osclib in coverage.

jberry-suse:
I always suspected based on percentage and change, but when working on #1207 I dug in when trying to get `check_source.py` to show up in report as covered. It's a bit interesting that no `--cover-package` argument will no result in what you would expect as documentation indicates current directory which one would assume is the same as `.`, but apparently not in nose land.

Makes sense to correct this before I finish up the fancy test PR so we can see the impact.

Summary: PITA + overly optimistic coverage reports.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ matrix:
         - pip install -r requirements.txt
         - pip install python-coveralls
       script:
-        - nosetests --with-coverage --cover-package=osclib --cover-inclusive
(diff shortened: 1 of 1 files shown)


jberry-suse:
```
       Stmts Miss Cover
before  3550 1874 47%
after  10916 7604 30%
```

jberry-suse:
Now things like the removal of obsolete tools from root directory will actually have an effect.

Issue #1211, opened 2017-10-20
Setup docker hub build of generalized `osc build` image


jberry-suse:
Now that I generalized the spec build process (#1126) and worked around the issues with using `osc build` the package dependencies are no longer required to be in the docker image, but are instead downloaded via `osc` during the build. This has a number of advantages one of which is the docker image is essentially static and only changes when `osc` dependencies change.

We can see how this works out for a few weeks, but likely this should be extracted as a stand-alone project for providing OBS pre-builds for use in CI in addition to OBS sync.

jberry-suse:
Anyone know if possible to use some common `openSUSE` account or if I should create a new on like `openbuildservice` account used for OBS docker images? I have personal one for a variety of images, but that seems less than ideal.

coolo:
We moved to obs built containers

PR #1215, opened 2017-10-21
Travis improve caching by installing realpath and enabling pip cache

jberry-suse:


jberry-suse:
d.o.o appears to be dead so I'll hold off.

jberry-suse:
Net saves ~3-5 minutes and #1211 should save another 5 minutes bringing a full build down to under a minute.

PR #1227, opened 2017-10-27
pkglistgen: don't export skip_releases

lnussel:
would spoil next iteration in loop


pkglistgen.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+# FIXME: 000package-groups is not frozen, osc up doesn't do anything
+# when updated in underlying project
```
coolo: works for me

PR #1229, opened 2017-10-27
repo_checker: store and compare build hash to skip more aggressively (similar check for project_only)

### LINKED ISSUE #973: repo_checker: speed up reviews by comparing has of staging project psuedometa
Building atop the new `repo_checker` provided in #964 it makes sense to keep a cache of hashes of each staging project pseudometa + `overall_state` from dashboard API and if nothing changed since last review do not bother performing an installcheck or cycle check again.

Since the `repo_checker` runs in a loop it will see non-acceptable, but not declined reviews over and over until the problem is resolved. This was the same with the previous code and even worse since it would look at non-staged requests which means the window for looping was larger.

### LINKED ISSUE #1029: repo_checker: project_only check for full build completed

jberry-suse:
- 95777e194f420d1a829af75ff11a3b3d0ed160cc:
    repo_checker: project_only place dashboard update behind dryrun check.

- b8d81e15f6540d63217ff8b8a8dccfb88b8daa4c:
    repo_checker: project_only ensure project repository is published.

- 9b7d12142c198a1075b6d3b858530eb9e6b6c085:
    repo_checker: store and compare build hash to skip more aggressively.

Fixes #973.
Fixes #1029.

### DIFF
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -69,6 +69,7 @@ class Cache(object):
     TTL_SHORT = 5 * 60
     TTL_DUPLICATE = 3
     PATTERNS = {
+        '/build/[^/]+/_result': TTL_DUPLICATE,
         # Group members cannot be guaranteed, but change rarely.
         '/group/[^/?]+$': TTL_SHORT,
         # Clear target project cache upon request acceptance.
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -3,13 +3,16 @@
 import cmdln
 from collections import namedtuple
 import hashlib
+from lxml import etree as ET
 import os
+from osc.core import show_results_meta
 import pipes
 import re
 import subprocess
 import sys
 import tempfile
 
+from osclib.comments import CommentAPI
 from osclib.core import binary_list
 from osclib.core import depends_on
 from osclib.core import package_binary_list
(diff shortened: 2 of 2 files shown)


jberry-suse:
`packagelists` is rather under-powered for this bot and the addition of `Leap:15.0` does not help. Effectively should skip re-reviewing stagings until their build changes. This should also allow the `project_only` mode to run without producing false-negatives during rebuilds. As such I will go ahead and enable it.

jberry-suse:
The down-side of this is that if the whitelist is updated repo-checker will not recheck the staging.

DimStar77:
For adi, there might also be changes in oS:F directly that alter the outcome of the check (think install deps were missing, then accepted through a different adi staging). The fact that setting confug won't be cinsidered for change/rescan is kinda a bummer

jberry-suse:
This caches all relevant repos so it would detect that change. Basically all the repos it downloads to check it includes in the hash. If anything I just need to add config hash.

jberry-suse:
Cleaned up code a tad so it can cache more, and included staging project meta revision in hash to detect config changes.

jberry-suse:
For reference the last `repo-checker` run on `packageslists`:

- Oct 31 18:24:18
- Oct 31 23:39:33

Over 5 hours. I would expect this to drop that by 80% at least.

PR #1230, opened 2017-10-28
devel-project: override osc.core.search() to include withfullhistory for get_review_list().

jberry-suse:
#1206 was a bit to optimistic as the deficiency was only half resolved.

### DIFF
--- a/devel-project.py
+++ b/devel-project.py
@@ -12,7 +12,6 @@ from osc.core import get_request_list
 from osc.core import get_review_list
 from osc.core import http_GET
 from osc.core import makeurl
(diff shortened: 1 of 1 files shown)


jberry-suse:
Resolves this error, seen on `packagelists`:

```
  File "./devel-project.py", line 238, in <module>
    sys.exit(args.func(args))
  File "./devel-project.py", line 134, in reviews
    age = request_age(request)
  File "./devel-project.py", line 92, in request_age
    date = dateutil.parser.parse(request.statehistory[0].when)
IndexError: list index out of range
```

PR #1233, opened 2017-10-31
accept_command: force to accept the staging if --force parameter is given

nilxam:
If --force parameter is given, it must to force to change the request state, otherwise --force does not work if still had pending review.

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -138,7 +138,8 @@ class AcceptCommand(object):
                 change_request_state(self.api.apiurl,
                                      str(req['id']),
                                      'accepted',
-                                     message='Accept to %s' % self.api.project)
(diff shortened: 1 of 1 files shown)


jberry-suse:
Is there a use-case for separating this out as a new flag? `--force` being used to skip openqa/build results and a new flag to for when looking to skip on obs level.

DimStar77:
Don't think an extra flag makes much sense: if i want to accept --force Letter, i want it to get accepted - regardles of what is not ready yet

nilxam:
Right, release manager should understand what is the risk if using `--force` and they have to review the staging before accepts it, therefor `--force` should accept Letter staging anyway regardless something is not ready.

jberry-suse:
Fair enough.

PR #1234, opened 2017-10-31
accept_command: remove the ring package if it is 2nd specfile is removed

nilxam:
may fixes https://github.com/openSUSE/osc-plugin-factory/issues/1232


DimStar77:
LGTM

PR #1249, opened 2017-11-09
repo_checker: review failed stagings with only openQA failures.

### LINKED ISSUE #1069: repo_checker: process stagings that failed due only to openqa
repochecker now uses the overall state that the staging plugin reports before reviewing:
https://github.com/openSUSE/osc-plugin-factory/blob/master/repo_checker.py#L114

Currently looking at M and apparmor the staging has actually finished building but fails in openQA. So the overall state is failed and repochecker doesn't check the package. I'd suggest to look at the build result only to speed things up.

jberry-suse:
Build failures may cause false negatives in repo_checker results, but openQA failures are not relevant and should be ignored.

Fixes #1069.


nilxam:
Is it deployed to packagelists?

jberry-suse:
Yes

PR #1270, opened 2017-11-22
check_source: Ignore download_files failure in some cases

nilxam:
Fixes https://github.com/openSUSE/osc-plugin-factory/issues/1264


nilxam:
CC @DimStar77 @lnussel

lnussel:
I think a cleaner approach (if there's any hope for clean here) would be to e.g os.putenv('IGNORE_DOWNLOAD_FAIL', '1') and then check for that in the perl part to not fail in the first place.

PR #1272, opened 2017-11-22
Split EDITOR environment variable to handle flags in prefix.

### LINKED ISSUE #1271: staging plugin: EDITOR set to "/usr/bin/gvim -f" breaks --interactive
if EDITOR environment variable is set to "/usr/bin/gvim -f" (to make sure gvim call is blocking ), call to plugin with --interactive fails with "[Errno 2] No such file or directory"

jberry-suse:
Fixes #1271.

@fcrozat: if you have the time could you confirm? I tested with `cat -n`.


fcrozat:
I can confirm the fix works fine for me

PR #1275, opened 2017-11-24
Add more details to openQA job group description about current status

okurz:
Example output from dry run:

```
2017-11-24 06:42:57,927 - totest-manager:490 DEBUG snapshotable: True
2017-11-24 06:42:57,927 - totest-manager:495 DEBUG no change in snapshot version
2017-11-24 06:42:58,415 - totest-manager:207 DEBUG Writing openQA status message: {'text': 'pinned-description: Ignored issues\r\n\r\n\r\n\r\ntag:20171123:testing:testing'}
```

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -170,7 +170,7 @@ class ToTestBase(object):
             logger.info('%s %s %s' %
                         (module['name'], module['result'], module['flags']))
 
-    def update_ignored_issues(self):
+    def update_openqa_status_message(self):
         url = makeurl(self.openqa_server,
                       ['api', 'v1', 'job_groups'])
         f = self.api.retried_GET(url)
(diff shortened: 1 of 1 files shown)


okurz:
@DimStar77 @nilxam WDYT?

DimStar77:
This PR should be a different branch than  the ttm-SLE15-enablement; No way to merge only the 2nd commit here

As it's a going to be a pinned comment, I'd assume the 'flag' just to move around and the history won;t show which snapshots were released and which were not, but that is a minor thing which would have been a nice to have for the feature.

okurz:
Yes, I tried to rebase and separate but that wasn't easy so I didn't do for the time being. If someone would approve the SLE support I don't need to invest the effort :)

okurz:
rebased after SLE support was merged. Should be easier to review now.

okurz:
As I don't have access to the machine running ttm I assume you will tell me if there are any problems around this :)

DimStar77:
> As I don't have access to the machine running ttm I assume you will tell me if there are any problems around this :)

you can bet on that

PR #1280, opened 2017-11-30
osclib/cycles: only cache target project build graph to avoid excessive cache.

### LINKED ISSUE #1265: repo_checker: excessive cache file
Just found that ~/.cache/opensuse-repo-checker/_builddepinfo was 63GB, making repochecker block in D all the time. Do we need the _buildepinfo persistent on disk?

jberry-suse:
- 824fc98afd276befc345bc8be2c0ba2b05a0c251:
    osclib/cycles: only cache target project build graph to avoid excessive cache.

- 57d9d2986bf96cc104fc413707780a2ae12dddf9:
    osclib/cycles: drop _get_builddepinfo and _get_builddepinfo_cycles since unused.

Fixes #1265.

### DIFF
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -66,10 +66,13 @@ class Cache(object):
 
     CACHE_DIR = os.path.expanduser('~/.cache/osc-plugin-factory')
     TTL_LONG = 12 * 60 * 60
+    TTL_MEDIUM = 30 * 60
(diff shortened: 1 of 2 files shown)


jberry-suse:
Refactored to use `osclib/cache.py` to avoid keeping graphs in memory and thus avoid memory issues.

PR #1281, opened 2017-11-30
repochecker: add --force and --limit-group options

lnussel:
useful for debugging to limit reviews to a specific group

### DIFF
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -456,6 +456,12 @@ class RepoChecker(ReviewBot.ReviewBot):
         # TODO Ignore tgt_project packages that depend on this that are part of
         # ignore list as and instead look at output from staging for those.
         what_depends_on = depends_on(self.apiurl, action.tgt_project, 'standard', [action.tgt_package], True)
+
+        # filter out dependency on package itself (happens with eg
+        # java bootstrapping itself with previous build)
+        if action.tgt_package in what_depends_on:
+            what_depends_on.remove(action.tgt_package)
+
         if len(what_depends_on):
             self.logger.warn('{} is still a build requirement of {}'.format(action.tgt_package, ', '.join(what_depends_on)))
 
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -39,6 +39,8 @@ class RepoChecker(ReviewBot.ReviewBot):
 
         # RepoChecker options.
         self.skip_cycle = False
+        self.force = False
+        self.limit_group = None
 
     def project_only(self, project, post_comments=False):
         # self.staging_config needed by target_archs().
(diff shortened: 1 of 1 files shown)


repo_checker.py
```diff
@@ -39,14 +39,16 @@ def __init__(self, *args, **kwargs):
...
+        self.limit_group = None
 
     def project_only(self, project, post_comments=False):
         # self.staging_config needed by target_archs().
         api = self.staging_api(project)
 
         root = ET.fromstringlist(show_results_meta(
             self.apiurl, project, multibuild=True, repository=['standard']))
-        if len(root.xpath('result[@state!="published"]')):
+        if not self.force and len(root.xpath('result[@state!="published"]')):
```
jberry-suse: I have suggest perhaps a "--skip-checks" or somesuch to skip all the checks in that section as sometimes helpful when manually running, but "--force" seems a bit ambiguous and ends up only forcing one check.

repo_checker.py
```diff
@@ -119,6 +121,9 @@ def prepare_review(self):
                 self.logger.debug('{}: not staged'.format(request.reqid))
                 continue
 
+            if self.limit_group and group != 'openSUSE:Leap:15.0:Staging:C':
```
jberry-suse: Assume this is not intentional to have hard-coded group?

repo_checker.py
```diff
@@ -476,6 +481,8 @@ def get_optparser(self):
         parser = ReviewBot.CommandLineInterface.get_optparser(self)
 
         parser.add_option('--skip-cycle', action='store_true', help='skip cycle check')
+        parser.add_option('--force', action='store_true', help='force review even if project is not ready')
```
jberry-suse: I have debating adding a `--skip-checks` or somesuch for skipping the normal checks on regular run. `--force` seems a bit ambiguous especially when it only skips certain checks.

repo_checker.py
```diff
@@ -476,6 +481,8 @@ def get_optparser(self):
...
+        parser.add_option('--limit-group', metavar='GROUP', help='only review requests in specific group')
```
jberry-suse: Is this relevant for `ReviewBot` as could want to run any bot against requests in a group to get a staging finished? Although currently this is likely the only bot to be waiting on besides legal.

jberry-suse:
Main concern being the hard-coded bit.

repo_checker.py
lnussel: oh, shit :)

repo_checker.py
lnussel: I don't mind, maybe I missed some cases that I didn't run into. Those should be covered by --force too of course. I already spotted the "build unchanged" case now.

repo_checker.py
lnussel: could be refactored into reviewbot indeed

Issue #1222, opened 2017-10-25
rebuildpacs.pl: seems to have major pain with _multibuild packages


DimStar77:
The HPC team has been converting most of their packages to _multibuild... 

rebuildpacs.pl seems to identify them constantly as 'having trouble' and retriggers them, even though it should not retrigger packages it already recorded in the problems list (on first trigger they are being recorded, then only new stuff should be attempted)

In order to get over the obstacle at all for now, rebuildpacs.pl is hotpacthed on the pkglist vm to ignore some of the packages I have seen being retriggered over and over, which of course bears the risk of masking actual issues

coolo:
it's gone now

PR #1283, opened 2017-12-02
rebuildpacs: Ignore some more multibuild packages

DimStar77:
Rebuildpacs has quite some issues when it comes to multibuild packages as
it can trigger packages over and over.

This works around the major pains for now, see https://github.com/openSUSE/osc-plugin-factory/issues/1222 for
more details.


DimStar77:
NOTE: this had been hot-patched on the packagelist VM ever since; until we can address #1222 we better have this merged correctly though

lnussel:
🤢

PR #1287, opened 2017-12-07
totest-manager: Make sure internal object for notifications is always valid

okurz:
Fixes a problem I saw in the logfiles 'ERROR 'ToTestFactory' object has no
attribute 'update_pinned_descr'' in case of a failed snapshot.


nilxam:
isn't it better to be defined while initialing in _init_?

okurz:
yes, true. Fixed

okurz:
yes, true. Fixed

PR #1290, opened 2017-12-07
systemd units: remove oneshot type

lnussel:
Those services are of type 'simple'

### DIFF
--- a/systemd/osrt-leaper-review.service
+++ b/systemd/osrt-leaper-review.service
@@ -2,7 +2,6 @@
 Description=openSUSE Release Tools: leaper review
 
 [Service]
-Type=oneshot
 User=osrt-leaper
 ExecStart=/usr/bin/osrt-leaper --verbose --manual-version-updates --manual-maintenance-updates review
 
--- a/systemd/osrt-metrics@.service
+++ b/systemd/osrt-metrics@.service
@@ -2,11 +2,9 @@
 Description=openSUSE Release Tools: metrics for %i
 
 [Service]
-Type=oneshot
 User=osrt-metrics
 SyslogIdentifier=osrt-metrics
 ExecStart=/usr/bin/osrt-metrics --debug -p "%i"
-TimeoutStartSec=8 hour
 
 [Install]
 WantedBy=multi-user.target
--- a/systemd/osrt-repo-checker-project_only@.service
+++ b/systemd/osrt-repo-checker-project_only@.service
@@ -2,11 +2,9 @@
 Description=openSUSE Release Tools: repo-checker project_only for %i
 
 [Service]
-Type=oneshot
 User=osrt-repo-checker
 SyslogIdentifier=osrt-repo-checker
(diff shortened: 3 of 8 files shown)


DimStar77:
huh? osrt-repo-checker is no 'simple' service. it is triggered by a timer and exits upon completion; it is not running in an endless loop.

a 'simple' service does not terminate on its own - if it does so, ths service is marked as 'failed'

a 'oneshot' service is supposedly exiting at one point

jberry-suse:
In fact I suggest the exact opposite. `osrt-pkglistgen@.service` should have oneshot type added.

lnussel:
If the process exits with exit code zero all is fine:

```
repochecker:/root # systemctl status osrt-pkglistgen@openSUSE:Leap:15.0 
* osrt-pkglistgen@openSUSE:Leap:15.0.service - openSUSE Release Tools: generate package lists for openSUSE:Leap:15
   Loaded: loaded (/usr/lib/systemd/system/osrt-pkglistgen@.service; disabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/osrt-pkglistgen@openSUSE:Leap:15.0.service.d
           `-letters.conf
   Active: inactive (dead) since Fri 2017-12-08 08:56:28 CET; 3min 57s ago
  Process: 32419 ExecStart=/usr/bin/osrt-pkglistgen-openSUSE:Leap:15.0-all (code=exited, status=0/SUCCESS)
 Main PID: 32419 (code=exited, status=0/SUCCESS)
```
Quoting systemd.service(5):
> Behavior of oneshot is similar to simple; however, it is expected   that the process has to exit before systemd starts follow-up units.

IOW systemd blocks waiting to the service to terminate. Ie systemd shows it as 'starting' instead of running. You can see that also in the journal where systemd logs 'started' when the service terminates.

jberry-suse:
My understanding is the "starting" state is intended to work that way. The documentation you quoted, that I also quoted in the issue, describes exactly how our services work. Starts, must exit before another, and then a timer kicks it off again. `oneshot` meaning that the service is not "good" until it exits vs simple which is good once it is running.

Seems to be two distinct forms of `oneshot`, those that remain and basically just start something else and scripts such as ours:

- https://gist.github.com/drmalex07/d006f12914b21198ee43
- https://coreos.com/os/docs/latest/scheduling-tasks-with-systemd-timers.html

That or `systemd` has managed to make the most misleading documentation ever.

lnussel:
we'll see what happens when we merge this

jberry-suse:
... not a matter or working or not as clearly it will "work" both ways...

jberry-suse:
Also not sure why the timeout was removed as I added it after the bs_mirrorfull script completely hung on occasion.

lnussel:
BSRPC::rpc has a timeout parameter, so if bs_mirrorful is prone to hanging we should add the timeout there

jberry-suse:
Either way that should have been discussed separately instead of just dropping things, but then again if going to ignore the consensus and merge what's the difference in the end.

lnussel:
sometimes I'm just mean :)

PR #1294, opened 2017-12-08
cleanup_rings: first attempt to fix for multibuild

lnussel:


osclib/cleanup_rings.py
```diff
@@ -120,16 +117,24 @@ def repo_state_acceptable(self, project):
...
-            b = bdep.attrib['name']
-            if b not in self.bin2src:
-                continue
-            b = self.bin2src[b]
-            self.pkgdeps[b] = 'MYdvd{}'.format(self.api.rings.index(project))
+        for dvd in ('000product:openSUSE-dvd5-dvd-{}'.format(arch), 'Test-DVD-{}'.format(arch)):
+            try:
+                url = makeurl(self.api.apiurl, ['build', project, 'images', arch, dvd, '_buildinfo'])
+                root = ET.parse(http_GET(url)).getroot()
+            except urllib2.HTTPError as e:
```
DimStar77: NameError: global name 'urllib2' is not defined

osclib/cleanup_rings.py
```diff
@@ -112,16 +118,24 @@ def repo_state_acceptable(self, project):
...
-        root = ET.parse(http_GET(url)).getroot()
-        for bdep in root.findall('bdep'):
-            if 'name' not in bdep.attrib:
-                continue
-            b = bdep.attrib['name']
-            if b not in self.bin2src:
-                continue
-            b = self.bin2src[b]
-            self.pkgdeps[b] = 'MYdvd{}'.format(self.api.rings.index(project))
+        for dvd in ('000product:openSUSE-dvd5-dvd-{}'.format(arch), 'Test-DVD-{}'.format(arch)):
```
jberry-suse: It would seem at minimum the `000product` should allow for config override since the new pkglistgen was setup with it as a variable and kept during refactoring. Having the defaults in one place would be nice, but at least not having this be gotcha would be nice.

https://github.com/openSUSE/osc-plugin-factory/blob/7eaed349b38ed3e482f543abb717390b5d6b9afe/pkglistgen.py#L1023-L1025

The dvd name likely should be as well as this won't do much for SLE.

PR #1297, opened 2017-12-11
Add support for non-x86_64 products (Leap 15.0/aarch64)

dirkmueller:


pkglistgen.py
```diff
@@ -69,7 +70,7 @@ def __init__(self, name, pkglist):
         self.solved = False
         self.not_found = dict()
         self.unresolvable = dict()
-        for a in self.architectures:
+        for a in ARCHITECTURES:
```
lnussel: should be the specified architecures here

pkglistgen.sh
```diff
@@ -2,14 +2,16 @@
 # FIXME: 000package-groups is not frozen, osc up doesn't do anything
 # when updated in underlying project
 
-set -e
+set -ex
```
lnussel: x would just spam the logs here

pkglistgen.py
dirkmueller: no, thats the fix.. in my case specified architectures is aarch64, but the package group contained x86_64, which resulted in keyerror.

nilxam:
please next time make sure travis passed before merging...

DimStar77:
https://github.com/openSUSE/osc-plugin-factory/pull/1307 fixes the issues introduced with this PR (one was caught by travis on the PR, the 2nd was test setup failure, where the package built was not based on the PR)

PR #1301, opened 2017-12-15
Ignore hdf5:serial

nilxam:


DimStar77:
Hope one day we get to get rebuildpacs to not mess up multibuild flavors :)

PR #1303, opened 2017-12-17
Add 15.0 Ports to totest-manager

dirkmueller:


totest-manager.py
```diff
@@ -725,6 +725,25 @@ def get_current_snapshot(self):
         return self.iso_build_version(self.project + ':ToTest', self.main_products[0])
 
 
+class ToTest150Ports(ToTestBaseNew):
+    main_products = [
+        '000product:openSUSE-cd-mini-aarch64',
+        '000product:openSUSE-dvd5-dvd-aarch64',
```
okurz: Wasn't there ppc64le as well?

dirkmueller: There is ppc64le as well, but it doesn't build yet. I will add it once it does in a new PR

PR #1305, opened 2017-12-19
Be more forgiving with ttm ignore comments

dirkmueller:
accept @ttm: ignore also as valid input

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -245,7 +245,7 @@ class ToTestBase(object):
                         refs.add(str(ref))
                     if comment['userName'] == 'ttm' and comment['text'] == 'label:unknown_failure':
                         labeled = comment['id']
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -245,7 +245,7 @@ def overall_result(self, snapshot):
                         refs.add(str(ref))
                     if comment['userName'] == 'ttm' and comment['text'] == 'label:unknown_failure':
                         labeled = comment['id']
-                    if comment['text'].find('@ttm ignore') >= 0:
+                    if re.search('\@ttm\:? ignore', comment['text']):
```
aplanas: re.search(r'@ttm:? ignore') can be more clear
(not tested)

DimStar77:
(closed/reopened for new travis run - issues should be solved in master)

PR #1306, opened 2017-12-19
freeze_command: use result call to get proper multibuild package information for aggregatepac

nilxam:
Side note: do not need to give it arch parameter as the result list should not architecture specific.

Fixes https://github.com/openSUSE/open-build-service/issues/4245


DimStar77:
LGTM

nilxam:
wait for https://github.com/openSUSE/osc-plugin-factory/pull/1307

PR #1307, opened 2017-12-19
Fix packaging after commit 3e5f3b137

DimStar77:


DimStar77:
For the packaging part, travis is not using my branch apparently - or at least not the right rev:

```
+<service name="tar_scm">
+                <param name="url">/usr/src/target</param>
+              <param name="changesrevision">c9a0027f8a42284c4bf5bbdcb36fe754ab1f7e0a</param></service></servicedata>
```

so, using the 'local' path to the git checkout, but then using the MASTER rev of the parent branch.

nilxam:
LGTM

jberry-suse:
Ah yeah! I have a local cleanup to improve this a variety of ways as well as to fix this which was introduced during the twice to `osc build` instead of direct rpm commands. Thanks.

PR #1308, opened 2017-12-20
rebuildpacs: rebuild perf

lnussel:
perf needs to be rebuilt when kernel-source changes


lnussel:
https://bugzilla.suse.com/show_bug.cgi?id=1033067

PR #1309, opened 2017-12-20
reochecker: avoid accessing empty status

lnussel:


nilxam:
nit: em**pt**y :wink:

PR #1318, opened 2018-01-03
 tests/comment: flesh out tests and fix OBSLocal same process authentication.

jberry-suse:
Since generally helpful and making a variety of changes of last couple months to this heavily relied upon code it seems worth a few more tests.


jberry-suse:
With the addition of tests against local OBS this should be full coverage. Even checks removing nested comments as that was bugged before which caused `osc staging accept` to crash as @DimStar77 noticed.

Issue #1221, opened 2017-10-24
tests: decide how to proceed installing additional dependencies on travis


jberry-suse:
In order to use `osc vc` to generate a changelog entry (rather than re-creating functionality in python) the `build` package is needed. This is available for ubuntu out of `openSUSE:Tools` repo, but will not be the last odd dep that ubuntu may not have. As such it may make sense to run the tests themselves in an `openSUSE` container or inside the `osc build` changeroot. The latter will be tricky to get the local OBS available there, but has the benefit that those tests can run on OBS.

Until this is resolved I have not commited the `check_source.py` tests and have disabled the test suite in `%check` section of spec. The spec may also end up just needing to filter out the integration tests and only run the mock tests. The method of filtering needs to be figured out as whitelisting or blacklisting seems a bit suboptimal.

coolo:
this is done using the docker container, no?

jberry-suse:
A large part is yes. For doing the package builds yes, but not for running the tests themselves which is the part that really needs it to do more.

lnussel:
so all this dance for osc vc? How about reimplementing the vc command in osc directly, in python?

jberry-suse:
> so all this dance for osc vc?

That being part of it, but there are other requirements. Additionally, the whole point is to make things a) maintainable, b) representative of real thing so re-inventing more code brings us back to mocks.

> IMO we should use docker-compose and run the tests on opensuse.

Running the tests in a docker container like I did for spec builds is a possibility and one I like, but introduces a bit of complexity surrounding using python versions travis supplies or having them available in openSUSE in addition to properly communicating to and from the container (like coverage results).

Also the possibility of installing the built packages and running the tests against those which would catch missing deps and such as well as installation problems.

jberry-suse:
The other big one at the time was getting the obs_factory plugin included in test setup, but that should be done now since included in OBS.

coolo:
you really should look at circleci. They don't even provide a VM at all, they only give you  container orchestration. As such you can only test python versions you have available in openSUSE, but I don't think that's a big problem.

coolo:
I went with running the tests in a docker container itself composed against the obs containers

PR #1320, opened 2018-01-04
osclib/stagingapi: detect baselibs.conf in adi package and ensure archs enabled.

### LINKED ISSUE #1304: adi staging: add i586 arch if there is a baselibs.conf
Especially for the case of wine: adi must build i586 so wine is completely installable, otherwise repo-checker will have to complain:

```
can't install wine-3.0~rc2-465.1.x86_64:
  nothing provides wine-32bit = 3.0~rc2 needed by wine-3.0~rc2-465.1.x86_64
can't install wine-nine-3.0~rc2-465.1.x86_64:
  nothing provides wine-nine-32bit = 3.0~rc2 needed by wine-nine-3.0~rc2-465.1.x86_64
```

IIRC, @jberry-suse already had this once patched and tested, but seems this was never pushed to git an activated for the bot? (I think we agreed then to enable i586 for everything that comes with a baselibs.conf)

jberry-suse:
Fixes #1304.

I have a test prototype, but too much local hax to run it pending #1221.

Tested again by running:

```
osc staging adi --move wine
```

Which moved [request 561156](https://build.opensuse.org/request/show/561156) and enabled `i586` in the new staging where it was not enabled in previous and had `repo-checker` comment.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1481,7 +1481,8 @@ class StagingAPI(object):
         :param package:  The package to quert
         :param filename: The filename to query
         """
-        url = self.makeurl(['source', project, package, '{}?expand=1'.format(filename)])
+        query = {'expand': 1}
+        url = self.makeurl(['source', project, package, filename], query)
         try:
             return http_GET(url).read()
         except urllib2.HTTPError:
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1257,6 +1257,13 @@ class StagingAPI(object):
         url = self.makeurl(['source', project, tar_pkg, '_link'])
         http_PUT(url, data=ET.tostring(root))
 
+        # If adi project, check for baselibs.conf in all specs to catch both
+        # dynamically generated and static baselibs.conf.
+        baselibs = False if self.is_adi_project(project) else None
+        if baselibs is False and 'baselibs.conf' in str(self.load_file_content(
+            src_prj, src_pkg, '{}.spec'.format(src_pkg), src_rev)):
+            baselibs = True
+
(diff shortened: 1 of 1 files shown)


jberry-suse:
Staged [request 561505](https://build.opensuse.org/request/show/561505) and noted that `i586` was not added.

DimStar77:
Do we really need to parse the .spec files? IIRC we already have an API call to just list the files, which should be cheaper (it allows to filter filenames serverside) - and in case of multiple spec files would not require multiple api requests

jberry-suse:
We discussed this and determined that some `baselibs.conf` are dynamically generated in `.spec` so this was the only reliable way of which I am aware.

DimStar77:
You are right - as usual :)

Then, LGTM

jberry-suse:
This discussion is probably a good reason to add a comment indicate as much which I will do.

jberry-suse:
```diff
diff --git a/osclib/stagingapi.py b/osclib/stagingapi.py
index 90983fc..eac803b 100644
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1257,7 +1257,8 @@ class StagingAPI(object):
         url = self.makeurl(['source', project, tar_pkg, '_link'])
         http_PUT(url, data=ET.tostring(root))
 
-        # Check for baselibs.conf in specs if adi project.
+        # If adi project, check for baselibs.conf in all specs to catch both
+        # dynamically generated and static baselibs.conf.
         baselibs = False if self.is_adi_project(project) else None
         if baselibs is False and 'baselibs.conf' in str(self.load_file_content(
             src_prj, src_pkg, '{}.spec'.format(src_pkg), src_rev)):
```

nilxam:
LGTM

PR #1331, opened 2018-01-09
Remove reference to the deleted pkglistgen.sh script

dirkmueller:


jberry-suse:
Perhaps include the description bits from `.sh` in the `.py` since it took over the role.

jberry-suse:
Well really it should be reworded further since the line before is:

```
The package list generator consists of two scripts:
```

Since I made the refactoring I can do so if you are not interested.

jberry-suse:
Closing since it has been 10 days for doc change. I create relevant PR.

PR #1333, opened 2018-01-10
pkglistgen fixes

lnussel:


lnussel:
dvd fix applied as hotfix

pkglistgen.py
```diff
@@ -1013,8 +1014,9 @@ def do_update_and_solve(self, subcmd, opts):
                 if api.rings:
                     opts_dvd = copy.deepcopy(opts)
                     opts_dvd.project += ':DVD'
-                    self.options.repos.insert(0, '/'.join([opts.project, main_repo]))
+                    self.options.repos.insert(0, '/'.join([opts_dvd.project, main_repo]))
```
jberry-suse: Yeah, I was guessing there was something wrong with staging bit after output, this makes sense.

PR #1341, opened 2018-01-11
new requestfinder tool

lnussel:
helps to mass approve, decline reopen reviews and requests

### DIFF
--- a/requestfinder.py
+++ b/requestfinder.py
@@ -1,0 +1,223 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 SUSE LLC
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from ConfigParser import ConfigParser
+from xdg.BaseDirectory import load_first_config
+from xml.etree import cElementTree as ET
+import sys
+import cmdln
+import logging
+import urllib2
+import osc.core
+import yaml
+import os
+
+import ToolBase
+
+makeurl = osc.core.makeurl
+
+logger = logging.getLogger()
+
+
+class Requestfinder(ToolBase.ToolBase):
+
+    def __init__(self):
+        ToolBase.ToolBase.__init__(self)
+        self.devel = None
+
+    def fill_package_meta(self, project):
+        self.package_metas = dict()
+        url = makeurl(self.apiurl, ['search', 'package'], "match=[@project='%s']" % project)
+        root = ET.fromstring(self.cached_GET(url))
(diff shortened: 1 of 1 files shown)


requestfinder.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018 SUSE LLC
```
jberry-suse: From what I understand the copyright and warranty at the top of every file is unnecessary (see "Berne Convention").

requestfinder.py
```diff
@@ -0,0 +1,226 @@
...
+class Requestfinder(ToolBase.ToolBase):
+
+    def __init__(self):
+        ToolBase.ToolBase.__init__(self)
+        self.devel = None
+
+    def fill_package_meta(self, project):
+        self.package_metas = dict()
+        url = makeurl(self.apiurl, ['search', 'package'], "match=[@project='%s']" % project)
+        root = ET.fromstring(self.cached_GET(url))
```
jberry-suse: Should be able to pass `withfullhistory` ever since https://github.com/openSUSE/osc/commit/d401eb2c691027e8913147029a3ef2c126b2e5c4 like I do in `metrics.py` and elsewhere.

https://github.com/openSUSE/osc-plugin-factory/blob/c32b59bd554556520bdc4334c8cb6dd1b2957a6b/devel-project.py#L25-L34

requestfinder.py
```diff
@@ -0,0 +1,226 @@
...
+        url = osc.core.makeurl(self.apiurl, ('search', 'request'), {"match": xquery})
+        root = ET.parse(osc.core.http_GET(url)).getroot()
+
+        self.requests = []
+
+        for request in root.findall('request'):
+            req = osc.core.Request()
+            req.read(request)
+#           with full history
+#           req = osc.core.get_request(self.apiurl, req.reqid)
```
jberry-suse: Should be able to drop with `search()` change detailed above.

requestfinder.py
```diff
@@ -0,0 +1,226 @@
...
+        if opts.query:
+            settings['query'] = opts.query
+
+        if not settings['query']:
+            raise Exception('please specify query')
+
+        rqs = self.tool.find_requests(settings['query'])
+        for r in rqs:
+            if r.actions[0].type == 'submit':
+                # only works with full history r.get_creator(),
```
jberry-suse: Also should resolve.

requestfinder.py
```diff
@@ -0,0 +1,226 @@
...
+        EOF
+        $ ${name} review --settings foo | tee doit.sh
+        ./doit.sh
+        """
+
+if __name__ == "__main__":
+    app = CommandLineInterface()
+    sys.exit(app.main())
+
+# vim: sw=4 et
```
jberry-suse: Is there no project level configuration file that can be added for this rather than including in all files? Still never get this as we could add one for every editor folks use.

requestfinder.py
```diff
@@ -0,0 +1,226 @@
...
+            settings['query'] = opts.query
+
+        if not settings['query']:
+            raise Exception('please specify query')
+
+        rqs = self.tool.find_requests(settings['query'])
+        for r in rqs:
+            if r.actions[0].type == 'submit':
+                # only works with full history r.get_creator(),
+                print '#', r.reqid, r.actions[0].type, r.actions[0].src_project, r.actions[0].src_package, r.actions[0].tgt_project
```
jberry-suse: Old-style (keyword instead of function) `print` is deprecated, but perhaps this code will never live to see python 3.

requestfinder.py
lnussel: nope, every single file should include copyright and license

requestfinder.py
lnussel: there are emacs tags too so this is just fair. There are only those two editors, the rest is toys anyways :p

jberry-suse: Religious debates aside is there no way to configure that for an entire directory? Like `.vimrc` or somesuch?

requestfinder.py
jberry-suse: If one wants to be overly verbose perhaps, but it isn't.

requestfinder.py
jberry-suse: Not sure what the plan is as you removed the other references to references to `withfullhistory`. This comment should be on the `['search', 'request']` line below instead.

requestfinder.py
jberry-suse: Even OBS doesn't. Can read about it at length, but at the most a single line reference to read `LICENSE` file. Every file is like putting the entire mailing address, name, and phone number embossed on every part of a consumer product. The convention is there and copyright granted automatically since 1800s. From my understanding the practice started as pandering to lawyers since common sense typically does not apply in that realm.

requestfinder.py
lnussel: with some extra code in .vimrc yes

requestfinder.py
lnussel: the code that required withfullhistory wasn't used anymore

requestfinder.py
jberry-suse: So we have something like 30 lines of boilerplate per file before we even get to python? Eww.

PR #1343, opened 2018-01-12
osclib/request_splitter: rework quick strategy to handle Leap 15.0 workflow

### LINKED ISSUE #1340: staging: update quick strategy to handle ReviewBots that were added to the 15.0 workflow
Likely the correct resolution to @lnussel's complaint about the `staging-bot` not being "aggressive enough with SLE requests". I noticed I had not seen a `quick` strategy for quite a while and confirmed my suspicion about the implementation and as to why.

Just need to decide if no reviews other than whitelisted (like bots) or look for legal/human project reviews.

jberry-suse:
The addition of factory-auto and repo-checker reviews need to be allowed
since they are always present and do not constitute a non-quick review.
In an attempt to future proof the strategy the allowed review groups are
taken from the project configuration and repo-checker from config.

Fixes #1340.

### DIFF
--- a/osclib/request_splitter.py
+++ b/osclib/request_splitter.py
@@ -3,6 +3,7 @@ import dateutil.parser
 import hashlib
 from lxml import etree as ET
 from osc import conf
+from osc.core import show_project_meta
 import re
 
 class RequestSplitter(object):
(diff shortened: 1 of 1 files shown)


jberry-suse:
Reworked, to improve robustness of xpath.

PR #1344, opened 2018-01-12
osclib/request_splitter: allow special packages to be empty.

jberry-suse:
@lnussel I added this variable override a while back, but seems we forgot about it since you asked in IRC the other day. Given it sounds like you would like no special packages for Leap I updated code to handle empty string variable override.

I'll go ahead and update project config for Leap 15.0 to set the special list to empty, but feel free to update.


jberry-suse:
[dashboard/config change](https://build.opensuse.org/package/rdiff/openSUSE:Leap:15.0:Staging/dashboard?linkrev=base&rev=263)

jberry-suse:
For now it will just keep the default list until this is deployed. Can also set the list to any random package name that it will never hit.

PR #1348, opened 2018-01-17
ttm: change Leap 15 product path in current_version

nilxam:
There is no openSUSE-cd-mini-x86_64 in Leap 15 main project's 000product, it
was only exist in :ToTest sub-project.


nilxam:
not really correct

PR #1349, opened 2018-01-17
pkglistgen: delete kiwis by scope

lnussel:
Replace config option pkglistgen-delete-kiwis by
pkglistgen-delete-kiwis-$scope. The main project needs all kiwis while
in Rings and Stagings we don't want the ftp tree for example.


lnussel:
hot patched on tortuga

PR #1350, opened 2018-01-17
news/web: allow dot in version.

### LINKED ISSUE #1325: factory-package-news-web: handle dot in build numbers
in Leap 15.0 build numbers have a dot in them because OBS thinks thats a good idea. factory-package-news-web can't handle that (digits_re)

jberry-suse:
Fixes #1325.

### DIFF
--- a/factory-package-news/factory-package-news-web.py
+++ b/factory-package-news/factory-package-news-web.py
@@ -28,7 +28,7 @@ import os
 import sys
 from urlparse import urlparse
 
-digits_re = re.compile('^[0-9]+$')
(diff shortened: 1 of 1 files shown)


lnussel:
I haven't thought through the code again, are we safe against '..'?

jberry-suse:
I'll look at it further, but believe we should be for the same reasons are before.

jberry-suse:
It seems the same logic applies from https://github.com/openSUSE/osc-plugin-factory/pull/826#discussion_r112223426.

Since `..81.1` would not cause it to go back a dir, but only `..` by itself. Which is interpreted the same way.

PR #1354, opened 2018-01-19
osc-staging: add project selector to unlock command in help output

krauselukas:
For me as a beginner it wasn't directly obvious that by default factory gets unlocked if no other project is selected. I know it's the same with the select command etc. but i thought it might be good to mention it in combination with the unlock command.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -302,7 +302,8 @@ def do_staging(self, subcmd, opts, *args):
     "unlock" will remove the staging lock in case it gets stuck or a manual hold
         If a command lock gets stuck while a hold is placed on a project the
         unlock command will need to be run twice since there are two layers of
-        locks.
+        locks. By default the Factory project gets unlocked if no other project
+        is selected with the "-p" parameter.
 
     "rebuild" will rebuild broken packages in the given stagings or all
         The rebuild command will only trigger builds for packages with less than
@@ -336,7 +337,7 @@ def do_staging(self, subcmd, opts, *args):
             [--merge] [--try-strategies] [--strategy]
             [STAGING...] [REQUEST...]
         osc staging unselect [--cleanup] [-m MESSAGE] [REQUEST...]
-        osc staging unlock
+        osc staging [-p PROJECT] unlock
         osc staging rebuild [--force] [STAGING...]
         osc staging repair [--cleanup] [REQUEST...]
         osc staging setprio [STAGING...]


osc-staging.py
```diff
@@ -336,7 +337,7 @@ def do_staging(self, subcmd, opts, *args):
             [--merge] [--try-strategies] [--strategy]
             [STAGING...] [REQUEST...]
         osc staging unselect [--cleanup] [-m MESSAGE] [REQUEST...]
-        osc staging unlock
+        osc staging [-p PROJECT] unlock
```
jberry-suse: Including a global option on one subcommand and does not make sense. The global options are documented first and the subcommand specific ones last. Having this documentation split up by subcommand might be nice, but the current implementation does not lend it self to that.

If anything perhaps some note about the `--project` option at the beginning of the documentation. Could also mention the feature I added to allow detecting project based on `.osc/project` from current directory (which is how I work along with PS1 to indicate as much).

jberry-suse: It may also make sense not to include the `osc staging` bit in front of the subcommands to make it more clear this does not show the global options.

krauselukas: Makes sense, i was a bit to fast with the PR. The reason for using the unlock command was that i wanted to select a package to one of the adi's. By accident i just pasted a few spaces and entered. After cancelling with control-c there was an lock acquired by me. After like 10-15 minutes i unlocked it manually.

jberry-suse: Another behavior possibly worth documenting is that pressing `ctrl-c` once will kill and allow unlocking, but pressing more than once will immediately kill and leave the lock.

DimStar77:
I don't see the reason to add [-p …] to the command help.

-p is a staging global parameter, not specific to unlock (it is valid to all staging commands) and is documented in the top part of ocs staging --help:

```
> osc staging --help
staging: Commands to work with staging projects

Options:
    -h, --help          show this help message and exit
    --clear             clear value
    --append            append to existing value
    --save              save the result to the dashboard container
    --no-color          strip colors from output (or add staging.color = 0 to
                        the .oscrc general section
    --strategy=STRATEGY
                        apply a specific strategy
    --try-strategies    apply strategies and keep any with desireable outcome
    --merge             propose merge where applicable and store details to
                        allow future merges
    -n, --non-interactive
                        do not ask anything, use default answers
    -i, --interactive   interactively modify selection proposal
    --group-by=GROUP_BY
                        xpath by which to group requests
    --filter-by=FILTER_BY
                        xpath by which to filter requests
    -m MESSAGE, --message=MESSAGE
                        message used by ignore command
    --wipe-cache        wipe GET request cache before executing
    --no-bootstrap      do not update bootstrap-copy when freezing
    --no-cleanup        do not cleanup remaining packages in staging projects
                        after accept
    --cleanup           cleanup after completing operation
    --no-freeze         force the select command ignoring the time from the
                        last freeze
    -v, --version       print the plugin version
    -o, --old           use the old check algorithm
    --force             force action, overruling internal checks (CAUTION)
    --add=PACKAGE       mark additional packages to be checked by repo checker
    -p PROJECT, --project=PROJECT
                        indicate the project on which to operate, default is
                        openSUSE:Factory
    -f FROMPROJECT, --from=FROMPROJECT
                        specify a source project when moving a request
    --supersede         replace staged requests when superseded
    --split             split the requests into individual groups
    --by-develproject   split the requests by devel project
    --move              force the selection to become a move
```

osc-staging.py
jberry-suse: To clarify how I work https://asciinema.org/a/d8jEi7ZzxjrvQbs88clK0WOB0 which may be worth documenting.

krauselukas:
@DimStar77 yes i don' t see the necessity of my PR anymore, aswell. But i will keep the control-c behavior which @jberry-suse mentioned in mind.

jberry-suse:
@krauselukas Based on your last comment I will close this PR. Feel free to take one of the alternative documentation routes.

Issue #1361, opened 2018-01-24
ci: docker-compose-obs fails due to osc change


jberry-suse:
Fails to become responsive, but works locally. Looked at last few OBS commits and nothing interesting. Need to take a look at `osc`.

jberry-suse:
Nothing for `osc` either. Did `travis-ci` change something or is this more just funky?

jberry-suse:
Upon further investigation this is indeed caused by `osc` https://github.com/openSUSE/osc/commit/1a92c8b995f762f9357fa9630d7981ab1ab3497c. I can reproduce locally using that commit and the one before works fine.

jberry-suse:
Created openSUSE/osc#372.

jberry-suse:
Resolved by upstream.

PR #1362, opened 2018-01-24
dist/ci/docker-compose-obs: print osc output on timeout for debugging.

jberry-suse:
To aid in future debugging of `osc` failures as seen in #1361.


jberry-suse:
Properly exposed `osc` crash after timeout. So this works, once issue is fixed a retest should pass.

PR #1367, opened 2018-01-25
osclib/accept: print todo from staging config after completion.

### LINKED ISSUE #1324: TODO list for staging accept
sometimes it's important to remember something for staging accept, e.g changing the rebuild mode, modifying the prjconf etc. Would be nice to have some way to store some free form text that is shown on accept.

jberry-suse:
I went for the simplest implementation, just `todo` key. If this ends up being heavily used in such a way that it makes sense to support `todo*` or similar then adding something like `--add` support to `config` command where that adds incremental `todo` keys and reads them back with `todo*` then it can be added to accept as well.

For now this seems sufficient. I went ahead and ran the following for `Leap:15.0`.

```
osc staging config C todo something
```

So this can be tested in production when C is accepted. Otherwise I tested the added code manually outside of `accept` command. The result should be the follow after the normal output.

```
todo = something
```

Fixes #1324.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -192,6 +192,7 @@ def do_staging(self, subcmd, opts, *args):
         The keys that may be set in staging configuration are:
 
         - repo_checker-binary-whitelist[-arch]: appended to target project list
+        - todo: text to be printed after staging is accepted
 
     "cleanup_rings" will try to cleanup rings content and print
         out problems
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -6,6 +6,7 @@ from xml.etree import cElementTree as ET
 from osc.core import change_request_state, show_package_meta, wipebinaries
 from osc.core import http_GET, http_PUT, http_DELETE, http_POST
 from osc.core import delete_package, search, set_devel_project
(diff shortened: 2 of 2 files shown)


DimStar77:
```
Acceptable projects K L
Accepting staging review for cryptsetup
Accepting staging review for open-iscsi
Accepting staging review for dracut
Accepting staging review for argon2
Accepting staging review for installation-images
todo = None
Accepting staging review for kernel-source
todo = None
```

Is the print of 'todo = None' intentional?

jberry-suse:
That's how the config command works. It may make more sense to either print them before hand for all staging to be accepted and/or add no print on none option to config.

DimStar77:
I'd actually prefer the 'todo's printed _after_ all the accepts (skipping the None values).

After accept makes more sense as when starting ```osc staging accept```, the top is most likely running out of the screen already by the time the command ends; so telling me "what else do you have to do now, that you accepted" makes more sense aftet all acceptance (incl accept_other_new and eventual rebuild triggers that happen)

jberry-suse:
Makes sense, this implementation was (as stated) simplest. More curious to see if actually used, but the changes shouldn't be hard.

PR #1372, opened 2018-01-29
fcc_submitter: skip revoked request

nilxam:


fcc_submitter.py
```diff
@@ -173,7 +173,7 @@ def get_source_packages(self, project, expand=False):
         return packages
 
     def get_request_list(self, package):
-        return osc.core.get_request_list(self.apiurl, self.to_prj, package, req_state=('new', 'review', 'declined'))
+        return osc.core.get_request_list(self.apiurl, self.to_prj, package, req_state=('new', 'review', 'declined', 'revoked'))
```
jberry-suse: Does it make sense to mix a permanent state in with 3 temporary states? A single revoked request should always be picked up?

nilxam: well, that just make sure fcc submitter do not submit the package again anyway. fcc submitter should not submit again if the request has been declined or revoked, that must have a good reason for it, maintainer or release manager don't want it(or atm), ppl have to submit it manually in any case, but not automatically.

jberry-suse: Yeah, I suppose. Was confusing this with update crawler.

PR #1378, opened 2018-01-30
Handling for NonFree: splits NonFree request to a own group always

nilxam:
Splits NonFree request to a own group always and enable :NonFree project build
in meta.

### DIFF
--- a/osclib/adi_command.py
+++ b/osclib/adi_command.py
@@ -87,9 +87,13 @@ class AdiCommand:
             print(Fore.YELLOW + (group if group != '' else 'wanted') + Fore.RESET)
 
             name = None
+            nonfree_repo_required = False
             for request in splitter.grouped[group]['requests']:
                 request_id = int(request.get('id'))
                 target_package = request.find('./action/target').get('package')
+                target_project = request.find('./action/target').get('project')
+                if self.api.cnonfree and self.api.cnonfree == target_project:
+                    nonfree_repo_required = True
                 line = '- {} {}{:<30}{}'.format(request_id, Fore.CYAN, target_package, Fore.RESET)
 
                 message = self.api.ignore_format(request_id)
@@ -108,7 +112,7 @@ class AdiCommand:
                 if name is None:
                     use_frozenlinks = group in source_projects_expand and not split
                     name = self.api.create_adi_project(None,
-                            use_frozenlinks, group)
+                            use_frozenlinks, group, nonfree_repo_required)
 
                 if not self.api.rq_to_prj(request_id, name):
                     return False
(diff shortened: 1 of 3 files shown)


osclib/adi_command.py
```diff
@@ -87,9 +87,13 @@ def create_new_adi(self, wanted_requests, by_dp=False, split=False):
             print(Fore.YELLOW + (group if group != '' else 'wanted') + Fore.RESET)
 
             name = None
+            nonfree_repo_required = False
```
jberry-suse: Why have a loop variable when it is simply:

```python
nonfree = target_project == self.api.cnonfree
```

As for variable name, same suggestion as on `create_adi_project()`.

osclib/adi_command.py
```diff
@@ -87,9 +87,13 @@ def create_new_adi(self, wanted_requests, by_dp=False, split=False):
...
             for request in splitter.grouped[group]['requests']:
                 request_id = int(request.get('id'))
                 target_package = request.find('./action/target').get('package')
+                target_project = request.find('./action/target').get('project')
+                if self.api.cnonfree and self.api.cnonfree == target_project:
```
jberry-suse: This condition is in both here and in `RequestSplitter`. Should be able to get away with only one rather than duplicate check.

osclib/stagingapi.py
```diff
@@ -1757,6 +1757,11 @@ def create_adi_project(self, name, use_frozenlinks=False, src_prj=None):
             linkproject = ''
             repository = '<repository name="standard">'
 
+        if nonfree_repo_required:
+            nonfree_repo = "<path project=\"{}\" repository=\"standard\"/>".format(self.cnonfree)
```
jberry-suse: nit: seems like `nonfree_path` would make more sense given it is a path inside repo not a new repo.

osclib/stagingapi.py
```diff
@@ -1739,7 +1739,7 @@ def update_adi_frozenlinks(self, name, src_prj):
         l = ET.tostring(flink)
         http_PUT(url, data=l)
 
-    def create_adi_project(self, name, use_frozenlinks=False, src_prj=None):
+    def create_adi_project(self, name, use_frozenlinks=False, src_prj=None, nonfree_repo_required=False):
```
jberry-suse: nit: perhaps `nonfree` since the rest is implementation detail the outside does not need to know.

nilxam:
I'm fine move to https://github.com/openSUSE/osc-plugin-factory/pull/1379 , but please do not group request from same devel project together if there is nonfree request.

PR #1382, opened 2018-01-31
build-faul-reminder: Small rewording of the 2nd mail

DimStar77:


DimStar77:
This is just a minor change to make the wording less aggressive; I was notified that somebody felt insulted by the original wording (which is valid feedback)

DimStar77:
Reworded as advised by @sysrich  after discussion

PR #1385, opened 2018-02-01
CreatePackageDescr: ignore debuginfo.build deps for i-i-debuginfodeps…

DimStar77:
…-Kubic

All -debuginfo dependencies are split to other repos, and are not able to be
tested here. Allow repo-checker to skip over those.

### DIFF
--- a/CreatePackageDescr.pm
+++ b/CreatePackageDescr.pm
@@ -127,10 +127,7 @@ sub package_snippet($) {
             && $prv eq "/usr/bin/mplayer" );
         next if ( $prv eq "this-is-only-for-build-envs" );
         next
-          if ( $name eq "installation-images-debuginfodeps"
-            && $prv =~ m/debuginfo.build/ );
-        next
-          if ( $name eq "installation-images-debuginfodeps-openSUSE"
+          if ( $name =~ "^installation-images-debuginfodeps.*"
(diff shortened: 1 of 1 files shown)


DimStar77:
This should allow us to remove one more from the staging/config white list (handling it the same as i-i-debbuginfodeps-openSUSE)

CreatePackageDescr.pm
```diff
@@ -132,6 +132,9 @@ sub package_snippet($) {
         next
           if ( $name eq "installation-images-debuginfodeps-openSUSE"
             && $prv =~ m/debuginfo.build/ );
+        next
+          if ( $name eq "installation-images-debuginfodeps-Kubic"
```
jberry-suse: Perhaps we should switch to a `.*` for that future `-DimStar` target?

Still rather ugly this needs to be done, but better than whitelisting whole package.

DimStar77: .* sounds reasonable to me; I'm sure any future/potential i-i-debuginfodeps-* will have debuginfo.build deps, which we won't be able to handle.. let me change that

jberry-suse:
I wanted to be sure this worked by testing locally without whitelisted, but took a second to realize that the meta cache needed to be cleared each time to rerun this bit. Anyway, after doing that it works as expected.

PR #1386, opened 2018-02-06
pkglistgen: integrate drop list creation into update_and_solve and project family methods

### LINKED ISSUE #1375: pkglistgen: integrate drop list generator
Per [poo#30225: enable drop list generator](https://progress.opensuse.org/issues/30225), it seems there are a few things that need to be reconciled.

- How to the `.solv` files in `trees/` (not in git, but on `packagelists` machine) relate to those in `create-drop-list`? From what I can surmise they represent the pre-published `standard` repo before it enters the _totest_ workflow. Meaning what is available from `https://api.opensuse.org/public/build/` instead of `http://download.opensuse.org/`.
- How are `do_create_droplist` and `do_dump_solv` intended to work together since they expect different file name formats?

  https://github.com/openSUSE/osc-plugin-factory/blob/5a472eb4006ba59ef5487eb3ab764a324fdae7b2/pkglistgen.py#L782-L784

  vs

  https://github.com/openSUSE/osc-plugin-factory/blob/5a472eb4006ba59ef5487eb3ab764a324fdae7b2/pkglistgen.py#L835-L844

  Which is something like:
  - `repo-openSUSE:Leap:15.0-standard-x86_64.solv`
  - `openSUSE-15.0-x86_64-Build109.3.solv`
- How does one generate the `repo-` (pre-released) files using `pkglistgen.py`?
- Is the `obsoletepackages.inc`, generated by `do_create_droplist`, to be placed into `000package-groups` and copied to `000product`? Or would it make sense to only keep in `000product` since it is generated just like the rest of solving process?
- The drop list is only to be done for target project and ports and not for stagings?
- How does one reconcile the `non-oss` and `oss` repos as the [current `package-lists` repo](https://github.com/openSUSE/package-lists/tree/master/create-drop-list/Leap:42.3) only has one `solv` file. Presumably that would mean `do_create_droplist` would also need to be able to handle multiple target projects to include `:NonFree` which it currently does not. Is this expected?
- Should the new code commit to the old repo, `package-lists`, the new `solv` files? It seems like it should be at least noted somewhere that the new code cannot generate `solv` files for < _Leap 15.0_. As such it clearly cannot generate a cache even if desired.

Overall, seems like a lot more to figure out and code than "enable drop list generator" would imply. Also yet another task that would be greatly aided by any level of detail about how all the parts come together to create a workflow. Even just a dump of commands to run in order for an example workflow. Trying to reverse engineer this from the two code-bases is not ideal especially when it appears what remains is not to simply wire up the two subcommands to be called as part of `do_update_and_solve`.

I have a local stab that abstracts the download baseurl and calls the commands, but definitely some details that need answers. @lnussel

jberry-suse:
- dc3b9e15e126e40c71757f544c504dec95ed7051:
    osclib/conf: include Leap for Factory during pkglistgen.

- c56c23f14808880e60bca633acba1428e20fbbdd:
    pkglistgen: include some hints about the difference between solv files.

- b58cb89641a7a23ebae3a5150c2c79efd1333014:
    **pkglistgen: integrate drop list creation into update_and_solve.**
    
    In order to generate a complete drop list the nonfree repository must also
    be cached for both current and published builds. The solv files are then
    merged and used to generate drop list.

- 5e35baac618bac8b106d0afc567360d51a976950:
    pkglistgen: do_dump_solv(): handle old-style product repo format.
    
    This allows for generating solv cache for Leap versions prior to 15.0
    which avoids the need to have them checked into git like previous
    package-lists repository.

- 0590162cc300a71fecd46de24918a4f939720d88:
    pkglistgen: do_create_droplist(): print to file when output_dir available.

- 4d98ff86c42171326108395b1b471784adb4a231:
    pkglistgen: provide do_update_merge() to combine free and nonfree solv.

- 9e77f2ccbccea77ea981e7e950f11a3ddab0e4a7:
    osclib/util: provide project_list_family* and project_version().

- 378470e8c9ece5d4d42754baa8c281ff8303b1a0:
    osclib/stagingapi: get_staging_projects(): utilize project_list_prefix().

- 42e698ec340513b4322cd25447c4b5d4efda032f:
    osclib/core: provide project_list_prefix().

- 5aa70e15466dfb30113adfe03410bc760d85de71:
    osclib/conf: add download-baseurl for Leap and Factory.

- 3f8cf29542e5f0d49378d51acad6e1a1881c347e:
    osclib/conf: support version in project pattern as value replacement.

The product family functions should be usable in `leaper.py` to avoid all the hard-coding and priority orders and projects that has to be updated each cycle.

Rather than checking `solv` files into git like `package-lists` this code creates `solv` files for the relevant products in the family in a local cache. Unless I am missing something there seems to be no reason to keep them in git since they can be re-created. As discussed in #1375 they can be sync'd somewhere for backup if desired (as perhaps _Tumbleweed_ may desire to keep all snapshots just in case). Then again, we can re-create them for at least the last three months from my _Tumbleweed Snapshots_ ;).

Locally, the following exist after running for _Leap 15.0_.

```
~/.cache/opensuse-packagelists/solv/
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.2
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.2/openSUSE-Addon-NonOss-FTP-x86_64-Build0073.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.2/openSUSE-42.2-x86_64-Build0283.merged.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.2/openSUSE-42.2-x86_64-Build0283.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.3
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.3/openSUSE-Addon-NonOss-FTP-x86_64-Build0082.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.3/openSUSE-42.3-x86_64-Build0331.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.3/openSUSE-42.3-x86_64-Build0331.merged.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:15.0
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:15.0/openSUSE-15.0-x86_64-Build115.1.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:15.0/openSUSE-Leap-15.0-Addon-NonOss-FTP-x86_64-Build115.1.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:15.0/openSUSE-15.0-x86_64-Build115.1.merged.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.1
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.1/openSUSE-42.1-x86_64-Build0265.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.1/openSUSE-Addon-NonOss-FTP-i586-x86_64-Build0039.solv
~/.cache/opensuse-packagelists/solv/openSUSE:Leap:42.1/openSUSE-42.1-x86_64-Build0265.merged.solv
```

`mergesolv` seems to create a `solv` file that is smaller than the size of the two source `solv` files. I haven't verified if that makes sense, but performing the same operation as `package-lists` version. I noticed that `steamcmd` is show as removed when it exists in Leap 15.0, but did not investigate further.

Once deployed the `obsoletepackages.inc` in `000package-groups` should be dropped to avoid confusion as this will generate one in `000product` each run.

In general the command line classes should be thin glue code that maps command line arguments to function calls instead of contain the bulk of code. With this and my previous PR to migrate wrapper scripts into `pkglistgen.py` it is ever more clear the misuse of the `ToolBase.CommandLineInterface` class. Since a variety of methods within that class need to be called by higher level wrapper methods they should be separated to allow for all arguments to be passed directly. Having everything in _CLI_ class results in the awquardness of having to rerun `setup_tool()` multiple times to change global arguments (like `repos`) and copy/pass around `opts`, not to mention `subcmd` (ie. `do_thing('thing', ...)`). For example see the extremely small `repo_checker` _CLI_ class.

https://github.com/openSUSE/osc-plugin-factory/blob/cd1fa5db15f01317f9a1c9893380e17d42122fc0/repo_checker.py#L500-L529

Fixes #1375.

### DIFF
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -180,6 +180,8 @@ class Config(object):
                         defaults[k] = v % {'project': project}
                     elif isinstance(v, basestring) and '%(project.lower)s' in v:
                         defaults[k] = v % {'project.lower': project.lower()}
+                    elif isinstance(v, basestring) and '%(version)s' in v:
+                        defaults[k] = v % {'version': match.group('version')}
                     else:
                         defaults[k] = v
                 break
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -52,6 +52,7 @@ DEFAULT = {
         'devel-project-enforce': 'True',
         'review-team': 'opensuse-review-team',
         'repo-checker': 'repo-checker',
+        'pkglistgen-product-family-include': 'openSUSE:Leap:N',
     },
     r'openSUSE:(?P<project>Leap:(?P<version>[\d.]+))': {
         'staging': 'openSUSE:%(project)s:Staging',
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -47,12 +47,13 @@ DEFAULT = {
         'lock-ns': 'openSUSE',
         'delreq-review': 'factory-maintainers',
         'main-repo': 'standard',
+        'download-baseurl': 'http://download.opensuse.org/tumbleweed/',
         # check_source.py
         'devel-project-enforce': 'True',
         'review-team': 'opensuse-review-team',
         'repo-checker': 'repo-checker',
     },
-    r'openSUSE:(?P<project>Leap:[\d.]+)': {
+    r'openSUSE:(?P<project>Leap:(?P<version>[\d.]+))': {
         'staging': 'openSUSE:%(project)s:Staging',
         'staging-group': 'factory-staging',
         'staging-archs': 'i586 x86_64',
@@ -67,6 +68,7 @@ DEFAULT = {
         'lock-ns': 'openSUSE',
         'delreq-review': None,
         'main-repo': 'standard',
+        'download-baseurl': 'http://download.opensuse.org/distribution/leap/%(version)s/',
         # check_source.py
         # review-team optionally added by leaper.py.
         'repo-checker': 'repo-checker',
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from dateutil.parser import parse as date_parse
 import re
 from xml.etree import cElementTree as ET
+from lxml import etree as ETL
 from urllib2 import HTTPError
 
 from osc.core import get_binarylist
@@ -173,3 +174,10 @@ def request_age(request):
         created = request.find('history').get('when')
     created = date_parse(created)
     return datetime.utcnow() - created
+
+def project_list_prefix(apiurl, prefix):
+    """Get a list of project with the same prefix."""
+    query = {'match': 'starts-with(@name, "{}")'.format(prefix)}
+    url = makeurl(apiurl, ['search', 'project', 'id'], query)
+    root = ETL.parse(http_GET(url)).getroot()
+    return root.xpath('project/@name')
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -49,6 +49,7 @@ from osc.core import streamfile
 
 from osclib.cache import Cache
 from osclib.core import devel_project_get
+from osclib.core import project_list_prefix
 from osclib.comments import CommentAPI
 from osclib.ignore_command import IgnoreCommand
 from osclib.memoize import memoize
@@ -333,17 +334,10 @@ class StagingAPI(object):
         :return list of known staging projects
         """
 
-        projects = []
+        projects = project_list_prefix(self.apiurl, self.cstaging + ':')
+        if not include_dvd:
+            projects = filter(lambda p: not p.endswith(':DVD'), projects)
 
-        query = "id?match=starts-with(@name,'{}:')".format(self.cstaging)
-        url = self.makeurl(['search', 'project', query])
-        projxml = http_GET(url)
-        root = ET.parse(projxml).getroot()
-        for val in root.findall('project'):
-            project = val.get('name')
-            if not include_dvd and project.endswith(':DVD'):
-                continue
-            projects.append(project)
         return projects
 
     def extract_staging_short(self, p):
--- a/osclib/util.py
+++ b/osclib/util.py
@@ -1,0 +1,78 @@
+from osclib.core import project_list_prefix
+
+
+def project_list_family(apiurl, project):
+    """
+    Determine the available projects within the same product family.
+
+    Skips < SLE-12 due to format change.
+    """
+    if project == 'openSUSE:Factory':
+        return [project]
+
+    count_original = project.count(':')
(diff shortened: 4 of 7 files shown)


jberry-suse:
This should [handle _Factory_ worklfow to include _Leap_ family for obsoletes drop list](https://github.com/openSUSE/osc-plugin-factory/pull/1386/files#diff-a9bd1d18e9e8e0198fe2f85c4e97bb38R55).

```python
'pkglistgen-product-family-include': 'openSUSE:Leap:N',
```

jberry-suse:
The brittleness of tests that re-create _OBS_ was shown again as it did not handle a properly encoded query string that my refactoring generated. Fixed that and a few minor `flake8` notes.

jberry-suse:
Created #1387 as a reminder to utilize the `project_list_family*()` methods.

jberry-suse:
Corrected edge-case with staging scope.

nilxam:
Hmm... AFAICS the update in 000product/000release-packages what removed old `obsoletepackage`  against the package were obsoleted in the version older than 42.1 ie. 13.2,etc., I'll ignore that dup failure on openqa for now, but we have to decide should we keep support those version upgrade case in Leap 15.0, and just in case we still have 13.2 dup test on openqa for TW and Leap 15.

ref. https://openqa.opensuse.org/tests/604949#step/zdup/144

@lnussel

DimStar77:
TW tests back to 13.1; 12.x was dropped as maintenance was getting too tedious (not from a product pov, but the disks were getting too small for the ever growing updates, and it was not worthy to produce new base disks)

For TW, we still test upgrades from 13.1, 13.2, 42.[123]

nilxam:
just noticed the discussion on https://github.com/openSUSE/osc-plugin-factory/issues/1375#issuecomment-361883365 , seems we can remove 13.x test from leap 15 openqa group.

jberry-suse:
Yeah, the current code only includes `42.{1,2,3}` and `15.0` pre-releases. If that needs to be different let me know.

lnussel:
If we only knew how much 13.2 users are left ... I wanted to drop support for upgrading from 13.2 already with 42.3 but the access statistics of download.o.o made me reconsider that.
In general I'm fine with limiting the upgrade calculations and tests to 42.* for Leap 15.0.

PR #1390, opened 2018-02-07
osc-staging: select: add filter-by example using hashtag in description.

jberry-suse:
Example based on my response to email from Alexander regarding CaaSP staging workflow.


jberry-suse:
@lnussel since you were mentioned and perhaps curious as to result.

jberry-suse:
Added some related documentation changes.

PR #1391, opened 2018-02-08
Fixes for openSUSE:*ARM rebuilding

dirkmueller:


jberry-suse:
Travis failure tracked in openSUSE/open-build-service#4462

pkglistgen.py
```diff
@@ -1120,6 +1121,9 @@ def update_and_solve_target(self, apiurl, target_project, target_config, main_re
         product_dir = os.path.join(cache_dir, product)
         release_dir = os.path.join(cache_dir, release)
 
+        conf_no_colon = conf.config['checkout_no_colon']
+        conf.config['checkout_no_colon'] = False
```
jberry-suse: Is this not the [default](https://github.com/openSUSE/osc/blob/5e3fe8ba08d7625140f06441c804e82790ea7db0/osc/conf.py#L149)? Are you explicitly setting to override local config or what?

pkglistgen.py
```diff
@@ -1120,6 +1121,9 @@ def update_and_solve_target(self, apiurl, target_project, target_config, main_re
         product_dir = os.path.join(cache_dir, product)
         release_dir = os.path.join(cache_dir, release)
 
+        conf_no_colon = conf.config['checkout_no_colon']
```
jberry-suse: `conf_no_colon` never used.

pkglistgen.py
jberry-suse: Considering in a commit title "Skip automatic rebuild for more pattern packages" assuming this was not intentional?

dirkmueller: sorry, this was unrelated. removed.

PR #1399, opened 2018-02-12
Support unneeded.yml as list of regular expressions

coolo:
UNWANTED is a bit heavy weapon to filter unsorted.yml, so leave
its role to packages we sure don't want (e.g. -mini and -bootstrap
packages). Packages that are there but would end in unsorted otherwise,
filter by means of package regular expressions in unneeded.yml

The format of that file is yml with a list of groups of regular
expressions.

```
global:
  - .*-32bit

texlive:
  - texlive.*

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -92,7 +92,7 @@ class Group(object):
 
         self.comment = ' ### AUTOMATICALLY GENERATED, DO NOT EDIT ### '
         self.srcpkgs = None
-        self.develpkgs = []
+        self.develpkgs = dict()
         self.silents = set()
         self.ignored = set()
         # special feature for SLE. Patterns are marked for expansion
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: pkglistgen.py [hunk 1]
```diff
@@ -639,8 +627,25 @@ def _collect_unsorted_packages(self, modules):
...
+                    reason = self._find_reason(p, modules)
+                    if reason:
+                        fh.write(' # ' + reason)
                 fh.write(" \n")
 
+    # give a hint if the package is related to a group
+    def _find_reason(self, package, modules):
+        # go through the modules multiple times to find the "best"
+        for g in modules:
+           if package in g.recommends:
```
jberry-suse: From `flake8` run:

```
./pkglistgen.py:639:12: E111 indentation is not a multiple of four
./pkglistgen.py:640:16: E111 indentation is not a multiple of four
./pkglistgen.py:642:12: E111 indentation is not a multiple of four
./pkglistgen.py:643:16: E111 indentation is not a multiple of four
./pkglistgen.py:645:12: E111 indentation is not a multiple of four
./pkglistgen.py:646:16: E111 indentation is not a multiple of four
```

coolo:
fixed the pep

PR #1404, opened 2018-02-13
pkglistgen: Treat stagings differently

coolo:
This is specific to SLE stagings, but won't hurt in openSUSE stagings
as the products there are hand made anyway

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -1156,6 +1156,7 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
         self.copy_directory_contents(group_dir, product_dir,
                                      ['supportstatus.txt', 'groups.yml', 'package-groups.changes'])
         self.change_extension(product_dir, '.spec.in', '.spec')
+        self.change_extension(product_dir, '.product.in', '.product')
 
         self.options.input_dir = group_dir
         self.options.output_dir = product_dir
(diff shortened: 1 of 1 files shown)


coolo:
these are the current changes I run

pkglistgen.py
```diff
@@ -70,7 +70,7 @@
 SCRIPT_PATH = os.path.dirname(os.path.realpath(__file__))
 ARCHITECTURES = ['x86_64', 'ppc64le', 's390x', 'aarch64']
 DEFAULT_REPOS = ("openSUSE:Factory/standard")
-PRODUCT_SERVICE = '/usr/lib/obs/service/create_single_product'
+PRODUCT_SERVICE = '/home/leaper/obs-service-product_converter/create_single_product'
```
jberry-suse: This would seem like a problem.

jberry-suse:
Is the drop list desired for SLE? If so merely configure the `download-baseurl` in `conf.py` to point to something the code can see when running.

pkglistgen.py
coolo: Just as your code is a problem for me - fair treat, no? :)

coolo:
In general I would say yes, but I would need to experiment with it to see the consequences

pkglistgen.py
jberry-suse: A joke I presume.

coolo:
I removed the constant for now - to reduce my diff to master

PR #1412, opened 2018-02-20
Tool to debug dependencies

lnussel:


deptool.py
```diff
@@ -0,0 +1,249 @@
...
+        for r in relation:
+            i = self.pool.str2id(r)
+            for s in self.pool.whatprovides(i):
+                print('- {}-{}@{}:'.format(s.name, s.evr, s.arch))
+
+if __name__ == "__main__":
+    app = DepTool()
+    sys.exit(app.main())
+
+# vim: sw=4 et
```
jberry-suse: :(

deptool.py
```diff
@@ -0,0 +1,249 @@
...
+        level = None
+        if self.options.debug:
+            level = logging.DEBUG
+        elif self.options.verbose:
+            level = logging.INFO
+
+        logging.basicConfig(level=level)
+
+        if level:
+            logger.setLevel(level)  # wtf?
```
jberry-suse: Presumably this is due to calling `getLogger()` before `basicConfig()` is invoked. For example `ReviewBot.py`:

https://github.com/openSUSE/osc-plugin-factory/blob/945bf4f58ff7c66e006cb5d756e252a59bc2ede8/ReviewBot.py#L559-L560

deptool.py
```diff
@@ -0,0 +1,249 @@
...
+from fnmatch import fnmatch
+from ConfigParser import SafeConfigParser
+import solv
+import rpm
+
+logger = logging.getLogger()
+
+REASONS = [i[14:] for i in dir(solv.Solver) if i.startswith('SOLVER_REASON_')]
+
+class DepTool(cmdln.Cmdln):
```
jberry-suse: At some point we really should stop introducing more non-osc plugin tools that use the deprecated `cmdln` in favor of `argparse`.

jberry-suse:
Flake8 also has some correct notes.

deptool.py
lnussel: yeah, the cmdln interface with the decorators is quite nice though

deptool.py
lnussel: it's my boilerplate. sorry for that :) didn't bother cleaning it up too much before showing the code to dimstar

deptool.py
lnussel: IIRC calling getLogger on module level is the recommended use. The way ReviewBot does it with self.logger is wrong, just me not knowing better and  not bothering to fix it after all.

Maybe the script needs to not initialize the global logger if __name__ == '__main__' so it can be used as module or program.

jberry-suse: That would be another way.

lnussel:
travis fail due to ring2 removal :(

PR #1413, opened 2018-02-21
Revert "staging: add exceptions for new 000* product files"

coolo:
This reverts commit 239c6314c7a8b0192c85181e7f64ab274e74cc31.

The whole purpose of freeze is not to have target changes affect the stagings. We didn't notice yet in Factory as we don't have the product builder there - but in SLE this is an ongoing problem

### DIFF
--- a/osclib/freeze_command.py
+++ b/osclib/freeze_command.py
@@ -257,8 +257,6 @@ class FreezeCommand(object):
         # If the package is an internal one (e.g _product)
         if package.startswith('_'):
             return None
-        if package.startswith('000'):
-            return None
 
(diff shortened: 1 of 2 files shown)


nilxam:
but it leads 000* package has defined in the two different ring error

nilxam:
maybe https://github.com/openSUSE/osc-plugin-factory/pull/1416 is something we needs to avoid list command fails to work.

jberry-suse:
The 'ol insta merge. Presumably staging-bot and other tools won't be effect since they'll still have the ring bit to ignore.

PR #1416, opened 2018-02-22
Do not complain 000* package has defined in the two different ring

nilxam:


coolo:
makes sense - just as we hardcoded Test DVD there

osclib/stagingapi.py
```diff
@@ -200,7 +200,7 @@ def _generate_ring_packages(self, checklinks=False):
                 # XXX TODO - Test-DVD-x86_64 is hardcoded here
                 if pkg in ret and not pkg.startswith('Test-DVD-'):
                     msg = '{} is defined in two projects ({} and {})'
-                    if checklinks and pkg in except_pkgs and prj == except_pkgs[pkg]:
+                    if pkg.startswith('000') or (checklinks and pkg in except_pkgs and prj == except_pkgs[pkg]):
```
jberry-suse: Would seem best to put this check together with the other exception.

PR #1424, opened 2018-03-01
fix a typo that currently breaks token authentication

kraih:
The `Authorization` value needs to be added to the `bot` object, not `self`.


kraih:
Yes... i made the typo originally, and just noticed it trying to deploy token authentication. 😞

PR #1425, opened 2018-03-02
[Very WIP] Automated release of TW docker images

Vogtinator:
This is just a very early dump of what I have, so that you can already complain about this hacky approach ;-)

Basically, to release the docker images, it:

* Gets the URL of openQA's repo mirror
* Parses the repo data to get the URLs of the packages
* Downloads the packages
* Extracts the root file system archive from the docker image contained within those
* Copies that file system archive into a git checkout (specified branch and filename)
* And pushes it!

The last two parts are not implemented yet, neither is any kind of error checking

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -10,13 +10,16 @@
 
 import cmdln
 import datetime
+import glob
 import json
 import os
 import re
 import sys
 import urllib2
 import logging
+import shutil
 import signal
+import subprocess
 import tempfile
 import time
 import zlib
@@ -490,9 +493,26 @@ class ToTestBase(object):
         else:
             self.api.retried_POST(url)
 
-    def _release_docker_image(self, name, snapshot, repo_url, primary_tree):
+    def _release_docker_image(self, pkg, snapshot, repo_url, primary_tree):
         """The docker image is wrapped in an RPM in the ToTest repo"""
-        print self._download_snapshot_rpm(name, repo_url, primary_tree)
+
+        # Download the RPM to a temporary location
+        rpm_path = self._download_snapshot_rpm(pkg['pkg'], repo_url, primary_tree)
+        # Create a temporary dir for extraction
+        tmp_path = tempfile.mkdtemp()
+        # Extract the .tar.xz inside the RPM into the dir
+        subprocess.call("rpm2cpio '%s' | cpio -i --to-stdout \*.tar.xz | tar -xJf - -C '%s'" % (rpm_path, tmp_path), shell=True)
+        # Remove the RPM file again
+        os.unlink(rpm_path)
+        # Parse the manifest to get the name of the root tar.xz
+        manifest_path = glob.glob(tmp_path + '/manifest.json')[0]
+        manifest = json.loads(open(manifest_path).read())
+        root_path = tmp_path + '/' + manifest[0]['Layers'][0]
+        # TODO: Copy it into the repo and push.
+        print root_path
+        # And remove the temporary directory
+        shutil.rmtree(tmp_path)
+
 
     def _release(self, snapshot, set_release=None):
         for product in self.ftp_products:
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -446,13 +525,19 @@ def _release(self, set_release=None):
...
+                self._release_docker_image(docker_pkg, snapshot, repo_url, primary_tree)
+
     def update_totest(self, snapshot=None):
         release = 'Snapshot%s' % snapshot if snapshot else None
         logger.info('Updating snapshot %s' % snapshot)
         if not self.dryrun:
             self.api.switch_flag_in_prj(self.test_project, flag='publish', state='disable')
-
-        self._release(set_release=release)
+        snapshot = "20180228"
```
DimStar77: And what is tomorrow?

totest-manager.py
```diff
@@ -446,13 +525,19 @@ def _release(self, set_release=None):
         for cd in self.main_products:
             self._release_package(self.project, cd, set_release=set_release)
 
+        if self.docker_packages:
+            repo_url = self._get_repo_mirror(snapshot) + "/suse"
```
DimStar77: This will break as soon as we switch to the new product builder

DimStar77:
Generic issue: 'release' is the point in time where things move from openSUSE:Factory/standard to openSUSE:Factory:ToTest - BEFORE things are in openQA and tested; this is not the point where you want to upload stuff into the hub (and certainly can't download files from openQA yet, as it is not even synced yet)

totest-manager.py
Vogtinator: I know - I'll probably introduce autodetection for that instead.

Vogtinator:
> Generic issue: 'release' is the point in time where things move from openSUSE:Factory/standard to openSUSE:Factory:ToTest - BEFORE things are in openQA and tested;

That's what I was afraid of - it's nowhere documented in the code nor visible in the --verbose output...

> this is not the point where you want to upload stuff into the hub (and certainly can't download files from openQA yet, as it is not even synced yet)

Where does it fit?

I plan on moving the actual rpm extraction and git push into a shell script, so moving the caller around isn't an issue. I'd prefer a cleaner solution than hooking the announcer though.

totest-manager.py
```diff
@@ -435,7 +493,28 @@ def _release_package(self, project, package, set_release=None):
         else:
             self.api.retried_POST(url)
 
-    def _release(self, set_release=None):
+    def _release_docker_image(self, pkg, snapshot, repo_url, primary_tree):
+        """The docker image is wrapped in an RPM in the ToTest repo"""
+
+        # Download the RPM to a temporary location
+        rpm_path = self._download_snapshot_rpm(pkg['pkg'], repo_url, primary_tree)
+        # Create a temporary dir for extraction
```
jberry-suse: nit: a summary comment about the overall process makes sense, but commenting every line would seem to indicate either the code needs to be refactored or unnecessary.

totest-manager.py
```diff
@@ -222,6 +230,56 @@ def update_openqa_status_message(self):
...
+
+    def _download_file(self, url):
+        try:
+            req = urllib2.Request(url=url)
+            f = urllib2.urlopen(req)
+            return f.read()
+        except urllib2.HTTPError:
+            return None
+
+    def _get_repo_primary(self, baseurl):
```
jberry-suse: We now have code like this in at least 2 places, presumably this was copied or based on one. Seems like a candidate for `osclib.util`.

Vogtinator:
I discussed this with lnussel and it seems to be the best option to put all of this into a different .py file (Python 3, yay!) and run it (until there's proper supporting infrastructure) as a separate bot.

As the next PR will be vastly different I'll close this one for now.

totest-manager.py
Vogtinator: I'll use `requests` instead, that's a one line call as well.

Issue #1181, opened 2017-10-10
check_source: replace factory-auto group reviewer on openSUSE:Factory with the user to be consistent with other review bots


jberry-suse:
Seems to be no reason for this, at least any more. All other bots review things assigned to them rather than a group. This requires `--group factory-auto` in service file rather than using `osc` user from `.oscrc` and thus will require special handling for SLE unless changed to user. This being relevant since the bot will be used on SLE and ends up adding `repo-checker`.

jberry-suse:
@DimStar77 @coolo @nilxam Thoughts on this?

jberry-suse:
Relevant to #1043.

DimStar77:
the reason there is/was a group around it was so that factory-maintainers could be added to the group if there was need to reopen/override on bots decisions; currently I do those actions directly on the pkglist VM as the respective user - and @coolo is the only user in for example factory-auto group (for historical reasons)

jberry-suse:
Makes sense. All the other bots are just users and overrides are done directly in the same way you do them. Is this something we want to change to be groups? My understanding given the more recent bots and majority are just user is that was the preference. Since they are all for review bots there seems to be do reason for them to be different.

Before we introduce this bot in SLE and Leap 15 it would seem best to decide if a reason to keep it this way.

DimStar77:
Let's list pro/cons

For Groups:
Pro: we can allow users to override a bot by adding the user to the group
Pro: The override does not seem to come from the bot, but actually visibly from a respective user
Pro: We could potentially allow this to users who do not have access to Pkglistvm/maintVM without sharnig credentials
Con: AFAIK, it is only possibly to create new groups by OBS Admins; administration of the group is then possible by 'group maintainers' though

For user:
Pro: A user is anyway needed in the end, to perform the action the group was assigned for
Pro: less administration, as we do not need user AND group

jberry-suse:
Sounds great, so lets set them all up that way? My primary point here is consistency. All the review bots act in the same way and should setup the same way.

jberry-suse:
I would suggest we take https://github.com/openSUSE/osc-plugin-factory/issues/1400#issuecomment-365408620 comment as a final decision on this and make all bots _users_ and drop the group.

lnussel:
with the bot control via comments I think the main reason for having groups in the first place is no longer valid so fine with me to remove the indirection level of the group.

jberry-suse:
Yeah!

PR #1427, opened 2018-03-05
Define comment commands and provide for ReviewBot review overrides and repo-checker skip-cycle.

### LINKED ISSUE #1365: repo_checker: consider adding staging config option to skip cycle check
It would seem to make sense to allow cycles to be _accepted_ at a staging level and thus allow the `repo-checker` to accept such reviews without having to do it manually or run `repo-checker` with the `--skip-cycle` flag. Perhaps `repo_checker-cycle-skip` or `repo_checker-cycle-accept`?

The usage would then be:

```
osc staging config A repo_checker-cycle-skip 1
```

### LINKED ISSUE #1400: ReviewBot: allow review overrides via comments by staging masters
via check_source_in_factory leaper has hardcoded logic to pass requests if they are only left waiting for openSUSE:Factory:Staging:*. This should be configurable to be able to also skip for e.g. repochecker.

jberry-suse:
- 68b93e1dbd4fbb395bb1ea4cf6aec35a45f4babb:
    repo_checker: provide comment command to skip-cycle for group.
    
    For example (on staging project):
    
      @repo-checker skip-cycle

- 4f4e6da7efc48fc43dd63138081b61d9df01c960:
    repo_checker: utilize ReviewBot.comment_api instead of new instance.

- e775d440494d39291b4ded228abb6ed3555463d3:
    ReviewBot: provide comment command override.
    
    For example:
    
      @leaper override accept
      @repo-checker override decline

- 3674191a41f4d6f82c530eee14ac85ce86150e3a:
    osclib/core: maintainers_get(): utilize new group_members() function.

- e2efbf55a90b98755f9e889bf501069ca5405ad1:
    osclib/core: provide group_members() function.

- eb19b827aa1b82dcf1880c98d761c5caed6ee2c7:
    osclib/comments: provide command_find() for comment commands.

Fixes #1400.
Fixes #1365.

Should partial work towards #1181, but would likely need to scan declined reviews to see if re-open or change workflow for bots like `factory-auto`.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -28,6 +28,7 @@ from collections import namedtuple
 from collections import OrderedDict
 from osclib.comments import CommentAPI
 from osclib.conf import Config
+from osclib.core import group_members
 from osclib.memoize import memoize
 from osclib.stagingapi import StagingAPI
 import signal
@@ -84,6 +85,8 @@ class ReviewBot(object):
         self.only_one_action = False
         self.request_default_return = None
         self.comment_handler = False
+        self.override_allow = True
+        self.override_group_key = '{}-override-group'.format(self.bot_name.lower())
 
         self.load_config()
 
@@ -141,7 +144,12 @@ class ReviewBot(object):
         for req in self.requests:
             self.logger.info("checking %s"%req.reqid)
             self.request = req
-            good = self.check_one_request(req)
+
+            override = self.request_override_check(req)
+            if override is not None:
+                good = override
+            else:
+                good = self.check_one_request(req)
 
             if self.review_mode == 'no':
                 good = None
@@ -154,6 +162,40 @@ class ReviewBot(object):
                 self._set_review(req, 'accepted')
             elif self.review_mode != 'accept-onpass':
                 self._set_review(req, 'declined')
+
+    @memoize(session=True)
+    def request_override_check_users(self, project):
+        """Determine users allowed to override review in a comment command."""
+        self.staging_api(project)
+        config = self.staging_config[project]
+
+        users = group_members(self.apiurl, config['staging-group'])
+
+        if self.override_group_key:
+            override_group = config.get(self.override_group_key)
+            if override_group:
+                users += group_members(self.apiurl, override_group)
+
+        return users
+
+    def request_override_check(self, request):
+        """Check for a comment command requesting review override."""
+        if not self.override_allow:
+            return None
+
+        comments = self.comment_api.get_comments(request_id=request.reqid)
+        users = self.request_override_check_users(request.actions[0].tgt_project)
+        for args, who in self.comment_api.command_find(
+            comments, self.review_user, 'override', users):
+            message = 'overridden by {}'.format(who)
+            override = args[1] or None
+            if override == 'accept':
+                self.review_messages['accepted'] = message
+                return True
+
+            if override == 'decline':
+                self.review_messages['declined'] = message
+                return False
 
     def _set_review(self, req, state):
         doit = self.can_accept_review(req.reqid)
--- a/legal-auto.py
+++ b/legal-auto.py
@@ -59,6 +59,7 @@ class LegalAuto(ReviewBot.ReviewBot):
             self.apinick = 'ibs#'
         else:
             self.apinick = 'obs#'
+        self.override_allow = False # Handled via external tool.
 
     def request_priority(self):
         prio = self.request.priority or 'moderate'
--- a/osclib/comments.py
+++ b/osclib/comments.py
@@ -112,6 +112,33 @@ class CommentAPI(object):
                 return c, info
         return None, None
 
+    def command_find(self, comments, user, command=None, who_allowed=None):
+        """
+        Find comment commands with the optional conditions.
+
+        Usage (in comment):
+            @<user> <command> [args...]
+        """
+        command_re = re.compile(r'^@(?P<user>[^ ]+) (?P<args>.*)$')
+
+        # Search for commands in the order the comment was created.
+        for comment in sorted(comments.values(), key=lambda c: c['when']):
+            if who_allowed and comment['who'] not in who_allowed:
+                continue
+
+            match = command_re.search(comment['comment'])
+            if not match:
+                continue
+
+            if match.group('user') != user:
+                continue
+
+            args = match.group('args').strip().split(' ')
+            if command and (args[0] or None) != command:
+                continue
+
+            yield args, comment['who']
+
     def add_marker(self, comment, bot, info=None):
         """Add bot marker to comment that can be used to find comment."""
 
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -67,6 +67,7 @@ DEFAULT = {
         'openqa': 'https://openqa.opensuse.org',
         'lock': 'openSUSE:%(project)s:Staging',
         'lock-ns': 'openSUSE',
+        'leaper-override-group': 'leap-reviewers',
         'delreq-review': None,
         'main-repo': 'standard',
         'download-baseurl': 'http://download.opensuse.org/distribution/leap/%(version)s/',
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -20,6 +20,15 @@ BINARY_REGEX = r'(?:.*::)?(?P<filename>(?P<name>.*?)-(?P<version>[^-]+)-(?P<rele
 RPM_REGEX = BINARY_REGEX + '\.rpm'
 BinaryParsed = namedtuple('BinaryParsed', ('package', 'filename', 'name', 'arch'))
 
+@memoize(session=True)
+def group_members(apiurl, group, maintainers=False):
+    url = makeurl(apiurl, ['group', group])
+    root = ETL.parse(http_GET(url)).getroot()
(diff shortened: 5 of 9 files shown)


jberry-suse:
This allows members of `staging-group` or `tool-maintainer-group`, but perhaps the later is note desired

jberry-suse:
The comment for `skip-cycle` is to be placed on staging project just like `repo-checker` comment. The individual request overrides on the requests themselves (for any bot).

jberry-suse:
Resolved test failures, by disabling `override_allow` during test.

jberry-suse:
The comment can be placed within a larger comment as long as the command is on it's own line.


Example:

```
Leaper is confused by XYZ and we should really do this differently, but accept for now.

$ osrt leaper override accept
```

lnussel:
Jimmy Berry wrote:
> This allows members of `staging-group` or `tool-maintainer-group`, but perhaps the later is note desired

What's the use case of the latter?

Would it make sense to be able to define a group per bot instead? Ie
leaper overrides are policy decisions (release management) whereas
repochecker is purely technical (release engineering).

cu
Ludwig

-- 
  (o_   Ludwig Nussel
  //\
  V_/_  http://www.suse.com/
SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard,
Graham Norton, HRB 21284 (AG Nürnberg)

lnussel:
Jimmy Berry wrote:
> The comment can be placed within a larger comment as long as the command is on it's own line.
> 
> 
> Example:
> 
> ```
> Leaper is confused by XYZ and we should really do this differently, but accept for now.
> 
> $ osrt leaper override accept
> ```

What about using "@leaper override accept" instead? The shell style
is geeky for sure, if we were talking to a human one would use the @
style to address the other person though. So I think that would be
more natural and requires less syntax to remember.

cu
Ludwig

-- 
  (o_   Ludwig Nussel
  //\
  V_/_  http://www.suse.com/
SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard,
Graham Norton, HRB 21284 (AG Nürnberg)

jberry-suse:
> What's the use case of the latter?

Having a technical group familiar with bot issues able to override was the idea.

> Would it make sense to be able to define a group per bot instead? Ie
> leaper overrides are policy decisions (release management) whereas
> repochecker is purely technical (release engineering).

Perhaps, although I imagine we have less than half a dozen potential unique members of all groups. I am not sure if the ideal scenario is for SLE differs or the same setup. Presumably they do not have the tools group at all currently.

> What about using "@leaper override accept" instead? The shell style
> is geeky for sure, if we were talking to a human one would use the @
> style to address the other person though. So I think that would be
> more natural and requires less syntax to remember.

I went back and forth on a number of styles. The original thinking was having a generic syntax that wasn't tide to bots, but obviously that is our first usage. Currently the code uses the `bot_name` which is based off the class to determine if commands for that bot. This works in the case where the bot is run as a different user (like locally) although that may or may not always be desired. I suppose at run-time the bot does know which user it is running as and could look for references to itself.

The idea behind the non-bot targeted commands was thinking about staging instructions or similar that should effect anyone running `osc staging` commands, but perhaps that should be done differently if ever.

jberry-suse:
Reworked:

- provide `<botname>-override-group` config key in addition to the `staging-group`. For example `leaper-override-group` is set to `leap-reviewers`.
- changed command syntax to `@user ...`
- dropped `release-tools-maintainers` group from overrides

```diff
diff --git a/ReviewBot.py b/ReviewBot.py
index 3e6be99..c515557 100644
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -86,6 +86,7 @@ class ReviewBot(object):
         self.request_default_return = None
         self.comment_handler = False
         self.override_allow = True
+        self.override_group_key = '{}-override-group'.format(self.bot_name.lower())
 
         self.load_config()
 
@@ -170,9 +171,10 @@ class ReviewBot(object):
 
         users = group_members(self.apiurl, config['staging-group'])
 
-        tool_group = config.get('tool-maintainer-group')
-        if tool_group:
-            users += group_members(self.apiurl, tool_group, maintainers=True)
+        if self.override_group_key:
+            override_group = config.get(self.override_group_key)
+            if override_group:
+                users += group_members(self.apiurl, override_group)
 
         return users
 
@@ -184,13 +186,15 @@ class ReviewBot(object):
         comments = self.comment_api.get_comments(request_id=request.reqid)
         users = self.request_override_check_users(request.actions[0].tgt_project)
         for args, who in self.comment_api.command_find(
-            comments, self.bot_name.lower(), 'override', users):
-            if args[2] == 'accept':
-                self.review_messages['accepted'] = 'overridden by {}'.format(who)
+            comments, self.review_user, 'override', users):
+            message = 'overridden by {}'.format(who)
+            override = args[1] or None
+            if override == 'accept':
+                self.review_messages['accepted'] = message
                 return True
 
-            if args[2] == 'decline':
-                self.review_messages['declined'] = 'overridden by {}'.format(who)
+            if override == 'decline':
+                self.review_messages['declined'] = message
                 return False
 
     def _set_review(self, req, state):
diff --git a/osclib/comments.py b/osclib/comments.py
index 0cede21..2d787c9 100644
--- a/osclib/comments.py
+++ b/osclib/comments.py
@@ -23,8 +23,6 @@ from osc.core import http_GET
 from osc.core import http_POST
 from osc.core import makeurl
 
-COMMAND_PREFIX = 'osrt'
-
 
 class CommentAPI(object):
     COMMENT_MARKER_REGEX = re.compile(r'<!-- (?P<bot>[^ ]+)(?P<info>(?: [^= ]+=[^ ]+)*) -->')
@@ -114,16 +112,17 @@ class CommentAPI(object):
                 return c, info
         return None, None
 
-    def command_find(self, comments, bot=None, command=None, who_allowed=None):
+    def command_find(self, comments, user, command=None, who_allowed=None):
         """
         Find comment commands with the optional conditions.
 
         Usage (in comment):
-            $ osrt <bot> <command> <args...>
+            @<user> <command> [args...]
         """
-        command_re = re.compile(r'^\$\s*{}(?P<args>.*)$'.format(COMMAND_PREFIX))
+        command_re = re.compile(r'^@{} (?P<args>.*)$'.format(user))
 
-        for comment in comments.values():
+        # Search for commands in the order the comment was created.
+        for comment in sorted(comments.values(), key=lambda c: c['when']):
             if who_allowed and comment['who'] not in who_allowed:
                 continue
 
@@ -132,10 +131,7 @@ class CommentAPI(object):
                 continue
 
             args = match.group('args').strip().split(' ')
-            if bot and bot != args[0].replace('-', ''):
-                continue
-
-            if command and command != args[1]:
+            if command and command != (args[0] or None):
                 continue
 
             yield args, comment['who']
diff --git a/osclib/conf.py b/osclib/conf.py
index 10b84b3..015bdf8 100644
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -45,7 +45,6 @@ DEFAULT = {
         'openqa': 'https://openqa.opensuse.org',
         'lock': 'openSUSE:%(project)s:Staging',
         'lock-ns': 'openSUSE',
-        'tool-maintainer-group': 'release-tools-maintainers',
         'delreq-review': 'factory-maintainers',
         'main-repo': 'standard',
         'download-baseurl': 'http://download.opensuse.org/tumbleweed/',
@@ -68,7 +67,7 @@ DEFAULT = {
         'openqa': 'https://openqa.opensuse.org',
         'lock': 'openSUSE:%(project)s:Staging',
         'lock-ns': 'openSUSE',
-        'tool-maintainer-group': 'release-tools-maintainers',
+        'leaper-override-group': 'leap-reviewers',
         'delreq-review': None,
         'main-repo': 'standard',
         'download-baseurl': 'http://download.opensuse.org/distribution/leap/%(version)s/',
diff --git a/repo_checker.py b/repo_checker.py
index 00e341b..5137d01 100755
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -190,7 +190,7 @@ class RepoChecker(ReviewBot.ReviewBot):
                 # Look for skip-cycle comment command.
                 users = self.request_override_check_users(request.actions[0].tgt_project)
                 for _, who in self.comment_api.command_find(
-                    comments, self.bot_name.lower(), 'skip-cycle', users):
+                    comments, self.review_user, 'skip-cycle', users):
                     self.logger.debug('comment command: skip-cycle by {}'.format(who))
                     self.groups_skip_cycle.append(group)
                     break
```

jberry-suse:
In order to pickup override comments locally one must set `--user` appropriately or `.oscrc`.

jberry-suse:
Tweaked regex in relation to user.

jberry-suse:
To be clear this is somewhat inconsistent although it is likely more straight-forward from a usage standpoint. Comments (reports) are discovered based on `bot_name` which is derived from class (ie per tool) and comment commands are discovered based on user running the tool. This means depending on invocation details one can get rather different results.

Changing the comment discovery to user would be consistent, but would break things like `osc staging select` comments which should be discovered regardless of user running the command.

Overall, we can keep it this way, but it may come bite us at some point.

lnussel:
so maybe comments need to be seen as group responsibility whereas command address individuals. In any case cool. Let's bring it live :)

PR #1435, opened 2018-03-08
create_test_Factory: No longer block kdelibs4-branding-upstream

DimStar77:
kdelibs4-branding-openSUSE has been dropped, since there is nothing useful
left in it, and kdebase4 is slowly dying.


DimStar77:
This is currently already deployed on pkglist VM, as it was needed for Staging:J, where the new branding package is included

PR #1436, opened 2018-03-08
leaper: fix can_accept_review for combined user/group usage

lnussel:


lnussel:
hotfix deployed on maintenance

PR #1438, opened 2018-03-08
Introduce a new docker-publisher bot

Vogtinator:
This bot runs every 15 minutes and compares the version of Leap 42.3, 15.0 and TW containers with the ones on the registry and pushes, if necessary.

### DIFF
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -113,6 +113,18 @@ BuildArch:      noarch
 %description check-source
 Check source review bot that performs basic source analysis and assigns reviews.
 
+%package docker-publisher
+Summary:        Docker image publishing bot
+Group:          Development/Tools/Other
+BuildArch:      noarch
+Requires:       python3-requests
+Requires:       python3-lxml
+Requires(pre):  shadow
+
+%description docker-publisher
+A docker image publishing bot which regularly pushes built docker images from
+several sources (Repo, URL) to several destinations (git, Docker registries)
+
 %package maintenance
 Summary:        Maintenance related services
 Group:          Development/Tools/Other
@@ -301,6 +313,14 @@ exit 0
 %postun check-source
 %{systemd_postun}
 
+%pre docker-publisher
+getent passwd osrt-docker-publisher > /dev/null || \
+  useradd -r -m -s /sbin/nologin -c "user for openSUSE-release-tools-docker-publisher" osrt-docker-publisher
+exit 0
+
+%postun docker-publisher
+%{systemd_postun}
+
 %pre maintenance
 getent passwd osrt-maintenance > /dev/null || \
   useradd -r -m -s /sbin/nologin -c "user for openSUSE-release-tools-maintenance" osrt-maintenance
@@ -372,6 +392,8 @@ exit 0
 %exclude %{_datadir}/%{source_dir}/check_maintenance_incidents.py
 %exclude %{_datadir}/%{source_dir}/check_source.py
 %exclude %{_datadir}/%{source_dir}/devel-project.py
+%exclude %{_datadir}/%{source_dir}/docker_publisher.py
+%exclude %{_datadir}/%{source_dir}/docker_registry.py
 %exclude %{_datadir}/%{source_dir}/metrics
 %exclude %{_datadir}/%{source_dir}/metrics.py
 %exclude %{_datadir}/%{source_dir}/metrics_release.py
@@ -408,6 +430,13 @@ exit 0
 %files check-source
 %{_bindir}/osrt-check_source
 %{_datadir}/%{source_dir}/check_source.py
+
+%files docker-publisher
+%{_bindir}/osrt-docker_publisher
+%{_datadir}/%{source_dir}/docker_publisher.py
+%{_datadir}/%{source_dir}/docker_registry.py
+%{_unitdir}/osrt-docker-publisher.service
+%{_unitdir}/osrt-docker-publisher.timer
 
 %files maintenance
 %{_bindir}/osrt-check_maintenance_incidents
--- a/docker_publisher.py
+++ b/docker_publisher.py
@@ -1,0 +1,475 @@
+#!/usr/bin/python3
+#
+# Copyright (c) 2022 SUSE LLC
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# This script's job is to listen for new releases of products with docker images
+# and publish those.
+
+import argparse
+import json
+import os
+import re
+import requests
+import subprocess
+import sys
+import tempfile
+from lxml import etree as xml
+
+import docker_registry
+
+REPOMD_NAMESPACES = {'md': "http://linux.duke.edu/metadata/common",
+                     'repo': "http://linux.duke.edu/metadata/repo",
+                     'rpm': "http://linux.duke.edu/metadata/rpm"}
+
+
+class DockerImagePublisher:
+    """Base class for handling the publishing of docker images.
+    This handles multiple architectures, which have different layers
+    and therefore versions."""
+
+    def releasedDockerImageVersion(self, arch):
+        """This function returns an identifier for the released docker
+        image's version."""
+        raise Exception("pure virtual")
+
+    def prepareReleasing(self):
+        """Prepare the environment to allow calls to releaseDockerImage."""
+        raise Exception("pure virtual")
+
+    def addImage(self, version, arch, image_path):
+        """This function adds the docker image with the image manifest, config layers
+        in image_path."""
+        raise Exception("pure virtual")
+
+    def finishReleasing(self):
+        """This function publishes the released layers."""
+        raise Exception("pure virtual")
+
+
+class DockerPublishException(Exception):
+    pass
+
+
+class DockerImageFetcher:
+    """Base class for handling the acquiring of docker images."""
+
+    def currentVersion(self):
+        """This function returns the version of the latest available version
+        of the image for the product."""
+        raise Exception("pure virtual")
+
+    def getDockerImage(self, callback):
+        """This function downloads the root fs layer and calls callback
+        with its path as argument."""
+        raise Exception("pure virtual")
+
+
+class DockerFetchException(Exception):
+    pass
+
+
+class DockerImagePublisherRegistry(DockerImagePublisher):
+    """The DockerImagePublisherRegistry class works by using a manifest list to
+    describe a tag. The list contains a manifest for each architecture.
+    The manifest will be edited instead of replaced, which means if you don't
+    call addImage for an architecture, the existing released image stays in place."""
+    MAP_ARCH_RPM_DOCKER = {'i586': ("386", None),
+                           'x86_64': ("amd64", None),
+                           'armv6l': ("arm", "v6"),
+                           'armv7l': ("arm", "v7"),
+                           'aarch64': ("arm64", "v8"),
+                           'ppc64le': ("ppc64le", None),
+                           's390x': ("s390x", None)}
+
+    def __init__(self, dhc, tag, aliases=[]):
+        """Construct a DIPR by passing a DockerRegistryClient instance as dhc
+        and a name for a tag as tag.
+        Optionally, add tag aliases as aliases. Those will only be written to,
+        never read."""
+        self.dhc = dhc
+        self.tag = tag
+        self.aliases = aliases
+        # The manifestlist for the tag is only downloaded if this cache is empty,
+        # so needs to be set to None to force a redownload.
+        self.cached_manifestlist = None
+        # Construct a new manifestlist for the tag.
+        self.new_manifestlist = None
+
+    def getDockerArch(self, arch):
+        if arch not in self.MAP_ARCH_RPM_DOCKER:
+            raise DockerPublishException("Unknown arch %s" % arch)
+
+        return self.MAP_ARCH_RPM_DOCKER[arch]
+
+    def _getManifestlist(self):
+        if self.cached_manifestlist is None:
+            self.cached_manifestlist = self.dhc.getManifest(self.tag)
+
+        return self.cached_manifestlist
+
+    def releasedDockerImageVersion(self, arch):
+        docker_arch, docker_variant = self.getDockerArch(arch)
+
+        manifestlist = self._getManifestlist()
+
+        if manifestlist is None:
+            # No manifest -> force outdated version
+            return "0"
+
+        for manifest in manifestlist['manifests']:
+            if docker_variant is not None:
+                if 'variant' not in manifest['platform'] or manifest['platform']['variant'] != docker_variant:
+                    continue
+
+            if manifest['platform']['architecture'] == docker_arch:
+                if 'vnd-opensuse-version' in manifest:
+                    return manifest['vnd-opensuse-version']
+
+        # Arch not in the manifest -> force outdated version
+        return "0"
+
+    def prepareReleasing(self):
+        if self.new_manifestlist is not None:
+            raise DockerPublishException("Did not finish publishing")
+
+        self.new_manifestlist = self._getManifestlist()
+
+        # Generate an empty manifestlist
+        if not self.new_manifestlist:
+            self.new_manifestlist = {'schemaVersion': 2,
+                                     'tag': self.tag,
+                                     'mediaType': "application/vnd.docker.distribution.manifest.list.v2+json",
+                                     'manifests': []}
+
+        return True
+
+    def getV2ManifestEntry(self, path, filename, mediaType):
+        """For V1 -> V2 schema conversion. filename has to contain the digest"""
+        digest = filename
+
+        if re.match(r"^[a-f0-9]{64}", digest):
+            digest = "sha256:" + os.path.splitext(digest)[0]
+
+        if not digest.startswith("sha256"):
+            raise DockerPublishException("Invalid manifest contents")
+
+        return {'mediaType': mediaType,
+                'size': os.path.getsize(path + "/" + filename),
+                'digest': digest,
+                'x-osdp-filename': filename}
+
+    def convertV1ToV2Manifest(self, path, manifest_v1):
+        """Converts the v1 manifest in manifest_v1 to a V2 manifest and returns it"""
+
+        layers = []
+        # The order of layers changed in V1 -> V2
+        for layer_filename in manifest_v1['Layers'][::-1]:
+            layers += [self.getV2ManifestEntry(path, layer_filename,
+                                               "application/vnd.docker.image.rootfs.diff.tar.gzip")]
+
+        return {'schemaVersion': 2,
+                'mediaType': "application/vnd.docker.distribution.manifest.v2+json",
+                'config': self.getV2ManifestEntry(path, manifest_v1['Config'],
+                                                  "application/vnd.docker.container.image.v1+json"),
+                'layers': layers}
+
+    def addImage(self, version, arch, image_path):
+        docker_arch, docker_variant = self.getDockerArch(arch)
+
+        manifest = None
+
+        with open(image_path + "/manifest.json") as manifest_file:
+            manifest = json.load(manifest_file)
+
+        manifest_v2 = self.convertV1ToV2Manifest(image_path, manifest[0])
+        # Upload blobs
+        if not self.dhc.uploadBlob(image_path + "/" + manifest_v2['config']['x-osdp-filename'],
+                                   manifest_v2['config']['digest']):
+            raise DockerPublishException("Could not upload the image config")
+
+        for layer in manifest_v2['layers']:
+            if not self.dhc.uploadBlob(image_path + "/" + layer['x-osdp-filename'],
+                                       layer['digest']):
+                raise DockerPublishException("Could not upload an image layer")
+
+        # Upload the manifest
+        manifest_content = json.dumps(manifest_v2).encode("utf-8")
+        manifest_digest = self.dhc.uploadManifest(manifest_content)
+
+        if manifest_digest is False:
+            raise DockerPublishException("Could not upload the manifest")
+
+        # Register the manifest in the list
+        replaced = False
+        for manifest in self.new_manifestlist['manifests']:
+            if 'variant' in manifest['platform'] and manifest['platform']['variant'] != docker_variant:
+                continue
+
+            if manifest['platform']['architecture'] == docker_arch:
+                manifest['mediaType'] = manifest_v2['mediaType']
+                manifest['size'] = len(manifest_content)
+                manifest['digest'] = manifest_digest
+                manifest['vnd-opensuse-version'] = version
+                if docker_variant is not None:
+                    manifest['platform']['variant'] = docker_variant
+
+                replaced = True
+
+        if not replaced:
+            # Add it instead
+            manifest = {'mediaType': manifest_v2['mediaType'],
+                        'size': len(manifest_content),
+                        'digest': manifest_digest,
+                        'vnd-opensuse-version': version,
+                        'platform': {
+                            'architecture': docker_arch,
+                            'os': "linux"}
+                        }
+            if docker_variant is not None:
+                manifest['platform']['variant'] = docker_variant
+
+            self.new_manifestlist['manifests'] += [manifest]
+
+        return True
+
+    def finishReleasing(self):
+        # Generate the manifest content
+        manifestlist_content = json.dumps(self.new_manifestlist).encode('utf-8')
+
+        # Push the aliases
+        for alias in self.aliases:
+            if not self.dhc.uploadManifest(manifestlist_content, alias):
+                raise DockerPublishException("Could not push an manifest list alias")
+
+        # Push the new manifest list
+        if not self.dhc.uploadManifest(manifestlist_content, self.tag):
+            raise DockerPublishException("Could not upload the new manifest list")
+
+        self.new_manifestlist = None
+        self.cached_manifestlist = None  # force redownload
+
+        return True
+
+
+class DockerImageFetcherURL(DockerImageFetcher):
+    """A trivial implementation. It downloads a (compressed) tar archive and passes
+    the decompressed contents to the callback.
+    The version number can't be determined automatically (it would need to extract
+    the image and look at /etc/os-release each time - too expensive.) so it
+    has to be passed manually."""
+    def __init__(self, version, url):
+        self.version = version
+        self.url = url
+
+    def currentVersion(self):
+        return self.version
+
+    def getDockerImage(self, callback):
+        """Download the tar and extract it"""
+        with tempfile.NamedTemporaryFile() as tar_file:
+            tar_file.write(requests.get(self.url).content)
+            with tempfile.TemporaryDirectory() as tar_dir:
+                # Extract the .tar.xz into the dir
+                subprocess.call("tar -xaf '%s' -C '%s'" % (tar_file.name, tar_dir), shell=True)
+                return callback(tar_dir)
+
+
+class DockerImageFetcherOBS(DockerImageFetcher):
+    """Uses the OBS API to access the build artifacts.
+    Url has to be https://build.opensuse.org/public/build/<project>/<repo>/<arch>/<pkgname>
+    If maintenance_release is True, it picks the buildcontainer released last with that name.
+    e.g. for "foo" it would pick "foo.2019" instead of "foo" or "foo.2018"."""
+    def __init__(self, url, maintenance_release=False):
+        self.url = url
+        self.newest_release_url = None
+        if not maintenance_release:
+            self.newest_release_url = url
+
+    def _isMaintenanceReleaseOf(self, release, source):
+        """Returns whether release describes a maintenance release of source.
+        E.g. "foo.2019", "foo" -> True, "foo-asdf", "foo" -> False"""
+        sourcebuildflavor = source.split(":")[1] if ":" in source else None
+        releasebuildflavor = release.split(":")[1] if ":" in release else None
+        return sourcebuildflavor == releasebuildflavor and release.startswith(source.split(":")[0] + ".")
+
+    def _getNewestReleaseUrl(self):
+        if self.newest_release_url is None:
+            buildcontainername = self.url.split("/")[-1]
+            prjurl = self.url + "/.."
+            buildcontainerlist_req = requests.get(prjurl)
+            buildcontainerlist = xml.fromstring(buildcontainerlist_req.content)
+            releases = [entry for entry in buildcontainerlist.xpath("entry/@name") if
+                        self._isMaintenanceReleaseOf(entry, buildcontainername)]
+            releases.sort()
+            # Pick the first one with binaries
+            for release in releases[::-1] + [buildcontainername]:
+                self.newest_release_url = prjurl + "/" + release
+                try:
+                    self._getFilename()
+                    break
+                except DockerFetchException:
+                    continue
+
+        return self.newest_release_url
+
+    def _getFilename(self):
+        """Return the name of the binary at the URL with the filename ending in
+        .docker.tar."""
+        binarylist_req = requests.get(self._getNewestReleaseUrl())
+        binarylist = xml.fromstring(binarylist_req.content)
+        for binary in binarylist.xpath("binary/@filename"):
+            if binary.endswith(".docker.tar"):
+                return binary
+
+        raise DockerFetchException("No docker image built in the repository")
+
+    def currentVersion(self):
+        """Return {version}-?({flavor}-)Build{build} of the docker file."""
+        filename = self._getFilename()
+        # Capture everything between arch and filename suffix
+        return re.match(r'[^.]*\.[^.]+-(.*)\.docker\.tar$', filename).group(1)
+
+    def getDockerImage(self, callback):
+        """Download the tar and extract it"""
+        filename = self._getFilename()
+        with tempfile.NamedTemporaryFile() as tar_file:
+            tar_file.write(requests.get(self.newest_release_url + "/" + filename).content)
+            with tempfile.TemporaryDirectory() as tar_dir:
+                # Extract the .tar into the dir
+                subprocess.call("tar -xaf '%s' -C '%s'" % (tar_file.name, tar_dir), shell=True)
+                return callback(tar_dir)
+
+
+def run():
+    drc_tw = docker_registry.DockerRegistryClient(os.environ['REGISTRY'], os.environ['REGISTRY_USER'], os.environ['REGISTRY_PASSWORD'],
+                                                  os.environ['REGISTRY_REPO_TW'])
+    drc_leap = docker_registry.DockerRegistryClient(os.environ['REGISTRY'], os.environ['REGISTRY_USER'], os.environ['REGISTRY_PASSWORD'],
+                                                    os.environ['REGISTRY_REPO_LEAP'])
+
+    config = {
+        'tumbleweed': {
+            'fetchers': {
+                'i586': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Tumbleweed/containers/i586/opensuse-tumbleweed-image:docker", maintenance_release=True),  # noqa: E501
+                'x86_64': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Tumbleweed/containers/x86_64/opensuse-tumbleweed-image:docker", maintenance_release=True),   # noqa: E501
+                'aarch64': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Tumbleweed/containers/aarch64/opensuse-tumbleweed-image:docker", maintenance_release=True),  # noqa: E501
+                'armv7l': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Tumbleweed/containers/armv7l/opensuse-tumbleweed-image:docker", maintenance_release=True),  # noqa: E501
+                'armv6l': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Tumbleweed/containers/armv6l/opensuse-tumbleweed-image:docker", maintenance_release=True),  # noqa: E501
+                'ppc64le': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Tumbleweed/containers/ppc64le/opensuse-tumbleweed-image:docker", maintenance_release=True),  # noqa: E501
+                's390x': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Tumbleweed/containers/s390x/opensuse-tumbleweed-image:docker", maintenance_release=True),  # noqa: E501
+            },
+            'publisher': DockerImagePublisherRegistry(drc_tw, "latest"),
+        },
+        'leap-15.3': {
+            'fetchers': {
+                'x86_64': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Leap:15.3/containers/x86_64/opensuse-leap-image:docker", maintenance_release=True),  # noqa: E501
+                'aarch64': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Leap:15.3/containers/aarch64/opensuse-leap-image:docker", maintenance_release=True),  # noqa: E501
+                'armv7l': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Leap:15.3/containers_armv7/armv7l/opensuse-leap-image:docker", maintenance_release=True),  # noqa: E501
+                'ppc64le': DockerImageFetcherOBS(url="https://build.opensuse.org/public/build/openSUSE:Containers:Leap:15.3/containers/ppc64le/opensuse-leap-image:docker", maintenance_release=True),  # noqa: E501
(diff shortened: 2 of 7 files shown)


docker_publisher.py
```diff
@@ -0,0 +1,299 @@
+#!/usr/bin/env python3
```
lnussel: please no /usr/bin/env :)

docker_publisher.py
```diff
@@ -0,0 +1,299 @@
...
+    except Exception as e:
+        print("Exception during version fetching - aborting:")
+        raise e
+
+    print("Available version: %s" % (current))
+    print("Released version: %s" % (released))
+
+    release = current != released
+
+    if "--release" in sys.argv[1:]:
```
jberry-suse: Take a look at `argparse` (used in this repo for quick examples) and easy to use.

docker_publisher.py
```diff
@@ -0,0 +1,299 @@
...
+                match = version_regex.match(line.decode('utf-8').strip())
+                if match:
+                    return match.group(1)
+
+        return "0"
+
+    def generateDockerFile(self, version, filename):
+        """Return the contents of a docker file for the image"""
+        return """FROM scratch
+MAINTAINER Fabian Vogt <fvogt@suse.com>
```
jberry-suse: Is this right for official image?

I see https://github.com/openSUSE/docker-containers/blob/master/openSUSE-Tumbleweed/config.kiwi#L15 has:

```
SUSE Containers Team <containers@suse.com>
```

docker_publisher.py
```diff
@@ -0,0 +1,299 @@
...
+# Version: %s
+ADD %s /
+""" % (version, filename)
+
+    def releaseDockerImage(self, version, rootfile):
+        # Try to delete the old branch
+        try:
+            subprocess.call(self.git_call + ["checkout", "-q", "origin/%s" % (self.git_branch)])
```
jberry-suse: I would suggest we replace all of this with something pre-built that already handles the cases like `git-sync`. I'll put in a PR to abstract the function in `issue-diff.py` so you can use it as well.

docker_publisher.py
```diff
@@ -0,0 +1,299 @@
...
+            raise DockerPublishException("Could not checkout git branch")
+
+        # Remove all old files
+        for file in glob.glob("%s/%s/*" % (self.git_path, self.path)):
+            os.remove(file)
+
+        # Re-compress it into the correct location and format
+        targetfilename = "openSUSE-Tumbleweed-%s-%s.tar.xz" % (self.arch, version)
+
+        if not recompress(rootfile, "%s/%s/%s" % (self.git_path, self.path, targetfilename)):
```
jberry-suse: For making paths us `os.path.join()`. Up to you, but I like `from os import path` and `path.join()`.

docker_publisher.py
```diff
@@ -0,0 +1,299 @@
...
+            with tempfile.TemporaryDirectory() as tar_dir:
+                # Extract the .tar.xz inside the RPM into the dir
+                subprocess.call("rpm2cpio '%s' | cpio -i --quiet --to-stdout \*.tar.xz | tar -xJf - -C '%s'" % (rpm_file.name, tar_dir), shell=True)
+                # Parse the manifest to get the name of the root tar.xz
+                manifest = json.loads(open(tar_dir + '/manifest.json').read())
+                layers = manifest[0]['Layers']
+                if len(layers) != 1:
+                    raise DockerFetchException("Unexpected count of layers in the image")
+
+                image_layer_file = tar_dir + '/' + layers[0]
```
jberry-suse: Lots of these as I noted above about `os.path.join()`.

docker_publisher.py
```diff
@@ -0,0 +1,299 @@
...
+
+        # Re-compress it into the correct location and format
+        targetfilename = "openSUSE-Tumbleweed-%s-%s.tar.xz" % (self.arch, version)
+
+        if not recompress(rootfile, "%s/%s/%s" % (self.git_path, self.path, targetfilename)):
+            raise DockerPublishException("Could not repackage the root fs layer")
+
+        # Update the version number
+        try:
+            with open("%s/%s/Dockerfile" % (self.git_path, self.path), "w") as dockerfile:
```
jberry-suse: Make a variable containing the full path combination and pass that `recompress` and `open` rather than create it multiple times.

```python
targetpath = path.join(self.git_path, self.path, targetfilename)
```

I believe this is done elsewhere. Generally makes it harder to refactor when lots of placing doing the same thing and more cpu time.

docker_publisher.py
jberry-suse: Reading through the git bits again it may not make sense given you purposely do not want to carry around history which makes sense.

docker_publisher.py
Vogtinator: Yes, the containers team moved the responsibility of those images to me.

I plan to move the kiwi description file away from git (already done for 15), so what's in there is not correct anymore anyway.

docker_publisher.py
Vogtinator: Why not? The internet recommends that strongly as it takes `$PATH` into account and I'm not aware of any downsides

docker_publisher.py
Vogtinator: Yup, the git repo is not really used git-like but more like a bucket of data.

docker_publisher.py
okurz: move responsibility from a team to a single person? Doesn't sound right. [bus-factor](https://en.wikipedia.org/wiki/Bus_factor)

docker_publisher.py
Vogtinator: Which is actually longer and more cumbersome to use, this is not going to work on windows anyway...

What about `pathlib`?
It allows syntax like `self.git_path / self.path / targetfilename`, which seems like a good compromise.

docker_publisher.py
Vogtinator: Will do, once there actually is a proper CLI interface.

docker_publisher.py
Vogtinator: It is. Good that I mostly travel by train :-)

It might make sense to use a more generic mail address there, maybe `opensuse-factory@opensuse.org`?

okurz: I guess

Vogtinator:
On second thought, I'll actually add a different publishing method just for tumbleweed.

For official images, it is required to have a git repo like `docker-containers-build`, but outside of automated builds, which don't support multi-arch, they can't be published properly.

The current image in `opensuse:tumbleweed` is part of the official library and has (kind-of) multi-arch support, so to avoid regressions the image in `opensuse/tumbleweed` should as well.

In the code this means that `DockerImagePublisherGit` will be used for Leap only (after releases and maintenace updates (?)) and for TW there'll be another class.

adrianschroeter:
we will have a registry.opensuse.org including proper signing of containers via notary ... it may be not a good idea to have another unofficial source beside ...

Vogtinator:
> we will have a registry.opensuse.org including proper signing of containers via notary ... it may be not a good idea to have another unofficial source beside ...

It would be an **incredibly bad** idea to not make the opensuse images available on the docker hub.
Docker itself does not support having multiple registries out of the box, so most current users won't be able to use the opensuse registry.

docker_publisher.py
coolo: because we package this repo and the require will be to /usr/bin/env - which is pointless. we want to require /usr/bin/python3. If you want python3 from your path, call it as python3 docker_publisher.py

Vogtinator: I thought the autoreq handler was intelligent enough to handle such cases? Apparently not.

coolo: you're welcome to improve that. it would solve quite some problems :)

adrianschroeter:
On Freitag, 9. März 2018, 09:36:16 CET wrote Fabian Vogt:
> > we will have a registry.opensuse.org including proper signing of containers via notary ... it may be not a good idea to have another unofficial source beside ...
> 
> It would be an **incredibly bad** idea to not make the opensuse images available on the docker hub.
> Docker itself does not support having multiple registries out of the box, so most current users won't be able to use the opensuse registry.

it is also bad to recommend a download without checking that it is a valid image like docker does by default.

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

sysrich:
@adrianschroeter while I agree that docker does lots of bad things, it's essential for the percieved relevance of the openSUSE project that we offer images in the official Docker hub

We need people using our images with a simple "docker pull opensuse"

We need people to be able to "use opensuse" in their dockerfiles to be used by others

These people need to be more than just openSUSE users

None of these are use cases which are possibly sated by the promised registry.opensuse.org

We need to ensure the images we put in the hub come from OBS in a streamlined way

Which means we need this pull request

We need registry.opensuse.org too, because we need an option to use something better and more reliable in our own tooling and in our default arrangements, but there is no way openSUSE can survive in this age of containers without official images on the hub, and Fabian's approach seems to be the best way we can accomplish that.

adrianschroeter:
On Freitag, 9. März 2018, 09:52:27 CET wrote Richard Brown:
> @adrianschroeter while I agree that docker does lots of bad things, it's essential for the percieved relevance of the openSUSE project that we offer images in the official Docker hub
> 
> We need people using our images with a simple "docker pull opensuse"
> 
> We need people to be able to "use opensuse" in their dockerfiles to be used by others
> 
> These people need to be more than just openSUSE users
> 
> None of these are use cases which are possibly sated by the promised registry.opensuse.org
> 
> We need to ensure the images we put in the hub come from OBS in a streamlined way
> 
> Which means we need this pull request
> 
> We need registry.opensuse.org too, because we need an option to use something better and more reliable in our own tooling and in our default arrangements, but there is no way openSUSE can survive in this age of containers without official images on the hub, and Fabian's approach seems to be the best way we can accomplish that.

you may need the images there and it is fine as long as everbody is aware that SUSE can not
guarantee that these images get not altered (at upload time, at any later time or only delivered for specific users).

as long no one is using them to store private data this might be fine, if everyone is aware.
However, I doubt and I doubt that we want to take responsibility by announcing these.

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

sysrich:
> you may need the images there and it is fine as long as everbody is aware that SUSE can not guarantee that these images get not altered (at upload time, at any later time or only delivered for specific users). 

The problems you describe are endemic of the docker hub ecosystem, so there should be no special consideration required for users of the _openSUSE_ images there

Obviously, once registry.opensuse.org is a reality we will favour it as our preferred registry, we should do a good job advertising all of it's benefits over the hub, and we're already looking at alternative OCI runtimes that would allow us to pull from such a registry by default while falling back to the Hub for image tags not hosted by the registry

docker_publisher.py
lnussel: in any case we're not some random ISV that repackages random python versions somewhere. This script will be for /usr/bin/python3 and nothing else.

adrianschroeter:
On Freitag, 9. März 2018, 10:36:20 CET wrote Richard Brown:
> > you may need the images there and it is fine as long as everbody is aware that SUSE can not guarantee that these images get not altered (at upload time, at any later time or only delivered for specific users). 
> 
> The problems you describe are endemic of the docker hub ecosystem, so there should be no special consideration required for users of the _openSUSE_ images there
> 
> Obviously, once registry.opensuse.org is a reality we will favour it as our preferred registry, we should do a good job advertising all of it's benefits over the hub, and we're already looking at alternative OCI runtimes that would allow us to pull from such a registry by default while falling back to the Hub for image tags not hosted by the registry

we can also publish from OBS to other external registries, even right now. We just would need to know 
the mapping.

This would also help us with maintenance updates later on. And at some point of time we 
could do propper signing (using official key once the skopeo/notary problems are sorted out). 
Also the publish states would not be broken.

But don't expect that we workaround via external scripts in future.

Also I see the possible additional trust and maintenance support as possible additional
value, which we can offer in openSUSE in future compared to others. It is IMHO therefore
bad that we just tell our user to compromise their systems ...

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

Vogtinator:
> we can also publish from OBS to other external registries, even right now. We just would need to know
the mapping.

Good to know. A bit late though.

Are multi-arch images from multiple repositories supported? i.e. `openSUSE:Factory:Containers/opensuse-tumbleweed -> opensuse/tumbleweed for x86_64 and `openSUSE:Factory:ARM:Containers/opensuse-tumbleweed -> opensuse/tumbleweed for aarch64?

Can OBS also publish to the official images? I don't think so, we'll need a script for that anyway.

adrianschroeter:
On Freitag, 9. März 2018, 14:00:25 CET wrote Fabian Vogt:
> > we can also publish from OBS to other external registries, even right now. We just would need to know
> the mapping.
> 
> Good to know. A bit late though.
> 
> Are multi-arch images from multiple repositories supported? i.e. `openSUSE:Factory:Containers/opensuse-tumbleweed -> opensuse/tumbleweed for x86_64 and `openSUSE:Factory:ARM:Containers/opensuse-tumbleweed -> opensuse/tumbleweed for aarch64?

in general yes, but afaik you can not push to same repo since you would overwrite the tags. registry has not
really an arch handling.

> 
> Can OBS also publish to the official images? I don't think so, we'll need a script for that anyway.

what is official and why not?

official would anyway mean that they are signed with official key, so if anyone can, obs can.

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

adrianschroeter:
On Freitag, 9. März 2018, 16:14:03 CET wrote Richard Brown:
> > Also I see the possible additional trust and maintenance support as possible additional value, which we can offer in openSUSE in future compared to others. It is IMHO therefore bad that we just tell our user to compromise their systems ...
> 
> We will tell our users to use our platforms (when they exist)
> 
> But such platforms do not exist right now, and yet we have entire projects that are dependant on openSUSE having a viable container story, as well as hundreds of thousands of users who are already pulling openSUSE containers from the docker hub
> 
> The docker hub is a wild west, and the openSUSE distributions are open source software.
> If WE do not put official openSUSE images into the docker hub, there is nothing preventing anyone else doing so
> 
> And as long as they do not alter the sources they use, they would be fully within their right to call it openSUSE. 
> Our Trademark Policy and license of the openSUSE Project prevents us from limiting the use of openSUSE in any other way.
> 
> Meaning someone other than us would be publishing openSUSE (legally) on the docker hub.
> 
> This would be fine, and legal, just as many hosting providers have openSUSE images which are not built by SUSE/openSUSE and not hosted on SUSE/openSUSE infrastructure, but are still openSUSE as long as they only use our sources.
> 
> Conversely, if we actively engage with a service like the Docker Hub, then we are in a better position to enforce our standards and enforce our trademark. There is solid ground to argue that we are actually legally COMPELLED to manage our openSUSE images in the Docker Hub, or else we could be seen to erode our enforceability of the openSUSE trademarks.
> 
> I think we have to accept the reality that we cannot ensure every openSUSE image out there is hosted on our infrastructure.
> 
> I'd much rather we have something like @Vogtinator's proposed solution so we at least make sure that the openSUSE images in the docker hub are as official and as sanely built and as valid as possible.
> 
> It might not be hosted on our infra, and that might bring risks, but at least Fabian's tooling will mean that we can be confident that it was built properly and came from an authoritative source of openSUSE software.

It is not about "our infrastructure", that is only a workaround for other problems in docker/notary/skopeo env.

The main point is that you can not trust the content and you need actively warn about the content you uploaded.
That is kind of stupid.

However, if you still want this, it is better to do in a way where we can at least convert to in a supportable
way for the future.

> That's better than leaving it to random-untrusted-person uploading random-untrustworthy-image from random-untrusted-machine using random-untrusted-tools
>
> Which is what we'll face if we do not properly manage the openSUSE images in the Docker Hub.
> 
> Of course, once we have nice better alternatives, we'll favour and focus on them, but reality is imperfect and I foresee the need to have a decent openSUSE presence on the Docker Hub being an unavoidable truth for years to come.

What I described exist since last year. You saw most likely our announcements in internal lists. There is no need to wait, except you really want it doing it
using your own tools and your own control and guarantee that others will have more work later and make it harder to get this
into a maintained stream.

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

sysrich:
> You saw most likely our announcements in internal lists. There is no need to wait

I don't understand are you saying that openSUSE has a working container registry that will work with everything we need from Kubic, Docker, Cri-O, etc, today?

Why would such news be kept to internal SUSE mailinglist? I'm talking about what openSUSE needs for openSUSE's software - in the case of containers we have very different requirements from SUSE with SLE, and I can assure you that no one has spoken to me trying to gather those requirements.

sysrich:
> The main point is that you can not trust the content and you need actively warn about the content you uploaded. That is kind of stupid.

Where is the 'active warning' about the fact that users cannot trust the ISO content from openSUSE's many http mirrors?

Sure, rpm's have their signatures, and our ISO's have their gpg signed checksums, but I would describe the documentation we have on this topic as part of software.opensuse.org as 'passive' warning at best.

We could put a similar warning on our docker hub page, i'm sure.

If that is not good enough, why are you trying to set a higher bar for the Docker Hub than we've had for the openSUSE infrastructure since the project began?

docker_publisher.py
jberry-suse: Does not appear to be in SLE SP3 repo which is where most of this code ends up running. Up to you, I never find:

```python
"%s/%s/%s" % (...)
```

more readable or shorter than

```python
path.join(...)
```

even

```python
'/'.join([...])
```

Vogtinator:
> On Freitag, 9. März 2018, 14:00:25 CET wrote Fabian Vogt:
>  >  > we can also publish from OBS to other external registries, even right now. We just would need to know
>  >  the mapping.
>  > 
>  >  Good to know. A bit late though.
>  > 
> >  Are multi-arch images from multiple repositories supported? i.e. `openSUSE:Factory:Containers/opensuse-tumbleweed -> opensuse/tumbleweed for x86_64 and `openSUSE:Factory:ARM:Containers/opensuse-tumbleweed -> opensuse/tumbleweed for aarch64?
>  
> in general yes, but afaik you can not push to same repo since you would overwrite the tags. registry has not really an arch handling.

It has - using manifest lists. I'm currently looking into a way to push multi-arch images easily using the registry API as the official tools (`docker push`, `skopeo`) don't support those yet.

>  Can OBS also publish to the official images? I don't think so, we'll need a script for that anyway.
> what is official and why not?
> 
> official would anyway mean that they are signed with official key, so if anyone can, obs can.

The "official images" on docker hub are images in the library namespace. To submit those, you need to make a PR to https://github.com/docker-library/official-images

adrianschroeter:
On Freitag, 9. März 2018, 17:04:59 CET wrote Richard Brown:
> > You saw most likely our announcements in internal lists. There is no need to wait
> 
> I don't understand are you saying that openSUSE has a working container registry that will work with everything we need from Kubic, Docker, Cri-O, etc, today?

I say we have working integrated upload to any registry server since last year. Even including signing,
but it has some open issues (which are outside of OBS code, but we are working on it).

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

adrianschroeter:
On Freitag, 9. März 2018, 17:07:22 CET wrote Richard Brown:
> > The main point is that you can not trust the content and you need actively warn about the content you uploaded. That is kind of stupid.
> 
> Where is the 'active warning' about the fact that users cannot trust the ISO content from openSUSE's many http mirrors?
> 
> Sure, rpm's have their signatures, and our ISO's have their gpg signed checksums, but I would describe the documentation we have on this topic as part of software.opensuse.org as 'passive' warning at best.
> 
> We could put a similar warning on our docker hub page, i'm sure.
>
> If that is not good enough, why are you trying to set a higher bar for the Docker Hub than we've had for the openSUSE infrastructure since the project began?

it is not the same, the docker tools is not enabled by default to check signatures (code exists). 
And if you would enable it there us not even the chance to check for a trusted signature atm. 


-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

sysrich:
> it is not the same, the docker tools is not enabled by default to check signatures (code exists). And if you would enable it there us not even the chance to check for a trusted signature atm.

I think it's exactly the same - all of openSUSE's browsers and download tools don't validate the signature checks for ISO downloads by default

I do not see how one sad reality with browsers is acceptable, but a similar sad reality with the docker client is unacceptable and is grounds for holding back Fabian's submission

Like it or not (I do not like it), Docker Hub has become the de facto source for linux containers in the world

If openSUSE doesn't have well built OBS containers there, the only people that hurts is openSUSE

Holding back our images from Docker because we object to how they run their ecosystem isn't going to help openSUSE find more users, contributors, and relevance in the container space..quite the contrary.

Can we take a pragmatic approach to this? I promise to champion the 'right' solution (our own openSUSE registry) once that option is available to the Project.

Vogtinator:
> I do not see how one sad reality with browsers is acceptable, but a similar sad reality with the docker client is unacceptable and is grounds for holding back Fabian's submission

This isn't holding back the PR - the code which would submit to the Hub isn't in here (yet).
We need this in any case for the official-library.

adrianschroeter:
On Freitag, 9. März 2018, 22:31:27 CET wrote Richard Brown:
> > it is not the same, the docker tools is not enabled by default to check signatures (code exists). And if you would enable it there us not even the chance to check for a trusted signature atm.
> 
> I think it's exactly the same - all of openSUSE's browsers and download tools don't validate the signature checks for ISO downloads by default

it is not, since you have the chance to validate and we don't deliver the tooling for downloading.

Also they are usually used to setup a new system and can not compromise an existing one (in the light of spectre/meltdown 
and give that one is likely putting own content inside).

> I do not see how one sad reality with browsers is acceptable, but a similar sad reality with the docker client is unacceptable and is grounds for holding back Fabian's submission
> 
> Like it or not (I do not like it), Docker Hub has become the de facto source for linux containers in the world

heck, I don't argue not to use docker hub. I argue that we should use it in a way which can get
integrated in our workflows. And this stuff exist already, no need for workarounds.

the OBS can publish directly, the stuff is there since last year. You can integrate that way 
the release mechanics, we can deliver maintenance updates and we can deliver new containers
dependening on the base container.

And we _will_ be able to do a real trusted signing in future.

We have learned what extra work it is causing when every single product image needs special
treatment.


> If openSUSE doesn't have well built OBS containers there, the only people that hurts is openSUSE
> 
> Holding back our images from Docker because we object to how they run their ecosystem isn't going to help openSUSE find more users, contributors, and relevance in the container space..quite the contrary.

Tell me on single thing what you win if you do it with external scripts (no, it is not availability
on docker hub).

> Can we take a pragmatic approach to this? I promise to champion the 'right' solution (our own openSUSE registry) once that option is available to the Project.

this is an additional thing which would give security with the current broken tooling. You don't
need that if you ignore signing and trusted images anyway.

-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

sysrich:
> heck, I don't argue not to use docker hub. I argue that we should use it in a way which can get integrated in our workflows. And this stuff exist already, no need for workarounds.

> Tell me on single thing what you win if you do it with external scripts (no, it is not availability
on docker hub).

openSUSE images for all currently available architectures, including x86_64 and aarch64

As far as Fabian and I can find, there is no provision for this in the OBS functionality you refer to from the last year.

thkukuk:
Adrian, containers cannot compromise your system if the admin does not give explicit rights to the container to do so. That was the design goal of containers.
Else: we need to become aware in the container environment. For this, we need to be visible in the places where container people look, which means docker hub. For a clean better solution we are years too late. We have to play the game according to todays rules, not our own which nobody else is interested in.
And this means, we have to provide openSUSE containers on docker hub, there is no way around this except, that we loose all visibility. And while the missing visibility has already a negative impact on us, not being visible at all would be the worst thing which can happen. So, please help to get openSUSE containers into docker hub in a smooth way. If we have an openSUSE registry later, we can still upload signed container images there additional, and people can use that if they care about security and want to check what they download. But today, this people care only about their comfort.

adrianschroeter:
On Montag, 12. März 2018, 11:48:46 CET wrote Thorsten Kukuk:
> Adrian, containers cannot compromise your system if the admin does not give explicit rights to the container to do so. That was the design goal of containers.

I explicit wrote "in the light of spectre/meltdown" and there you can fetch for example your private ssh key and get on your system.

But this is not really my main point here, it is bad enough if you can not trust the container itself at all.
What are you gonna using for it if you can not put any kind of data inside which you need to trust?

> Else: we need to become aware in the container environment. For this, we need to be visible in the places where container people look, which means docker hub. For a clean better solution we are years too late. We have to play the game according to todays rules, not our own which nobody else is interested in.

again (third time), I don't argue against putting stuff on docker hub. I would prefer if we do it in a way where it would be signed,
so that you can use them for anything serious. But if you wont to ignore this, I am also fine to upload it in
a not trusted way.

I was arguing to use our built-in mechanics, which do exist for that. We can later on doing the real signing
and we don't need to break our workflows and schemas again.

However, I don't like to be in this situation where we just want to do the same broken stuff as others do.
We should do it better, we should highlight that and have an advantage. We should really start thinking in
this way..

> And this means, we have to provide openSUSE containers on docker hub, there is no way around this except, that we loose all visibility. And while the missing visibility has already a negative impact on us, not being visible at all would be the worst thing which can happen. So, please help to get openSUSE containers into docker hub in a smooth way. If we have an openSUSE registry later, we can still upload signed container images there additional, and people can use that if they care about security and want to check what they download. But today, this people care only about their comfort.
> 
> 


-- 

Adrian Schroeter
email: adrian@suse.de

SUSE Linux GmbH, GF: Felix Imendörffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nürnberg) 
 
Maxfeldstraße 5                         
90409 Nürnberg 
Germany

sysrich:
>  However, I don't like to be in this situation where we just want to do the same broken stuff as others do. We should do it better, we should highlight that and have an advantage. We should really start thinking in this way

But with the situation regarding multi-arch, we can't really say that we're doing it better - that approach would totally prevent the use of any openSUSE docker container is any (normally arch neutral) k8s manifest for example

Which means no one in their right mind would use an openSUSE image compared to any other distribution, all of which offer multi-arch support out of the box

This is especially important when you consider that the officially supported arch's of Docker include s390x, ARMv7 32-bit as well as x86-64

So if openSUSE ends up being the only mainline distro that doesn't support those architectures easily, making it unsuitable for use in any solution needing images on those architectures..then how exactly is any claim that "we do it better" going to be seen with ANY credibility at all?

I honestly would not stand up in front of anyone and make such a claim at the moment - the Status Quo is an embarrassment and making it more difficult for everyone isn't going to be any better.

This also risks being problematic from a SUSE perspective - am I meant to explain to the world that openSUSE will not have solid multi-arch containers because of SUSE's OBS team and the decisions they made about how the world should handle container images?

do we really want SUSE to be seen as lacking credibility in a multi-arch container story? I would imagine we need quite the inverse, we need to be out there, and strong, in every arch which SUSE makes it's money on, which includes ppc and aarch64.

Let's worry about doing it better after we've at least caught up with the rest of the world already is please.

docker_publisher.py
jberry-suse: Why wait, add two lines and you have a real arg parsing.

Vogtinator:
Next step committed.

Remaining WIP items are:

- Add a configuration file for docker registry credentials? I'd use a systemd unit with `EnvironmentFile=`, is that ok?
- Maybe add a check that the `docker-image` check for TW passed

jberry-suse:
> Add a configuration file for docker registry credentials? I'd use a systemd unit with EnvironmentFile=, is that ok?

Sounds fine, see the `config` directory and how that is setup in related systemd templated services and `Makefile`. Will also need to decide under what user this runs and add or create relates sub-package for deployment.

docker_publisher.py
jberry-suse: Something more generic would definitely seem better.

docker_publisher.py
```diff
@@ -0,0 +1,547 @@
...
+        # Re-compress it into the correct location and format
+        targetfilename = "openSUSE-Tumbleweed-%s-%s.tar.xz" % (arch, version)
+
+        # Parse the manifest to get the name of the root tar.xz
+        manifest = json.loads(open(image_path + '/manifest.json').read())
+        layers = manifest[0]['Layers']
+        if len(layers) != 1:
+            raise DockerPublishException("Unexpected count of layers in the image")
+
+        image_layer_file = image_path + '/' + layers[0]
```
jberry-suse: nit: mixing and matching `'` vs `"` in same function without any visible reason. Not a big deal, but consistency is nice.

docker_registry.py
```diff
@@ -0,0 +1,212 @@
...
+        if resp.status_code != 200:
+            return False
+
+        return resp.json()
+
+    def getManifestDigest(self, reference):
+        """Return the digest of the manifest with the given reference.
+        If the manifest doesn't exist or the request fails, it returns False."""
+        resp = self.doHttpCall("HEAD", "/v2/%s/manifests/%s" % (self.repository, reference),
+                               headers={'Accept': "application/vnd.docker.distribution.manifest.list.v2+json,application/vnd.docker.distribution.manifest.v2+json"})
```
jberry-suse: nit: same thing on quotes, but the same line even :(

docker_registry.py
```diff
@@ -0,0 +1,212 @@
...
+        content = None
+        with open(filename, "rb") as blob:
+            content = blob.read()
+
+        # First request an upload "slot", we get an URL we can PUT to back
+        upload_request = self.doHttpCall("POST", "/v2/%s/blobs/uploads/" % self.repository)
+        if upload_request.status_code == 202:
+            location = upload_request.headers["Location"]
+            upload = self.doHttpCall("PUT", location + "&digest=" + digest,
+                                     headers={'Content-Length': str(len(content))},
```
jberry-suse: This really has to be done manually? Not overly familiar with python's `requests`, but that seems rather basic.

docker_publisher.py
jberry-suse: Still not seeing the allure in manually writing out `%s/` for each part over either other solution, but the variable is nice.

jberry-suse:
Will need to either exclude these files or figure out how to configure nose for different python versions in same repo.

```
======================================================================
ERROR: Failure: ImportError (No module named parse)
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/openSUSE/osc-plugin-factory/docker_registry.py", line 33, in <module>
    import urllib.parse
ImportError: No module named parse
```

Since you have no tests at the moment excluding would seem best.

jberry-suse:
Also make sure to squash commits:

- Fix docker arch name for ARM
- Revert "Fix docker arch name for ARM"

etc is not helpful.

docker_registry.py
Vogtinator: I always use `'` for keys and `"` for values.

docker_registry.py
Vogtinator: I just put it there to be safe - apparently it's not necessary:

`Content-Length headers will be overridden when we can determine the length of the content.`

docker_publisher.py
Vogtinator: Indeed, fixed.

Vogtinator:
> Also make sure to squash commits:
> 
> Fix docker arch name for ARM
> Revert "Fix docker arch name for ARM"
> etc is not helpful.

I won't do that - when merging just select "Squash and merge" from the merge button selector.

Vogtinator:
> Will need to either exclude these files or figure out how to configure nose for different python versions in same repo.
>
> Since you have no tests at the moment excluding would seem best.

I would like to change that, but as 90% of the code is just a client to some API that's not really that practical - a full server would be necessary to test that :-/

Ideas welcome.

jberry-suse:
> I won't do that - when merging just select "Squash and merge" from the merge button selector.

Unfortunately it is disabled. Generally we control our commits as many times we would like more than one.

> I would like to change that, but as 90% of the code is just a client to some API that's not really that practical - a full server would be necessary to test that :-/
> 
> Ideas welcome.

Indeed, which is why I bring up a full OBS in container (even on travis) and load it with a decent facsimile of Factory via `obs_clone`. Still needs someone to setup `obs_factory` and likely would need to clone more things for your use-case, but I'd have to look at it closer to be sure how much.

docker_registry.py
jberry-suse: my eyes!

Vogtinator:
> Unfortunately it is disabled.

Currently it's disabled because the state is red.

> Indeed, which is why I bring up a full OBS in container (even on travis) and load it with a decent facsimile of Factory via obs_clone. Still needs someone to setup obs_factory and likely would need to clone more things for your use-case, but I'd have to look at it closer to be sure how much.

That's a lot of setup for a few tests :-/

jberry-suse:
> Currently it's disabled because the state is red.

More than just red.

image

> That's a lot of setup for a few tests :-/

So is faking all of OBS queries for a few tests. Rather easy if everything is just the way you expect it. If you mean for your script sure, but that's the case I find myself in and did the heavy work already. You shouldn't need `obs_factory` plugin. But for copying images around and such you'll likely need to setup fake binaries and such which may or may not be fun, but that's the life of testing....harder than writing the code typically.

jberry-suse:
Unless you are feeling in the mood I would just configure nose to ignore your scripts.

Vogtinator:
> More than just red.

Indeed - I didn't even know you could disable that.

> So is faking all of OBS queries for a few tests. Rather easy if everything is just the way you expect it. If you mean for your script sure, but that's the case I find myself in and did the heavy work already. You shouldn't need obs_factory plugin. But for copying images around and such you'll likely need to setup fake binaries and such which may or may not be fun, but that's the life of testing....harder than writing the code typically.

The copying of images is only done in the fetcher from the repo - which doesn't need to be mocked to be tested.
The publishing is the harder issue, with read-modify-write interaction over HTTP and doing destructive git checkouts...

> Unless you are feeling in the mood I would just configure nose to ignore your scripts.

Yes, how can I do that? I never heard about nose before.

jberry-suse:
> Yes, how can I do that? I never heard about nose before.

I do not know of the top of my head, but likely has some method for excluding what files it loads.

https://github.com/openSUSE/osc-plugin-factory/blob/3bd6833f89583dea77285a1d2da8c39071e6aa24/.travis.yml#L50-L75

Vogtinator:
I also adjusted the Makefile - let's see what travis says.

.travis.yml
```diff
@@ -70,7 +70,8 @@ matrix:
         # Needs python prefix to use the correct interpretor.
         - python ./obs_clone.py --cache --debug --apiurl-target local
       script:
-        - nosetests --with-coverage --cover-package=. --cover-inclusive
+        # Ignore the Python3 docker_*.py scripts
+        - nosetests --with-coverage --cover-package=. --cover-inclusive --ignore-files 'docker_.+\.py' --ignore-files 'docker_.+\.py'
```
jberry-suse: If I remember correctly, there is a config format for nose. If this can be placed there (after seeing that it works) that would be preferred. One place to manage and local runs during development won't require flags.

Vogtinator: I'll try that - hopefully it supports multiple mentions, which seems counter-intuitive for ini-style files

jberry-suse: Seems to have resolved the failure on travis although you have packaging fun now :)

systemd/osrt-docker-publisher.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=openSUSE Release Tools: Docker image publisher
+
+[Service]
+User=opensuse-relmgr-bot
+EnvironmenFile=~/.config/osrt-docker_publisher
```
jberry-suse: Would seem to make more sense in in sysconfig dir. Similar to:

https://github.com/openSUSE/osc-plugin-factory/blob/3bd6833f89583dea77285a1d2da8c39071e6aa24/systemd/osrt-leaper-manager%40.service#L7

Can also include the base config in `config` dir to have it installed there.

Vogtinator: No, the file contains a password which must only be readable by that user.

dist/package/openSUSE-release-tools.spec
```diff
@@ -298,6 +310,14 @@ exit 0
 %postun check-source
 %systemd_postun
 
+%pre docker-publisher
+getent passwd opensuse-relmgr-bot > /dev/null || \
```
jberry-suse: The pattern is `osrt-docker-publisher`. When deployed move away from any existing users on packagelists which also works well during migration period where things run half and half.

Vogtinator: I don't quite understand - this .spec file is only useful for packagelists AFAICT and we already agreed on `opensuse-relmgr-bot` for this on packagelists

jberry-suse: This spec produces packages deployed on 4 machines and part of the point was to standardize names since users, services, and everything else were all over the place before.

Vogtinator: Ok - then I'm not sure why @DimStar77 and @lnussel made this proposal...

Once this is deployed I guess I'll move it over?

docker_registry.py
```diff
@@ -0,0 +1,211 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# This is a very basic client for the Docker Registry V2 API.
+# It exists for a single reason: All clients either:
+# - Don't work
+# - Don't support uploading
+# - Don't support multi-arch images (manifest lists)
+# and some even all three.
```
jberry-suse: hehe, sounds about right.

docker_publisher.py
jberry-suse: Seems like this still needs to be addressed. Seems like Factory is best if not perhaps opensuse-releaseteam@o.o?

docker_publisher.py
```diff
@@ -0,0 +1,623 @@
...
+
+        # Remove all old files
+        for file in glob.glob(target_dir + "/*"):
+            os.remove(file)
+
+        # Re-compress it into the correct location and format
+        targetfilename = "openSUSE-Tumbleweed-%s-%s.tar.xz" % (arch, version)
+
+        # Parse the manifest to get the name of the root tar.xz
+        manifest = json.loads(open(image_path + '/manifest.json').read())
```
jberry-suse: nit: seemingly inconsistent quote usage based on your pattern.

docker_publisher.py
```diff
@@ -0,0 +1,623 @@
...
+        self.versioned_redir = versioned_redir
+        self.repourl = repourl
+        self.packagename = packagename
+        self.arch = arch
+
+    def currentVersion(self):
+        # For TW we ask the mirrorbrain server about the -Current redirection target
+        meta4_xml = requests.get(self.versioned_redir + ".meta4")
+        meta4 = xml.fromstring(meta4_xml.content)
+        filename = meta4.xpath("//m:metalink//m:file//@name", namespaces={'m': 'urn:ietf:params:xml:ns:metalink'})[0]
```
jberry-suse: nit: another single-quote value? Seems like a reason not to mix as it makes it easy for this to happen.

docker_publisher.py
```diff
@@ -0,0 +1,623 @@
...
+        return re.search('Snapshot(\d+)-', filename).group(1)
```
jberry-suse: nit: perhaps use regex string `r"pattern"` to avoid issues in future

Vogtinator: Indeed, fixed.

docker_publisher.py
Vogtinator: Well - it's always easy to violate styles by accident... Fixed.

docker_publisher.py
Vogtinator: Indeed, fixed.

Vogtinator:
For some reason GitHub doesn't let me comment on the email address thread...

I won't change it here as it has to be the same value as in the .kiwi files. The value here is only relevant for the official library images, which aren't important anymore anyway.

If you prefer releaseteam@o.o, I can change it everywhere.

jberry-suse:
> If you prefer releaseteam@o.o, I can change it everywhere.

If it doesn't bother others, but otherwise one of the two would seem better. Perhaps factory since that's where most issues are discussed.

Vogtinator:
I don't think a mailing list should be used as maintainer of an image.

jberry-suse:
Looking at a number of the official images from other projects included in library I see no maintainer. Can obviously do as we please, but the benefit of mailing list is they don't change, people do. Either way as likely no one will look at it anyway, they'll just submit a bug or factory list mail as they do for everything else.

docker_registry.py
okurz: I suggest to use the style checks provided by flake8 (drop-in replacement for pep8) and follow these :)

systemd/osrt-docker-publisher.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=openSUSE Release Tools: Docker image publisher
+
+[Timer]
+OnBootSec=120
+OnUnitInactiveSec=15 min
```
okurz: Wouldn't it be about time to move to event-based triggering based on OBS/openQA events? See https://rabbit.opensuse.org/ and https://openbuildservice.org/help/manuals/obs-admin-guide/obs.cha.administration.html#_message_bus for details

Vogtinator: That alone won't be sufficient - if somehow the TW release message gets lost (slow network, packet loss, downtime...), the bot would never release the containers.

The script itself can already be called to only release one image, listening for events should be done externally then.

Vogtinator:
GH does not let me reply to your comment inline:

@okurz: My IDE already invokes pycodestyle, except for line too long (which makes no sense for this line, it's just a long hardcoded string) there's no major style offense in there.

okurz:
ok, good enough for me :)

systemd/osrt-docker-publisher.timer
coolo: If you rebase you can drop the packaging and the systemd units and replace it with a gocd trigger on the Factory release pipeline.

coolo:
What's your plan here?

Vogtinator:
> What's your plan here?

Once I'm less busy with more important stuff, migration to botmaster. Should be relatively simple, after the secrets are copied over.

coolo:
... after fixing the python offenses

Vogtinator:
Merging to get this working again - feel free to review still, I'll try to address comments with a new PR.

Issue #1415, opened 2018-02-21
pkglistgen: IOError: Not a gzipped file (intermittent)


lnussel:
I see that quite often in the log:
```
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]: -> do_dump_solv for openSUSE:Leap:42.2:Update/non-oss
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]: Traceback (most recent call last):
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:   File "/usr/bin/osrt-pkglistgen", line 1358, in <module>
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:     sys.exit(app.main())
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:   File "/usr/lib/python2.7/site-packages/cmdln.py", line 261, in main
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:     return self.cmd(args)
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:   File "/usr/lib/python2.7/site-packages/cmdln.py", line 284, in cmd
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:     retval = self.onecmd(argv)
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:   File "/usr/lib/python2.7/site-packages/cmdln.py", line 422, in onecmd
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:     return self._dispatch_cmd(handler, argv)
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:   File "/usr/lib/python2.7/site-packages/cmdln.py", line 1123, in _dispatch_cmd
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:     return handler(argv[0], opts, *args)
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:   File "/usr/bin/osrt-pkglistgen", line 1037, in do_update_and_solve
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:     self.do_update_and_solve(subcmd, copy.deepcopy(opts))
Feb 21 16:17:40 repochecker osrt-pkglistgen[26851]:   File "/usr/bin/osrt-pkglistgen", line 1056, in do_update_and_solve
[...]

PR #1442, opened 2018-03-08
pkglistgen: do_dump_solv(): include hash in update solv to avoid download.

jberry-suse:
- f9b7832e8ba1f8f09e69ef52ae4cfde6ab3520f9:
    pkglistgen: do_dump_solv(): include hash in update solv to avoid download.

- 28b467f4e8d9a8c40177b4d477729cabce850ba4:
    pkglistgen: do_dump_solv(): utilize os.path.join().

Should be rebased on #1439.


jberry-suse:
This should fix #1415 assuming production has a copy of it already.

jberry-suse:
Rebased.

PR #1444, opened 2018-03-09
metrics: provide release metrics ingestion and dashboard.

jberry-suse:
Per my plans and factory mailing list requests this will ingest metrics data about TW snapshots and review data.

image


jberry-suse:
[Live view](https://metrics.opensuse.org/d/osrt_release/osrt-release?orgId=1) post #1451 deployment along with Grafana 5.x upgrade.

PR #1447, opened 2018-03-09
pkglistgen: Fix reading boolean values

lnussel:


osclib/conf.py
```diff
@@ -168,6 +168,8 @@
 # lock = openSUSE:Factory:Staging
 #
 
+def str2bool(v):
+  return v.lower() in ("yes", "true", "t", "1")
```
coolo: conf.py:172:3: E111 indentation is not a multiple of four

nilxam:
`check-source-add-review-team` does not add to Factory config in conf.py, you prefer put it on remote config?

check_source.py
```diff
@@ -131,7 +133,7 @@ def check_source_submission(self, source_project, source_package, source_revisio
             self.review_messages['accepted'] += "\n\nOutput of check script (non-fatal):\n" + output
 
         if not self.skip_add_reviews:
-            if self.review_team is not None:
+            if self.add_review_team and self.review_team is not None:
```
jberry-suse: Presumably intending leaper should be in charge of this?

Issue #1185, opened 2017-10-10
packagelists: migrate services to those provided by package


jberry-suse:
As services are included in the packaging as part of #1006 the services should be re-deployed on packagelists to run out of package. Once completed maintenance machine can be migrated and new services deployed for SLE.

Before any services can be switched the primary code-base (the one used globally by `osc`) must be switched to be that of the one provided by package.

List of services to migrate based on old name.

- [ ] `build-fail-reminder@.service`
- [ ] `create-test-dvds.service`
- [x] `check-duplicate-binaries@.service`
- [x] `factory-announcer.service`
- [ ] `factory-auto-submit.service`
- [x] `factory-dvl-prj-list.service`
- [x] `factoryauto.service`
- [x] `gnome-review-bot.service`
- [x] `leap-423-announcer.service`
- [x] `leaperbot.service`
- [x] `maintbot.service`
- [x] `manager_42.service`
- [x] `openqa-comments@.service`
- [x] `osc-check_dups.service`
- [x] `pkglist@.service`
- [x] `repo-checker@.service`
- [x] `repochecker.service`
- [ ] `source-checker.service`
- [x] `staging-bot-daily@.service`
- [x] `staging-bot-regular@.service`
- [x] `staging-bot-reminder.service`
- [x] `suppkg_rebuild@.service`
- [ ] `trigger-rebuilds.service`
- [x] `ttm-arm.service`
- [x] `ttm-leap.service`
- [x] `ttm-s390.service`
- [x] `ttm-tumbleweed.service`
- [x] `update-crawler.service`

jberry-suse:
```
sudo zypper mr --disable openSUSE:Tools
sudo zypper ar http://download.opensuse.org/repositories/openSUSE:/Tools/SLE_12_SP2 openSUSE:Tools-1169-workaround
sudo zypper ref
sudo zypper in openSUSE-release-tools-repo-checker osc-plugin-staging osc-plugin-check_dups

sudo mv /var/lib/osc-plugins/ /var/lib/osc-plugins-old/
```

First step is complete, the global source used by `osc` is now provided by package. I will monitor logs to see what issues arise.

jberry-suse:
```
sudo less /home/_opensuse.org-repo-checker/.oscrc
# copy credentials
sudo -u osrt-repo-checker $(which osc)
# fill in credentials

sudo systemctl enable osrt-repo-checker.timer
sudo systemctl start osrt-repo-checker.timer
```

[...]

PR #1452, opened 2018-03-11
services: Add Install section to totest-manager service

DimStar77:
Without an install section, systemctl enable osrt-totest-manager@openSUSE:Factory has no effect, which in turn results in the service not running upon reboot (unless manually being symlinked, whichis definitively not the way to go)

### DIFF
--- a/systemd/osrt-totest-manager@.service
+++ b/systemd/osrt-totest-manager@.service
@@ -6,3 +6,7 @@ Type=simple
 User=osrt-totest-manager
 WorkingDirectory=~
 ExecStart=/usr/bin/osrt-totest-manager --verbose run --interval 5 "%i"
+
+[Install]
+WantedBy=multi-user.target
+


jberry-suse:
Did someone migrate to the package services? As #1185 does not reflect the `ttm*` services having changed (ie I didn't do it). Presumably that is when this issue was discovered.

systemd/osrt-totest-manager@.service
```diff
@@ -6,3 +6,7 @@ Type=simple
 User=osrt-totest-manager
 WorkingDirectory=~
 ExecStart=/usr/bin/osrt-totest-manager --verbose run --interval 5 "%i"
+
+[Install]
+WantedBy=multi-user.target
+
```
jberry-suse: nit: extra newline compared to all the other files

DimStar77:
> Did someone migrate to the package services? 

Yes, @nilxam  and me migrated the totest-managers to the osrt-package version

jberry-suse:
Great!

PR #1453, opened 2018-03-12
Add create_sle_weakremovers command

coolo:
It's similiar but not the same as the droplist for openSUSE, so I created
a fork of the openSUSE command

To be called with the target project as first argument and then a
list of old projects

e.g. create_sle_weakremovers SUSE:SLE-15:GA SUSE:SLE-12-SP3:GA \
SUSE:SLE-12-SP3:Update SUSE:SLE-12-SP2:GA SUSE:SLE-12-SP2:Update ...


lnussel:
worth refactoring at some point I guess

PR #1454, opened 2018-03-12
ReviewBot: make interval work without tty

lnussel:


lnussel:
so @bugfinder doesn't have to use screen :)

coolo:
poor Rudi!

jberry-suse:
Still not sure why using screen instead of systemd + journal as is supported by this repo.

coolo:
because of the check that lnussel just patched out? And Rudi doesn't want systemd timers so he can implement a monitor check if the service is running - so prefers the interval check... that didn't work without screen until this PR

jberry-suse:
IT worked out something for such services on packagelists that use timers since you really only care if it ever exited with failure, but seems like it go a bit messed up when lnussel changed the service type as I get more emails.

PR #1455, opened 2018-03-12
leaper: allow comment override also on openSUSE

lnussel:


leaper.py
```diff
@@ -404,6 +404,13 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
                 self.needs_legal_review = True
                 return True
 
+        if origin_same is False and self.override_allow:
```
jberry-suse: Based on where `origin_same` is set this could be called when it is not set.

jberry-suse:
Currently, one should be able to re-open request after making comment to achieve the same thing.

lnussel:
yeah one could but this risks that things slip through

leaper.py
lnussel: ? you lost me here

jberry-suse: If the package is not in lookup file (presumably new package) `origin_same` will not be set and this condition will cause crash.

```python
if package in self.lookup_150:
    origin = self.lookup_150[package]

# ...

if origin:
    origin_same = self._check_same_origin(origin, src_project)
    # ...
else:
    # ...., but not always return

# ...

# origin_same amay not be set if not origin
if origin_same is False and self.override_allow:
```

Unless there is some larger condition combination that is not obvious when looking at the code that would prevent this from happening, but given how long this function is that should likely be protected against in-case of future change even if it exists.

PR #1459, opened 2018-03-15
Avoid boolean dependencies for repo checker

coolo:
SLE uses them for some module recommending hacks

### DIFF
--- a/CreatePackageDescr.pm
+++ b/CreatePackageDescr.pm
@@ -141,6 +141,8 @@ sub package_snippet($) {
 
     $out .= "+Rec:\n";
     foreach my $prv ( @{ $qq{5046} || [] } ) {
+        # ignore boolean dependencies
+        next if $prv =~ m/^\(/;
         $out .= "$prv\n";
     }
     $out .= "-Rec:\n";


mlschroe:
I think ignoring them when writing the repo data is a bad idea. You should ignore them when they are read/processed.

coolo:
That's done within libsolv-tools. And I only ignore them for recommends, that don't have much relevance for installcheck anyway

mlschroe:
So where are they hurting you? Why not ignore them there?

coolo:
This is how they hurt:
```
could not add repo /tmp/repochecker-K4GL7aV/packages: susetags: line 1115880: unknown relation: 'if'

mlschroe:
Oh, so you're using a very old libsolv version? How about switching to something more recent?

coolo:
you're free to convince maintenance to release a newer version for sp3 - until then I'm fine with having recommends ignored in repo checker ;)

jberry-suse:
This was encountered with _Factory_ already and while on an SP3 machine I created [`libsolv` with richdeps enabled](https://build.opensuse.org/package/show/home:jberry:packagelists/libsolv). The `repo-checker` was since moved to a Tumbleweed container as suggested.

Either option works, but this would mean nuking richdeps checks in _Factory_.

jberry-suse:
In general, given the `repo-checker` is used to check pre-released products it would seem to make sense to run it on either updated tool stacks or in a container like we do for openSUSE.

coolo:
I don't disagree, but this PR still doesn't harm

jberry-suse:
> I don't disagree, but this PR still doesn't harm

Does this not `# ignore boolean dependencies` which will now do so for _Factory_ and _Leap_?

coolo:
Yeah - but Factory can't use boolean dependencies at all as it's using kiwi to build products (according to @DimStar77) and Leap doesn't use this module for anything than repo checker. And again: repo checker doesn't care for recommends. So even if tons of boolean dependencies in recommends pop up: it doesn't matter.

jberry-suse:
That makes sense, but first I've heard mention of _recommends_ as the title and comment would lead one to believe it is all boolean dependencies. Presumably something to do with `$qq{5046}` means recommends.

So this actually "repo_checker: avoid boolean recommends" or similar. Thanks for clarifying.

coolo:
well, recommends are dependencies too :)

http://rpm.org/user_doc/boolean_dependencies.html

jberry-suse:
Sure, but you are indicating it does not ignore all boolean dependencies? Either it does or doesn't the underlying word choice is not particularly relevant.

coolo:
I only ignore 5046 dependencies - which are recommends :)

jberry-suse:
I am not sure what is unclear about my comments. I understand what it does now that you mentioned (I wrote 5046 above), but glancing at code and reading commit message/PR title/PR description do not indicate that. Reading those one would assume all since it does not denote a subset.

PR #1461, opened 2018-03-15
Use opensuse/tumbleweed in the Dockerfile

### LINKED ISSUE #1460: Update CI Dockerfile for openSUSE dockerhub namespace changes
Per [Changes to the openSUSE images on Docker Hub](https://lists.opensuse.org/opensuse-factory/2018-03/msg00389.html) from `opensuse-factory` list:

> - In a week, opensuse:tumbleweed and opensuse:42.2 (EOL) will be removed
> - You'll have to use opensuse/tumbleweed instead of opensuse:tumbleweed

https://github.com/openSUSE/osc-plugin-factory/blob/24ea059e15a7da08f52d869a5cec68e4ef316632/dist/ci/Dockerfile#L4

Vogtinator:
opensuse:tumbleweed will be removed and we want dogfooding anyway.

Fixes #1460. (added by jberry-suse)

### DIFF
--- a/dist/ci/Dockerfile
+++ b/dist/ci/Dockerfile
@@ -1,7 +1,7 @@
 # https://docs.docker.com/engine/reference/builder/
 # Used for TEST_SUITE=distribution and deployment to OBS.
 
-FROM opensuse:tumbleweed
+FROM opensuse/tumbleweed
 MAINTAINER Jimmy Berry <jberry@suse.com>
 
 # https://github.com/openSUSE/obs-service-set_version/issues/44 python-packaging


jberry-suse:
From travis log:

```
latest: Pulling from opensuse/tumbleweed
Status: Downloaded newer image for opensuse/tumbleweed:latest
```

Which would seem to indicate it works.

jberry-suse:
Plus old output:

```
20180122-0 -> 20180313-0
```

vs new

```
20180312-0 -> 20180313-0
```

jberry-suse:
Perhaps we should wait until after #1460 is merged in-case of any changes since there is no rush.

Vogtinator:
> Perhaps we should wait until after #1460 is merged in-case of any changes since there is no rush.

I don't see any advantages there - the pushed images are supposed to always work.
If they break and we find it this way, it's easier to fix it than after I deleted the `opensuse:tumbleweed` images.

jberry-suse:
Referring to name changes unless `opensuse/tumbleweed` has been around a while, but likely won't matter.

Vogtinator:
> Referring to name changes unless opensuse/tumbleweed has been around a while, but likely won't matter.

It's been around since Tuesday :-)

The new name was part of the official announcement, so changing it would be a bad idea.

jberry-suse:
> so changing it would be a bad idea.

Indeed, but stranger things happen all the time.

PR #1467, opened 2018-03-16
Change references to osc-plugin-factory to openSUSE-release-tools.

jberry-suse:
The abichecker is not deployed via package so in theory these changes will not effect it, but would need to verify.

- rename [issue-db](https://github.com/jberry-suse/openSUSE-release-tools-issue-db) repo
- travis-ci seems to have new name automatically
- coveralls.io does not so I left badge
- caches are fine if lost and refreshed
- usage of `save_cache_path()` either did not include page name or is also fine. for instance `save_cache_path('opensuse-packagelists'...)`
- spec setup with this in mind so only one variable to change
- rest is user facing documentation which works as is, but should be updated
- updated openSUSE-release-tools package to point to new repo

No rush as I'll be out next week, but presumably no fallout unless deployed.


jberry-suse:
It could be a good idea to remove old cache dir in a relatively common tool like `osc-staging` that way all staging masters have auto clean.

PR #1471, opened 2018-03-23
totest-manager: Tumbleweed moved to the new product builder

DimStar77:
Since openSUSE:Factory moved to the new prod builder, we need this now for TTM

I've been running it from the osc-plugin-factory for the last few days to get it sorted (osrt-totest-manager was disabled)

as TW was the last product on the old builder, I went the simple route of ```sed /_product/000product/g``` to get going quickly with TW snapshots again.

We can start looking into some cleanups/simplifications after that

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -100,7 +100,7 @@ class ToTestBase(object):
     def get_current_snapshot(self):
         """Return the current snapshot in the test project"""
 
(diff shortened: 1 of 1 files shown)


okurz:
LGTM

okurz:
https://progress.opensuse.org/issues/20890 looks like the related ticket in the issue tracker about "openSUSE Release Process". Care to comment there what is missing?

PR #1474, opened 2018-03-24
repo-checker: check delete requests for runtime deps

DimStar77:
I took a stab at this old P1 issue

Tested with the current gegl-unstable delete request:

```
$ ./repo_checker.py --dry  --skip-cycle --debug --force id 590049
[D] 590049: openSUSE:Factory:Staging:G ready
[D] requests: 0 skipped, 1 queued
[I] checking 590049
[W] gegl-unstable is still a build requirement of:

- bundle-lang-gnome
- gimp
- gnome-photos
- openttd-opengfx
[W] gegl-unstable provides runtime dependencies to:

- gimp
- gnome-photos
[D] removing previous comment on 590049
[D] adding comment to 590049: <!-- RepoChecker state=seen result=failed -->

gegl-unstable is still a build requirement of:

- bundle-lang-gnome
- gimp
- gnome-photos
- openttd-opengfx

gegl-unstable provides runtime dependencies to:

- gimp
- gnome-photos
[I] 590049 ignored
```

In many cases, build deps and runtime deps will have some overlap, but as can already be seen: not all build deps result in runtime deps - and we know from the past that not all runtime deps stem from build deps.

### DIFF
--- a/osclib/cleanup_rings.py
+++ b/osclib/cleanup_rings.py
@@ -2,6 +2,7 @@ from xml.etree import cElementTree as ET
 
 from osc.core import makeurl
 from osc.core import http_GET
+from osclib.core import fileinfo_ext_all
 
 import urllib2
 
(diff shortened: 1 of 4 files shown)


DimStar77:
screenshot from 2018-03-24 13-55-43

nilxam:
LGTM :+1:

fcrozat:
LGTM too and it works fine, I tested on a runtime only dependency delete request on SLE-15 on IBS.

PR #1477, opened 2018-03-28
rebuildpacs: kernel checkin needs to trigger installation-images

lnussel:


DimStar77:
That's a band aid which is far away from complete - all yast modules inside i-i need to trigger a rebuild of i-i, like anything else too.

lnussel:
right, I guess we need to remember to explicitly trigger it when checkin in and have rebuild=local

PR #1478, opened 2018-03-29
conf: make sure variable is not None before do lower()

nilxam:
Fix AttributeError: 'NoneType' object has no attribute 'lower' error


nilxam:
I hot-patched it on tortuga btw.

nilxam:
I merge it now in case SLE needs it as well.

PR #1480, opened 2018-03-29
Change remaining usage of _product to 000product

DimStar77:
All openSUSE distros in developments switched to the new product builder, meaning
they no longer use _product as the package container, but operate on the standardized
name 000product.

Addresses https://progress.opensuse.org/issues/20890


DimStar77:
@lnussel AFAIK, this impacts mainly tools that are used by Leap - things like biarch tool I was actually surprised to see not fixed, as this means this can't have been working for Leap 15 - since you switched to 000product long ago. Is this still in use? Otherwise we might also consider cleaning out stuff here

lnussel:
yes it is in use but in the worst case it just treated 000* as regular package which didn't harm in this case

PR #1488, opened 2018-04-16
totest manager fixes for Factory:ARM 

dirkmueller:


totest-manager.py
```diff
@@ -707,7 +707,11 @@ def jobs_num(self):
 
 class ToTestFactoryARM(ToTestFactory):
     main_products = ['000product:openSUSE-cd-mini-aarch64',
-                     '000product:openSUSE-dvd5-dvd-aarch64']
+                     '000product:openSUSE-dvd5-dvd-aarch64',
+                     '000product:openSUSE-ftp-ftp-armv7hl',
+                     '000product:openSUSE-ftp-ftp-armv6hl',
+                     'JeOS'
```
DimStar77: Does that work? JeOS is not built in repo images (product_repo) arch local (product_arch)

```is_snapshottable``` is presumably never returning true in this case, as openSUSE:Factory:ARM/JeOS/images/local is never 'succeeded', but rather 'excluded'

dirkmueller: good point, this is correct. I'll postpone this to a followup commit, first I need to remodel it into a  :live subproject

DimStar77:
Please adjust the commit message of e247655: it still mentions the addition of JeOS; other than that, :+1:

dirkmueller:
@DimStar77 done

PR #1487, opened 2018-04-13
unmaintained: tool to find umaintained binaries sourced from SLE.

jberry-suse:
- 96f46dab5bb903af86418037ecb38c811b97aa9e:
    unmaintained: tool to find umaintained binaries sourced from SLE.

- 920478b62eec80c42d8074b90f3da28470dd6a94:
    osclib/core: provide source_file_load() from StagingAPI.

Fixes [poo#28815](https://progress.opensuse.org/issues/28815)

PR #1489, opened 2018-04-16
Makefile: package all executable files in bindir.

jberry-suse:
- 6122aa4b6a4f6d4969a353699b5ceafdcc14a453:
    Makefile: package all executable files in bindir.
    
    Rather than only expose the tools deployed via package it makes sense to
    provide all tools to aid use by others via package.

- 722041a0e5b64a7eff383436f28fe82ae73bd94b:
    metrics_release: remove executable flag left over from development.
    
    Ignore metrics from nosetests loader since it is no longer executable.

The following are the new bins provided. Remove the executable bit to drop any undesriables, but the list seems fine on inspection. Some likely to be moved to subpackages if deployment is worked out for them, but better something rather than nothing.

This should also allow maintenance to easily use #1487 by installing main package.

```diff
+./biarchtool.py
+./bs_mirrorfull
+./build-fail-reminder.py
 ./check_maintenance_incidents.py
+./checknewer.py
+./check_source_in_factory.py
 ./check_source.py
+./check_tags_in_requests.py
+./compare_pkglist.py
+./create_test_dvds.sh
 ./devel-project.py
+./fcc_submitter.py
+./findfileconflicts
+./issue-diff.py
 ./leaper.py
+./legal-auto.py
 ./manager_42.py
 ./metrics.py
+./metrics_release.py
+./obs_clone.py
+./openqa-maintenance.py
 ./pkglistgen.py
+./rebuildpacs.pl
 ./repo_checker.py
+./requestfinder.py
+./status.py
 ./suppkg_rebuild.py
+./sync-rebuild.py
+./umaintained.py
 ./totest-manager.py
 ./update_crawler.py
```


jberry-suse:
Rebased after #1148 to resolve conflict.

PR #1493, opened 2018-04-17
publish appliances for ARM alongside of normal product

dirkmueller:


DimStar77:
The error message in the logs was:

```
2018-05-31 14:31:55,347 - totest-manager:1009 ERROR HTTP Error 404: repository 'openSUSE Factory ARM Live/images' has no architecture 'i586'
```

PR #1495, opened 2018-04-18
Rework opensuse maintenance

coolo:
Readd openSUSE support to the openqa bot - but this is so far only taking care of the incidents


coolo:
Fixed both

PR #1502, opened 2018-04-23
devel-project: reword notifier

lnussel:


jberry-suse:
Tested package arg locally.

devel-project.py
```diff
@@ -118,7 +120,22 @@ def notify(args):
     subject = 'Packages you maintain are present in {}'.format(args.project)
     for userid, package_identifiers in maintainer_map.items():
         email = entity_email(apiurl, userid)
-        message = 'The following packages you maintain are in {}:\n\n- {}'.format(
+        message = """This is a friedly reminder about your packages in {}.
```
jberry-suse: `s/friedly/friendly/`

devel-project.py
```diff
@@ -118,7 +120,22 @@ def notify(args):
...
+
+Please verify that the included packages are working as intended and
+have versions appropriate for a stable release. Please submit any
+changes until April 26th latest.
```
jberry-suse: > Please submit any changes until April 26th latest.

This is a bit odd. Perhaps

```
Changes may be submitted until April 26th [at the latest].
```

jberry-suse:
Since emails already sent the changes only matter for the future I suppose. :)

jberry-suse:
Also note the idea was the tool could be used for SLE or Factory, but without adding templates they'd just need to adjust message.

lnussel:
travis fail :(

jberry-suse:
#1575

PR #1512, opened 2018-04-26
Begin porting to python3 (osc branch)

coolo:
This is just the start to experiment with the osc3 branch

### DIFF
--- a/oqamaint/cli.py
+++ b/oqamaint/cli.py
@@ -55,7 +55,7 @@ class CommandLineInterface(ReviewBot.CommandLineInterface):
                 elif i.startswith('openSUSE'):
                     project[i] = openSUSEUpdate(j)
                 else:
-                    raise "Unknown openQA", i
+                    raise Exception("Unknown openQA", i)
         return project, target, api
 
     def postoptparse(self):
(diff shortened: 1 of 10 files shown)


jberry-suse:
So there is movement from osc now? Presumably the `python3` branch on `osc` repo?

coolo:
Yes. But it looks like we can't port to python3 branch and keep sane python2 :(

At least I get this:

```
('adi:1', '\x1b[37mreview:', '\x1b[33msle-release-managers\x1b[39m for \x1b[36mpolicycoreutils\x1b[39m[163031]')
('adi:5', '\x1b[35mbuilding')
('adi:6', '\x1b[35mbuilding')
('adi:10', '\x1b[35mbuilding')

**Comment by jberry-suse** [comment]
That is our code and the change from

```python
print "foo" "bar"
```

to

```python
print("foo", "bar")
```

jberry-suse:
Looks like already replaced with string concatenation which should do the trick.

coolo:
I'm using this for quite a while now (using python2), so I think we can accept it as first step :)

coolo:
I stalled the porting btw as the python3 branch also changed the config parser - and that was a bit too much for me to cover

PR #1519, opened 2018-04-29
[openqa-maintenance] Use to_xml methot of Request object

mimi1vx:
Forgotten method call caused a problem in check_request and killed bot
too early before proper incident results resolution

### DIFF
--- a/oqamaint/openqabot.py
+++ b/oqamaint/openqabot.py
@@ -373,12 +373,13 @@ class OpenQABot(ReviewBot.ReviewBot):
         return msg.rstrip('\n')
 
     def check_one_request(self, req):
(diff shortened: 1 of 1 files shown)


okurz:
Typo in subject line: methot

oqamaint/openqabot.py
```diff
@@ -373,12 +373,13 @@ def summarize_openqa_jobs(self, jobs):
         return msg.rstrip('\n')
 
     def check_one_request(self, req):
+        ret = None
```
coolo: I solved this differently - see the other PR

coolo:
merged in 1520 - and fixed it up in the process

PR #1520, opened 2018-04-30
Fix openqa crash

coolo:


oqamaint/update.py
```diff
@@ -56,15 +56,14 @@ def get_max_revision(self, job):
     def settings(self, src_prj, dst_prj):
         s = self._settings.copy()
 
+        patch_id = src_prj.split(':')[-1]
```
mimi1vx: this is equal to `def patch_id(repo):`? In INCIDENT patch is needed a real name of patch of which is incident number only small part

oqamaint/openqabot.py
```diff
@@ -408,7 +411,7 @@ def check_one_request(self, req):
             self.logger.error(traceback.format_exc())
             return None
 
-        return ret
+        return
```
mimi1vx: define ret=None at beginning of function and not needed explicit return on every if branch

oqamaint/update.py
coolo: hmm, not for openSUSE. that's what fooled me.

The actual problem is that https://download.opensuse.org/repositories/openSUSE:/Maintenance:/8006/openSUSE_Leap_15.0_Update/repodata/ does not have an updateinfo.xml - which might be a bug.

coolo: CC @andreasstieger

oqamaint/openqabot.py
coolo: I know, but I like it better to have the returns more local. That's why I changed it to begin with - I just forgot about one branch

oqamaint/update.py
coolo: So I checked and you are right - my patch only works for openSUSE.

And http://download.opensuse.org/repositories/openSUSE:/Maintenance:/8027/openSUSE_Leap_15.0_Update/repodata/ also doesn't have updateinfo, but it must have had at some point.

coolo:
29234921e34c9199fc65363a1a50f2c522eae914 is the next best thing

coolo:
there is always a default return value in python

mimi1vx:
@coolo but not default variable `ret`

so `return` returns `None`
but `return ret` without defined ret throws `NameError`

coolo:
I see now - I managed to confuse myself with branches.

coolo:
(git branches not python if/else branches - that statement wasn't clear :)

PR #1522, opened 2018-05-02
Fix target project for :Ports

dirkmueller:


pkglistgen.py
```diff
@@ -1146,6 +1146,7 @@ def do_update_and_solve(self, subcmd, opts):
             # TODO Continue supporting #1297, but should be abstracted.
             main_repo = 'ports'
             opts.project += ':Ports'
+            target_project += ':Ports'
```
jberry-suse: When `target_project` is then passed to `project_list_family_prior()` the following invalid output is returned since `:Ports` isn't a base product name.

```python
['openSUSE:Leap:15.0:Ports', 'openSUSE:Leap:15.0:Rings', 'openSUSE:Leap:15.0:SLE-workarounds', 'openSUSE:Leap:15.0:Staging', 'openSUSE:Leap:15.0:ToTest']
```

Whereas without suffix:

```python
['openSUSE:Leap:15.0', 'openSUSE:Leap:42.3', 'openSUSE:Leap:42.2', 'openSUSE:Leap:42.1']
```

Presumably you want this so that the drop list functions and retrieves `:Ports` solv files for prior Leap releases? If so the drop list generation will need to be worked on a bit as the download urls come from the config for the product release and do not contain the `:Ports` paths.

As such this should be resolved differently as `target_project` is intended to be the "main release product" just like it is `openSUSE:Leap:15.0` even for `openSUSE:Leap:15.0:Staging:A`.

Basically the lines in `solv_cache_update()` where it gets the `baseurl` need to figure out the `:Ports` path. Otherwise I imagine it works.

jberry-suse:
After analyzing further some questions arise given the way `:Ports` are setup vs the parent project. Is the 000product generation to be done separately per architecture since they are handle in separate repos and likely have rather different sets of packages? If so would this not require multiple projects/copies of that package or delineation within the package?

To explain further, I went to tweak the drop list generation to handle the `:Ports` URLs, but realized they are separated by arch, but pkglistegen only runs once for the whole `:Ports` project. As such for which arch would it be generating a drop list? I was thinking I had to add a loop within `solv_cache_update()`, but that is just wrong since there is nowhere to store the different lists. As such it would seem to make sense to loop over ports-archs in `do_update_and_solve()` similar to how stagings are treated, but again still no where to store the different runs.

Unless I am missing something it seems either a "master" arch needs to be chosen which pkglistgen bases everything off of (which seems like bad idea). Or there should be a separate project per port which would fit the paradigm much better where they could be treated as separate runs with separate output.

dirkmueller:
I believe I don't actually need this change, as you correctly identified (thanks for the detailed writeup!)

PR #1523, opened 2018-05-05
Handle armv6/armv7 with totest manager as well

dirkmueller:
Recently we decided to incorporate armv6 and armv7 publishing into
ToTest as well.


okurz:
LGTM

PR #1526, opened 2018-05-09
manager_42: check if SLE-workarounds are no longer utilized and request removal.

jberry-suse:
Created to resolve [poo#36012](https://progress.opensuse.org/issues/36012).

Should be fine to include as part of regular run and since:

```python
            if project.endswith(':SLE-workarounds'):
                self.sle_workarounds = project
```

should never be true for SLE itself this code will never execute there.

### DIFF
--- a/manager_42.py
+++ b/manager_42.py
@@ -21,6 +21,8 @@
 # SOFTWARE.
 
 import argparse
+from dateutil.parser import parse as date_parse
+from datetime import datetime
 import itertools
 import logging
 import sys
@@ -28,7 +30,10 @@ from xml.etree import cElementTree as ET
 
 import osc.conf
 import osc.core
+from osc.core import get_commitlog
+from osc.core import get_request_list
 import urllib2
+import subprocess
 import time
 import yaml
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


jberry-suse:
Invoked with `--all` and the following diff.

```diff
diff --git a/manager_42.py b/manager_42.py
index 5af9044..1816811 100755
--- a/manager_42.py
+++ b/manager_42.py
@@ -176,6 +176,9 @@ class Manager42(object):
         """Main method of the class that runs the crawler."""
 
         for package in sorted(packages):
+            if package not in self.packages[self.sle_workarounds]:
+                continue
+
             try:
                 self.check_one_package(package)
             except urllib2.HTTPError as e:
```

Generated 47 deletes requests out of 51 packages in SLE-workarounds.

The following packages remain:

- still sourced from SLE-workarounds
  - patterns-base
- updated in SLE-workarounds in last 24 hours
  - MozillaFirefox
  - go1.7
  - netcdf

jberry-suse:
I spot checked 10 plus the remaining ones to verify the output makes sense.

jberry-suse:
Requests were accepted so seems the code should as well.

PR #1532, opened 2018-05-10
dist/spec: exclusively build on x86_64 to avoid needless builds and OBS bug.

jberry-suse:
Related to openSUSE/open-build-service#4373 as disabling s390x leaves old binaries in repo-md while publishing new ones :(((((((.


jberry-suse:
Created openSUSE/open-build-service#4968 as another `disabled` state related OBS bug.

jberry-suse:
Workaround two bugs in one line! https://bugzilla.redhat.com/show_bug.cgi?id=1298668

PR #1534, opened 2018-05-11
systemd: ensure all services have a relaxed TimeoutStartSec, SyslogIdentifier, and retry as needed.

jberry-suse:
- 6faaf4226059830a0eb106cedd37ca201469b414:
    systemd: add service retry to support-rebuild which 502s regularly.

- 7b1ff6e2f6eb75db1ec72bc6c1f6f432db995714:
    systemd: include SyslogIdentifier in each service to improve journal.

- 1a3f344d418b0911b16db843f34da0ae53a13c2a:
    systemd: ensure all services have a relaxed TimeoutStartSec.
    
    The value should allow ample time for seldom larger runs and such and is
    purely a safe-guard from a service being stuck for weeks until noticed. If
    any values turn out to be too small simply increase.

- e3d89e525e80bd7d1d8c818accc81f6028133a0f:
    systemd: drop Type=simple to be consistent since default not required.

My timeouts from `repo-checker-*` services were removed, but we really need some values on these as we get a runaway every couple months.

```
● openqa-comments@Factory.service - Comment openQA/Build results on staging project Factory
   Loaded: loaded (/etc/systemd/system/openqa-comments@.service; disabled; vendor preset: disabled)
   Active: active (running) since Fri 2018-02-23 10:09:54 CET; 2 months 15 days ago
```

I noticed when migrating (#1185) and enabled for _Leap 15.0_. Not sure if/how it was disabled.

I added a retry for service that regularly fails to at minimum improve monitoring as I will require IT add all these services.


jberry-suse:
Since everything was changed to `Type=simple` instead of `oneshot` :'( this parameter needs to change too.

```
$ sed -i 's/TimeoutStartSec/RuntimeMaxSec/' systemd/*.service factory-package-news/systemd/*.service
```

jberry-suse:
For the curious, every service file in this repo is now actively being used. :)

PR #1539, opened 2018-05-14
ttm: Allow to ignore job failures without bug reference

coolo:
This is supposed to be used for temporary failures that do not deserve
a carry over as they are sporadic but should not stop a snapshot release
nor are they worth it to wait a job duplicate

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -255,7 +255,8 @@ class ToTestBase(object):
                         labeled = comment['id']
                     if re.search(r'@ttm:? ignore', comment['text']):
                         to_ignore = True
-                ignored = len(refs) > 0
+                # to_ignore can happen with or without refs
+                ignored = True if to_ignore else len(refs) > 0
                 for ref in refs:
                     if ref not in self.issues_to_ignore:
                         if to_ignore:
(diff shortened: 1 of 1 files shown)


DimStar77:
only travis is not happy:
```./totest-manager.py:278:32: E121 continuation line under-indented for hanging indent```

File: totest-manager.py [hunk 1]
```diff
@@ -268,13 +269,17 @@ def overall_result(self, snapshot):
...
                         data = {'text': 'label:unknown_failure'}
-                        self.openqa.openqa_request(
-                            'POST', 'jobs/%s/comments' % job['id'], data=data)
+                        if self.dryrun:
+                            logger.info("Would label {} as unknown".format(job['id']))
+                        else:
+                            self.openqa.openqa_request(
+                                'POST', 'jobs/%s/comments' % job['id'], data=data)
                 elif labeled:
+                    text = 'Ignored issue' if len(refs) > 0 else 'Ignored failure'
```
okurz: How about putting a real openQA label in here as well so that we can distinguish these in openQA easier and potentially even visualize differently?

okurz: label:ignored_failure

PR #1542, opened 2018-05-14
Use OBS Attribute for ignored issues

### LINKED ISSUE #1403: totest-manager: put issue file to dashboard
the issue file is currently local which makes it hard to run ttm locally. Should be stored in the dashboard too.

coolo:
Fixes #1403

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -33,8 +33,6 @@ from osc.core import makeurl
 
 logger = logging.getLogger()
 
-ISSUE_FILE = 'issues_to_ignore'
-
 # QA Results
 QA_INPROGRESS = 1
 QA_FAILED = 2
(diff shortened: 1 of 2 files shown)


coolo:
This also includes #1541

totest-manager.py
```diff
@@ -257,17 +274,28 @@ def overall_result(self, snapshot):
...
-                if not ignored:
+                if ignored:
+                    if labeled:
+                        text = 'Ignored issue' if len(refs) > 0 else 'Ignored failure'
+                        # remove flag - unfortunately can't delete comment unless admin
+                        data = {'text': text}
+                        self.openqa.openqa_request(
+                            'PUT', 'jobs/%s/comments/%d' % (job['id'], labeled), data=data)
+
+                    logger.info("job %s failed, but was ignored", jobname)
```
okurz: Could you extend this to "job %s failed, but was ignored due to '@ttm ignore' comment in job" to be explicit?

totest-manager.py
```diff
@@ -62,15 +63,31 @@ def __init__(self, project, dryrun=False, api_url=None, openqa_server='https://o
...
-        if os.path.isfile(self.issuefile):
-            with open(self.issuefile, 'r') as f:
-                for line in f.readlines():
-                    self.issues_to_ignore.append(line.strip())
+        self.load_issues_to_ignore()
         self.project_base = project.split(':')[0]
         self.update_pinned_descr = False
 
+    def load_issues_to_ignore(self):
+        url = self.api.makeurl(['source', self.project, '_attribute', 'OSRT:IgnoredIssues'])
```
okurz: Does this also exist in IBS? We are about to use ttm for SLE as well so I hope this does not break it

coolo: it could exist - it doesn't yet

totest-manager.py
coolo: we don't know at this point if the comment contained any action - or if this was a carry over who lead to the ignore. So no.

coolo:
Actually on the way to the office one thing crossed my mind: we won't see any comment from ttm in those jobs without bug reference. Because ttm only comments 'label:unknown_failure' once there is a issue reference from carry over or reviewer. And those labels are then updated once the issue is ignored. Adding labels for ignored *jobs* is actually a different thing then.

PR #1544, opened 2018-05-15
Add product scan to legal-auto

coolo:


kraih:
I'll be testing this in production on the legaldb over the next few days.

legal-auto.py
```diff
@@ -178,6 +208,108 @@ def _set_review(self, req, state):
                                      by_user=self.review_user, message=self.message)
         self.delete_from_db(req.reqid)
 
+    def _pkl_path(self):
+        CACHE_DIR = os.path.expanduser('~/.cache/osc-plugin-factory')
```
jberry-suse: nit: use the new project name and perhaps xdg as done elsewhere.

coolo: yeah - these are changes done in june 2017 (and used in production since then ;)

kraih:
Seems to be working smoothly so far. Running it once a day around midnight with a systemd timer.

legal-auto.py
lnussel: could you please fix this nevertheless?

jberry-suse: why merge?

PR #1548, opened 2018-05-16
Add AMQP support to TTM

asdil12:
See https://progress.opensuse.org/issues/27735

Example Message:

```json
'opensuse.ttm.build.failed': '{
    "project": "openSUSE:Factory",
    "build": "20180516",
    "failed_jobs": {
        "relevant": [677715, 677717, 677721, 677734, 677744, 677746, 677749, 677750, 677776, 677797, 677895, 677896, 677925, 677976, 677977, 678017],
        "ignored": [677719, 677722, 677725, 677742, 677770, 677772, 677773, 677873, 677889, 677900, 677972, 677978, 677979, 677985]
    }
}'

```

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -6,6 +6,7 @@
 # (C) 2014 aplanas@suse.de, openSUSE.org
 # (C) 2014 coolo@suse.de, openSUSE.org
 # (C) 2017 okurz@suse.de, openSUSE.org
+# (C) 2018 dheidler@suse.de, openSUSE.org
 # Distribute under GPLv2 or GPLv3
 
 from __future__ import print_function
@@ -21,6 +22,7 @@ import logging
 import signal
 import time
 import yaml
+import pika
 
 from xml.etree import cElementTree as ET
 from openqa_client.client import OpenQA_Client
@@ -66,6 +68,7 @@ class ToTestBase(object):
         self.load_issues_to_ignore()
         self.project_base = project.split(':')[0]
         self.update_pinned_descr = False
+        self.amqp_url = osc.conf.config.get('ttm_amqp_url')
 
     def load_issues_to_ignore(self):
         url = self.api.makeurl(['source', self.project, '_attribute', 'OSRT:IgnoredIssues'])
@@ -247,16 +250,17 @@ class ToTestBase(object):
 
         jobs = self.find_openqa_results(snapshot)
 
+        self.failed_relevant_jobs = []
+        self.failed_ignored_jobs = []
+
         if len(jobs) < self.jobs_num():  # not yet scheduled
             logger.warning('we have only %s jobs' % len(jobs))
             return QA_INPROGRESS
 
-        number_of_fails = 0
         in_progress = False
         for job in jobs:
             # print json.dumps(job, sort_keys=True, indent=4)
             if job['result'] in ('failed', 'incomplete', 'skipped', 'user_cancelled', 'obsoleted', 'parallel_failed'):
-                jobname = job['name']
                 # print json.dumps(job, sort_keys=True, indent=4), jobname
                 url = makeurl(self.openqa_server,
                               ['api', 'v1', 'jobs', str(job['id']), 'comments'])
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -895,6 +926,7 @@ def get_optparser(self):
         parser.add_option("--dry", action="store_true", help="dry run")
         parser.add_option("--debug", action="store_true", help="debug output")
         parser.add_option("--verbose", action="store_true", help="verbose")
+        parser.add_option("--config", default="~/.ttm.cfg", help="path to config file")
```
lnussel: load_first_config from xdg.BaseDirectory please

totest-manager.py
```diff
@@ -541,6 +547,31 @@ def totest(self):
             # with a new release
             can_release = False
 
+        if self.amqp_url:
```
coolo: https://refactoring.com/catalog/extractMethod.html

asdil12:
Updated patch.

totest-manager.py
```diff
@@ -942,7 +979,18 @@ def _setup_totest(self, project):
                 project, ', '.join(self.totest_class))
             raise cmdln.CmdlnUserError(msg)
 
-        return self.totest_class[project](project, self.options.dry, self.options.obs_api_url, self.options.openqa_server)
+        amqp_url = None
+        configpath = os.path.expanduser(self.options.config)
```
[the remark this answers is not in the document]
asdil12: Because this is completely unrelated to osc.

asdil12:
Updated.

asdil12:
> why not take that from the oscrc? Avoids one more config file

@coolo ttm will now read the `ttm_amqp_url` from `oscrc`.

totest-manager.py
```diff
@@ -575,6 +582,35 @@ def totest(self):
             self.update_totest(new_snapshot)
             self.write_version_to_dashboard("totest", new_snapshot)
 
+    def send_amqp_event(self, current_snapshot, current_result):
+        if not self.amqp_url:
+            logger.info('No ttm_amqp_url configured in oscrc - skipping amqp event emission')
```
okurz: That would trigger a lot but ok

totest-manager.py
```diff
@@ -575,6 +582,35 @@ def totest(self):
...
+            return
+
+        logger.info('Sending AMQP message')
```
okurz: I consider this too verbose for "info" and not adding much value. How about a `logger.debug` with at least the msg_topic reflected? So one line further below

totest-manager.py
```diff
@@ -575,6 +582,35 @@ def totest(self):
...
+        msg_topic = '%s.ttm.build.%s' % (self.project_base.lower(), self._result2str(current_result))
+        msg_body = json.dumps({
+            'build': current_snapshot,
+            'project': self.project,
+            'failed_jobs': self.failed_jobs,
+            'ignored_failed_jobs': self.ignored_failed_jobs,
+        })
```
okurz: Can you give us an example message in the PR description?

totest-manager.py
coolo: that sounds indeed like a .debug

totest-manager.py
asdil12: done

okurz:
+1

DimStar77:
> spamming IRC? :)

I hope we don't get every 5 minutes (bot run time) that the snapshot 'still failed'

asdil12:
> I hope we don't get every 5 minutes (bot run time) that the snapshot 'still failed'

`suse_msg` has can be setup so that it will not spam similar messages

totest-manager.py
```diff
@@ -21,6 +22,7 @@
 import signal
 import time
 import yaml
+import pika
```
DimStar77: Needs to be represented on the rpm package as a dependency (and needs to be provided on SLE12SP3, current OS of packagelist vm - currently ttm is down as the update got deployed)

PR #1552, opened 2018-05-18
dist: Require python2-pika by totest-manager (new amqp dependency)

DimStar77:


dist/package/openSUSE-release-tools.spec
```diff
@@ -202,6 +202,7 @@ BuildArch:      noarch
 # TODO Update requirements.
 Requires:       osclib = %{version}
 Requires:       python-openqa_client
+Requires:       python2-pika
```
DimStar77: Depending on what version package we use, we MIGHT need to require python-pika (in case the package we build is not py-singlespec)

DimStar77:
created https://build.opensuse.org/request/show/610282 to openSUSE:Tools (where ttm lives) - imho the missing dep should be living next to the package

dist/package/openSUSE-release-tools.spec
dirkmueller: even if it isn't singlespeced we should rather have a python provide

PR #1555, opened 2018-05-19
openqabot: Quote the module name

coolo:
module names may contain # - especially if the module is loaded twice
and openQA appends #1 and #2


okurz:
LGTM

PR #1565, opened 2018-05-31
Update announcer.py to ask people to report bugs

simotek:
Rather then replying to the generated email we would like people to create new issues on bugzilla.opensuse.org

### DIFF
--- a/factory-package-news/announcer.py
+++ b/factory-package-news/announcer.py
@@ -55,7 +55,7 @@ Please check the known defects of this snapshot before upgrading:
 https://openqa.opensuse.org/tests/overview?distri=opensuse&groupid=1&version=Tumbleweed&build={version}
 
 Please do not reply to this email to report issues, rather file a bug
-on bugzilla.opensuse.org for more information on filing bugs please
+on bugzilla.opensuse.org. For more information on filing bugs please
 see https://en.opensuse.org/openSUSE:Submitting_bug_reports
 
 {text}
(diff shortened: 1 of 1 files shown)


factory-package-news/announcer.py
```diff
@@ -54,9 +54,9 @@
 Please check the known defects of this snapshot before upgrading:
 https://openqa.opensuse.org/tests/overview?distri=opensuse&groupid=1&version=Tumbleweed&build={version}
 
-When you reply to report some issues, make sure to change the subject.
-It is not helpful to keep the release announcement subject in a thread
-while discussing a specific problem.
+Please do not reply to this email to report issues, rather file a bug
+on bugzilla.opensuse.org for more information on filing bugs please
```
DimStar77: Feels like there is a full stop missing after 'bugzilla.opensuse.org'

simotek: I thought about putting a comma, but I tend to not like punctuation after url's especially full stops, they tend to break things in places

DimStar77: actually, a full stop would be no problem, as the 'technically correct' name would actually be ```bugzilla.com.``` (the last dot indicating 'DNS root')

simotek: Done, didn't put one at the end of the paragraph though because that does screw up the wiki.

jberry-suse:
RIP review.tumbleweed.boombatower.com

jberry-suse:
Still not sure about this as there is no clear way to designate show-stopping bugs that everyone would like to be aware of via bugzilla. Just get mixed in with the tasks.

simotek:
Yeah I'm still hoping / expecting that people will report really bad issues to the mailing list once there is a bug report, I tried to imply that this is still ok. If the new approach doesn't work the board has discussed creating another list for reporting tumbleweed / leap beta breakages but others wanted to see how well this approach worked first.

PR #1566, opened 2018-06-01
[openqa-maintenance] remove incidents from repohash

mimi1vx:
In all cases 'incidents' = {}

### DIFF
--- a/data/repos.json
+++ b/data/repos.json
@@ -40,12 +40,12 @@
             "TCM" : "SUSE:Maintenance:Test:SLE-Module-Toolchain:12:x86_64"
          },
          "repos" : [
-            "http://download.suse.de/ibs/SUSE/Updates/SLE-SERVER/12-SP3/x86_64/update/",
-            "http://download.suse.de/ibs/SUSE/Updates/SLE-SDK/12-SP3/x86_64/update/",
-            "http://download.suse.de/ibs/SUSE/Updates/SLE-WE/12-SP3/x86_64/update/",
-            "http://download.suse.de/ibs/SUSE/Updates/SLE-Module-Web-Scripting/12/x86_64/update/",
-            "http://download.suse.de/ibs/SUSE/Updates/SLE-Module-Toolchain/12/x86_64/update/",
-            "http://download.suse.de/ibs/SUSE/Updates/SLE-Module-HPC/12/x86_64/update/"
+            "http://download.suse.de/ibs/SUSE:/Maintenance:/Test:/SLE-SERVER:/12-SP3:/x86_64/update/",
+            "http://download.suse.de/ibs/SUSE:/Maintenance:/Test:/SLE-SDK:/12-SP3:/x86_64/update/",
+            "http://download.suse.de/ibs/SUSE:/Maintenance:/Test:/SLE-WE:/12-SP3:/x86_64/update/",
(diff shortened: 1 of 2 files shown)


coolo:
No. for opensuse incidents is not empty - and I also for SLE don't want to depend on the :Test aggregation script, which is yet another intransparent script.

So better let's find out why incidents is empty for sle.

mimi1vx:
1) calculate_repo_hash is called before calculate_incidents
2) original data schedules aggregate jobs if is update released... Not much interesting moment for qam. -> can be solved using aggregate repositories for incidents under test or add incidents to repohash.
3) for openSUSE Is this codepath same as for SLE. So I guess no incidents used in openSUSE REPOHASH.

coolo:
ok, let me rephrase that: the incidents were correctly passed when I tested it for openSUSE :)

We want to reschedule a repo if the incidents under test changes - but we indeed don't care for incidents that were released (and disappeared as such from being under test). But a simple hash won't give you such options. I considered appending the incident IDs to make it easier to debug at least.

BTW, the repohash of :Test repos also changes if updates are released - as the aggregate script removes it.

coolo:
I stopped the openSUSE bot to check

coolo:
This is the content of the hash for openSUSE:
````
{"8184": 1527523511, "8188": 1527539010, "8193": 1527595136, "8194": 1527595140, "8198": 1527877009, "8203": 1527699631}

mimi1vx:
Hmmm. I bet in tmux instance? So I'd of previous incidents?

Dne ne 3. 6. 2018 18:27 uživatel Stephan Kulow <notifications@github.com>
napsal:

coolo:
tmux? there is no such thing anymore. This is the content if I run the bot locally

mimi1vx:
hmm, with inter loop or without `-n` parameter?

coolo:
Hmm, I only ran it as id X so it calculates the incidents and publishes the update repos. I can check if review -n makes a difference - but I can't see how.

coolo:
See https://github.com/openSUSE/openSUSE-release-tools/pull/1568 - I ran this with review -n 1 and it worked reliable.

Can't test for SLE atm as openqa.suse.de is down :(

mimi1vx:
`-n 1` means every minute right? :D

#1568 looks interesting but SLE has all issues covered in `_TEST_ISSUES` variables + still incident can be totally changed without change of ID ... so repohash from aggregate repository is the most secure way to get job scheduled in time

coolo:
well, -n 1 was only for my debugging :)

And once openqa is up again, I'll check how it works for SLE. The incident hash you're removing contains the repohash per incident. So the digest changes - the additional IDs are only to ease understanding.

mimi1vx:
we run SLE without `-n`, so every run is a clean instance.

  *  https://github.com/openSUSE/openSUSE-release-tools/blob/master/oqamaint/openqabot.py#L463

`incident_repos` is filled with Incidents NR from '-Incidents` tests which in SLE case can be also stagging incidents, not only in test and different than is scheduled for **Updates** 

relevant tests for **Updates** are calculated in https://github.com/openSUSE/openSUSE-release-tools/blob/master/oqamaint/openqabot.py#L213 

-->  move repohash calculation after this + small change in merging incidents can return correct repohash for us

coolo:
I just ran the -n because of your comment.

I'll check SLE and append to 1568. Let's discuss there

PR #1568, opened 2018-06-04
oqamaint: Make debuging of incident list in repohash easier

coolo:
By appending the IDs of incidents considered, e.g.
 'REPOHASH': '9b9ff324e95d25dc6bb216cb45bdd855:8198,8200,8201,8203',

### DIFF
--- a/oqamaint/openqabot.py
+++ b/oqamaint/openqabot.py
@@ -105,7 +105,11 @@ class OpenQABot(ReviewBot.ReviewBot):
             m.update(cs.text)
         # now add the open incidents
         m.update(json.dumps(incidents, sort_keys=True))
-        return m.hexdigest()
+        digest = m.hexdigest()
+        open_incidents = sorted(incidents.keys())
+        if open_incidents:
+          digest += ':' + ','.join(open_incidents)
+        return digest
 
     def is_incident_in_testing(self, incident):
         # hard coded for now as we only run this code for SUSE Maintenance workflow
--- a/oqamaint/openqabot.py
+++ b/oqamaint/openqabot.py
@@ -108,7 +108,7 @@ class OpenQABot(ReviewBot.ReviewBot):
         digest = m.hexdigest()
         open_incidents = sorted(incidents.keys())
         if open_incidents:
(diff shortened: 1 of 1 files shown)


mimi1vx:
still , for sle can be hundreds Incidents + and different than really included in TEST + incidents are stored in `*-TEST_ISSUES` variables ..

coolo:
hundreds per project?

coolo:
if it's so many, we might want to show the highest 10 - this would still make debugging differences easier without exceeding limits

mimi1vx:
now is in staging area over 250 incidents, but I don't understand how duplication of `OS_TEST_ISSUES` helps with debugging issues

plus small complication is INCIDENT's hanging during time ( rarely in openSUSE but in SLE is this common thing)

coolo:
well, for now it helps me debugging if the incidents were correctly added to the repohash :)

coolo:
Found the problem. Let's leave the IDs in the repohash for a moment, it's 'e92bb8e7fcecf2f8da13c9a5d9d29ef9:5092,5364,5584,6682,6689,7058,7194,7426,7433,7451,7452,7453,7454,7455,7456,7457,7458,7459,7460,7461,7462,7509,7529,7621' for sp3, this is bearable

coolo:
deployed to see it in action :)

coolo:
Good thing I did - https://openqa.suse.de/tests/1743213#settings shows that e.g. 5092 is part of the repohash but not part of the ISSUES. I need to look at that later - the state is stale

coolo:
Same for 4826 on sp3 - it's stale too.

mimi1vx:
As I said before  you mixed two different things:

for aggregate jobs:

https://github.com/openSUSE/openSUSE-release-tools/blob/master/oqamaint/openqabot.py#L123 - calculates what will be in OS_TEST_ISSUES etc...

and `self.incidents_repos` from per INCIDENT jobs ( which is used also for staging updates, not only in Maintenance queue with release request)

coolo:
yeah, we need to filter the test_issues out of the incidents_repo hash. If openqa hadn't exploded again, I would have looked into it :)

PR #1569, opened 2018-06-05
TTM: Add norelease option for use with OSD

asdil12:
See https://progress.opensuse.org/issues/27735


okurz:
Uhm, what *is* the difference to "--dry-run" then?

asdil12:
`--dry-run` will not create comments in openQA and also will not save the ignored issues in obs.

okurz:
alright, please add this explanation to the help string of the option then

okurz:
fine :)

PR #1573, opened 2018-06-11
Use attribute to store staging config

coolo:
My goal was to setup SP1 without a dashboard package


osclib/conf.py
```diff
@@ -258,4 +213,4 @@ def apply_remote(self, api):
             self.populate_conf()
         elif config is None:
             # Write empty config to allow for caching.
-            api.dashboard_content_save('config', '')
+            api.attribute_value_save('Config', '')
```
jberry-suse: The caching rules need to be updated to make this useful. Also this will now insert blank configs in gnome project and similar. I don't think the remote-config toggle should be removed.

osclib/conf.py
```diff
@@ -161,7 +119,6 @@
         'lock-ns': None,
         'delreq-review': None,
         'main-repo': 'openSUSE_Factory',
-        'remote-config': False,
```
jberry-suse: This will break the gnome review bot unless either `attribute_value_load()` is changed to handled `404`.

osclib/conf.py
```diff
@@ -246,10 +203,8 @@ def read_section(self, section, defaults):
 
     def apply_remote(self, api):
         """Fetch remote config and re-process (defaults, remote, .oscrc)."""
-        if not conf.config[self.project].get('remote-config', True):
-            return
 
-        config = api.dashboard_content_load('config')
+        config = api.attribute_value_load('Config')
```
jberry-suse: This logic will not work the same since a `404` will be raised and not caught.

jberry-suse:
Personally, I do not see a large benefit from this especially in the light of potential changes we have in the works. Additionally, it breaks metrics parsing.

jberry-suse:
Also, since when did

image

become more readable and useful than

image

osclib/conf.py
coolo: Actually writing the empty value is pointless anyway - and reading attributes is like 800% faster than reading files. So possibly we just don't cache the config.

I just don't see the value of not having remote config (read)

coolo:
Well, we need to move it as we can't rely on staging subprojects to exist. That the webui doesn't render attributes nicely is something a junior can fix.

Handling 404 and not writing back is something I'll follow up on.

coolo:
haem - this is to fill the emptiness of https://metrics.opensuse.org/d/osrt_dashboard/osrt-dashboard?orgId=1 ?

coolo:
I looked into this metrics topic, but simply have no idea how to test this. And the attribute history is in osc log -M openSUSE:Factory _project - but it's all attributes and other changes to e.g. the project config.

And in case we'd move lock attribute also to the main project (which sounds fair), it would be *very* noisy. So I'd rather look into emiting an event over the bus for attribute changes, which you could subscribe to for metrics.

osclib/stagingapi.py
```diff
@@ -524,7 +529,7 @@ def superseded_request(self, request, target_requests=None):
                 replace_old = request_old.find('state').get('name') in ['revoked', 'superseded']
 
                 if (request_new.find('action').get('type') == 'delete' and
-                    request_old.find('action').get('type') == 'delete'):
+                        request_old.find('action').get('type') == 'delete'):
```
jberry-suse: I don't find this beneficial or relevant to this change (nor the rest of these white-space changes).

osclib/obslock.py
```diff
@@ -74,7 +74,7 @@ def _read(self):
         return signature
 
     def _write(self, signature):
-        url = makeurl(self.apiurl, ['source', self.lock, '_attribute', '%s:LockedBy' % self.ns])
+        url = makeurl(self.apiurl, ['source', self.lock, '_attribute'])
```
jberry-suse: - should be mirrored in `_read()`?
- should likely be removed from config if no longer used

osclib/conf.py
```diff
@@ -244,18 +203,32 @@ def read_section(self, section, defaults):
...
+            config = self._migrate_staging_config(api)
         if config:
             cp = ConfigParser()
             config = '[remote]\n' + config
             cp.readfp(io.BytesIO(config))
             self.remote_values = dict(cp.items('remote'))
             self.populate_conf()
-        elif config is None:
-            # Write empty config to allow for caching.
-            api.dashboard_content_save('config', '')
```
jberry-suse: To be clear, this will mean a remote request looking for config every time for projects that do not have one since 404's are not cached. This was avoided with remote-config option locally. Obviously, if opening this up to any project it is likely not desired to write empty config.

tests/obs.py
```diff
@@ -479,41 +489,63 @@ def request_search(self, request, uri, headers):
...
+        if not ns in hash and not name in hash[ns]:
+            return response
+
+        root = ET.fromstring('<attributes><attribute name="" namespace="">' +
+                             '<value/></attribute></attributes>')
+        root.find('./attribute').set('name', name)
+        root.find('./attribute').set('namespace', ns)
+        root.find('./attribute/value').text = hash[ns][name]
+
+        response = (200, headers, ET.tostring(root))
```
jberry-suse: May be of interest in future, but could write against dockerized OBS that is already available with a test hardness and tests using it. Avoids needed to rewrite OBS.

jberry-suse:
Emitting is nice in theory, but it means it needs to be done with the switch to avoid data loss and also depends on persistent queue since things die. metrics.o.o had an outage due to influxdb killing itself by writing too much to its internal metrics (I grow less impressed by influxdb by the day). I would hope everyone has seen the data that is normally there and still there for non-Factory projects.

Overall, I am not terribly happy about breaking things as I have had to heavily refactor all sorts of things and handle edge-cases of zero interest to me to avoid breaking other use-cases. To say nothing of figuring out poorly and incorrectly documented messes all over. Share the love. Testing does not even require grafana or influxdb to be available even though just a matter of `systemctl start ...`, just see that the history can be parsed and equivalent point data created.

tests/obs.py
coolo: Actually the docker-compose of OBS works quite nicely - we'd need to fill the database with Factory data though. But this is required anyway soonish to test https://github.com/openSUSE/open-build-service/pull/5140 properly

coolo:
Persistent queues aren't a problem - we would need to define a TTL though.

osclib/stagingapi.py
coolo: autopep...

osclib/obslock.py
coolo: No, in read it works as filter. In write it's just ignored by OBS and it only takes what's in XML. The API for attributes is most likely the worst in OBS.

coolo:
I implemented some parsing of attribute history - but there are multiple questions with it. As I wrote earlier I would split out the repo checker stuff from the config, but this would also mean to migrate the datapoints.

metrics.py
```diff
@@ -464,6 +469,74 @@ def ingest_dashboard_revision_get():
...
+        for v in root.findall('.//attribute[@namespace="OSRT"]'):
+            attribute_name = v.get('name')
+            last_values.setdefault(attribute_name, None)
+            if last_values[attribute_name] == v.find('value').text:
+                continue
+            # TODO: no idea what fields to make
+            fields = {revision: revision }
+
+            points.append({
+                'measurement': 'attribute_osrt_{}'.format(attribute_name),
```
jberry-suse: The influxdb database in OSRT specific so no need to prefix measurements. Otherwise, all should be prefixed.

metrics.py
```diff
@@ -351,10 +352,14 @@ def ingest_release_schedule(project):
...
-            get_commitlog(api.apiurl, api.cstaging, 'dashboard', None, format='xml')))
-
         revision_index.index = {}
+
+        try:
+            root = ET.fromstring(''.join(
+                get_commitlog(api.apiurl, api.cstaging, 'dashboard', None, format='xml')))
+        except HTTPError as e:
+            return revision_index.index
+
```
jberry-suse: Perhaps a comment indicating, fallback to config in attribute.

metrics.py
```diff
@@ -464,6 +469,74 @@ def ingest_dashboard_revision_get():
 
     return None
 
+def ingest_attributes(api):
```
jberry-suse: Would seem better to order after `iingest_dashboard()` since all the map functions above are related and named as such.

metrics.py
```diff
@@ -464,6 +469,74 @@ def ingest_dashboard_revision_get():
...
+        root = ET.parse(osc.core.http_GET(url))
+        #print revision, time, ET.tostring(root)
+
+        for v in root.findall('.//attribute[@namespace="OSRT"]'):
+            attribute_name = v.get('name')
+            last_values.setdefault(attribute_name, None)
+            if last_values[attribute_name] == v.find('value').text:
+                continue
+            # TODO: no idea what fields to make
+            fields = {revision: revision }
```
jberry-suse: The same as before:

```python
map_func = globals()['ingest_dashboard_{}'.format(filename)]
fields = map_func(content)
```

The files parsed before are (presumably only migrating `config`):

```python
    filenames = ['config', 'repo_checker', 'version_snapshot']
    if api.project == 'openSUSE:Factory':
        filenames.append('devel_projects')
```

metrics.py
```diff
@@ -478,7 +551,9 @@ def ingest_dashboard(api):
 
     count = 0
     points = []
+    max_revision = 0
```
jberry-suse: Perhaps instead the revisions should be parsed by same bit of code since they are really rather similar (ie revision with this content for config) and need to be turned into points in the same way.

This also naturally works if further parsed files re migrated. The reason this was was setup with revision_at() and such is that in the future I wanted to parse the psudo-state ignored list to then create segments when the requests are ignored and filter them out of the various buckets like _backlog_. This would also be useful going forward since that information will likely also move.

tests/obs.py
jberry-suse: `obs_clone.py` is already doing that. It does not copy everything (I have local unifinish attribute copy), but it already provided skeleton project with a devel package.

jberry-suse:
> I implemented some parsing of attribute history - but there are multiple questions with it. As I wrote earlier I would split out the repo checker stuff from the config, but this would also mean to migrate the datapoints.

Can certainly be done in follow-up. I don't know what question that leaves, just means yet another branch in metrics code and changing `repo_checker` usage for virtually no gain.

jberry-suse:
After I cleaned up influxdb on metrics.o.o the ingest job ran over the weekend and the data parsed from dashboard files can be seen again at https://metrics.opensuse.org/d/osrt_dashboard/osrt-dashboard.

coolo:
The gain of splitting out repo checker is about permissions - and avoiding that the repo checker kills the config (as it did in the past).

coolo:
I will not continue working on metrics - as such I will merge. Feel free to redo all of it in a followup

coolo:
I will work on OBS attribute events - which most likely makes all metrics revisison parsing obsolete (actually you could do the same for source commits to dashboard)

jberry-suse:
> I will not continue working on metrics - as such I will merge. Feel free to redo all of it in a followup

I guess that's how we got the giant mess I spent a year and half cleaning.



> I will work on OBS attribute events - which most likely makes all metrics revisison parsing obsolete (actually you could do the same for source commits to dashboard)

Historical data (a huge part of the usefulness of metrics) is not available using events, but sure as I've said before that's nice going forward, but having a TTL and dropping data of this sort is definitely not helpful.

Overall, I am extremely impressed by:

https://github.com/openSUSE/openSUSE-release-tools/blob/67121e297e72b3c6c8de3dea22055358d47dc83f/metrics.py#L510-L511

When the code required to create the fields was already written and you have all the arguments necessary to call it...as I wrote.

coolo:
Historical data is hopefully in your influxdb.

jberry-suse:
> I will work on OBS attribute events - which most likely makes all metrics revisison parsing obsolete (actually you could do the same for source commits to dashboard)

Talking about priorities, but rather that utilize already written code which functions more than adequately we'll opt for a completely new route with no benefits aside from commit diffstat.

> I will not continue working on metrics - as such I will merge. Feel free to redo all of it in a followup

In general do not commit broken crap. It's one thing to make code stop functioning...it is quite another to commit code that corrupts data. Just drop the code if you're not going to make it work, much less well.

jberry-suse:
In general this is rude at best. This entire request moves an optional feature to being a required one for SLE. A feature that was envision and implemented entirely by myself. All the necessary code was refactored to make it possible to utilize a remote config and now all that has to be done is to move the location from which it is read. In addition I spent the time to thoroughly review and catch all the breakages and regressions the original submission would create. Rather than be grateful, much less just a hint of respect I get shit on and beyond that my priorities questioned. How does one question the priorities of someone while they elevate a feature they built alone? Amusing.

PR #1575, opened 2018-06-14
dist/ci/Dockerfile: add python2-pyOpenSSL as dependency.

jberry-suse:
Pending a proper resolution to openSUSE/osc#429.

This has been failing for a few days...

```
  File "/usr/bin/osc", line 10, in <module>
    from osc import commandline, babysitter
  File "/usr/lib/python2.7/site-packages/osc/commandline.py", line 9, in <module>
    from . import conf
  File "/usr/lib/python2.7/site-packages/osc/conf.py", line 44, in <module>
    import ssl
ImportError: No module named ssl
```

### DIFF
--- a/dist/ci/Dockerfile
+++ b/dist/ci/Dockerfile
@@ -12,6 +12,7 @@ RUN zypper -n ref && zypper -n dup && zypper -n install \
   obs-service-* \
   osc \
   python-packaging \
+  python2-pyOpenSSL \
   sudo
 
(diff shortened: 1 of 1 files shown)


jberry-suse:
See if response on `osc` issue, otherwise we can merge this until they resolve.

coolo:
Even if they fix it - it will take a while for a new TW snapshot to fix our CI. So let's merge it right away

PR #1576, opened 2018-06-14
TTM: Remove project-base option

asdil12:
Derive project-base from project argument.
This removes some redundancy from cli.

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -920,11 +920,11 @@ class CommandlineInterface(cmdln.Cmdln):
         }
         self.openqa_server = {
             'openSUSE': 'https://openqa.opensuse.org',
-            'SLE': 'https://openqa.suse.de',
+            'SUSE': 'https://openqa.suse.de',
         }
         self.api_url = {
             'openSUSE': 'https://api.opensuse.org',
-            'SLE': 'https://api.suse.de',
+            'SUSE': 'https://api.suse.de',
         }
 
     def get_optparser(self):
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -936,13 +936,10 @@ def get_optparser(self):
...
-        parser.add_option(
-            "--openqa-server", help="""Full URL to the openQA server that should be queried, default based on '--project-base' selection, e.g.
+            "--openqa-server", help="""Full URL to the openQA server that should be queried, default based on project selection, e.g.
             'https://openqa.opensuse.org' for 'openSUSE'""")
         parser.add_option(
-            "--obs-api-url", help="""Full URL to OBS instance to be queried, default based on '--project-base' selection, e.g.
+            "--obs-api-url", help="""Full URL to OBS instance to be queried, default based on project selection, e.g.
             'https://api.opensuse.org' for 'openSUSE'""")
         return parser
         return parser
```
okurz: this double return looks wrong, not your change of course but please fix

totest-manager.py
```diff
@@ -920,11 +920,11 @@ def __init__(self, *args, **kwargs):
         }
         self.openqa_server = {
             'openSUSE': 'https://openqa.opensuse.org',
-            'SLE': 'https://openqa.suse.de',
+            'SUSE': 'https://openqa.suse.de',
```
okurz: Can you explain why this can be done now?

asdil12: For some reason you introduced the args `SLE` and `openSUSE` for the `--project-base` argument and defaulted to `openSUSE`.
`SLE` doesn't fit nicely, because the IBS project actually starts with `SUSE:`

okurz: ok, don't know anymore why I did that, fine ;)

okurz:
LGTM

PR #1578, opened 2018-06-15
Metrics access log parsing tool for ingesting download.o.o logs.

jberry-suse:
- 26ab99bfb38e68a5a2d7106bcb2667e4601a03cf:
    metrics/access: add summary README.md.

- baad36e12320dce59cc5b11d1b90bd534af5856e:
    dist: package metrics/access tool and related.

- 8490877fce978628550ed5b304079c83333f1fc9:
    metrics/grafana: add access dashboard and data source.
    
    Utilized by metrics/access tool.

- 93530a7eefd1584659f8776e8934a73eab1ef740:
    metrics/access: initial aggregate script.

- d073daf5e6de9e72683c3cc66a6370762c31c282:
    metrics/access: initial ingest script.

I need to confirm the distribution bits work (likely will fail on travis) and [setup the vhost on pontifex](https://progress.opensuse.org/issues/36862) so I can run the most recent data. Overall, this should be rather close and will provide time for anyone interested in reviewing. A better solution for determining real products would be nice as there is a lot of completely garbage data and the current approach requires manually update when new product is released. However, it does not require re-parsing data and as such is not time pertinent.

The generated graphs on my local setup can be seen below.

image


jberry-suse:
I plan to do a write-up on all the interest data I discovered and possibly some room to run further analysis looking for such data. So far all the data marked invalid by the tool that I reviewed is exactly that...invalid. The graph will make it clear if the parsing becomes broken in the future as there would be a huge spike.

jberry-suse:
I'll also add a bit more to README.

jberry-suse:
If it's no clear from picture the "Aggregation Frequency" can be changed to:

- day
- week
- month

As the data has been pre-processed in those intervals. All graphs are thus refreshed from the new source.

jberry-suse:
A tried a variety of parsing approaches, but as I've seen before and come to expect regex turned out to be fastest...likely becomes of native implementation instead of string operations done in scripting language.

jberry-suse:
Made a few small tweaks, including dist fix and DNS name suggested in progres.o.o issue. Added a couple extra sections to README.

jberry-suse:
Added annotations for the last three Leap releases based on the data already imported from schedules.

jberry-suse:
For example, with `42.2` and `42.3` annotations enabled.

image

jberry-suse:
Or `42.2` show by itself we see `6.7k` users by `Beta 1`:

image

jberry-suse:
Deployed and metrics.o.o and completed initial data import with pontifex ingest disabled (pending vhost setup).

PR #1580, opened 2018-06-15
request_splitter: gcc8 is the default compiler now

nilxam:


osclib/request_splitter.py
```diff
@@ -505,7 +505,7 @@ class StrategySpecial(StrategyNone):
     # Configurable via splitter-special-packages.
     PACKAGES = [
         'gcc',
-        'gcc7',
+        'gcc8',
```
jberry-suse: This can also be done via `splitter-special-packages` if you want it to go into effect sooner.

PR #1584, opened 2018-06-19
systemd/metrics-access.timer: skip Monday to avoid conflict with metrics.timer.

jberry-suse:
Since:

> weekly → Mon *-*-* 00:00:00

per https://www.freedesktop.org/software/systemd/man/systemd.time.html and verified logs for length. Hasn't happened yet, but would rather avoid.


jberry-suse:
Likely won't be an issue since the aggregate process never uses more than `~120MB` of RAM, but better safe than sorry since the other one uses a most of system RAM.

PR #1586, opened 2018-06-20
Revert crap from 'Use attribute to store staging config' that create corrupt data

jberry-suse:
As explained in my reviews #1573 original broke metrics parsing going forward, but was upgraded to create corrupt data. I would suggest the author use the toilet at home as this prevents it from being deployed to metrics.o.o.

Additionally, as noted and noted before, the update contained lots of style changes for code that was otherwise untouched which is very annoying at best. The worse two I reverted.

Since reviews are generally ignored there is no sense to wait.

### DIFF
--- a/metrics.py
+++ b/metrics.py
@@ -549,9 +549,9 @@ def main(args):
     Cache.PATTERNS['/source/[^/]+/dashboard/[^/]+\?rev=.*'] = sys.maxint
     Cache.init()
 
-    c = Config(args.project)
+    config = Config(args.project)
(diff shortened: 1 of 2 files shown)


coolo:
If it helps your mental state, feel free to do this more often

jberry-suse:
> If it helps your mental state, feel free to do this more often

That's what you get out of having to revert `Output Attribute metrics (not well tested)`....a commit that says it's garbage in the message and creates corrupt data? I'm sure OBS and openQA would like me to merge such pull requests for them. Everybody likes corrupt data.

PR #1588, opened 2018-06-20
Metrics access followup for live data final location and small cleanups

jberry-suse:
- ef31dc41aa9a76adfd8ef72e80ff6c55fabe8a68:
    metrics/grafana/access: add explanation about loss of custom log format.

- 8a59f4a862496a52e9c61321e9e458f87a35fb4c:
    metrics/access/aggregate: change PONTIFEX base URL based on poo#36862.

- 138facc9610a9d9dd8953d16a05950f58c028cca:
    metrics/grafana/access: remove broken datasource templating from export.
    
    Unresolved upstream issue that makes the export unusable on a new system.

- 2e93b6c3bd6fe8123a6135e2be0bd825f83fdcd7:
    metrics/access/aggregate: remove left over sprintf() from PONTIFEX lines.
    
    The directory was broken out from PONTIFEX and LANGLEY which made sprintf()
    no longer necessary. Still works, but a waste.

- 625fa382fa3ab5833f8d1a2d74cb99b9fd02852b:
    metrics/access/aggregate: skip the current day since the logs are incomplete.

- 409f02194877495f2c55729ee4744d0aed884161:
    dist/spec: move access.json dashboard to metrics-access package.

The explanation in access dashboard will need to be updated once data is parsed, but want to see if travis has any issues.


jberry-suse:
Updated message about hole in data with dates.

jberry-suse:
Will follow-up once more to write zeros for previous data since we have essentially no data for 6 months which ends up writing no records. Explicit zero for products previously seen would make that fact more clear.

PR #1589, opened 2018-06-21
metrics/access/aggregate: write zeros for empty product data.

jberry-suse:
Since the aggregation is not done by influxdb a fill option is not
available and instead needs to be performed manually. As such the order of
writing must be from oldest to newest so that the list of product seen
before can be kept and missing products for a given interval can be filled.


jberry-suse:
This is especially needed since we have 6 months of missing data. Without this there is no data written for that period for these measurements and the lines connect.

jberry-suse:
Before:

image

After:

image

PR #1592, opened 2018-06-25
Leaper: adjust for 15.1

lnussel:
preliminary adjustments

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -83,8 +83,8 @@ class Leaper(ReviewBot.ReviewBot):
             self.lookup_sle15 = self.factory.lookup.copy()
             return
 
-        self.factory.parse_lookup('openSUSE:Leap:15.0')
-        self.factory.parse_lookup('openSUSE:Leap:15.0:NonFree')
+        self.factory.parse_lookup('openSUSE:Leap:15.1')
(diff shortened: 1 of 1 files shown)


leaper.py
```diff
@@ -347,15 +347,6 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
...
 
-            # new package submitted from Factory. Check if it was in
-            # 42.3 before and skip maintainer review if so.
-            subprj = src_project[len('openSUSE:Factory'):]
-            if self.source_in_factory and target_project.startswith('openSUSE:Leap:15.0') \
-                and self.is_package_in_project('openSUSE:Leap:42.3'+subprj, package):
-                    self.logger.info('package was in 42.3')
-                    self.do_check_maintainer_review = False
-                    return True
-
```
jberry-suse: Is this not something that will be desired to have in the future? If so I'd rather disable it like other conditions as it will make it a lot better if/when we decide to refactor this whole thing to know what the desired features actually are.

jberry-suse:
Fine with me, but I'd stick with using `if False` as it keeps diff smaller and doesn't loose real blame history.

lnussel:
included in #1595

PR #1596, opened 2018-06-26
Fix JeOS snapshotable detection for Factory:ARM

dirkmueller:
In order to release the live images for Factory:ARM we need
to check if they finished building. Currently JeOS is multibuild
for armv7l and aarch64 but the main package is just building for armv7l,
so thats what we check for now. Without this change it was checking
for the default architectures (i586, x86_64) which did never pass.


ggardet:
LGTM but, cannot we handle aarch64 too?

PR #1603, opened 2018-06-29
Add support for Backports

lnussel:


osclib/conf.py
```diff
@@ -107,13 +107,35 @@
         'mail-noreply': 'noreply@opensuse.org',
         'mail-release-list': 'opensuse-releaseteam@opensuse.org',
     },
+    r'openSUSE:(?P<project>Backports:(?P<version>[^:]+))': {
+        'staging': 'openSUSE:%(project)s:Staging',
+        'staging-group': 'factory-staging',
+        'staging-archs': 'x86_64',
+        'lock': 'openSUSE:%(project)s:Staging',
+        'lock-ns': 'openSUSE',
```
jberry-suse: Perhaps add:

```python
r'openSUSE:(?P<project>(?:Backports|Factory|Leap)(?::.*)?)'
```

With all the common properties for all `openSUSE` products with a priority of `1`.

osclib/request_splitter.py
```diff
@@ -140,6 +140,8 @@ def supplement(self, request):
         request.set('postponed', 'False')
 
     def ring_get(self, target_package):
+        if self.api.conlyadi:
+            return None
```
jberry-suse: This does not handle:

https://github.com/openSUSE/openSUSE-release-tools/blob/43558590825d2a9ecfdadde999a32c3aa82644c4/osclib/request_splitter.py#L130-L132

which will likely need `!self.api.conlyadi and ...`.

osclib/conf.py
jberry-suse: Can decide if you want it to match sub projects as currently all patterns do even without the optional `(?::.*)?` on the end. As such you can drop that. `re.match()` is basically `re.startswith()`. Alternatively the patterns could be restricted by placing `$` on the end.

lnussel: there is a break in the loop, so only one setting gets applied atm

PR #1604, opened 2018-06-29
No longer have Ring2 and staging :DVD subproject

nilxam:


DimStar77:
Part of it is also covered in https://github.com/openSUSE/openSUSE-release-tools/pull/1602

lnussel:
this one is better :)

osclib/stagingapi.py
```diff
@@ -85,8 +85,7 @@ def __init__(self, apiurl, project):
         if self.crings:
             self.rings = (
                 '{}:0-Bootstrap'.format(self.crings),
-                '{}:1-MinimalX'.format(self.crings),
-                '{}:2-TestDVD'.format(self.crings)
+                '{}:1-MinimalX'.format(self.crings)
```
jberry-suse: nit: no need to remove trailing comma for cleaner diffs.

PR #1605, opened 2018-07-03
pkglistgen: Fixes required for SP1

coolo:


nilxam:
:+1: I've hot-patched the similar change of https://github.com/openSUSE/openSUSE-release-tools/pull/1605/commits/562a5c389cff78336fe288015bc12c6777851ad9 on tortuga make it working for 15.1

lnussel:
flake8 doesnt like it

coolo:
that's what I call a mutual agreement :)

PR #1608, opened 2018-07-04
stagingapi: subprojects method has been removed from dashboard

nilxam:
ref. https://github.com/openSUSE/open-build-service/pull/5254/commits/e80915d0c3d68c2f3988344aa282475f216e3f26

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -911,19 +911,18 @@ class StagingAPI(object):
 
     def rebuild_broken(self, status, check=True):
         """ Rebuild broken packages given a staging's status information. """
-        for project in self.project_status_walk(status):
-            for package in project['broken_packages']:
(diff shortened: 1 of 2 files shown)


osclib/stagingapi.py
```diff
@@ -1004,10 +1004,6 @@ def project_status(self, staging=None, aggregate=False):
 
     def project_status_walk(self, project):
         yield project
-        for sub in project['subprojects']:
```
coolo: if there are no subprojects, you don't need a walk

nilxam: I also see some other places need to update, will re-do this PR

DimStar77:
Change is incomplete (already told @nilxam )

```osc staging check A``` results in ```KeyError: 'subproject'```

PR #1609, opened 2018-07-04
pkglistgen: remove default-support-status option

coolo:
The OUTPUT groups need to specify the status now or it will default to
unsupported (which should still be fine for leap/factory)

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -676,7 +676,7 @@ class PkgListGen(ToolBase.ToolBase):
 
 
 class CommandLineInterface(ToolBase.CommandLineInterface):
-    SCOPES = ['all', 'arm', 'target', 'rings', 'staging', 'ports']
+    SCOPES = ['all', 'arm', 'target', 'rings', 'ports', 'staging']
 
     def __init__(self, *args, **kwargs):
         ToolBase.CommandLineInterface.__init__(self, args, kwargs)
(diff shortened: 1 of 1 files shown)


pkglistgen.py
```diff
@@ -1135,6 +1135,12 @@ def do_update_and_solve(self, subcmd, opts):
 
         if not opts.project:
             raise ValueError('project is required')
+        opts.staging_project = None
+        if opts.scope.startswith('staging:'):
```
lnussel: how about not doing this here but instead use a few lines further down
if opts.scope.startswith("staging"), then try the match and set letters accordingly?

pkglistgen.py
```diff
@@ -1177,6 +1183,8 @@ def do_update_and_solve(self, subcmd, opts):
         elif opts.scope == 'staging':
```
lnussel:
```
  if opts.scope.startswith('staging'):
    if ':' in opts.scope:
       letters = opts.scope.split(':')[1]
    else:
       letters = api.get_staging_projects_short()
```

pkglistgen.py
coolo: the problem is the line in between those you commented - staging:X is not in SCOPES

coolo:
The feedback of Ludwig won't work and I'm running this for 16 days already off branch

PR #1612, opened 2018-07-05
manager42: don't consider ignored packages vanished

lnussel:


coolo:
perhaps this script should be renamed? 42 is a little old :)

PR #1618, opened 2018-07-17
Remove debug code accidentally part of #1615

dirkmueller:
Apparently accidentally did a commit --amend --all on last commit,
revert.


nilxam:
LGTM

PR #1620, opened 2018-07-19
repo_checker: Fix the review loop

coolo:
The break on openQA failures is a left over from 93ee829260a6abf094cbbc31e26eb21bf45e8f15
where we stopped the loop over subprojects. We don't want to break out of
the review loop on openQA failures

### DIFF
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -225,6 +225,17 @@ class RepoChecker(ReviewBot.ReviewBot):
         self.logger.debug('requests: {} skipped, {} queued'.format(
             count_before - len(self.requests), len(self.requests)))
 
+    def expand_project_repo(self, api, project, repo, repos):
+        repos.append([project, repo])
+        url = api.makeurl(['source', project, '_meta'])
+        meta = ET.parse(api.retried_GET(url)).getroot()
+        for path in meta.findall('.//repository[@name="{}"]/path'.format(repo)):
+            self.expand_project_repo(api, path.get('project', project), path.get('repository'), repos)
+        return repos
+
+    def expand_repos(self, api, project, repo):
+        return self.expand_project_repo(api, project, repo, [])
+
     def ensure_group(self, request, action):
         project = action.tgt_project
         group = self.requests_map[int(request.reqid)]
@@ -254,8 +265,12 @@ class RepoChecker(ReviewBot.ReviewBot):
                 continue
 
             # Only bother if staging can match arch, but layered first.
-            directories.insert(0, self.mirror(project, arch))
-
+            repos = self.expand_repos(self.staging_api(project), project, 'standard')
+            for layered_project, repo in reversed(repos):
+                if repo != 'standard':
+                    raise "We assume all is standard"
+                directories.insert(0, self.mirror(layered_project, arch))
+                
             whitelist = self.binary_whitelist(project, arch, group)
 
             # Perform checks on group.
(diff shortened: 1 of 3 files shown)


nilxam:
oh right, not in the loop block anymore. LGTM.

nilxam:
maybe squashing `Move expanded_repos into staging api` and `repo_checker: Expand repositories of target project`?

coolo:
I actually did this in 2 steps and would like to keep it this way in case I broke pkglist with it (there is another PR for pkglist open)

repo_checker.py
```diff
@@ -257,7 +254,11 @@ def ensure_group(self, request, action):
                 continue
 
             # Only bother if staging can match arch, but layered first.
-            directories.insert(0, self.mirror(project, arch))
+            repos = self.staging_api(project).expanded_repos('standard')
+            for layered_project, repo in reversed(repos):
+                if repo != 'standard':
+                    raise "We assume all is standard"
+                directories.insert(0, self.mirror(layered_project, arch))
```
jberry-suse: `repo_checker.pl` only handles a single base directory and then a list of directories stacked on top. If `expand_repos()` returns additional directories it will not function properly. If nothing else perhaps a note since this was not possible with the previous code.

coolo: hmm, it did something for me. Before my change it complained about missing glibc and now it is satisfied.

the staging projects don't really have much content atm, but they are the base now and just have way more layers than those for factory.

What will not work in general is project mode - as most problematic packages aren't in the target project, but are inherited (through maintenance)

jberry-suse: It will pass all directories beyond index `0` as stacking so any packages present with override those from 0, but override between them are not handled. Presumably a package may be written out multiple times and the last one wins. Not ideal, but might cause it to do something resembling correct in your case.

coolo: I was looking into providing a service to get a solv file for a repo including its dependencies. But when looking, I found the repo checker mangles the whitelist into the creation of the solv file. So we would need to fix that by removing these hacks and move them into code using the libsolv bindings

jberry-suse: If a service provides the solv file including everything than the whitelist could be used in `.pl` or `.py` to ignore errors. That part should be easy. The `.py` code already has a parse routine that could be extended assuming the `.pl` code could be entirely replaced by the service.

coolo: First step I had in mind is providing a full solv file that can be used by pkglistgen and repo checker. Both have similiar requirements. If we can align these, we can outsource the implementation :)

jberry-suse: It would certainly be a big improvement. I see nothing holding that back. The `.pl` file always mixed _check_ with _solv generation_. If one outsources the solv....then just leave the check bits where they are.

coolo: I'm mostly worried about https://github.com/openSUSE/openSUSE-release-tools/blob/master/CreatePackageDescr.pm#L126 and following

coolo: or let's say those hold me back

jberry-suse: Those errors could also be ignored like any other whitelist entry when parsing the result right? That particular case does not appear to be present in TW and the `debuginfodeps` should really be resolved by publishing in the debug repo along with the dependencies, but that same style rule could be placed in `repo_checker.py` as it can already be done with the exact package name.

Not suggesting this hold up this PR.

coolo: yeah, right. Let's fix the repo checker loop ... :)

PR #1624, opened 2018-07-30
WIP: Simplify the repo_checker handling overlays

coolo:
So far we had one repo directory (having the staging prj) and other
directories (having the target project). Additionally the staging prj's
binaries were added to to_ignore, which led to improper overlay handling
for service packs where we have 4 projects to look at - for sp1 (2 more
for sp2...)

Now passing the projects in expansion order as in OBS, so we can ignore
overlayed binaries later on and stop handling staging prj any different.

### DIFF
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -90,7 +90,7 @@ class Cache(object):
         '/source/([^/]+)/_meta$': TTL_LONG,
         '/source/([^/]+)/(?:[^/]+)/(?:_meta|_link)$': TTL_LONG,
         '/source/([^/]+)/dashboard/[^/]+': TTL_LONG,
-        '/source/([^/]+)/_attribute/[^/]+': TTL_MEDIUM,
+        '/source/([^/]+)/_attribute/[^/]+': TTL_SHORT,
         # Handles clearing local cache on package deletes. Lots of queries like
         # updating project info, comment, and package additions.
         '/source/([^/]+)/(?:[^/?]+)(?:\?[^/]+)?$': TTL_LONG,
--- a/repo_checker.pl
+++ b/repo_checker.pl
@@ -19,24 +19,10 @@ require CreatePackageDescr;
 my $ret = 0;
 my $arch = shift @ARGV;
 my @directories = split(/\,/, shift @ARGV);
-my %toignore;
-my $repodir;
 my %whitelist;
 while (@ARGV) {
     my $switch = shift @ARGV;
-    if ( $switch eq "-f" ) {
-        my $toignore = shift @ARGV;
-        open( TOIGNORE, $toignore ) || die "can't open $toignore";
-        while (<TOIGNORE>) {
-            chomp;
-            $toignore{$_} = 1;
-        }
-        close(TOIGNORE);
-    }
-    elsif ( $switch eq "-r" ) {
-        $repodir = shift @ARGV;
-    }
-    elsif ( $switch eq "-w" ) {
+    if ( $switch eq "-w" ) {
         %whitelist = map { $_ => 1 } split(/\,/, shift @ARGV);
     }
     else {
@@ -47,9 +33,8 @@ while (@ARGV) {
 
 my %targets;
 
-sub write_package($$) {
-    my $ignore  = shift;
-    my $package = shift;
+sub write_package {
+    my ($package, $packages_fd, $written_names) = @_;
 
     my $name = basename($package);
     if ($name =~ m/^[a-z0-9]{32}-/) { # repo cache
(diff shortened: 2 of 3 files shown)


coolo:
OK, it's not as easy.

Running this on a staging request will report all problems of Factory too. This is not what we want - just as we don't want to know the Update problems in SP1 Staging.

But I have the feeling this ignoring of problems should be done on the level up - as we discussed previously to outsource the installcheck logic

jberry-suse:
> Running this on a staging request will report all problems of Factory too.

Just like when reviewing adi stagings with packages not in the rings. If this change is made it will mean the staging master would need to maintain a much larger whitelist. The whitelist could be inherited using the repo expansion chain or could be required to be maintained in the target project. I see reasons for both being desirable.

It seems unavoidable if removing the staging project as filter of packages to care about.

coolo:
But wouldn't this long whitelist make the project mode toothless? I would rather move the filtering into the .py part. I just didn't come around to do it - for SP1 I went the 'long whitelist' road, but that's not practical IMO for Factory.

jberry-suse:
> But wouldn't this long whitelist make the project mode toothless?

`project_only` run still shows you just the problems in the target project as otherwise they're mixed in with staging project problems.

> I would rather move the filtering into the .py part. I just didn't come around to do it - for SP1 I went the 'long whitelist' road, but that's not practical IMO for Factory.

Where the filtering is done is not a big deal and does not change the problem at hand. I agree the long-whitelist could be rather annoying to deal with in Factory/Leap.

See #1630 for a proposal to resolve whitelist issue.

jberry-suse:
One item to note since you are trying to do repo stacking which can involve greater than two layers. The current approach is not 100% correct as it only ignores binaries from the lower layers produced with the same name as those in the upper layers (staging). If I remember correctly (a few existing issues related to this), there are a couple edge-cases specially around old-style multibuild packages and renaming sub-packages that are not handled the same as OBS which does package level overrides of binaries.

It appears your reworking continues the same pattern as before, but since you introduce more layers that span multiple years of a project the likelyhood of hitting such edge-cases likely increases. This is why this really should be handled on OBS end.

osclib/cache.py
```diff
@@ -90,7 +90,7 @@ class Cache(object):
         '/source/([^/]+)/_meta$': TTL_LONG,
         '/source/([^/]+)/(?:[^/]+)/(?:_meta|_link)$': TTL_LONG,
         '/source/([^/]+)/dashboard/[^/]+': TTL_LONG,
-        '/source/([^/]+)/_attribute/[^/]+': TTL_MEDIUM,
+        '/source/([^/]+)/_attribute/[^/]+': TTL_SHORT,
```
jberry-suse: Not particularly related.

jberry-suse: Nothing wrong with the change. Just not related to commit "Simplify the repo_checker handling overlays".

jberry-suse:
Can't it still be assumed the top layer is the only one being reviewed and as such filtered the same was before?

coolo:
I think so. I just want to raise the filtering to the python level - but it's a little warm for real work :(

jberry-suse:
Raising the filter should likely be scoped for separate work. Basically, just keep the existing filtering and use all rpms in top layer as that list instead of requiring it to be passed in.

jberry-suse:
Assuming that is changed, I would see nothing preventing this from being merged (assuming it works) as the behavior would not be changed.

jberry-suse:
`toignore` should basically be populated with all the packages from highest level directory. Since python never reads the rpms it makes sense for that to remain here. Alternatively, those lists can be built base on OBS query like before and passed like before, but no need to change all that in order to achieve the additional layering.

jberry-suse:
Fixing the scoping should be relatively easy as it is basically just not removing that logic from perl end. I'll end up needing this or something similar to handle maintenance requests since they are layered on top of project. As such I would prefer to use the same code. Please let me know what your timeline looks like.

jberry-suse:
After asking in IRC it seems @coolo will not be working on this until after several weeks of vacation.

I was able to bring back to old behavior of only mentioning issues for RPMs in top layer, but as I suspected this does not work the same as before nor mimic the effect of merging.

With the following change:

```diff
diff --git a/repo_checker.pl b/repo_checker.pl
index 9537483d..562710a1 100644
--- a/repo_checker.pl
+++ b/repo_checker.pl
@@ -65,15 +65,17 @@ open( my $packages_fd, ">", $pfile ) || die 'can not open';
 print $packages_fd "=Ver: 2.0\n";
 
 my $written_names = {};
+my $first_layer = 1;
 
 foreach my $directory (@directories) {
     @rpms = glob("$directory/*.rpm");
     foreach my $package (@rpms) {
         my $name = write_package( $package, $packages_fd, $written_names );
-        if ($name && !exists($whitelist{$name})) {
+        if ($first_layer && $name && !exists($whitelist{$name})) {
             $targets{$name} = 1;
         }
     }
+    $first_layer = 0;
 }
 
 close($packages_fd);
```

The only issue I see in _Factory:K_ is the following:

```
found conflict of xorg-x11-Xvnc-1.8.0-14.3.x86_64 with xorg-x11-Xvnc-module-1.9.0-1.1.x86_64:
  - /etc/X11/xorg.conf.d/10-libvnc.conf
  - /usr/lib64/xorg/modules/extensions/libvnc.so
```

In this case _K_ has 1.8 of `tigervnc` while _Factory_ has 1.9. This is due to _K_ not behing rebased post merge of update. When evaluating this request with current code the problem does not arise. This is because the layer is done incorrectly in that all binaries produced by `K/tigervnc` do not override `Factory/tigervnc`, but instead all the binaries in `K` override all the binaries in `Factory`. Since the subpackage name changed we get both the old and new binary.

This is also true for delete requests which are completely broken by this since the binaries from parent project would still be present.

The python code removed from `repo_checker.py` exists for this reason and I would suggest this be un-deployed if being used for SP1 currently.

jberry-suse:
This will need to be heavily reworked to handle this close to correctly and since you indicated you won't work on it until September I'll end up making it work so closing this issue.

jberry-suse:
Based on IRC discussion, evaluating a letter staging along for Factory results in a [giant mess](https://gist.github.com/jberry-suse/649f1b465700141097a1eb0d6a6529c6) that is not really considerable.

jberry-suse:
For posterity, the following diff was used to test staging only.

```diff
diff --git a/repo_checker.py b/repo_checker.py
index 11b850ec..bb34d93c 100755
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -251,13 +251,6 @@ class RepoChecker(ReviewBot.ReviewBot):
             if not len(stagings):
                 continue
 
-            # Only bother if staging can match arch, but layered first.
-            repos = self.staging_api(project).expanded_repos('standard')
-            for layered_project, repo in repos:
-                if repo != 'standard':
-                    raise "We assume all is standard"
-                directories.append(self.mirror(layered_project, arch))
-
             whitelist = self.binary_whitelist(project, arch, group)
 
             # Perform checks on group.
```

PR #1627, opened 2018-07-31
pkglistgen: Support multiple scopes

coolo:
For SLE I need -s target -s staging, but it only did the later - as
I didn't figure the concept of commandline :)

please review @lnussel @jberry-suse

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -1123,7 +1123,7 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
 
     @cmdln.option('-f', '--force', action='store_true', help='continue even if build is in progress')
     @cmdln.option('-p', '--project', help='target project')
-    @cmdln.option('-s', '--scope', default='all', help='scope on which to operate ({}, staging:$letter)'.format(', '.join(SCOPES)))
+    @cmdln.option('-s', '--scope', action='append', default=['all'], help='scope on which to operate ({}, staging:$letter)'.format(', '.join(SCOPES)))
     @cmdln.option('--no-checkout', action='store_true', help='reuse checkout in cache')
     @cmdln.option('--stop-after-solve', action='store_true', help='only create group files')
     def do_update_and_solve(self, subcmd, opts):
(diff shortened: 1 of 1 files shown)


pkglistgen.py
```diff
@@ -1123,7 +1123,7 @@ def do_solve(self, subcmd, opts):
 
     @cmdln.option('-f', '--force', action='store_true', help='continue even if build is in progress')
     @cmdln.option('-p', '--project', help='target project')
-    @cmdln.option('-s', '--scope', default='all', help='scope on which to operate ({}, staging:$letter)'.format(', '.join(SCOPES)))
+    @cmdln.option('-s', '--scope', action='append', help='scope on which to operate ({}, staging:$letter)'.format(', '.join(SCOPES)))
```
aplanas: do makes sense to maintain the 'all' as `default`? Maybe needs to be changed to `[all]` or something

pkglistgen.py
```diff
@@ -1138,20 +1138,24 @@ def do_update_and_solve(self, subcmd, opts):
...
-            raise ValueError('scope must be one of: {}'.format(', '.join(self.SCOPES)))
-
-        if opts.scope == 'all':
-            for scope in self.SCOPES[1:]:
-                opts.scope = scope
-                self.do_update_and_solve(subcmd, copy.deepcopy(opts))
-            return self.error_occured
 
+        # special case for all
+        if len(opts.scope) == 0 or opts.scope == ['all']:
```
aplanas: With the default this first check can be avoided

lnussel:
lgtm

pkglistgen.py
coolo: like this?

```
diff --git a/pkglistgen.py b/pkglistgen.py
index f72e24f..4f0cbf0 100755
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -1123,7 +1123,7 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
 
     @cmdln.option('-f', '--force', action='store_true', help='continue even if build is in progress')
     @cmdln.option('-p', '--project', help='target project')
-    @cmdln.option('-s', '--scope', action='append', help='scope on which to operate ({}, staging:$letter)'.format(', '.join(SCOPES)))
+    @cmdln.option('-s', '--scope', action='append', default=['all'], help='scope on which to operate ({}, staging:$letter)'.format(', '.join(SCOPES)))
     @cmdln.option('--no-checkout', action='store_true', help='reuse checkout in cache')
     @cmdln.option('--stop-after-solve', action='store_true', help='only create group files')
     def do_update_and_solve(self, subcmd, opts):
@@ -1140,7 +1140,7 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
         opts.staging_project = None
 
         # special case for all
-        if len(opts.scope) == 0 or opts.scope == ['all']:
+        if opts.scope == ['all']:
             opts.scope = self.SCOPES[1:]
 
         for scope in opts.scope:

aplanas: Cool, yes

coolo:
squashed

jberry-suse:
fix (read `support`)

coolo:
True that - but I was using multiple scopes already for weeks, without realizing it's not supported. That's the reason I'm refering to fix :)

coolo:
Fixed the commit log

PR #1629, opened 2018-08-01
Rename superseded_by_id to superseded_by

bgeuken:
OBS is going to drop the superseded_by_id attribute soon as it is just
an alias for superseded_by.


bgeuken:
If there are other places that rely on superseded_by_id please let me know.

coolo:
thanks.

PR #1631, opened 2018-08-02
pkglistgen: respect main_repo setup for other variant

nilxam:
Respect main_repo setup for other variant eg. ARM. Also removed ports from SCOPES. Passing `apiurl` to `update_and_solve_target_wrapper()` also wrong here.

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -524,8 +524,8 @@ class PkgListGen(ToolBase.ToolBase):
         for e in excludes:
             g.ignore(self.groups[e])
 
-    def expand_repos(self, project, repo):
-        return StagingAPI(self.apiurl, project).expanded_repos('standard')
+    def expand_repos(self, project, repo='standard'):
+        return StagingAPI(self.apiurl, project).expanded_repos(repo)
 
     def _check_supplements(self):
         tocheck = set()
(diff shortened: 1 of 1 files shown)


lnussel:
:+1:

coolo:
The removal of ports from SCOPES is a little undermotivated IMO. So you call it as target now?

nilxam:
`:Ports` is obsoleted from 15.1, that is following what we had in Tumbleweed now, ie. `:ARM`, ':PowerPC' etc.

coolo:
so why do we have an arm scope but no powerpc?

nilxam:
@dirkmueller not yet submit a PR to support PowerPC I would say.

nilxam:
Note: another commit has been added https://github.com/openSUSE/openSUSE-release-tools/pull/1631/commits/f62f3305074a8395907319df1e7ef3b3eec1dc80

pkglistgen.py
```diff
@@ -525,7 +525,9 @@ def solve_module(self, groupname, includes, excludes):
             g.ignore(self.groups[e])
 
     def expand_repos(self, project, repo):
-        return StagingAPI(self.apiurl, project).expanded_repos('standard')
+        if not repo:
+            repo = 'standard'
+        return StagingAPI(self.apiurl, project).expanded_repos(repo)
```
jberry-suse: This would seem like another usecase for `osclib.core` (ie functions that would otherwise fit in `osc.core` and not not specific to staging). They are expected to follow the same format (ie. first parameter of `apiurl`): `expanded_repos(apiurl, project, repo)`.

So the `repo` argument was never used before?!?!? Nice. Perhaps (..., repo = 'standard') as that is a) self-documenting, and b) standard. Unless we explicitely need to support passing `None` which I do not see an example of currently.

jberry-suse: Yeah, I understand the code is pre-existing, but definitely a case for osclib.core since StagingAPI is being initialized solely to pass `project`.

PR #1632, opened 2018-08-02
Whitelist 000* packages in repo checker

coolo:
000check-modules in SP1 is just a check - not a package, the same is true
for broken product builds. We can still repo check the requests in those
staging projects

### DIFF
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -129,6 +129,31 @@ class RepoChecker(ReviewBot.ReviewBot):
             self.comment_write(state='seen', result=reference, bot_name_suffix=bot_name_suffix,
                                project=comment_project, package=comment_package, message=message)
 
+    def is_group_ready(self, api, group):
+       status = api.project_status(group, True)
+       # Corrupted requests may reference non-existent projects and will
+       # thus return a None status which should be considered not ready.
+       if not status:
+           return False
+
+       if self.force:
+           return True
+
+       overall_state = str(status['overall_state'])
+       if overall_state in ('testing', 'review', 'acceptable'):
+          return True
+
+       if overall_state != 'failed':
+          return False
+
+       # if the project is in failed state, it's ready if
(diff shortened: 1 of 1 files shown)


repo_checker.py
```diff
@@ -166,7 +166,8 @@ def prepare_review(self):
                 if status and str(status['overall_state']) == 'failed':
                     # Exception to the rule is openQA only in failed state,
                     # Broken packages so not just openQA.
-                    openQA_only = (len(status['broken_packages']) == 0)
+                    broken_packages = [a for a in status['broken_packages'] if not a['package'].startswith('000') ]
```
jberry-suse: This is really ugly, at least toss a comment in there so future generations know why this code was so bludgeoned. Seems more like it should be ignored on API end since likely useless to display in dashboard as failure as well.

jberry-suse: Meh, I guess the intention is the check should block staging, but you want repo checker to proceed. At minimum a comment.

coolo:
while extending that inner loop even more, I took the opportunity to refactor this out into its own function

repo_checker.py
```diff
@@ -129,6 +129,31 @@ def package_comments(self, project):
...
+          return True
+
+       if overall_state != 'failed':
+          return False
+
+       # if the project is in failed state, it's ready if
+       # the broken packages only contain meta packages like
+       # 000product or 000check-modules
+       # it can also be failed, if there are openQA failures,
+       # but we don't care here
```
jberry-suse: nit: wrapping to 58 chars? Rather awkward wording. Perhaps:

```
# Allow failed state if only openQA failures and/or meta package failures.
```

jberry-suse:
Breaking out as separate function and logic breakup looks nice.

jberry-suse:
Closing in favor of #1656 which completely reworks the code and makes this change unnecessary.

PR #1633, opened 2018-08-03
repo_checker: determine existing problems based on project_only output.

### LINKED ISSUE #1630: repo_checker: proposal to abandon project level whitelist
Based on my thinking for some time and the desire for stacking the repo chain in #1624 which exposes this issue more directly it seems now may be the time to make a change.

Currently, there exists two levels to the `repo_checker` whitelist.

1) target project (in `:Staging/dashboard/config/repo_checker-binary-whitelist*`)
2) staging project (project psuedo-meta set via `osc staging config repo_checker-binary-whitelist*`)

The second only applies to the staging project on which it is set and the target project is applied to all.

The whitelist was introduced when I rewrote the `repo_checker` to look at staging projects instead of devel project binaries. The previous approach directly inserted an individual or group of package container binaries into the solv information for the target project. This had the benefit of automatically scoping to only the packages being changed, but would miss problems occuring due to rebuilds and such occurring in staging, not to mention missing everything that happened in staging since it didn't look at it.

The new design reviewed a staging project as a whole, but this also meant it was unclear which problems were "known" or "pre-existing" and which were caused by the changes in question. As such the whitelist was introduced to _acknowledge_ problems that were _known_ to be in the target project. The assumption was the problems would eventually reduce down to near nothing which was true, but given the staging projects only cover ring packages or the packages being changed (adi) a large number of existing problems never had to be placed in the whitelist.

The alternative approach I had envisioned, which I think is theoretically closer to what you want, is to diff the result of the staging project as if it was accepted and part of the target project (ie everything) and diff against the output from the target project to see what is new. Based on the assumption that the problems in the target project would be kept minimal as they are undesirably and considered problems it seemed the whitelist approach as a initial phase tool was far simpler (and it was).

Diffing is a bit complex due to very subtle changes making waves in the solver output and of course any rebuild changing all the numbers. I later added `install_check_parse()` which parses the output to determine the packages involved. This was used to compute a hash to avoid spamming essentially the same comment on packages in devel projects caused by rebuilds.

The same function was later used to automate removing unnecessary entries from the whitelist in `install_check_parse()`. Essentially, if we boil the whitelist down to:

> Everything that made it past staging and into the target project is a known issue.

That can be entirely automated now. This means, during the `project_only` mode the whitelist could be generated by a slight tweak to `whitelist_clean` and thus becomes a "_cache_ of binary names involved in solver problems". At which point it could just be cached locally or computed every time `repo_checker` runs based on the latest `project_only` output in `:Staging/dashboard/repo_checker`. 

Effectively, removes the level 1 whitelist and leaves only the staging project whitelist. This would then allow non-filtered/scoped reviews of staging projects as if they were part of target project and effectively diffed against target project. The caveat (which already exists) based on the nature of the whitelist is that new issues involving binaries with existing issues will be ignored. This is outside the base-scope of this problem and can be solved separately by a more detailed parsing of solver output or using some more direct communication with solver to acquire structured data.

This also makes usage of the `repo_checker` for devel projects much more useful since one could review an individual request against devel project and automatically ignore existing issues since no one to maintain the whitelist. Plus staging masters no longer have to deal with it.

Assuming no one opposed I will implement this before #1624 since it can be done independently and then we can merge that without wreaking havoc in Factory.

jberry-suse:
- a61ebeef608d3cc0e01761d2537007d1d3c281f1:
    repo_checker: drop whitelist_clean sub-command.
    
    No longer needed since the whitelist is no longer managed manually.

- 3f043e968dd8445604546f5408deb61eb8517a6c:
    repo_checker: determine existing problems based on project_only output.
    
    The project level whitelist stored in config is no longer utilized, but
    staging level overrides are still available.

Fixes #1630.

Note that this effectively runs with no `whitelist` for SLE != 15:GA since the `project_only` run is no longer invoked and there is no dashboard container. The following is printed in log:

```
[W] no project_only run from which to extract existing problems
```

As the dashboard container is a requirement, liked or not, it should be provided and the service enabled until it is no longer required by the tools. I have thoughts on changing that bit, but out of the scope of this change.

### DIFF
--- a/repo_checker.py
+++ b/repo_checker.py
@@ -91,7 +91,6 @@ class RepoChecker(ReviewBot.ReviewBot):
         text = '\n'.join(comment).strip()
         if not self.dryrun:
             api.dashboard_content_ensure('repo_checker', text + '\n', 'project_only run')
-            self.whitelist_clean(project)
         else:
             print(text)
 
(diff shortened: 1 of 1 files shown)


jberry-suse:
This also effectively, builds an architecture neutral whitelist.

coolo:
I created that container for sp1 now - but I guess you will find other obstacles to run project_only on SP1

As such I would rather check with libsolv in what layer the problematic package is located - and ignore it if it's in target project or below (and ignore project_only output).

But give it a try for SP1.

jberry-suse:
> As such I would rather check with libsolv in what layer the problematic package is located - and ignore it if it's in target project or below (and ignore project_only output).

The layering work would block SP1, but that's the only difference that would have any impact. Even if we re-write would still need a project_only run (ie two runs to compare against). A lot more work at this point for the same thing. Had I built that from scratch, sure.

jberry-suse:
There is nothing wrong with this PR, but since SP1 repo_checker is not really working due to stack as it is and the `project_only` mode was not included in previous @coolo modifications they are not consistent and thus aggravated by this change. As such I'll hold off on this until the layering is fixed for SLE.

jberry-suse:
Using #1640 and #1641 I was able to complete a project only run for `SLE-15-SP1` to which I save the results in _dashboard_ container. I than test this code which then produced the same result as existing comment. As such this should be good to go.

jberry-suse:
It likely also makes sense to trigger a local run if the build hash differs even though the project_run has not completed since the target is still building, otherwise we need to wait for both before evaluating stagings. Correctness says wait, but not sure what staging masters think.

Issue #1635, opened 2018-08-03
leaper: adapt for maintenance workflow


jberry-suse:
- [x] handle `maintenance_incident` request
- [x] appropriately source lookup file from `:Update` project
- [x] basic logic branch look for no origin change
- [x] appropriate text in report comment

jberry-suse:
A separate issue would be to fix manager_42 for `:Update` projects and potentially include more search output, but really those have more to do with rewrite and merge of leaper and manager_42 than the maintenance workflow.

PR #1636, opened 2018-08-03
leaper: provide maintenance_incident support.

jberry-suse:
- aad8e611ef15066b0552c4f804d22ea94e76dac0:
    ReviewBot: check_one_request(): correct logic to return "lowest" result.
    
    Without this change if two actions return True and None, True is returned
    which is not desirable.

- f4460ebb146bdd8bccb44ea9eaf9ea8be5bc6548:
    leaper: remove only_one_action limitation.
    
    Leaper will now be used on maintenance incidents which commonly have
    multiple actions. A similar change was done to check_source.py where the
    restriction was moved to a review setting which is more appropriate rather
    than a hard limitation. Going forward it makes sense to lift the limitation
    and have only one bot enforce the rule where appropriate.

- 5e06287ac8a5e5fb5aec1bb6dd4be01808899fe5:
    leaper: provide maintenance_incident support.

Initial stab at #1635, but some questions are outstanding.

- Due to maintenance workflow submissions are rarely, if ever?, from `origin`. Is that actually what we are expecting? Running on a variety of the requests in queue they all fail (except for new package submissions).
- Confirm that `manager_42` handled `:Update` projects properly and does not need an update?
- Not going to advertise override functionality like we discussed?

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -48,7 +48,6 @@ class Leaper(ReviewBot.ReviewBot):
         ReviewBot.ReviewBot.__init__(self, *args, **kwargs)
 
         # ReviewBot options.
-        self.only_one_action = True
         self.request_default_return = True
         self.comment_handler = True
 
--- a/leaper.py
+++ b/leaper.py
@@ -225,6 +225,27 @@ class Leaper(ReviewBot.ReviewBot):
 
             return review_result
 
+        if target_project.endswith(':Update'):
+            self.logger.info("expected origin is '%s' (%s)", origin,
+                             "unchanged" if origin_same else "changed")
+
+            if origin_same:
+                return True
+
+            good = self._check_matching_srcmd5(origin, target_package, src_srcinfo.verifymd5)
+            if good:
+                self.logger.info('submission source found in origin ({})'.format(origin))
+                return good
+            good = self.factory._check_requests(origin, target_package, src_srcinfo.verifymd5)
+            if good or good == None:
+                self.logger.info('found pending submission against origin ({})'.format(origin))
+                return good
+
+            return None
+        elif self.action.type == 'maintenance_incident':
+            self.logger.debug('unhandled incident pattern (targetting non :Update project)')
+            return True
+
         # obviously
         if src_project in ('openSUSE:Factory', 'openSUSE:Factory:NonFree'):
             self.source_in_factory = True
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -330,7 +330,7 @@ class ReviewBot(object):
         if self.comment_handler is not False:
             self.comment_handler_add()
 
-        overall = None
+        overall = True
         for a in req.actions:
             # Store in-case sub-classes need direct access to original values.
             self.action = a
@@ -340,8 +340,14 @@ class ReviewBot(object):
                 fn = 'check_action__default'
             func = getattr(self, fn)
             ret = func(req, a)
-            if ret == False or overall is None and ret is not None:
-                overall = ret
+
+            # In the case of multiple actions take the "lowest" result where the
+            # order from lowest to highest is: False, None, True.
+            if overall is not False:
+                if ((overall is True and ret is not True) or
+                    (overall is None and ret is False)):
+                    overall = ret
+
         return overall
 
     @staticmethod


leaper.py
```diff
@@ -225,6 +225,27 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
...
+                self.logger.info('submission source found in origin ({})'.format(origin))
+                return good
+            good = self.factory._check_requests(origin, target_package, src_srcinfo.verifymd5)
+            if good or good == None:
+                self.logger.info('found pending submission against origin ({})'.format(origin))
+                return good
+
+            return None
+        elif self.action.type == 'maintenance_incident':
+            self.logger.debug('incident not yet asigned a target release project')
```
lnussel: does that happen in practice?

lnussel:
Ok, by adding a new branch this code turns out surprisingly simple. Does it make sense to do it that way or would refactoring the other branches work?

packages with origin SLE should actually come from SLE. They'd come from the update project rather than GA though,

manager42 does not run for :Update I think. maintenance maintainers the lookup file manually so far. would be good if it could be used there too I guess.

lnussel:
thinking about it more it would be good if the bot could handle the origin :GA with submission coming from :Update. A real change like to a fork or Factory should be approved by leap-reviewers. So we get a handle in that process.

leaper.py
jberry-suse: No it does not. I was thinking about the silly `target_package` not being set until it is assigned to an incident.

jberry-suse:
> Ok, by adding a new branch this code turns out surprisingly simple. Does it make sense to do it that way or would refactoring the other branches work?

This means rewriting the entire thing from scratch. Considering we haven't work out how exacting the maintenance bit works I would prefer we do that first so we have a clear view of the full requirements before rewriting. But perhaps if has to handle the `:GA` -> `:Update` conversion and such it might warrant some.

> manager42 does not run for :Update I think. maintenance maintainers the lookup file manually so far. would be good if it could be used there too I guess.

Does this apply to SLE updates as well or is that handled differently? And is this bot expect to work there?

jberry-suse:
My primary concern is the submissions are not from `origin` as I stated. For example, the current open requests.

```
~:branches:OBS_...tained:libebml
~:branches:OBS_...:minetest-game
~:branches:OBS_...ained:minetest
~:branche...kde-l10n
~:branche...d:kopete
~:HaskellUpdate
~:branche...0:Update
~:branches:ope...p:42.3:Update
```

Obviously, one can search for source md5 in `origin` or we have to fallback to same searching logic used for Leap. Obviously, can just add the "is in factory" check, but what is the actual goal and expectation. The few times I have submitted maintenance requests they ended up being forks that were never present in any product.

jberry-suse:
> also missing clear statements if sources are not in factory and whether they are in sle, eg 627379

The original requirements provided where that it should always match the `origin` in lookup which means that is the only place that needs to be checked. As such that is what it done, but the requirements seem to be a bit more.

jberry-suse:
Pushed two additional commits related to multi-action support.

The following comment is produced as an example:

```
[D] adding comment to 628296: <!-- Leaper state=seen result=None -->

home:susnux:branches:OBS_Maintained:libebml/libebml.openSUSE_Leap_15.0_Update@22005d596f747f22b82339d18c46d3ae -> openSUSE:Leap:15.0:Update/libebml

expected origin is 'openSUSE:Factory' (changed)

submission source found in origin (openSUSE:Factory)

home:susnux:branches:OBS_Maintained:libebml/libebml.openSUSE_Leap_42.3_Update@58d44f1274425e9d715fa392544443fe -> openSUSE:Leap:42.3:Update/libebml

expected origin is 'openSUSE:Leap:42.2' (changed)
```

As far as I understood the original requirements this meets them. We need to resolve what we actually want going forward.

lnussel:
Jimmy Berry wrote:
>     manager42 does not run for :Update I think. maintenance
>     maintainers the lookup file manually so far. would be good if it
>     could be used there too I guess.
> 
> Does this apply to SLE updates as well or is that handled differently? And is this bot expect to work there?

I think SLE just doesn't care about origin after the release (and
doesn't have to). For Leap it's important as we do not want
lighthearted origin changes from SLE to something else.

lnussel:
Jimmy Berry wrote:
> The few times I have submitted maintenance requests they
> ended up being forks that were never present in any product.

That's the problem. If there's a chance code should be identical to
other code streams. A version update for maintenance for example should
come from Factory rather than crafting some fork.
A new package introduced to SLE should come from Factory or Leap
rather than created from scratch etc.
So far it was a blind spot so nobody cared. As soon as the
information is presented things start to change.

jberry-suse:
> If there's a chance code should be identical to
> other code streams.

So in this case, look for source from current products. Whereas the Leap logic also includes bits about previous products.

PR #1638, opened 2018-08-14
check_source: allow for dot in package name during in-air rename check.

jberry-suse:
Based on a report by mgerstner regarding IBS requests 170009 and 170008.

The logic passes for both requests in question, although one has other issues, and tested on a factory request. Also simulated update request to ensure regex logic matches properly.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 
 import os
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


jberry-suse:
Previously, this was likely not an issue since SLE ran with `'check-source-in-air-rename-allow': 'True'` which disabled this check, but the default config was removed by @coolo (fa2e4f7b750fdb5cb2181af73c4e4aa0ec991c87) and the setting is no longer present in remote configs for the three most recent SLE versions. Should be fixed regardless, but this explains why it would not have been noticed before now.

jberry-suse:
CI failing due to #1639.

PR #1640, opened 2018-08-14
repo_checker: properly support multi-layer projects

jberry-suse:
- a1cf089572535b7ea5e839d32c649c33b35d9f3d:
    repo_checker: support multi-layer projects during project_only run.
    
    Include problems from all the layers instead of just the top layer as this
    is effectively what the end-user would see.

- 9e862efeb7488fb81f1c14bb60d402fe8feacf13:
    repo_checker: properly support multi-layer projects during request mode.
    
    Previously, the additional layers supported added by coolo treated them
    like staging projects which meant that everything except the bottom
    override the bottom. Obviously, for SLE service packs this is not correct.
    
    This rewrorks the underlying perl script to support a stack of directories
    where the second directory is assumed the be the target project for the
    purposes of the toignore (-f) argument and the top layer is the only
    one for which problems are reported.

Proper implementation of #1624 and add support during `project_only`.

### DIFF
--- a/repo_checker.pl
+++ b/repo_checker.pl
@@ -21,6 +21,7 @@ my $arch = shift @ARGV;
 my @directories = split(/\,/, shift @ARGV);
 my %toignore;
 my %whitelist;
+my $filter = 1;
 while (@ARGV) {
     my $switch = shift @ARGV;
     if ( $switch eq "-f" ) {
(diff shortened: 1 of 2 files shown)


jberry-suse:
To make it clear, a follow-up will be needed to incorporate the repo state hash of each layer into overal state hash as OBS only updates the top hash if changes from below layers affect building, not run-time requirements.

jberry-suse:
Was going to do another set of verification runs against SLE, but IBS is completely dead.

PR #1646, opened 2018-08-16
osclib/core: maintainers_get(): rework to properly supports groups.

### LINKED ISSUE #1293: factoryauto: delete reques by maintainer requires maintainer approval
I've seen a couple delete requests recently that, imho, wrongly ask for maintainer approval.

e.g.: https://build.opensuse.org/request/show/554585

The request was filed by dimstar_suse, member of factory_maintainers group. This group in turn is listed at https://build.opensuse.org/package/users/filesystems/fusecompress as (inherited) maintainer (no local maintainers assigned to the package, only a bug owner)

Yet, factory-auto asked for a review by 'the package' (hence, its maintainers), with this comment:

> Submission for fusecompress by someone who is not maintainer in the devel project (filesystems).

jberry-suse:
- ffcee027c011a2870eeb2dfb7e18e809ffc04d1e:
    osclib/core: maintainers_get(): rework to properly supports groups.
    
    Previously, group members were only added if no maintainers were present
    whereas group members should always be considered. Additionally, project
    queries never included group members. Now groups are included in both.

- 038bec6113c6a1a0b9653ab3e4402beb3a7d663d:
    osclib/core: provide groups_members() to expand muliple groups users.

- 600a48674544034ea84affc30318042a9f43dc4f:
    Utilize ET.fromstringlist() intead of joining strings.

Fixes #1293.

Verified new logic using:
- `server:php:applications`: `['jbroedelGB', 'weberho', 'ralflangb1', 'poeml', 'computersalat', 'lrupp', 'arclyde', 'maxlin_factory', 'elvigia', 'dstoecker', 'asemen', 'lnussel_factory', 'dimstar_suse', 'factory-maintainer', 'aeneas_jaissle', 'peternixon', 'Marcus_H']`
- `server:php:applications/apache-rpm-macros`: `['poeml', 'factory-maintainer', 'maxlin_factory', 'elvigia', 'sr_hosseini', 'lnussel_factory', 'pgajdos', 'dimstar_suse', 'anicka', 'draht', 'kstreitova', 'darix']`
- `server:php:applications/adminer`: `['boombatower']`
- request `627699` for `libconfuse0`: `['darix']`

The discrepancy between `server:php:applications` packages is odd, but that's what OBS returns. Short of reading the meta directly (as is done for project only) the `/search/owner` call returns those values.

### DIFF
--- a/devel-project.py
+++ b/devel-project.py
@@ -92,7 +92,7 @@ def maintainer(args):
     apiurl = osc.conf.config['apiurl']
     devel_projects = devel_projects_load(args)
     for devel_project in devel_projects:
-        meta = ET.fromstring(''.join(show_project_meta(apiurl, devel_project)))
+        meta = ET.fromstringlist(show_project_meta(apiurl, devel_project))
         groups = meta.xpath('group[@role="maintainer"]/@groupid')
         intersection = set(groups).intersection(desired)
         if len(intersection) != len(desired):
@@ -207,7 +207,7 @@ def maintainers_get(apiurl, project, package=None):
                 meta = show_project_meta(apiurl, project)
     else:
         meta = show_project_meta(apiurl, project)
-    meta = ET.fromstring(''.join(meta))
+    meta = ET.fromstringlist(meta)
 
     userids = []
     for person in meta.findall('person[@role="maintainer"]'):
--- a/fcc_submitter.py
+++ b/fcc_submitter.py
@@ -187,7 +187,7 @@ class FccSubmitter(object):
         """
 
         f = osc.core.show_prj_results_meta(self.apiurl, project)
-        root = ET.fromstring(''.join(f))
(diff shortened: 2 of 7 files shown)


jberry-suse:
This should resolve related problems seen now and again in `leaper` since it uses the same code.

jberry-suse:
The only place I can think of that would cause an issue is `devel-project.py` and interestingly it has it's own implementation that only looks at users not groups and directly using project meta. So in theory it should ignore factory-maintainers for the purposes of say the notification emails.

The example in the original issue was in allowing factory-maintainers to submit a delete request for a package which would be directly impact by this change and thus allowed as was desired.

Adding a reviewing if not is present is done by adding the project/package combo so it is not effected by this change ... other than when that review is added since the initial check for is it needed will include a larger pool based on this change.

PR #1650, opened 2018-08-17
osclib/conf: exclude sub-projects from matching project patterns (and Leap:N:Update defaults)

jberry-suse:
- 5b44ea188d11415b1dc6a4f1de7536ccb769a0cb:
    osclib/conf: provide Leap:N:Update default settings.

- 77e27fae9eea59e945034efc58ab02f1d207f06a:
    osclib/conf: exclude sub-projects from matching project patterns.
    
    Otherwise, :Update projects match the primary product settings which
    gives the wrong impression to tools as things like :Update:Staging do
    not exist. Since :Update and other sub-projects are in a different phase
    of development they should really have separate settings. :Ports however
    does have stagings and in theory should act like parent product. If it
    becomes desirable to split that can be done in the future.
    
    Some policy settings related to ReviewBots likely are desirable to carry
    over, but that can be resolved in follow-up as they are not currently used.

@lnussel We've discussed the topic a few times, but would you like to be involved in leaper overrides for maintenance (ie. `leaper-override-group` setting included in this PR)?

I'll need to double check some of staging tools to make sure they are not accessing the target project's settings via `:Staging:...` anywhere or somesuch.

### DIFF
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -108,6 +108,11 @@ DEFAULT = {
         'mail-noreply': 'noreply@opensuse.org',
         'mail-release-list': 'opensuse-releaseteam@opensuse.org',
     },
+    r'openSUSE:(?P<project>Leap:(?P<version>[\d.]+):Update)$': {
+        'main-repo': 'standard',
+        'leaper-override-group': 'leap-reviewers',
+        'repo_checker-arch-whitelist': 'x86_64',
+    },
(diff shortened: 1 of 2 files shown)


jberry-suse:
I can also see an argument for `:Ports` projects having separate settings since things like `staging-group` may want to be separate, but not clear if and how they are being used currently.

PR #1655, opened 2018-08-18
osclib/core: provide project_pseudometa_*() functions (and related)

### LINKED ISSUE #1652: RFC: For non-products default to 00Meta as the psudometa (previous dashboard) container
This would allow devel projects and `:Update` to store the `repo_checker` output and similar in `00Meta` package. Re-using name since `Leap:N` and `Leap:N:Update` already contain it and has builds disabled and only stores meta data.

The location will be configurable via a project config settings so if devel projects want it somewhere else they can do so.

This is primarily necessary to resolve maintenance repo_checker workflow and they definitely need to see the output as many problems have been introduced post-release of _15.0_.

Thoughts?

jberry-suse:
- 1fe03c16f0b836db70d0bd4145161a1199952a60:
    osc-staging: update "config" documentation to point to OSRT:Config.

- 6d60708b5979424faef55f0794aa9744168df798:
    osclib/core: source_file_save(): always include host in message.

- 536e17cb790628fadaf3d2c30a4bbbe87b8100f0:
    osclib/stagingapi: rename dashboard_content_* methods to pseudometa_file_*.
    
    No need to be confusing and use the deprecated dashboard name.

- 24b15d6b0bf2a161a8c9dc9ca64738d54b02bb67:
    Replace relevant 'dashboard' package references with pseudometa lookup.
    
    Skipping some of the broader once which have no tangible impact, but are
    expensive to change.

- 5e6df089a6ff8438d877f9e2ad7b0d7b504ea9ef:
    osclib/stagingapi: drop {load,save}_file_content() methods.
    
    Since the implementations have been made generic and dropped to osclib.core
    there is not point in having wrappers in StagingAPI with the same args.

- 26884da4586b6f6254d3a8628b7977c8da9feabb:
    osclib/core: provide project_pseudometa_*() functions.
    
    These methods provide for a generic place to store meta data related to
    a project. For the time being, keep the original :Staging/dashboard
    location for openSUSE products.

- 89ab151a59ed044ab77b07d4cfd706494637c7b7:
    osclib/core: provide source_file_save() adapted from StagingAPI.

Just as before the expectation is that the package is created and writable.

Fixes #1652.

### DIFF
--- a/devel-project.py
+++ b/devel-project.py
@@ -73,11 +73,11 @@ def list(args):
 
         if args.write:
             api = staging_api(args)
-            api.dashboard_content_ensure('devel_projects', out, 'devel_projects write')
+            api.pseudometa_file_ensure('devel_projects', out, 'devel_projects write')
 
 def devel_projects_load(args):
     api = staging_api(args)
-    devel_projects = api.dashboard_content_load('devel_projects')
+    devel_projects = api.pseudometa_file_load('devel_projects')
 
     if devel_projects:
         return devel_projects.splitlines()
--- a/devel-project.py
+++ b/devel-project.py
@@ -267,7 +267,7 @@ if __name__ == '__main__':
 
     parser_list = subparsers.add_parser('list', help='List devel projects.')
     parser_list.set_defaults(func=list)
-    parser_list.add_argument('-w', '--write', action='store_true', help='write to dashboard container package')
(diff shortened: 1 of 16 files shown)


jberry-suse:
Not as hard as last time since I already refactored 99% of the 'dashboard' usages to use the common functions.

PR #1660, opened 2018-08-22
ReviewBot: add devel_project_review_*() methods adapted from maintbot (and port users)

### LINKED ISSUE #1645: check_maintenance_incidents: evaluate origin based logic and overlap with leaper
https://github.com/openSUSE/openSUSE-release-tools/blob/fa1f929d19a57f2e1035e1a4d062e36125e45741/check_maintenance_incidents.py#L99-L112

Seems like a) this is/should be covered by `leaper` now assuming maintenance switches to it, b) leaper calls this code to add maintainer review so should likely move there if not already and disable the "maintbot" by itself. Eventually this entire file should be dropped and the couple function just made library functions.

This should happen before new set of tools deployed for maintenance.

jberry-suse:
- dab65356d755746c93269adba1d5cc7e19637b3e:
    leaper: replace maintbot usage with ReviewBot methods.
    
    The maintenance_incident logic is always updated to mimic the maintbot
    behavior in regards to when to add devel project review.

- 31d6a872c7a3a09ccb24f93284e374ba989375fe:
    check_source: replace maintbot usage with ReviewBot methods.

- a1c13348037a82d37356b97d7042808d852a263b:
    ReviewBot: add devel_project_review_*() methods adapted from maintbot.
    
    The owner logic surrounding a package removed from Factory does not appear
    to make sense as the current behavior of OBS never returns another owner
    pair for such packages. As such the existing devel project lookup makes
    more sense and is more straight forward.

This code causes headaches and takes far longer to figure out how to change than to write from scratch. :(((((((((

This should bring the maintenance workflow fully into line on the new tools. Any workflow changes should be done in leaper. Since the maintbot is not ported and deployed out of package it will continue running out of separate checkout and should be disabled once happy with new tools.

Fixes #1645.

### DIFF
--- a/leaper.py
+++ b/leaper.py
@@ -39,7 +39,6 @@ from osclib.core import devel_project_get
 import urllib2
 import yaml
 import ReviewBot
-from check_maintenance_incidents import MaintenanceChecker
 from check_source_in_factory import FactorySourceChecker
 
 class Leaper(ReviewBot.ReviewBot):
@@ -53,7 +52,6 @@ class Leaper(ReviewBot.ReviewBot):
 
         self.do_comments = True
 
-        self.maintbot = MaintenanceChecker(*args, **kwargs)
         # for FactorySourceChecker
         self.factory = FactorySourceChecker(*args, **kwargs)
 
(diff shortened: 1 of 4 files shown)


jberry-suse:
Rebased to include appropriate dist/* changes.

leaper.py
```diff
@@ -125,6 +123,26 @@ def _check_same_origin(self, origin, project):
...
+        # really desired a "post check source submission" not
+        # check_one_request() which applies to all action types and all actions
+        # at once. The follow-ups need the same context that the main method
+        # determining to flip the switch had. For maintenance incidents the
+        # ReviewBot code does a fair bit of mangling to the package and project
+        # values passed in which is not present during check_one_request().
+        # Currently the only feature used by maintenance is
+        # do_check_maintainer_review and the rest of the checks apply in the
+        # single action workflow so this should work fine, but really all the
+        # processing should be done per-action instead of per-request.
```
lnussel: I don't get your rant. If leaper has code in check_one_request() that needs to be moved to check_source_submission() or even to action specific functions then what prevents us from doing that? Also clearly the existing check_source_submission() in leaper is a mess and needs refactoring.

jberry-suse: It's mixing overall (ie all action) follow-ups and per-action toggle follow-ups. It could be moved, but for example it had to be moved for devel project review stuff since the loop on actions was in the removed code so it needed to be in the check_source_submission() since that is per-action.

This was always a problem, but since we enforce single-actions for the non-maintenance projects it is never exposed.

jberry-suse: To be clear, this means for a request with > 1 action it adds the devel reviews for each action an additional time per action. Meaning `add_review()` is called `len(actions) * len(actions)`. Obviously if `len(actions) = 1` it's not so much of an issue and since I made `add_review()` avoid duplicates it also would hide the negative side effects, but for maintenance where they commonly have >10 actions you would see 100 reviews added. :)

jberry-suse: As an side note, this whole process seems wrong since we do not really want to add "devel project" review, we want to add "expected origin" review. Which in many cases in devel, but not always. This was attempted to be handled via `/search/owner`, but that just falls back to priority list of OBS, in the case where the product origin differs from Factory the reviewer is wrong. This has complicated some of the weird results seen by leaper. I'll make an issue under Origin Manager milestone to consider this.

jberry-suse:
Split out maintbot removal into #1664 for merging later.

PR #1674, opened 2018-08-24
LICENSE: add proper GPLv2 license and drop unnecessary headers.

jberry-suse:
- 6069245350d057dc0f4ea2bcc9b69fb238932a38:
    Remove SUSE copyright, warranty, and license headers.
    
    Distinct copyrights were left as I do not wish to track down commit
    history to ensure it properly documents the copyright holders. Also left
    non-GPLv2 licenses and left bs_copy untouched as a mirror from OBS.
    
    Already have a mix of with and without headers and even OBS does not place
    on majority of files. If SUSE lawyers have an issue it will come up in
    legal review for Factory.

- 36988ed1d32096a8fc1f6a5a660b6c4587508ed8:
    LICENSE: add proper GPLv2 license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,339 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 2, June 1991
+
+ Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
+Foundation's software and to any other program whose authors commit to
+using it.  (Some other Free Software Foundation software is covered by
+the GNU Lesser General Public License instead.)  You can apply it to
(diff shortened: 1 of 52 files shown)


bmwiedemann:
You could have a look at the GPL's section "How to Apply These Terms to Your New Programs"
E.g. now it is not clear if GPLv2-or-later applies or just GPLv2-only.

jberry-suse:
Feel free to submit a PR, but per Linux's example there is no v2 only clause which based on common practice means v2+. I like v2-only on my projects, but given it was even more ambiguous before since most was unlicensed and some mentioned v2 this is an improvement.

The spec matches this (v2+):

https://github.com/openSUSE/openSUSE-release-tools/blob/541b94d1d022e94edb70fbc897422bfa8cb22d8f/dist/package/openSUSE-release-tools.spec#L26

as small bits or explicitly MIT.

jberry-suse:
I'll also be sure to go around stamping every part in my car, kitchen appliances, bike, etc instead of just stating it for the whole.

coolo:
As a rule of thumb: put an ID on every thing you're likely to borrow to someone else. This is the same with the open source: if you want your source to be borrowed/branched/extended - make sure it contains the rules.

A one liner does though (after the copyright):
```
SPDX-License-Identifier: GPL-2.0+
```

https://spdx.org/sites/cpstandard/files/pages/files/using_spdx_license_list_short_identifiers.pdf

jberry-suse:
Exactly, I'd put a label on my bike or laptop, but not on every screw, transistor, and part within them. Hence the LICENSE file.

PR #1678, opened 2018-08-29
metrics: provide initial telegraf setup for ingesting historical snapshots

jberry-suse:
After deployment of [poo#36766](https://progress.opensuse.org/issues/36766) the pretty storage consumption graph reading from AWS Cloudwatch will no longer be relevant (long-term). The re-creates essentially the same thing, plus snapshot count, but using download.o.o as a source.


jberry-suse:
Preview (data is super boring):

image

jberry-suse:
Deployed as https://metrics.opensuse.org/d/osrt_history/osrt-history.

Issue #1681, opened 2018-08-30
CI: odd failure of factory_source_tests


jberry-suse:
I cannot reproduce this locally and started occurring after `README.md` update. I assume something on travis-ci side, a dependency, or something else external to this repo is causing the issue.

```
======================================================================
FAIL: test_accept_request (tests.factory_source_tests.TestFactorySourceAccept)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py", line 165, in test_accept_request
    self.assertTrue(result['status'])
AssertionError: None is not true
-------------------- >> begin captured logging << --------------------
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: INFO: checking 770001
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: INFO: Base:System/timezone@481ecbe0dfc63ece3a1f1b5598f7d96c -> openSUSE:13.2/timezone
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: INFO: Checking in project openSUSE:Factory
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: DEBUG: checking timezone in openSUSE:Factory
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: DEBUG: new package
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: DEBUG: checking requests
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: ERROR: caught exception while checking openSUSE:Factory/timezone
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: INFO: timezone failed source submission check
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: INFO: 770001 ignored
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: INFO: checking 770001
/home/travis/build/openSUSE/openSUSE-release-tools/tests/factory_source_tests.py: INFO: Base:System/timezone@481ecbe0dfc63ece3a1f1b5598f7d96c -> openSUSE:13.2/timezone
[...]

PR #1682, opened 2018-08-31
dist/spec: include as requirement of check-source sub package.

jberry-suse:
Discovered missing while building container from scratch.


jberry-suse:
CI failure due to #1681.

PR #1686, opened 2018-09-04
Cache migrate prune

### LINKED ISSUE #1583: Remove reference to old repo name (osc-plugin-factory) re-introduced in #1544
#1544

jberry-suse:
- b59f384732a01d0c02a305925b1342a3a3ad9cb9:
    osclib/cache: provide mechanism to init multiple times.
    
    Useful for staging plugin to cache the remote config before StagingAPI
    is initialized.

- c278d9c25c11116779d6c57da9aab01a6e5edcad:
    Utilize CacheManager for all existing caches.

- 5d87145a246ebafb7b5eeb0c63131c39bc7afd54:
    osclib/cache_manager: provide general cache manager for all tools.

- f49b8d114971f3935b00e67da51e3b45cb4b749d:
    osclib/common: provide project name and version constants.
    
    Not specific to staging plugin, but previously only place to provide
    version.

The caches currently slowly build up as things like the major Leap/SLE products are releases and the tools disabled for them. The old caches remain on disk and are never removed. This requires either manual pruning or manual wiping of the caches. Pruning is not trivial since things like pkglistgen need to access repository meta caches for prior versions to produce drop list, but do not require any of the staging caches. Additionally, metrics caches data that never changes (like revisions and access log summaries) which should be kept forever, but are still caches since they can be re-created if desired. As such the only consistent and reliable metric for pruning is the last accessed time since things like metrics will continually read such files, but never update them. This still allows for individual pruning and cache expiration mechanisms, but provides an overall pruning mechanism.

This also moves _all_ caches under `openSUSE-release-tools` parent directory instead of the variety of other directories with name prefixes that changed 2-3 times. A migration routine is provided to move or remove old caches. Thus it also fixes #1583.

The prune is currently configured to run once a week (whenever a cache is accessed) and remove anything that has not been accessed in over a month.

This is also useful for determining accurate disk usage of `tortuga` for gather requirements for new hosting location.

The one cache that is accessed entirely outside of python (and thus cache manager) is the old pkglistgen bits surrounding `rebuildpacs.pl` which will need to be migrated to new directory manually when deployed or just `rm -r` the old directory later. This is deployed outside of OSRT package since deprecated so whoever decide to update that should just remove the directory later.


jberry-suse:
This also makes the cache path generation consistent and based on xdg package instead of mix of that and `os.path.expanduser`.

jberry-suse:
Locally on development user, this dropped 50GB of caches, but on my factory user only around 50MB. Will depend on tool used and the scope of projects.

PR #1691, opened 2018-09-12
check_source: reset skip_add_reviews for each request

lnussel:
skip_add_reviews needs to be reset, otherwise a single
maintenance_incident request affects the rest of the work queue.


jberry-suse:
Yeah, as I didn't like adding it, but based on the target projects having some of these config items it makes it really awkward.

PR #1693, opened 2018-09-13
check_source: ignore _product package check

nilxam:


check_source.py
```diff
@@ -66,8 +66,8 @@ def check_source_submission(self, source_project, source_package, source_revisio
             self.review_messages['declined'] = 'Only one action per request allowed'
             return False
 
-        if target_package.startswith('00'):
-            self.review_messages['accepted'] = 'Skipping all checks for 00* packages'
+        if target_package.startswith('00') or target_package.startswith('_'):
```
jberry-suse: Perhaps it's time to standardize these prefixes or make it a configuration option.

PR #1695, opened 2018-09-14
metrics: add image totals by product to access dashboard.

jberry-suse:
- 5c410b7c76aa85b5a0eb6b0859717f1e0aefe9c4:
    metrics/grafana/access: add new row and graphs for image totals by product.
    
    Includes a new variable for switching which product is displayed.
    Unfortunately, there is no way to hide the variable within a row and it
    seems like over-kill to create a separate dashboard.

- 7492b109e5f370988a2e95b9c22eded81e6d734b:
    metrics/access/aggregate: process and summarize image totals by product.
    
    Accessible in new measurement prefixed by 'image'.

- c84add0bf7081ce839f3c95a5e4d66f26b5cf28a:
    metrics/access/aggregate: extract merge_product_plus_key() from merge().

- dc9afe2adf16e1375001eca9bc73c2211f73e27c:
    metrics/access/ingest: detect product image paths and included in dump.

- 6203b52a420c66bb8c6f3b6ee0d4a4bf36803d50:
    metrics/grafana/access: remove left-over raw queries.
    
    For some reason Grafana does not remove these from data structure even
    though they no longer represent current state.

- f80d3ff03fe0ed4e4e9444b250bf4303105fc8f4:
    metrics/access/aggregate: include Leap 15.1 in published metrics.

Rough data from last week or so only parsing first million log entries:

image

I plan to reparse everything available from pontifex as that is trivial (just takes CPU). As such that is back to `2017-12-04` although that is 3 days before the _black hole_ so might as well just start at `2018-03-08`. Can re-parse langley if desired, but that requires copying data across networks (or super slow over VPN to my machine and back up).

### DIFF
--- a/metrics/access/aggregate.php
+++ b/metrics/access/aggregate.php
@@ -248,20 +248,25 @@ function merge(&$data1, $data2)
     $data1['total_product'][$product] += $data2['total_product'][$product];
   }
 
-  foreach ($data2['unique_product'] as $product => $unqiue) {
-    if (empty($data1['unique_product'][$product]))
-      $data1['unique_product'][$product] = [];
-
-    foreach ($unqiue as $uuid => $count) {
-      if (empty($data1['unique_product'][$product][$uuid]))
-        $data1['unique_product'][$product][$uuid] = 0;
-
-      $data1['unique_product'][$product][$uuid] += $data2['unique_product'][$product][$uuid];
-    }
-  }
+  merge_product_plus_key($data1['unique_product'], $data2['unique_product']);
 
   $data1['total_invalid'] += $data2['total_invalid'];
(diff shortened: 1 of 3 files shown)


jberry-suse:
Updated package on `metrics` and ran:
```
rm /home/osrt-metrics/.cache/openSUSE-release-tools/metrics-access/ipv{4,6}/2018-0{3..9}-{01..31}.json
```

Started the `osrt-metrics-access` service.

jberry-suse:
The raw (non-unique) download counts for images are available in the `Image Media - $product`section at the bottom of https://metrics.opensuse.org/d/osrt_access/osrt-access and are controllable via the product selector at the top of page (collapse the `Product` for easy switching).

PR #1697, opened 2018-09-17
check_source: decline all delete project requests and ReviewBot sub-delete check types.

### LINKED ISSUE #1696: check_source should decline a delete request which with no target package
People submit a delete request against a project eg. openSUSE:Factory https://build.opensuse.org/request/show/635926 which is disallowed, but it will staying in backlog until Factory maintainer declining this submission, the worst case is this kind of delete request might break down staging dashboard in case this submission is on top of backlog request and dashboard try to show it in the interface. https://errbit-opensuse.herokuapp.com/apps/5620ca2bdc71fa00b1000000/problems/5b9e0ec689299f00140e91ad

check_source should decline a delete request with no target package defined.

jberry-suse:
- 406fc41a851a594f3cbcfa11b7afe9d793de3203:
    ReviewBots: utilize delete_package check where it was already assumed.

- 658137b9e973b3fab66b0060d98acf44d71d9105:
    check_source: split out delete repository check to new sub action check.

- ecbc806b2d480ee1d92b64fb1b20b733a10e51e4:
    check_source: decline all delete project requests.

- d1b630b5dab3132f6dfaaf6febd825e59a190f8c:
    ReviewBot: break down sub-types of delete requests.
    
    Their is very little overlap in the types of reviews one typically wants
    to perform on delete requests. Already all of the existing code will crash
    if a delete project request is reviewed.

Fixes #1696, by producing:

```
$ ./check_source.py --debug --dry id 635926
[I] checking 635926
[I] 635926 declined: Deleting the openSUSE:Factory project is not allowed.
[D] 635926 review not changed
```

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -344,10 +344,7 @@ class ReviewBot(object):
             # Store in-case sub-classes need direct access to original values.
             self.action = a
 
-            fn = 'check_action_%s'%a.type
-            if not hasattr(self, fn):
-                fn = 'check_action__default'
-            func = getattr(self, fn)
+            func = getattr(self, self.action_method(a))
             ret = func(req, a)
 
             # In the case of multiple actions take the "lowest" result where the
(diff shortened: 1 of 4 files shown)


jberry-suse:
This also prevents all three from crashing on delete repository and project requests of which only check_source handled before.

PR #1702, opened 2018-09-20
pkglistgen: compare repository state to determine when to update solv.

### LINKED ISSUE #1377: pkglistgen: do_update() skips generating new solv when no updates indicated by bs_mirror
pkglistgen doesn't update the solv file if bs_mirrorfull didn't do anything. Since repochecker also calls bs_mirrorfull it would fool pkglistgen to work on an outdated solv file

jberry-suse:
Since both pkglistgen and repo_checker share the same rpm metadata cache
either may update it first and thus bs_mirrorfull will only contain output
for one of the tools. As such the check in pkglistgen could cause the solv
file to remain stale. Instead write a secondary file containing the
repository state hash and check to see if the file already exists.

Fixes #1377.

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -22,6 +22,7 @@ from osc import conf
 from osclib.cache_manager import CacheManager
 from osclib.conf import Config, str2bool
 from osclib.core import repository_path_expand
+from osclib.core import repository_state
 from osclib.stagingapi import StagingAPI
 from osclib.util import project_list_family
 from osclib.util import project_list_family_prior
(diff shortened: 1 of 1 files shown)


pkglistgen.py
```diff
@@ -732,21 +733,32 @@ def do_update(self, subcmd, opts):
                 if not os.path.exists(d):
                     os.makedirs(d)
 
+                # Fetch state before mirroring in-case it changes during download.
+                state = repository_state(self.tool.apiurl, project, repo)
```
coolo: this is buggy - it's pretty non-obvious, but it turns out the repository_state is not the state of the repository's binaries, but just the state of the build statuses. So we will skip package list generation of all packages go from succeeded to succeeded in between 2 checks.

This lead to outdated caches on SP1. I reverted this change locally, but we need a proper fix.

One possibility is always mirroring and then calculating a state hash based on the mirrored directory ourselves. 
@mlschroe suggested to have another view parameter on the repository to get e.g. the mtime of the solv file, so that's another possibility.

jberry-suse: My first thought was to hash the mirrored file list, but figured repo state hash was better. Hmm, having the extra view parameter would seem better as it would work for repo-checker as well, but if that isn't going to happen can go ahead with hash of file list.

PR #1709, opened 2018-09-26
Move creation of the Solver object out of loop

coolo:
It's not the state of the binaries, but about the build status - which
is irrelevant for package list generation. So just keep mirroring as
we used to, but instead of trusting the download we check the generated
solv file before and after. This is actually rather fast (on a VM in NBG)
and we can still skip the expensive package list generation

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -44,6 +44,7 @@ import traceback
 import random
 import shutil
 import string
+import time
 
 import ToolBase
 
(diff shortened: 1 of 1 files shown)


jberry-suse:
#1710 continues to allow skipping solv and repository sync while correcting the implementation for `repo-checker` as well. My vote would be to drop the `Don't trust the repository_state` commit and keep `Move creation of the Solver object out of loop`.

PR #1710, opened 2018-09-26
osclib/core: repository_state(): rework to utilize binarylist view.

jberry-suse:
For whatever reason the state hash only includes the overall publish states and not the binaries within the repository. As such, rebuilding state before a  rebuild and after may be the same.

Follow-up after #1702. The fix is better than #1709 since it prevents the needless repo sync and re-generation of solv file when the state has not changed. It also correct any issues that `repo-checker` might encounters as it also makes uses of the hash.


coolo:
This might sound like the cooler fix, but I had to abort
osc api '/build/openSUSE:Factory/_result?view=binarylist&multibuild=1&repository=standard'
after 4 1/2 minutes 

While updating the synced repository and creating a solv file from it takes less 20 seconds.

coolo:
/build/openSUSE:Factory/standard/x86_64/_repository?view=binaryversions is the route that bs_mirrorfull fetches and that takes 10-12 seconds here. I guess I will come up with a 3rd solution

coolo:
#1711 should work for repo checker *and* be quick

jberry-suse:
#1711 just took a minute and a half which is longer than I've ever seen `_result` take. Neither of these are ideal. hmm.

PR #1711, opened 2018-09-27
Jberry suse osclib core repository state

coolo:


osclib/core.py
```diff
@@ -382,9 +382,22 @@ def repository_path_search(apiurl, project, search_project, search_repository):
...
 def repository_state(apiurl, project, repository):
-    return ET.fromstringlist(show_results_meta(
-        apiurl, project, multibuild=True, repository=[repository])).get('state')
+    # Unfortunately, the state hash reflects the published state and not the
+    # binaries published in repository. As such request binary list and hash.
+    url = makeurl(apiurl, ['build', project, repository])
+    root = ETL.parse(http_GET(url)).getroot()
+    combined_state = ''
+    for arch in root.findall('entry'):
+        combined_state += repository_arch_state(apiurl, project, repository, arch.get('name'))
```
jberry-suse: Is it faster to make multiple calls instead of one?

osclib/core.py
```diff
@@ -382,9 +382,22 @@ def repository_path_search(apiurl, project, search_project, search_repository):
...
+                  {'view': 'binaryversions'})
+    from osclib.util import sha1_short
+    return sha1_short(http_GET(url).read())
+
 def repository_state(apiurl, project, repository):
-    return ET.fromstringlist(show_results_meta(
-        apiurl, project, multibuild=True, repository=[repository])).get('state')
+    # Unfortunately, the state hash reflects the published state and not the
+    # binaries published in repository. As such request binary list and hash.
+    url = makeurl(apiurl, ['build', project, repository])
```
jberry-suse: Already a `target_archs` function.

osclib/core.py
```diff
@@ -382,9 +382,22 @@ def repository_path_search(apiurl, project, search_project, search_repository):
...
+    root = ETL.parse(http_GET(url)).getroot()
+    combined_state = ''
```
jberry-suse: Perhaps utilize the list feature of `sha1_short`?

jberry-suse:
To avoid client-side caching and what not and hopefully allow for OBS side caching.

```bash
#!/bin/bash

time_multiple()
{
  counter=0
  while [ $counter -lt $1 ] ; do
    time $2 | wc -l
    let counter=counter+1
  done
}

repository_state_both()
{
  osc api '/build/openSUSE:Factory/_result?multibuild=1&repository=standard&view=binarylist'
}

repository_state_seperate()
{
  osc api '/build/openSUSE:Factory/standard'
  osc api '/build/openSUSE:Factory/standard/x86_64/_repository?view=binaryversions'
  osc api '/build/openSUSE:Factory/standard/i586/_repository?view=binaryversions'
}


echo warmup
time_multiple 2 repository_state_both
time_multiple 2 repository_state_seperate

echo real
time_multiple 4 repository_state_both
time_multiple 4 repository_state_seperate
```

Unfortunately, I got 502 Bad Gateway multiple times.

coolo:
Actually for the sake of repositoryhash, I don't even think we need binaryversions at all. Because mtime changes should be good enough to retrigger a resync

coolo:
And for pkglistgen the change to per arch most definitely gains us a lot - as we call repository_state per arch.

coolo:
warming up OBS side cache is pretty non meaningful - as we run this every half an hour once per staging, so you will most likely hit cold OBS caches anyway

coolo:
To give you the results of your script:
repository_state_both: minimum 20.392s, maximum 21.946s
repository_state_seperate: minimum 15.019s, maximum: 19.092s

If we ignore the binaryversions, we end up with:
minimum 10,556s, maximum 12,957s

osclib/core.py
coolo: yes, it is

osclib/core.py
coolo: adapted

PR #1718, opened 2018-10-01
pkglistgen: Make sure we don't calculate non-existant architectures i…

coolo:
…n stagings

Otherwise the package list is not staging specific

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -25,6 +25,7 @@ from osclib.conf import Config, str2bool
 from osclib.core import repository_path_expand
 from osclib.core import repository_arch_state
 from osclib.core import source_file_ensure
+from osclib.core import target_archs
 from osclib.stagingapi import StagingAPI
 from osclib.util import project_list_family
 from osclib.util import project_list_family_prior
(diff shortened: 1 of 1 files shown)


pkglistgen.py
```diff
@@ -1209,6 +1210,12 @@ def real_update_and_solve(self, opts):
                 if opts.staging_project and letter != opts.staging_project:
                     continue
                 opts.project = api.prj_from_short(letter)
+                filtered_archs = []
+                # make sure we only calculcate existant architectures
+                for arch in target_archs(apiurl, opts.project, main_repo):
+                    if arch in self.options.architectures:
+                        filtered_archs.append(arch)
+                self.options.architectures = filtered_archs
```
coolo: damn, this does not work. It won't create the group we need later :(

coolo:
That goes a little too deep, but it works. Suggestions how to improve this are welcome

pkglistgen.py
```diff
@@ -1233,6 +1235,12 @@ def update_and_solve_target(self, api, target_project, target_config, main_repo,
         product = target_config.get('pkglistgen-product', '000product')
         release = target_config.get('pkglistgen-release', '000release-packages')
 
+        opts.filtered_architectures = []
```
jberry-suse: Could use python set intersection like `repo_checker`, but obviously this works fine as well.

https://github.com/openSUSE/openSUSE-release-tools/blob/39a3f83180e0e54658454da91662d1eabbba2c4e/repo_checker.py#L109-L118

PR #1719, opened 2018-10-04
Do no longer CC submitters in request comments

coolo:
Submitters complain more and more about the spam they're getting from
staging projects - and we rather leave that weapon for actual feedback.

One especially annoying case is if a package is added to one staging project
and then later moved to another, the submitter will receive notifications
of all kind of bots and actions for both staging projects.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1555,7 +1555,7 @@ class StagingAPI(object):
             requests = meta['requests']
 
         for req in requests:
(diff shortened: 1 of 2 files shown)


coolo:
RFC @DimStar77 @nilxam @lnussel @fcrozat

coolo:
discussed in meeting and no objections there.

fcrozat:
fine for me

PR #1720, opened 2018-10-05
Add helper bot to listen to rabbit bus and feed OBS with openQA Status

coolo:
There is an API gap on OBS side atm, so this won't complete:
https://github.com/openSUSE/open-build-service/issues/6035

### DIFF
--- a/rabbit-openqa.py
+++ b/rabbit-openqa.py
@@ -7,18 +7,21 @@ import sys
 import json
 import osc
 import re
+from time import sleep
 from osc.core import http_GET, http_POST, makeurl
+from M2Crypto.SSL import SSLError as SSLError
 from osclib.conf import Config
 from osclib.stagingapi import StagingAPI
 from lxml import etree as ET
 from openqa_client.client import OpenQA_Client
+from openqa_client.exceptions import ConnectionError
 from urllib import quote_plus
 import requests
 try:
-    from urllib.error import HTTPError
+    from urllib.error import HTTPError, URLError
 except ImportError:
     # python 2.x
-    from urllib2 import HTTPError
+    from urllib2 import HTTPError, URLError
 from PubSubConsumer import PubSubConsumer
 
 
@@ -259,7 +262,11 @@ if __name__ == '__main__':
     for entry in root.findall('project'):
         l.add(Project(entry.get('name')))
 
-    try:
-        l.run()
-    except KeyboardInterrupt:
-        l.stop()
+    while True:
+        try:
+            l.run()
+        except KeyboardInterrupt:
+            l.stop()
+        except (HTTPError, URLError, ConnectionError, SSLError):
+            # OBS/openQA hickup
+            sleep(10)
--- a/rabbit-openqa.py
+++ b/rabbit-openqa.py
@@ -194,13 +194,10 @@ class Listener(object):
     def on_message(self, unused_channel, method, properties, body):
         if method.routing_key == '{}.obs.repo.published'.format(amqp_prefix):
             self.on_published_repo(json.loads(body))
-
-        if method.routing_key == '{}.openqa.job.done'.format(amqp_prefix):
+        elif re.search(r'.openqa.', method.routing_key):
             self.on_openqa_job(json.loads(body).get('ISO'))
-        if method.routing_key == '{}.openqa.job.create'.format(amqp_prefix):
-            self.on_openqa_job(json.loads(body).get('ISO'))
-        if method.routing_key == '{}.openqa.job.restart'.format(amqp_prefix):
-            self.on_openqa_job(json.loads(body).get('ISO'))
+        else:
+            print("unknown rabbitmq message {}".format(method.routing_key))
 
     def listen(self):
         print(' [*] Waiting for logs. To exit press CTRL+C')
--- a/rabbit-openqa.py
+++ b/rabbit-openqa.py
@@ -11,7 +11,13 @@ from osclib.conf import Config
 from osclib.stagingapi import StagingAPI
 from lxml import etree as ET
 from openqa_client.client import OpenQA_Client
-
+from urllib import quote_plus
+import requests
+try:
+    from urllib.error import HTTPError
+except ImportError:
+    #python 2.x
+    from urllib2 import HTTPError
 
 class Project(object):
     def __init__(self, name):
@@ -71,8 +77,12 @@ class Project(object):
                                         self.map_openqa_result(job),
                                         job['settings']['TEST'] + '@' + job['settings']['MACHINE'])
             url = self.api.makeurl(['status_reports', 'published', staging, 'images', 'reports', buildid])
-            http_POST(url, data=xml)
-            print(url, xml)
+            try:
+                http_POST(url, data=xml)
+            except HTTPError:
+                # https://github.com/openSUSE/open-build-service/issues/6051
+                print('failed to post status to ' + url)
+                print(xml)
 
     def update_staging_status(self, project):
         for iso in self.staging_projects[project]['isos']:
@@ -119,7 +129,6 @@ class Project(object):
         se.text = name
         return ET.tostring(check)
 
-
 class Listener(object):
     def __init__(self, amqp_prefix, amqp_url, openqa_url):
         self.projects = []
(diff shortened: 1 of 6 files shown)


coolo:
The checks section is what this bot feeds:

ss

coolo:
CC @ChrisBr

rabbit-openqa.py
```diff
@@ -0,0 +1,209 @@
...
+
+        if method.routing_key == '{}.openqa.job.done'.format(amqp_prefix):
+            self.on_openqa_done(json.loads(body))
+        if method.routing_key == '{}.openqa.job.create'.format(amqp_prefix):
+            self.on_openqa_create(json.loads(body))
+        if method.routing_key == '{}.openqa.job.restart'.format(amqp_prefix):
+            self.on_openqa_restart(json.loads(body))
+
+    def listen(self):
+        self.channel.start_consuming()
```
jberry-suse: Curious, how does this handle disconnects? When experimenting with a rabbit concept last week the connection would periodically close. Would imagine graceful reconnects would be important.

rabbit-openqa.py
```diff
@@ -0,0 +1,209 @@
...
+    l = Listener(amqp_prefix, amqp_url)
+    if amqp_prefix == 'opensuse':
+
+        class Leap15(Project):
+            def __init__(self, name):
+                super(name)
+
+        l.add(Leap15('openSUSE:Leap:15.1'))
+    else:
+        class Sle15(Project):
```
jberry-suse: Would seem this would be fit better in remote config rather than essentially hard-coding.

rabbit-openqa.py
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/python
```
jberry-suse: Should be executable, right?

rabbit-openqa.py
```diff
@@ -0,0 +1,209 @@
...
+        connection = pika.BlockingConnection(pika.URLParameters(amqp_url))
+        self.channel = connection.channel()
+
+        self.channel.exchange_declare(exchange='pubsub', exchange_type='topic', passive=True, durable=True)
+
+        result = self.channel.queue_declare(exclusive=True)
+        queue_name = result.method.queue
+
+        self.channel.queue_bind(exchange='pubsub',
+                                queue=queue_name,routing_key='#')
```
jberry-suse: Binding to events of interest (listed in `on_message`) should seem preferable.

rabbit-openqa.py
coolo: This is basically the basic rabbitmq tutorial code - but you can have callbacks for on_disconnect also and reconnect. 

For not getting disconnects, it's rather important to define a heartbeat_interval. And also we would need to define a server side message queue instead of using pubsub directly.

rabbit-openqa.py
coolo: Yes, but I'm super afraid of writing one regexp to rule it all. I guess it's possible, but having a function was easier in the end.

We would still need to define somewhere what projects to look at.

rabbit-openqa.py
coolo: Yes. As the title says: WIP

coolo: And this queue bind would happen server side (see above) and we would consume everything of *that* queue

ChrisBr:
@coolo jfyi I'm out of office until next week Monday so won't have much time to look into this ...

rabbit-openqa.py
coolo: Fixed

rabbit-openqa.py
coolo: Now that I implemented a cold start, I decided against using a custom queue and just bind the 4 routing keys on pubsub.

rabbit-openqa.py
coolo: https://pika.readthedocs.io/en/0.9.14/examples/asynchronous_consumer_example.html is most likely the better base.

coolo:
Hmm, the way I did it now we might miss events happening during startup phase. Not sure how relevant it is, but the problem exists.

rabbit-openqa.py
coolo: I think I found a way: 111aabb

Question remains how to avoid hardcoding project lists. I guess through an attribute. And if I need an attribute anyway, I can store the replacement pattern in it right away

coolo:
This runs now for sp1, sp4, leap and factory - and to me the only problem remaining is catching disconnects (which happen to me very seldomly, but it's still a problem)

jberry-suse:
> which happen to me very seldom, but it's still a problem

Interesting, so it does happen. Eventually probably worth trying to figure out why they happen in general (perhaps server issue).

Since you indicated waiting until OBS absorbs rest of staging API assume that postpones this?

coolo:
I had a Server Shutdown this morning - I assume maintenance window. And I had a connection closed on o3 the day before. That's 3 disconnects in 3 days - which is worrysome enough.

I have now the async consumer example modified (ae7f35c), we should make a generic listener out of it. Because we will have most likely more services that glue one service with another.

And about the OBS parts: the actual data in OBS is currently unused and will be for a while, but I want to have this running stable and sort out problems while OBS team hacks the staging workflow. The actual API will most likely be a sprint further down the lane.

coolo:
(in case you wonder how I make 3 out of that - the maintennace window affected both instances)

coolo:
I had both consumer.py and rabbit-openqa.py running over night and one survived, one didn't. But the consumer.py leaked FDs, but I take that as a pika bug.

coolo:
This now works to my personal liking

coolo:
So this came a little unexpected:

```
ERROR:pika.adapters.base_connection:Read empty data, calling disconnect
INFO:pika.connection:Disconnected from RabbitMQ at rabbit.opensuse.org:5671 (-1): EOF
WARNING:__main__:Channel 1 was closed: (-1) EOF
WARNING:pika.connection:Suppressing close request on <SelectConnection CLOSED socket=None params=<URLParameters host=rabbit.opensuse.org port=5671 virtual_host=/ ssl=True>>
WARNING:__main__:Connection closed, reopening in 5 seconds: (-1) EOF
INFO:__main__:Connecting to amqps://opensuse:opensuse@rabbit.opensuse.org
INFO:pika.adapters.base_connection:Connecting to 2620:113:80c0:8::19:5671 with SSL
INFO:__main__:Connection opened   
INFO:__main__:Binding opensuse.openqa.job.restart to amq.gen-beK-yG-taiRWYpr1eopIVw
INFO:__main__:Binding opensuse.openqa.job.create to amq.gen-beK-yG-taiRWYpr1eopIVw
INFO:__main__:Binding opensuse.openqa.job.done to amq.gen-beK-yG-taiRWYpr1eopIVw
INFO:__main__:Binding opensuse.obs.repo.published to amq.gen-beK-yG-taiRWYpr1eopIVw
INFO:__main__:Fetching ISOs of openSUSE:Factory
ERROR:pika.callback:Calling <bound method Listener.on_bindok of <__main__.Listener object at 0x7fb3c29aa210>> for "1:Queue.BindOk" failed
  File "/usr/lib/python2.7/site-packages/pika/callback.py", line 236, in process
    callback(*args, **keywords)
  File "/space/prod/openSUSE-release-tools/PubSubConsumer.py", line 274, in on_bindok
    self.start_consuming()
  File "./rabbit-openqa.py", line 180, in start_consuming
    project.init()
  File "./rabbit-openqa.py", line 35, in init
    for p in self.api.get_staging_projects():
  File "/space/prod/openSUSE-release-tools/osclib/stagingapi.py", line 336, in get_staging_projects
    return project_list_prefix(self.apiurl, self.cstaging + ':')
  File "/space/prod/openSUSE-release-tools/osclib/core.py", line 213, in project_list_prefix
    root = ETL.parse(http_GET(url)).getroot()
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 3360, in http_GET
    def http_GET(*args, **kwargs):    return http_request('GET', *args, **kwargs)
  File "/space/prod/openSUSE-release-tools/osclib/cache.py", line 39, in http_request
    ret = osc.core._http_request(method, url, headers, data, file)
  File "/usr/lib/python2.7/site-packages/osc/core.py", line 3349, in http_request
    fd = urlopen(req, data=data)
  File "/usr/lib64/python2.7/urllib2.py", line 154, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 429, in open
    response = self._open(req, data)
  File "/usr/lib64/python2.7/urllib2.py", line 447, in _open
    '_open', req)
  File "/usr/lib64/python2.7/urllib2.py", line 407, in _call_chain
    result = func(*args)
  File "/usr/lib/python2.7/site-packages/osc/oscssl.py", line 221, in https_open
    raise M2Crypto.m2urllib2.URLError(err)
URLError: <urlopen error [Errno 111] Connection refused: 'https://api.opensuse.org/search/project/id?match=starts-with%28%40name%2C+%22openSUSE%3AFactory%3AStaging%3A%22%29'>

```

So I guess in case of OBS exceptions we need to handle the same reset as with PIKA exceptions. Or we generally die and let systemd restart the service.

coolo:
It survived the night now - with 3 disconnects from OBS and rabbit. So ready for deployment

PR #1731, opened 2018-10-15
metrics: Fetch openqa jobs in influxdb format

coolo:


coolo:
building the packages is broken, but this is not the problem of this PR :(

jberry-suse:
And preferably name the dashboard similar to others like `OSRT: openQA`: `openqa.json`.

coolo:
I pushed a .json with the changes you mention blindly applied.

jberry-suse:
Since I resolved the CI failure in #1734 the package was updated and I deployed it to metrics.o.o. After restarted `osrt-metrics-telegraf` service the [dashboard](https://metrics.opensuse.org/d/osrt_openqa/osrt-openqa) is populated with initial data.

jberry-suse:
Something that might be beneficial would be to configure the fill mode in grafana to use previous so when zooming in one doesn't get lots of N to 0 jumps over and over.

coolo:
I tried a lot of things with this - and none work :(

previous doesn't work as we don't write real zeros, we stick with ones (or more) and create fake data.

coolo:
If you have other ideas, let me know - but to me it looks like I need to output *every* worker and *every* job group *every* time to be able to use other fill modes.

jberry-suse:
Yeah, if not always outputting previous will not work so well. For access metrics the code aggregating log data keeps track of what it has previous written and ensures to write zeros if not in current data.

coolo:
can't have that with polling routes

jberry-suse:
Wasn't suggesting that.

PR #1738, opened 2018-10-18
osc-staging: select: replace --from with --filter-from.

### LINKED ISSUE #1708: select --from tries to select packages from a different staging as well
With `xdg-desktop-portal-kde` in an adi staging, `osc staging select --from D K xdg-desktop-portal-kde` tries to move it as well:
`(1/1) Moving "636539" from "openSUSE:Factory:Staging:D" to "openSUSE:Factory:Staging:K"`

It doesn't actually move the request though AFAICT.

jberry-suse:
- 41dc359580d3def3dcac84d2499a028ae62d8e9e:
    osc-staging: select: document --move and --filter-from options.

- ef9ff661bb7b08f1a5743a174cc33e53e2619d0a:
    osc-staging: select: replace --from with --filter-from.
    
    --from no longer serves a purpose since requests should never live in
    more than one staging (apparent from linking).

Since no one indicated a reason for `--from` in the issue. @Vogtinator please verify this works as you expect and does not break other workflows as you are doing more staging than I at the moment.

Fixes #1708.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -285,6 +285,25 @@ def do_staging(self, subcmd, opts, *args):
         These concepts can be combined and interactive mode allows the proposal
         to be modified before it is executed.
 
+        Moving requests can be accomplished using the --move flag. For example,
+        to move already staged pac1 and pac2 to staging B use the following.
+
+        select --move B pac1 pac2
+
+        The staging in which the requests are staged will automatically be
+        determined and the requests will be removed from that staging and placed
+        in the specified staging.
+
+        Related to this, the --filter-from option may be used in conjunction
+        with --move to only move requests already staged in a specific staging.
+        This can be useful if a staging master is responsible for a specific set
+        of packages and wants to move them into a different staging when they
+        were already placed in a mixed staging. For example, if one had a file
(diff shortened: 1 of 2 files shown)


Vogtinator:
Currently I don't have anything to try this out with, but looking at the code it does exactly what I needed. Thanks!

coolo:
Well, we obviously have no QA team for this. So shit happens

jberry-suse:
I am simply suggested we don't merge stuff when I say that. Perhaps I should put `[WIP]` on it or something else meaning someone should run it through it's paces.

coolo:
I'm not a native speaker, but I'm curious: with what phrase did you say it shouldn't be merged? You asked Fabian to verify if it works as expected for him - which in my (non native) ears sounds like a *2nd* line of test.

jberry-suse:
That was it, and since he hadn't had an opportunity I intended to wait since now major change. Next time I'll use the `WIP` to make it clear.

DimStar77:
Caused a regression - see https://github.com/openSUSE/openSUSE-release-tools/issues/1750

PR #1739, opened 2018-10-19
metrics/grafana/openqa: set minimum interval of 10s to avoid gaps.

jberry-suse:
Since data is collected at the same interval it avoids grouping at smaller
intervals when zoomed in which reveals gaps between the data points.

@coolo Resolves the small issue noted in #1731.

Goes from:

image

to:

image

like magic.

I have used this in other dashboards, just had to remember.

### DIFF
--- a/metrics/grafana/openqa.json
+++ b/metrics/grafana/openqa.json
@@ -62,6 +62,7 @@
         "y": 0
       },
       "id": 2,
+      "interval": "10s",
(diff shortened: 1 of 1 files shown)


jberry-suse:
In edit mode you find the option on the right side of data source.

image

coolo:
cool, will add this to my suse dashboard too

PR #1744, opened 2018-10-24
[WIP] Add auth for influx to telegraf config

ChrisBr:
We need to open directly connect to the influxdb from OBS, therefore we need to enable authentication. @jberry-suse not sure how you want to store / handle the PW, this is just a starting point for a discussion. We probably also need to add the auth to the metrics.py scripts, however, I didn't look into detail into it so I would need some pointers from your side. 

As said, this should just be a starting point for a discussion.

cc @hennevogel

### DIFF
--- a/metrics/telegraf/agent.conf
+++ b/metrics/telegraf/agent.conf
@@ -4,3 +4,5 @@
 [[outputs.influxdb]]
  urls = ["http://localhost:8086"]
  database = "osrt_telegraf"
+ username = "influx"
+ password = "nottherealpassword"


jberry-suse:
Alternatively, I believe influxdb can handle host restrictions so could allow local connects while requiring auth on remote?

jberry-suse:
Hmm, not seeing that in the documentation. Otherwise, a config file for the osrt-metrics users (similar to `~/.oscrc`) would seem best. Then adjust the tools to use those credentials. For the OSRT telegraf agent config I would assume we can extend it with a local file added to the directory which extends agent config to include credentials. Alternatively could setup an additional directory to search in service file which can then be used to add credentials.

jberry-suse:
Curious, telegraf has a rabbitmq input is there no way to achieve what you want without direct access to influxdb?

ChrisBr:
Sorry for the late response, still recovering from the onsite...

> Curious, telegraf has a rabbitmq input is there no way to achieve what you want without direct access to influxdb?

Yes, we already do that for manually triggering events. However, we want to use tools which directly put data into the influxdb otherwise we would need to rewrite / hack these tools to go via rabbitmq. One example is https://github.com/influxdata/influxdb-rails

ChrisBr:
> For the OSRT telegraf agent config I would assume we can extend it with a local file added to the directory which extends agent config to include credentials. Alternatively could setup an additional directory to search in service file which can then be used to add credentials.

Yo, I think it should work if we place another config with the credentials to 
```/usr/share/openSUSE-release-tools/metrics/telegraf/``` (configuration dir). Would that work for you? Then we can drop the changes in ``agent.conf`` proposed in this PR.

> Otherwise, a config file for the osrt-metrics users (similar to ~/.oscrc) would seem best. Then adjust the tools to use those credentials.

Ok, which tools need to get adapted then? Quick search in this repository resulted in:

* https://github.com/openSUSE/openSUSE-release-tools/blob/3ad87ff320a8fd4927042d7e9a13880fa7f6e6bd/metrics.py
* https://github.com/openSUSE/openSUSE-release-tools/blob/38e2d3fa1c5cb72c91686aa91e70d89650fee0f6/metrics/access/aggregate.php

Is this all or do I miss something?

Thanks for your help.

**Comment by jberry-suse** [comment]
> Yo, I think it should work if we place another config with the credentials to
> `/usr/share/openSUSE-release-tools/metrics/telegraf/` (configuration dir). Would that work for you? Then we can drop the changes in `agent.conf` proposed in this PR.

That's what I was suggesting so yes.

> Is this all or do I miss something?

That should be all that writes to influxdb, but the [grafana data source setup](https://github.com/openSUSE/openSUSE-release-tools/blob/master/metrics/grafana/provisioning/datasources.yaml) needs read access which does not appear to have a simple clean solution.

**Comment by jberry-suse** [comment]
In order to keep to automatic, about the best I could imagine is on package update/install run script that looks for credentials file and if found updates yaml to include credentials.

**Comment by ChrisBr** [comment]
> In order to keep to automatic, about the best I could imagine is on package update/install run script that looks for credentials file and if found updates yaml to include credentials.

Ok, I will look into it.

**Comment by ChrisBr** [comment]
So looked more into it:

The ``aggregate.php`` needs to get updated here:
https://github.com/openSUSE/openSUSE-release-tools/blob/38e2d3fa1c5cb72c91686aa91e70d89650fee0f6/metrics/access/aggregate.php#L410
to something like:
```
$database = InfluxDB\Client::fromDSN(sprintf('influxdb://user:pass@%s:%s/%s', $host, $port, $dbname));
``
I'm not really into php so wondering what would be the best method to do that? Reading a file with the credentials with every request probably not? Any better suggestion?

Looking into

https://github.com/openSUSE/openSUSE-release-tools/blob/3ad87ff320a8fd4927042d7e9a13880fa7f6e6bd/metrics.py#L532

it already uses username and password supplied by args. I'm just wondering where this script gets called? I searched through the repository / systemd files but did not find anything? Any hint?

I'm also wondering if maybe an easy solution would be to just run another influxdb server on the same host (probably in a docker container like the telegraf) which is open to the internet and we just use this instead of adapting all the tools? What do you think?

jberry-suse:
> I'm just wondering where this script gets called? I searched through the repository / systemd files but did not find anything? Any hint?

https://github.com/openSUSE/openSUSE-release-tools/blob/master/systemd/osrt-metrics%40.service

But, really should make both it and PHP read a config file rather than putting user/pass in a systemd file (although systemd could read env file and pass that way). Env file is probably fine since standarized format that would work for both instances, but rather it be optional as otherwise if included in systemd service file the env file must always be present.

> I'm also wondering if maybe an easy solution would be to just run another influxdb server on the same host (probably in a docker container like the telegraf)

I suppose. We already have two instances of telegraf, one on host (used started by OSRT) and the one you run in docker. Not sure if you needed a different version or why host version was not sufficient. Two instances is likely still the best approach for telegraf since configs are different.

A separate influxdb instance is most definitely simpler. Assume you would move the existing telegraf database used by OBS to the new instance as well?

ChrisBr:
Closing, we run now a dedicated influxdb on the server in dir ``/var/lib/influxdb-rails``. The port is not yet public available but I created a ticket for Heroes.

cc @hennevogel

jberry-suse:
Good to know. I look forward to the additional metrics and potential insight.

PR #1745, opened 2018-10-24
osc-staging: handle None value for --filter-from.

jberry-suse:
Fixes #1738.

I'll move some adi packages around if no one tests this before me.


jberry-suse:
Verified this seems to resolve the issue at hand.

jberry-suse:
CI failures are all flake8 related which is resolved in #1747.

PR #1747, opened 2018-10-24
CI: ignore W504 and resolve other new flake8 detected issues

### LINKED ISSUE #1746: CI: flake8 update from 3.5.0 to 3.6.0 introduced new checks that fail
Compare:

- [last merge run](https://travis-ci.org/openSUSE/openSUSE-release-tools/jobs/444981387)
- [fresh merge request](https://travis-ci.org/openSUSE/openSUSE-release-tools/jobs/445816353)

jberry-suse:
- 0449d1f5a1615228f06ed9174cda00a6581ff4da:
    manager_42: remove extra newline to avoid flake8 E301.

- 4c8ade1a93dc7d028e9e7d6793272b7716b13a57:
    flake8: utilize raw strings to avoid invalid escape character warnings.

- e68e348781b7c9a7e44482ce50dd60c62255f002:
    flake8: ignore W504.

Fixes #1746.


metrics.py
```diff
@@ -331,7 +331,7 @@ def ingest_release_schedule(project):
 
         # Extract Factory "release schedule" from Tumbleweed snapshot list.
         command = 'rsync rsync.opensuse.org::opensuse-full/opensuse/tumbleweed/iso/Changes.* | ' \
-            'grep -oP "Changes\.\K\d{5,}"'
+            'grep -oP "' + r'Changes\.\K\d{5,}' + '"'
```
coolo: this isn't really a python regexp, so why mark it?

jberry-suse: https://docs.python.org/2.0/ref/strings.html

```
Unlike Standard C, all unrecognized escape sequences are left in the string unchanged, i.e., the backslash is left in the string.
```

Which is why it works, but flake just warns about it. We can choose to ignore those as well, but seems like a decent best practice.

coolo: so r is for raw not for regexp? Wow :)

jberry-suse: Yeah, the way it is used is basically for regex. hehe :)

PR #1754, opened 2018-10-30
[ARM] Skip build number comparison

dirkmueller:
The LiveCD products (JeOS) for ARM will never follow
the same iso build numbers (and don't produce iso's to begin with)
than the main project. In lack of a "ignore mismatch for livecd
products only" option we just skip the syncing globally. that should
be fine as hopefully the main products always produce the same build.


jberry-suse:
@nilxam is on vacation.

PR #1757, opened 2018-10-31
Don't report that a package is ok if the resultlist is empty

Vogtinator:


totest-manager.py
```diff
@@ -399,13 +399,19 @@ def package_ok(self, project, package, repository, arch):
         url = self.api.makeurl(['build', project, '_result'], query)
         f = self.api.retried_GET(url)
         root = ET.parse(f).getroot()
+        succeeded = False
```
jberry-suse: Perhaps just utilize xpath (or in this case even just `ET.find()`) instead of the loop and variable. Something like the following:

https://github.com/openSUSE/openSUSE-release-tools/blob/0d9cde5a9089f11e9747ffecfb5b1c034a248388/osclib/core.py#L429-L431

```suggestion
        succeeded = len(root.findall('result')) and not len(root.findall('result[@status!="succeeded"]))
```

totest-manager.py
```diff
@@ -399,12 +399,18 @@ def package_ok(self, project, package, repository, arch):
...
         f = self.api.retried_GET(url)
         root = ET.parse(f).getroot()
-        for repo in root.findall('result'):
-            status = repo.find('status')
-            if status.get('code') != 'succeeded':
-                logger.info(
-                    '%s %s %s %s -> %s' % (project, package, repository, arch, status.get('code')))
-                return False
+        succeeded = root.findall('result/status[@code="succeeded"]')
+        # [@code!='succeeded'] is not supported by ET
```
jberry-suse: If you wanted the `!=` support you can use `lxml` as done all over the code now.

https://github.com/openSUSE/openSUSE-release-tools/blob/33a4a3b3484e536062a1e75d0398a99f8c8b7c91/osclib/core.py#L7

Otherwise, fine by me.

Vogtinator: For now this should be enough - this is supposed to be a simple fix only.

jberry-suse: This makes it simpler, not more complex, but feel free to merge.

Vogtinator: > feel free to merge.

I don't have the permissions to do that myself.

coolo: You have now

PR #1762, opened 2018-11-03
userscript/staging-move-drag-n-drop: provide initial version.

jberry-suse:
Since it has been mentioned a few times before as very desirable I took a stab at it for a couple hours. I would wager the result is well worth the effort.

Could follow this up with a server (hosted locally or remote with a bit more thought) to actually run the commands. In fact depending on how it is desired the remote server could be configured to allow for community members to be given selective move writes where the actually commands are run under `staging-bot` user. Essentially giving them limited staging access. I would expect the server to be simple to build.

For the initial version it just collects the moves and generates commands.

image

Activate via `ctrl+m` (for move) as indicated in legend once installed. Enjoy!

### DIFF
--- a/userscript/README.md
+++ b/userscript/README.md
@@ -1,0 +1,5 @@
+# User Scripts
+
+The scripts may be installed in one's browser (using Greasemonkey or Tampermonkey) to provide additional features using OBS via the web.
+
+- [Staging Move Drag-n-Drop](https://github.com/openSUSE/openSUSE-release-tools/raw/master/userscript/staging-move-drag-n-drop.user.js)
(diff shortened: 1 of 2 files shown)


jberry-suse:
Could also make project portion toggleable as either Factory or one using a directory with `.osc/project` configured does not need that, but figured better safe than sorry for now.

jberry-suse:
@lnussel @nilxam @DimStar77 ;)

jberry-suse:
Other user scripts perhaps worth considering: https://github.com/jberry-suse/opensuse-staging-tools.

jberry-suse:
I can also see a quick staging filter to only show the few staging one wants to work on to make drag-n-drop distance shorter.

jberry-suse:
Be sure to try click and drag a box around requests too, just don't have to select directly. Plus hold ctrl while doing so works as expected.

image

coolo:
I don't think users mentioning drag & drop in staging projects meant it to create osc command lines, but to have obs move the requests right away - under the user dragging. 

Because while this is a great technology showcase and usability research, this would be staging dashboard engine story reloaded.

PR #1763, opened 2018-11-03
Fail rabbit-openqa and have systemd restart it

coolo:
There were still exceptions not caught - and I have the suspicion
that the memory leaks we see in production are caused by the reconnects
(if you google for python memory leaks you end up with pika and lxml
examples - both we use here, so a restart every couple of hours can't
hurt)


coolo:
Also prefixed the check name

PR #1766, opened 2018-11-05
Port more code to run under python3

coolo:
(python2 compat kept)

### DIFF
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -11,10 +11,10 @@ import sys
 try:
     from urllib.parse import unquote
     from urllib.parse import urlsplit, SplitResult
-    from urllib.request import URLError, HTTPError
+    from urllib.error import URLError, HTTPError
     from io import StringIO
 except ImportError:
-    #python 2.x
+    # python 2.x
     from urlparse import urlsplit, SplitResult
     from urllib import unquote
     from urllib2 import URLError, HTTPError
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -7,9 +7,19 @@ import osc.core
 import re
 import shutil
 import sys
-import urlparse
-import urllib
-from StringIO import StringIO
+
+try:
+    from urllib.parse import unquote
+    from urllib.parse import urlsplit, SplitResult
+    from urllib.request import URLError, HTTPError
+    from io import StringIO
+except ImportError:
+    #python 2.x
+    from urlparse import urlsplit, SplitResult
+    from urllib import unquote
+    from urllib2 import URLError, HTTPError
+    from StringIO import StringIO
+
 from osc import conf
 from osc.core import urlopen
 from osclib.cache_manager import CacheManager
@@ -122,7 +132,7 @@ class Cache(object):
 
     @staticmethod
     def get(url):
-        url = urllib.unquote(url)
+        url = unquote(url)
         match, project = Cache.match(url)
         if match:
             path = Cache.path(url, project, include_file=True)
@@ -176,7 +186,7 @@ class Cache(object):
 
     @staticmethod
     def put(url, data):
-        url = urllib.unquote(url)
+        url = unquote(url)
         match, project = Cache.match(url)
         if match:
             path = Cache.path(url, project, include_file=True, makedirs=True)
@@ -189,6 +199,7 @@ class Cache(object):
             # be replaced with urlopen('file://...') to be consistent, but until
             # the need arrises StringIO has less overhead.
             text = data.read()
+            text.decode('utf-8')
             data = StringIO(text)
 
             if conf.config['debug']: print('CACHE_PUT', url, project, file=sys.stderr)
@@ -200,7 +211,7 @@ class Cache(object):
 
     @staticmethod
     def delete(url):
-        url = urllib.unquote(url)
+        url = unquote(url)
         match, project = Cache.match(url)
         if match:
             path = Cache.path(url, project, include_file=True)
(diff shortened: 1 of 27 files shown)


coolo:
I'm stuck here:

```
  File "/space/prod/osc/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/space/prod/osc/osc/cmdln.py", line 501, in onecmd
  File "/space/prod/osc/osc/cmdln.py", line 1232, in _dispatch_cmd
  File "/suse/coolo/.osc-plugins/osc-staging.py", line 418, in do_staging
    Config(opts.apiurl, opts.project)
  File "/suse/coolo/.osc-plugins/osclib/conf.py", line 179, in __init__
    self.remote_values = self.fetch_remote(apiurl)
  File "/suse/coolo/.osc-plugins/osclib/conf.py", line 251, in fetch_remote
    cp.readfp(io.BytesIO(config))
TypeError: a bytes-like object is required, not 'str'

**Comment by aplanas** [comment]
@coolo try this:

```python
cp.readfp(io.BytesIO(config.encode()))
```

coolo:
moves the problem only a little bit :(
```
  File "/suse/coolo/.osc-plugins/osclib/conf.py", line 251, in fetch_remote
    cp.readfp(io.BytesIO(config.encode()))
  File "/usr/lib64/python3.6/configparser.py", line 763, in readfp
    self.read_file(fp, source=filename)
  File "/usr/lib64/python3.6/configparser.py", line 718, in read_file
    self._read(f, source)
  File "/space/prod/osc/osc/OscConfigParser.py", line 242, in _read
    if line.strip() == '' or line[0] in '#;':
TypeError: 'in <string>' requires string as left operand, not int

**Comment by aplanas** [comment]
@coolo right. In this case I would use StringIO to see how far we can go:

```python
# If you do not have StringIO
try:
    from StringIO import StringIO
except ImportError:
    from io import StringIO

...

            config = '[remote]\n' + config
            cp.readfp(StringIO(config))
```

coolo:
And I broke the tests with this - which is not good

coolo:
found the problem with the tests.

coolo:
So I can run rabbit-openqa under python3 now

coolo:
More specifically: rabbit-openqa runs on python3 and python2 (and the staging plugins don't interfer with osc in both versions).

osclib/conf.py
```diff
@@ -206,11 +206,11 @@ def populate_conf(self):
                 for k, v in DEFAULT[prj_pattern].items():
                     if k.startswith('_'):
                         continue
-                    if isinstance(v, basestring) and '%(project)s' in v:
```
coolo: what can v be if not string/basestring? @jberry-suse - remember what this is about?

jberry-suse: It should always be, but I think this may have had something to do with integer defaults and such. Really not sure why it is needed.

coolo: @aplanas introduced it in 
c9d1e06ca80f231a87703ec24cbe5c6dd9386ed7 - and the only other type involved there is None.   

The main question is: is `str` good enough or do we have to expect unicode. I don't think so.

aplanas: As comes from DEFAULT, it will be a string if the user do not use an explicit `u''`, that is why it can be safe.

If someday (unlikely) the user change one string to be unicode, the comparison will fail and `default[k]` will never be set, so will fail silently. And the user will have a fun day debugging it.

A dirty hack, write this somewere at the beginning this:

```
try:
    basestring
except NameError:
    basestring = str
```

and maintain the basestring comparison.

I used a variation of this here:

https://github.com/mozilla/python-spidermonkey/pull/12/files#diff-c95ffd29054e4e197c392068293e5213R8

jberry-suse: What is wrong with just having it crash if the DEFAULT is changed in that way? We are trying to maintain compatibility with osc config which means all the defaults must be strings. This is what I figured it was for, but not sure why we want to support that at all.

aplanas: @jberry-suse I am OK of dropping basestring, but you cannot make it fail if DEFAULT uses unicode.

The irony is that u'' in Python 2 is a string in Python 3, so u'' in Python 3 is a no-op. Basically dropping basestring will force to use bytes in Python 2 and unicode in Python 3, both are called string in each language. This is what can present problems.

In any case drop it and go pull Python 3 once osc is ported, and use this half backed solution meanwhile.

jberry-suse:
Triggered rebuild to see how far the port branch jobs make it on travis for curiosity.

coolo:
You can just push the master commit into my branch

osclib/core.py
```diff
@@ -316,7 +316,7 @@ def attribute_value_load(apiurl, project, name, namespace='OSRT'):
     if not len(value):
         return None
 
-    return value[0]
+    return str(value[0])
```
jberry-suse: Still like this will be almost everywhere (anything returned from API), is there a better general solution for this?

osclib/cache.py
```diff
@@ -273,7 +285,7 @@ def path(url, project, include_file=False, makedirs=False):
             os.makedirs(directory)
 
         if include_file:
-            parts.append(hashlib.sha1(url).hexdigest())
+            parts.append(hashlib.sha1(url.encode('utf-8')).hexdigest())
```
jberry-suse: Something is wrong with caching here, if I create caching without this PR and then switch to this is utilizes it properly, but if I wipe the request cache and then allow it to re-create it fails to find cache entries repeatedly.

jberry-suse:
I'd need to look at how/why test pass, but assume it's because the API call returns the u'' string and the mocking test loads a normal string. Another reason why mocking tests are garbage.

coolo:
Using BytesIO fixes this (4dec2234be0efc5fb5fb7d5604bd5cc75f17d123)

coolo:
this breaks python3 again - but for now I prefer a working python2. looks like we need test cases first that check this part before porting it for real.

coolo:
rebased to revert all the bytesio changes in conf.py (the one in cache.py should be good for both).

So the goal for now is just preparations - and being able to run the python3 branch of osc with python2.

coolo:
this is basically down to changes I consider harmless, so I'll merge to give others a chance to base changes on it

PR #1767, opened 2018-11-06
Support OBS checks instead of openqa

coolo:


coolo:
fixing the test cases is a little above my head, sorry

coolo:
I looked into removing stagings that were not adding coverage - but it's just a little too much work for a little too little gain. 1MB json is not the end of the world

coolo:
filtered now using:
cat openSUSE:Factory.json | jq "[ .[] | select(.name==\"openSUSE:Factory:Staging:$i\" | not) ] " > t && mv t openSUSE:Factory.json

okurz:
what does the " | not" do?

coolo:
the same as != in other languages :)

jberry-suse:
> 1MB json is not the end of the world

It's bigger than all the code combined. All that way needed was copying the checks arrays into the JSON where the prior checks had been. Should be a 40 line diff and no need to change the unselect cleanup test. This is the problem with mock style testing, leads to laziness and giant dumps of data.

Using the real OBS available could just write the status checks as the real code does and then read them back.

coolo:
If OBS goes ahead and drops the JSON api, we possibly can. But creating fake data in OBS to have all the cases covered is a lot of work and very time consuming, so I don't think its worth it.

So I dismiss your review

jberry-suse:
s/personal oppinion/technical taste/

What changed? A single nested key and its underlying data.
What needed to be changed in test data? The single nested key and its underlying data.
What was changed? The entire JSON data structure was copy/pasted from Factory (the largest project no less). Which changed all sorts of unrelated values and required updating unrelated tests in addition to complete output change of the test in question.
What could have been done? Copy/paste or manually update the couple array structures that changed.

A low standard as the lazy method required changed all sorts of unrelated stuff and produced a very non-concise commit. Just messy.

The alternative approach has a large relative cost (since the existing one has already been paid for), but if far better for maintenance and actually testing something real.

PR #1769, opened 2018-11-06
check_source: check if member of staging group before attempting to skip

### LINKED ISSUE #1483: check_source: SLE can't skip staging review
Factory-auto user has no permissions to skip staging reviews, so it's crashing if trying to do so. Not sure why it can for Leap actually :)

jberry-suse:
SLE likes to be different for the sake of it.

Fixes #1483.


jberry-suse:
Will still skip repo-checker review which is fine.

Issue #1770, opened 2018-11-07
CI: tests fail for the last three days, but passed pre-merge


jberry-suse:
As normally happens, tests failed, the person who merged got an e-mail from travis and nothing was done. :( I noticed when I went to merge one of my pull requests and so I am filing an issue.

PR #1755 
- [passed pre-merge](https://travis-ci.org/openSUSE/openSUSE-release-tools/builds/448219411)
- [failed post-merge](https://travis-ci.org/openSUSE/openSUSE-release-tools/builds/450752654)

Would seem to point in the direction of upstream.

coolo:
httppretty was updated that day: https://github.com/openSUSE/openSUSE-release-tools/pull/1771 works around it.

lnussel:
huh? #1755 has no indication at all that anything went wrong with travis.

coolo:
@jberry-suse is saying you should have got an email after merge

jberry-suse:
Makes sense as it did not appear to be the fault of any changes in this repo.

jberry-suse:
Created config change to notify list so those interested can be notified in the future.

PR #1773, opened 2018-11-08
travis: configure notification e-mail to internal list.

jberry-suse:
I setup this internal list a few weeks ago, but haven't used it for anything yet. The idea being for more spamming topics really only relevant to us in addition to potentially using for third-party account e-mail address so everyone on team has access instead of bus factor 1.

Related #1770.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,3 +94,8 @@ cache:
 before_cache:
   - sudo rm -rf .docker-tmp/build-root
   - sudo chmod -R o+rw .docker-tmp
+
+notifications:
+  email:
+    recipients:
+      - opensuse-releaseteam@suse.de


jberry-suse:
Indeed. Since travis is not configured for forks short of a fork owner setting it up the notifications should only be sent on those in this repository. My assumption is PRs never get e-mails since they not not with the default configuration which is said to send on failure and does not on failed PRs.

Obviously, if someone configured travis on their fork and started pushing bad commits we would get e-mails, but that's a lot of work just to spam us. Overall, just looking for a better way to keep up on CI failures so that they are resolved in a timely fashion..

jberry-suse:
Will revert if becomes a problem.

PR #1774, opened 2018-11-08
No longer build disable non-ring packages on select

coolo:


coolo:
As discussed with @lnussel and @DimStar77 today

osclib/stagingapi.py
```diff
@@ -836,12 +836,11 @@ def rm_from_prj(self, project, package=None, request_id=None,
 
         orig_project = project
         self._remove_package_from_prj_pseudometa(project, package)
-        project = self.map_ring_package_to_subject(project, package)
         if self._supersede:
             self.is_package_disabled(project, package, store=True)
 
         for sub_prj, sub_pkg in self.get_sub_packages(package, project):
-            sub_prj = self.map_ring_package_to_subject(project, sub_pkg)
```
coolo: note that get_sub_packages always returns project as first element now - but simplifying that was a little too dangerous for me atm :)

PR #1777, opened 2018-11-09
Reduce cache time of attributes

coolo:
Especially caching the lockedby attribute makes it basically
impossible to work with multiple persons

### DIFF
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -91,7 +91,7 @@ class Cache(object):
         r'/source/([^/]+)/_meta$': TTL_LONG,
         r'/source/([^/]+)/(?:[^/]+)/(?:_meta|_link)$': TTL_LONG,
         r'/source/([^/]+)/dashboard/[^/]+': TTL_LONG,
-        r'/source/([^/]+)/_attribute/[^/]+': TTL_MEDIUM,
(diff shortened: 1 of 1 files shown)


osclib/cache.py
```diff
@@ -91,7 +91,6 @@ class Cache(object):
         r'/source/([^/]+)/_meta$': TTL_LONG,
         r'/source/([^/]+)/(?:[^/]+)/(?:_meta|_link)$': TTL_LONG,
         r'/source/([^/]+)/dashboard/[^/]+': TTL_LONG,
-        r'/source/([^/]+)/_attribute/[^/]+': TTL_MEDIUM,
```
jberry-suse: Definitely not for `TTL_MEDIUM`. The trouble was introducing when trying to cache config as attribute. Likely makes more sense to cache just config using a pattern that looks for it with `OSRT:Config` or similar in pattern.

With remote config in dashboard it was cached more explicitly as can still be seen. Either way, just another API call on everything. If nothing else caching attributes with `TTL_DUPLICATE` would at least solve tools like ReviewBots that load target project config, but also different caching mechanism used now.

coolo: ok, then DUPLICATE - that's still short enough not to get in your way as stale lock

PR #1786, opened 2018-11-14
dist/kubernetes: provide OBS Operator component (and related changes)

jberry-suse:
- 6b02225c75b7c26c2e7d900ce6430d474f7ebdb5:
    dist/kubernetes: utilize new image home in openSUSE:Tools:Images.

- 2e50be7557f80c3c0f60d3efe9b53d1199e73878:
    dist/package: include worker kiwi to allow extraction via _service.
    
    Since the kiwi file will reside in the source package, it needs to be
    referenced in order to appease Factory rpmlint.

- a75f4fbabbaff83ad047973bb7eb1ce0343a7bf1:
    dist/kiwi/worker: include obs-operator subpackage.

- 6b1003fb062aabc542103ee3b91103d18f85d9e4:
    obs_operator: use unbuffered output to work properly in systemd.

- 3ac0086154c85807a56c5be9914ef4caba87bf82:
    dist/kubernetes: provide OBS Operator component.

Should accept [OBS request 649089](https://build.opensuse.org/request/show/649089) after merged to allow for kiwi file to be updated. Created [temporary branched package](https://build.opensuse.org/package/show/openSUSE:Tools:Images/openSUSE-release-tools) in openSUSE:Tools:Images to verify everything works. The package should be re-branched from openSUSE:Tools once merged.

WIP until verify k8s approach with @tampakrap.

### DIFF
--- a/dist/kiwi/osrt-worker-obs.kiwi
+++ b/dist/kiwi/osrt-worker-obs.kiwi
@@ -21,6 +21,7 @@
   <packages type="image">
     <package name="openSUSE-release-tools-check-source"/>
     <package name="openSUSE-release-tools-leaper"/>
+    <package name="openSUSE-release-tools-obs-operator"/>
     <package name="openSUSE-release-tools-pkglistgen"/>
     <package name="openSUSE-release-tools-repo-checker"/>
     <package name="openSUSE-release-tools-staging-bot"/>
--- a/dist/kubernetes/app.yaml
+++ b/dist/kubernetes/app.yaml
@@ -7,9 +7,9 @@ environments:
     k8sVersion: v1.8.0
     path: heroes
     targets:
-    - check-source
+    #- check-source
     - obs-operator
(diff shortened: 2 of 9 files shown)


jberry-suse:
Currently deployed and working via this setup, except for final image location which needs to have this merged.

coolo:
So the setup is fixed again - there is nothing to change in k8s though, the image name on registry is the same.

jberry-suse:
The reason I avoided that approach is now kiwi file changes have to be deployed manually. Ideally we could just build the image in the same project as package. The whole setup is rather cumbersome, but that seems to be the rule not the exception.

coolo:
why 'kiwi file changes have to be deployed manually'? What is manual about it?

jberry-suse:
The new package is not updated on merge. Hence changes to kiwi require updating the package manually. Can add that to CI as well.

PR #1787, opened 2018-11-15
travis: add test jobs against osc python3 branch for both python 2 and 3.

jberry-suse:
Allows for incremental work towards support while being able to see tests results and not have them mark every build as failing. Had to drop urlgrabber from python 3 install and adjust requirements.txt in travis since the real file should still point at `osc@master`.


jberry-suse:
Once merged #1766 can be re-triggered to compare test results.

PR #1788, opened 2018-11-16
fetch string return value '

lethliel:
can be a string containing 'none'.

This will be fetched with isinstance(data, str)


lethliel:
opened to soon

PR #1789, opened 2018-11-16
fetch string value 'none' of data

lethliel:
In some cases yaml.load(description_text) can return a string value 'none'. 

This will be fetched with isinstance(data, str)

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -698,7 +698,7 @@ class StagingAPI(object):
     def load_prj_pseudometa(self, description_text):
         try:
             data = yaml.load(description_text)
-            if data is None:
+            if isinstance(data, str) or data is None:
                 data = {}
         except (TypeError, AttributeError):
             data = {}


coolo:
Background: this was on freshly created staging prjs

osclib/stagingapi.py
```diff
@@ -698,7 +698,7 @@ def get_prj_meta(self, project, revision=None):
     def load_prj_pseudometa(self, description_text):
         try:
             data = yaml.load(description_text)
-            if data is None:
+            if isinstance(data, str) or data is None:
```
aplanas: I am not sure that is correct, as if I understand correctly data will be the string 'none', that is also a string.

lethliel: The problem is that `data.get('requests', [])` does not work with strings. 
So we have to initialize `data = {}` if data is not present or a string (which it should not be either)

coolo: I wonder where the 'none' is coming from though

lethliel: > I am not sure that is correct, as if I understand correctly data will be the string 'none', that is also a string.
> 
> Edit:
> Maybe is more secure this? `if not data or data == 'none':`, can this be OK?

It will fail with all strings. If description_text contains no valid yaml format data will always be string. 
If someone initializes a staging project with the description "This is my awesome staging project" data will be that string. 

Therefore I think the staging projects were created with the description 'none'

aplanas: You are right.

jberry-suse:
Something like this would have benefited from a comment as that is odd logic in general, but makes sense in that we expect structured yaml data only.

PR #1790, opened 2018-11-16
Checked with aplanas: using stringio is preferred

coolo:
@aplanas and me checked - and the config still works with StringIO. As unicode and string have the same hash value for all keys we pass


coolo:
with python2, right?

jberry-suse:
> with python2, right?

Yes.

PR #1793, opened 2018-11-17
Port pkglistgen (and some more)

coolo:


okurz:
Do you want to preserve full python 2 + 3 compatibility? I have good experiences with using the standard lib aliases as seen in https://github.com/okurz/openqa_review/blob/master/openqa_review/openqa_review.py. Using print from future is a good idea nevertheless

coolo:
No, the roadmap is basically:
- run with osc's python3 port
- have osc release 1.0
- require osc 1.0
- switch to python3 only

So the 2+3 compat is just temporary.

PR #1794, opened 2018-11-19
stagingapi: No longer build disable to sub packages

nilxam:
Follow up with https://github.com/openSUSE/openSUSE-release-tools/pull/1774/commits/7a3b86fed17566f0680ee4eb8ad64238a70b77b9 , a regression happens, fixes UnboundLocalError: local variable 'disable_build' referenced before assignment error.

Hot-patched on packagelists.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1213,9 +1213,7 @@ class StagingAPI(object):
             # Skip inner-project links for letter staging
             if not self.is_adi_project(project) and sub_prj == project:
                 continue
-            if self._supersede:
-                disable_build = self._package_disabled.get('/'.join([sub_prj, sub_pkg]), False)
-            self.create_package_container(sub_prj, sub_pkg, disable_build=disable_build)
+            self.create_package_container(sub_prj, sub_pkg)
 
             root = ET.Element('link', package=tar_pkg, project=project)
             url = self.makeurl(['source', sub_prj, sub_pkg, '_link'])


coolo:
> > UnboundLocalError: local variable 'disable_build' referenced before assignment error.
> 
> Which indicates the original code was never run. :'-(

And then left unnoticed for another 11 days as noone bothered to check staging-bot

jberry-suse:
> And then left unnoticed for another 11 days as noone bothered to check staging-bot

Which was updated by you after updating the machine to Leap 15. Whenever I update I monitoring things for a few hours minimum. The monitoring I setup with IT appears to also be broken post Leap 15.0 update as e-mail should have been sent out due to repeated failure.

Of course as if any of this is an excuse for never running code and deploying it to production.

DimStar77:
> And then left unnoticed for another 11 days as noone bothered to check staging-bot

I'll take that one on me - we (@coolo and I) agreed that he'd be doing the update to Leap 15 and I'd be checking the used bots - as such, if anybody is to be blamed for not noticing a bot doing the right job, it is on me in this case.

As for the time line
* Update was executed on Nov 12
* PR was submitted on Nov 19 (makes 7 days, not 11)

This was after the first weekend - with the weekend being the time the bot has most chance to actually perform work without us staging masters having grouped things around (and as not many stagings were empty in the week before, there was basically no work for the bot to be done - thus 'hiding' the fact that its not doing anything - of course I could have observed that in the logs)

As such: apologies for you guys having to fight out who neglected doing what

jberry-suse:
I appreciate you taking responsibility, but these are all secondary problems causes by not executing code that is merged and deployed to production. Not even once. This is not an isolated incident which is why it is so bothersome (has happened dozens of times at least). It is baffling everyone just acts as if it is normal (although in this context I suppose it is). Max should be annoyed he had to debug and figure this out of determining the result.

Oh well, I just need to stop looking at other PRs to avoid needing to point this out.

coolo:
But you realize the broken code only runs for adi projects? So yes, I did not run the broken code - but to claim the code was pushed untested is a little over the top.

PR #1796, opened 2018-11-19
leaper: sle tuning

lnussel:


lnussel:
not sure if that makes things better or worse, will try hot patch for now

PR #1798, opened 2018-11-20
[WIP] pkglistgen: let product converter creating data if product converter service is enabled

nilxam:
for Tumbleweed, we want to keep product converter service enabled, ie. generating data is on OBS instead of external.

### DIFF
--- a/pkglistgen.py
+++ b/pkglistgen.py
@@ -24,6 +24,7 @@ from osclib.conf import Config, str2bool
 from osclib.core import repository_path_expand
 from osclib.core import repository_arch_state
 from osclib.core import source_file_ensure
+from osclib.core import source_file_load
 from osclib.core import target_archs
 from osclib.stagingapi import StagingAPI
 from osclib.util import project_list_family
@@ -774,6 +775,11 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
             tool.architectures = ARCHITECTURES
         return tool
 
+    def _product_converter_enabled(self, api, project, product):
+        servicefile = ET.fromstring(source_file_load(api.apiurl, project, product, '_service'))
+
+        return not servicefile.findall('./service[@name="product_converter"][@mode="disabled"]')
+
     def do_list(self, subcmd, opts):
         """${cmd_name}: list all groups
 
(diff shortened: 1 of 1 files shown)


pkglistgen.py
```diff
@@ -770,6 +770,16 @@ def setup_tool(self):
             tool.architectures = ARCHITECTURES
         return tool
 
+    def _product_converter_enabled(self, api, project, product):
+        url = api.makeurl(['source', project, product, '_service'])
+        servicefile = ET.parse(http_GET(url)).getroot()
+
+        for service in servicefile.findall('./service[@name="product_converter"]'):
+            if 'mode' in service.attrib and service.attrib["mode"] == "disabled":
```
coolo: I think you can write this easier as 
```
return service.find('./service[@name="product_converter" and @mode="disabled"]') is None

File: pkglistgen.py [hunk 2]
```diff
@@ -1368,10 +1380,11 @@ def update_and_solve_target(self, api, target_project, target_config, main_repo,
         delete_products = target_config.get('pkglistgen-delete-products', '').split(' ')
         self.tool.unlink_list(product_dir, delete_products)
 
-        print('-> product service')
-        for product_file in glob.glob(os.path.join(product_dir, '*.product')):
-            print(subprocess.check_output(
-                [PRODUCT_SERVICE, product_file, product_dir, opts.project]))
+        if not product_converter_enabled:
```
coolo: as this now is the 4th level of indentation, can you please extract this as a function (not to be called for product_converter_enabled)

pkglistgen.py
nilxam: just changed it to `return not servicefile.findall('./service[@name="product_converter"][@mode="disabled"]')`

pkglistgen.py
nilxam: done

coolo:
I think I just convinced @DimStar77 he doesn't want this anymore :)

DimStar77:
_convinced_ if we can't find a way for the stagings to be based on the same group.yml - stay tuned, let's keep the PR around for now to see if we use it, or not, in the end

coolo:
just saying! :)

pkglistgen.py
```diff
@@ -1324,11 +1339,11 @@ def update_and_solve_target(self, api, target_project, target_config, main_repo,
             repos_ = self.repos
             opts_nonfree = copy.deepcopy(opts)
             opts_nonfree.project = nonfree
-            self.repos = self.tool.expand_repos(nonfree, main_repo)
+            self.tool.repos = self.tool.expand_repos(nonfree, main_repo)
             self.tool.update_repos(opts_nonfree)
 
             # Switch repo back to main target project.
-            self.repos = repos_
+            self.tool.repos = repos_
```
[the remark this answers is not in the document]
nilxam: rebased

jberry-suse:
So do we want this or is this not "want this anymore"? Planning deployment this afternoon.

DimStar77:
It says ```[WIP]``` - hence, do not merge/deploy; It's possible that in the end we will retract this PR, or not - the experiments will tell.

coolo:
this looks obsolete

PR #1799, opened 2018-11-20
pkglistgen: for nonefree repo handling should update to tool's repos rather than commandline's repos

nilxam:
Fix a bug that NonFree repo solv data has not get updated.


jberry-suse:
It looks like that was original setup to alter arguments and changed in 121cec768bb4a3894a15f4b7b24bd5f96041935c to change the values directly in the wrong place according to this change. ;)

coolo:
I don't like this whole tools vs cmdline thing in pkglistgen and would like to tear it apart - once I have my other changes for CI done.

jberry-suse:
I very much agree and said as much when I first worked on it.

PR #1802, opened 2018-11-21
Disable caching project _meta

Vogtinator:
latest_updated is broken for _meta:
https://github.com/openSUSE/open-build-service/issues/6323

### DIFF
--- a/osclib/cache.py
+++ b/osclib/cache.py
@@ -99,16 +99,17 @@ class Cache(object):
         r'/source$': TTL_LONG,
         # Sources will be expired with project, could be done on package level.
         r'/source/([^/?]+)(?:\?.*)?$': TTL_LONG,
-        # Project will be marked changed when packages are added/removed.
-        r'/source/([^/]+)/_meta$': TTL_LONG,
         r'/source/([^/]+)/(?:[^/]+)/(?:_meta|_link)$': TTL_LONG,
         r'/source/([^/]+)/dashboard/[^/]+': TTL_LONG,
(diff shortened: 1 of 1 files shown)


coolo:
We use TTL_DUPLICATE for that - aka very short caching.

Vogtinator:
Sure.

coolo:
Not sure what you mean with 'Sure.', but having disabled code is no solution.

Vogtinator:
By "Sure." I mean that I changed it and pushed the change.

osclib/cache.py
```diff
@@ -99,16 +99,17 @@ class Cache(object):
...
-        # Handles clearing local cache on package deletes. Lots of queries like
-        # updating project info, comment, and package additions.
-        r'/source/([^/]+)/(?:[^/?]+)(?:\?[^/]+)?$': TTL_LONG,
         # Presumably users are not interweaving in short windows.
         r'/statistics/latest_updated': TTL_SHORT,
+        # Use TTL_DUPLICATE for project _meta as only description changes are listed in latest_updated:
+        # https://github.com/openSUSE/open-build-service/issues/6323
+        r'/source/([^/]+)/_meta$': TTL_DUPLICATE,
+        ## Handles clearing local cache on package deletes. Lots of queries like
+        ## updating project info, comment, and package additions.
```
jberry-suse: Assume the double comment was unintentional?

Vogtinator: It was intentional - I commented out the entire block. The comment refers to the commented out code, so IMO this makes it obvious that it's not relevant to the active code anymore.

PR #1805, opened 2018-11-24
[WIP] Gitlab experiments

coolo:
There are still too many problems:
- ~~gitlab runners are pretty unreliable (took me half a day to find a valid workaround for https://gitlab.com/gitlab-org/gitlab-runner/issues/3144)~~ *moved to eng cloud to avoid kubernetes*
- ~~I don't get a build finished if the buildid didn't change (fix in https://github.com/openSUSE/open-build-service/pull/6360, but needs to be deployed before I can continue)~~, deployed to IBS
- ~~the required checks for build repositories disappear all the time due to https://github.com/openSUSE/open-build-service/issues/6341 - but the building repositories are the only place where I can store reliably the changing gitlab status, as the published repository's buildid only changes on publishing, which means we would accept outdated status.~~ https://github.com/openSUSE/open-build-service/pull/6370 was deployed to IBS
- We don't get a crashing pipeline into OBS. Gitlab only supports webhooks, so unless I want to poll, I need a webhook->rabbitmq translator. Fortunately plugging a pika publisher into gitlab-webhook-receiver is reasonable. The hosting problem is haunting us there again though. Possibly rabbit host itself (secured with a token required on gitlab, similiar to what we require for OBS)

### DIFF
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -623,6 +623,8 @@ exit 0
 %files rabbit-openqa
 %defattr(-,root,root,-)
 %{_bindir}/osrt-rabbit-openqa
+%{_bindir}/osrt-rabbit-build
+%{_bindir}/osrt-rabbit-watch-pipelines
 %{_datadir}/%{source_dir}/rabbit-openqa.py
 %{_unitdir}/osrt-rabbit-openqa.service
 
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -625,6 +625,7 @@ exit 0
 %{_bindir}/osrt-rabbit-openqa
 %{_bindir}/osrt-rabbit-build
 %{_bindir}/osrt-rabbit-watch-pipelines
+%{_bindir}/osrt-report-status
 %{_datadir}/%{source_dir}/rabbit-openqa.py
 %{_unitdir}/osrt-rabbit-openqa.service
 
--- a/rabbit-build.py
+++ b/rabbit-build.py
@@ -1,0 +1,162 @@
+#!/usr/bin/python
+
+from __future__ import print_function
+
+import argparse
+import datetime
+import json
+import logging
+import os
+import sys
+
+import osc
+from osc.core import http_GET, http_POST, makeurl
+
+from osclib.core import target_archs
+from lxml import etree as ET
+try:
+    from urllib.error import HTTPError
+except ImportError:
+    # python 2.x
+    from urllib2 import HTTPError
+
+from PubSubConsumer import PubSubConsumer
+
+
+class Listener(PubSubConsumer):
+    def __init__(self, apiurl, amqp_prefix, project, repository):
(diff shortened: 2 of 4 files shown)


coolo:
For pkglistgen the repo stack isn't the only variable. Input files and attributes are another, so whenever the repo started to build, there is reason enough to double check.

The bigger concern I have with this approach is that the build_standard job can run basically unlimited as it's very low on CPU consumption. So I set the concurrent jobs to a high value because these build_standard jobs run potentially for a very long time - but I can't have many pkglistgen (or repochecker) jobs in parallel.

So I guess there needs to be some manual limiting on this level.

Conan-Kudo:
> Gitlab only supports webhooks, so unless I want to poll, I need a webhook->rabbitmq translator

@coolo So... Pagure supports emitting events, though we currently only have [fedmsg](http://fedmsg.com/), MQTT, and STOMP. Would having AMQP 0-9-1 (RabbitMQ AMQP) help here?

(cc: @pypingou)

coolo:
don't get confused because the PR summary says gitlab that this would be about git.
This is about https://about.gitlab.com/product/continuous-integration/

Conan-Kudo:
@coolo Then I'm confused what you're trying to accomplish here. What is the goal with all this, and what are you trying to do?

Conan-Kudo:
To clarify why I'm asking, it looks like the jobs are actually happening in OBS, not GitLab CI. If the goal is to trigger OBS to do things as part of merge requests to test things (like create an SR and mirror the workflow happening in GitLab), then this is definitely something we could replicate in Pagure. After all, we're currently doing something akin to this with Fedora Dist-Git and Koji.

If the goal is to run stuff in the GitLab Runner before triggering actions in OBS, then that might be slightly more difficult to pull off, but it's doable, too.

Conan-Kudo:
For example, we have [this simple Koji CI service](https://pagure.io/fedora-ci/simple-koji-ci) that interacts with Koji and Pagure through [fedmsg](https://fedmsg.readthedocs.io/en/stable/). Given that OBS emits events via AMQP and can be controlled through its API too, it'd be easy enough to develop a similar service, provided Pagure was extended to AMQP too.

coolo:
your simple CI service is the analogy of https://github.com/openSUSE/obs-tools/tree/master/pull_request_package and both are about git changes, but here we're completely unrelated to git. The only thing in git is the ci configuration.

This is basically starting a bunch of scripts whenever staging projects change and run in gitlab CI runners to report repository status to OBS.

coolo:
I have to run that 'stuff' somewhere and gitlab CI happens to be easy to host :)

Conan-Kudo:
@coolo Why not just make it into a system service and have it spawn celery tasks to do stuff when it gets requests?

coolo:
One of the goals is having a UI with the job status - which gitlab CI offers, celery not so. But this is an experimental branch, that shows several problems with this approach, so I'm looking for something new.

Conan-Kudo:
@coolo Are you looking for [something like this](https://taskotron.fedoraproject.org/resultsdb/results)?

coolo:
Closing this one - FYI @Conan-Kudo, I'm using gocd.org now

PR #1810, opened 2018-11-29
travis: use ecrypted e-mail address to avoid spam from forks.

jberry-suse:
Rather silly there is not better solution, but as noted by others this is the only relatively sane way to do it.

- travis-ci/travis-ci#1094
- travis-ci/travis-ci#5063
- mozilla/build-tools@c13a297

An issue since coolo enabled testing in a fork a few weeks ago.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -156,4 +156,6 @@ before_cache:
 notifications:
   email:
     recipients:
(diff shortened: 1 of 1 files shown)


coolo:
I warned you it would be an issue. But the solution looks reasonable enough.

PR #1812, opened 2018-12-03
Remove 'ports' scope as powerpc is using target

coolo:
The 'ports' scope just didn't work for @hramrach, so we fixed the project setup to work as target project. To avoid tons of merge conflicts, I cherry-picked fixes I run for SP1 (from gitlab-experiments, but are unrelated to gitlab)


coolo:
To be exected (from packagelist vm I assume):
```
./pkglistgen.py --apiurl https://api.opensuse.org update_and_solve -s target -p openSUSE:Leap:15.1:PowerPC

coolo:
@dirkmueller as discussed - also removed arm scope. The nonfree code is disabled automatically if the target project config doesn't have a non-oss config, so deal done

PR #1814, opened 2018-12-04
pkglistgen: Big refactoring of how packages are resolved

coolo:
No further refactoring, but some cleanup of imports (which obviously has
dangers in itself)


coolo:
Now it's refactored and the fallout from implementing Tumbleweed support was meanwhile fixed in SLE. So next step: Leap and droplist

lnussel:
*shrug*. you break it, you fix it :)

PR #1821, opened 2018-12-11
ARM: don't set the snapshot version

dirkmueller:
Apparently this is different to how x86 does it, and
we don't want to be different


ggardet:
LGTM

nilxam:
LGTM

PR #1822, opened 2018-12-11
Fix check for "no change in snapshot version"

Vogtinator:
It compared against the version in openQA and not the version in :ToTest,
which can result in ttm getting stuck if the previous release did not complete.

Ironically `current_snapshot` comes from openQA while `current_qa_version` is from OBS.


coolo:
then I suggest to remove the irony

Vogtinator:
> then I suggest to remove the irony

That should be done in a separate refactoring PR IMO.

nilxam:
how did 'current_snapshot' is from openQA? isn't it from ToTest project?

Vogtinator:
> how did 'current_snapshot' is from openQA? isn't it from ToTest project?

Yes, the description is incorrect.

current_qa_version is `version_totest` in the pseudometa and `current_snapshot` is usually derived using `self.iso_build_version` in `ToTest`.

coolo:
is this still relevant? I guess so, but would like to verify

Vogtinator:
Yup.

PR #1829, opened 2019-01-09
Only decline delete requests that affect linked packages within the target

### LINKED ISSUE #1824: factory-auto bot doesn't like delete request on SP1
When issueing a delete request on a package in SLE15-SP1, factory-auto rejects the request (the package would become inherited back from SLE15).

See https://build.suse.de/request/show/179811 as example.

coolo:
This decline reason came from https://github.com/openSUSE/openSUSE-release-tools/pull/603
without much more context, so I'm guessing it's about delete requests for
e.g. kernel-default that should delete kernel-source

Packages that link to *other* projects basically do not exist in openSUSE
distributions but are very common in SLE service packs, so we need to diff
on this.

Fixes #1824


coolo:
[I] 181104 accepted: ok

PR #1831, opened 2019-01-13
Revert "Add salt packages to DVD for Staging tests"

coolo:
Reverts openSUSE/openSUSE-release-tools#1623

### DIFF
--- a/create_test_Factory_dvd-1.testcase
+++ b/create_test_Factory_dvd-1.testcase
@@ -96,7 +96,3 @@ job install name mariadb
 
 ## we newly want to start glxgears in Staging tests (sync up with SLE)
 job install name Mesa-demo-x
-
-## required in Staging tests
-job install name salt-master
-job install name salt-minion


nilxam:
Fine with me. @DimStar77 ?

dirkmueller:
I thought it works with python 3.7? Wasn't there a comment that it works after fixing requirements?

DimStar77:
Only in the combination of removing salt from the staging tests - but this is counter the initiatives that are planned around salt. We would basically break salt 'until further notice'

> I thought it works with python 3.7? Wasn't there a comment that it works after fixing requirements?

@scarabeusiv indeed did mention something like that last week

scarabeusiv:
Well it should work with python 3.7. At least basic operations worked here localy, but the developers set buildrequire to  < 3.7 (i suppose they can revert that) and then we can carry on.

The only major incompatibility was the tornado 5 not being compatible with salt rather than python 3.7 not being compatible.

aplanas:
The salt package from factory have a patch for 3.7 compatibility. Maybe @isbm have more data about the Tornado situation in upstream.

I suggest to remove in Factory the < 3.7 BR and maintain the tornado one, unless there is a temporal patch for Tornado

DimStar77:
There was indeed a Tornado 4.x submitted that support Python 3.7 - so this should make things smoother

coolo:
I see no movement on the < 3.7 front

dirkmueller:
Patches have been submitted, can we get them checked in?

DimStar77:
> Patches have been submitted, can we get them checked in?

Where has the salt submission for it supporting python 3.7 done? I get this sub, we're basically ready for it.

scarabeusiv:
Well I suppose dirk means https://build.opensuse.org/request/show/665721

coolo:
superseded and accepted as https://build.opensuse.org/request/show/666509 - leaves 3 steps more :)

DimStar77:
https://build.opensuse.org/request/show/666637 is the SR en route to TW - so all in order

Issue #1800, opened 2018-11-20
Have staging accept put the TW version in an attribute


coolo:
To support pkglist workflow, we need to stop modifying product - but have pkglistgen write a 000product with the right version.

So we discussed to have staging accept (additionally for now) write the TW version in a OSRT:ProductVersion attribute.

This would then be patched into the generated product by pkglistgen.

coolo:
@nilxam is this something you can look at?

coolo:
Created OSRT:ProductVersion modifiable by target maintainer and maxlin_factory on obs and ibs.

nilxam:
I think we can close this issue? or is there something else undone?

coolo:
no

PR #1832, opened 2019-01-15
accept_command: add additional version number to OSRT:ProductVersion

nilxam:
see https://github.com/openSUSE/openSUSE-release-tools/issues/1800


osclib/accept_command.py
```diff
@@ -279,6 +281,11 @@ def create_new_links(self, project, pkgname, oldspeclist):
                     source_file_save(self.api.apiurl, project, package, '_link', link)
         return True
 
+    def update_version_attribute(self, project, version):
+        version_attr = attribute_value_load(self.api.apiurl, project, 'ProductVersion')
```
coolo: you don't have to be that careful with attributes, they don't trigger any rebuilds. The only thing to worry about is potentially polluting the project history. It doesn't hurt to be careful though :)

PR #1842, opened 2019-01-21
pkgistgen: Update manager

coolo:
This consists of 2 parts: 
 - handle_update_repos command taking a project as argument and updates it's 000update-repos with *new* packages in a minimal susetags format (saving space - previously we added 15MB to git every day, now it's ~80KB)
 - update_and_solve overwriting weakremovers.inc in 000release-packages


coolo:
The calculated list is a little hard to compare as the methologies used are so greatly different, but @DimStar77 agreed to trying it out

PR #1845, opened 2019-01-22
[WIP] Add rabbitmq msg for announcer

### LINKED ISSUE #1834: [proposal] factory-package-news could send a rabbitmq message
I would like to hook an CI action after the release and publish of a new tumbleweed image. I have discussed this a bit with @DimStar77 and he suggested that the factory-package-news script could also fire a rabbitmq message.

I would propose that the `factory-package-news/announcer.py` also sends a message over the opensuse rabbitmq server.

I'd be happy to help out with this, but would need some guidance what exactly needs to be done. I guess the following should be addressed:
- Should this actually be added or some other mechanism be used?
- What message topic should be used and what data should be sent alongside? Maybe the changelog?
- Add this to the documentation on https://rabbit.opensuse.org/ and in this repo. Where else?

dcermak:
This fixes #1834.

The whole code is untested, since I wasn't able to find any test code for `announcer.py`. I would be more than happy to add some tests to this, as I have absolutely no idea whether this works.


jberry-suse:
This would seem to be backwards. OBS should be emitting an event that the announcer listens for to send out an e-mail.

dcermak:
In this case OBS is not sufficient, the factory-package-news script is invoked by openQA after a new changelog is created (I think). But I tend to agree with you, this process seems the wrong way around.

 @DimStar77 you probably know more about this and what the right procedure is.

coolo:
OBS sends out an event if openSUSE:Factory:ToTest/images is published - containing this ISO. I wonder what's your use case.

dcermak:
I would like to run a build using the fresh Tumbleweed image, once it got released. Originally I wanted to listen for the rabbitmq event that ttm sends around, but Dominique told me that they are probably firing too soon and that OBS still has to do the release afterwards.

DimStar77:
> OBS sends out an event if openSUSE:Factory:ToTest/images is published - containing this ISO. I wonder what's your use case.

the problem at this time is that download.o.o will only be updated in like 2 - 3 hours, as we hold off the switch for mirrors to catch up. Hence the idea of the announcer also creating an event - as it actually monitors when the publishing is actually really done and ready for users to consuime download.o.o again

Preferably this event would come from another place, like the "publisher holding off"-script for example, at the time when it does the switch - but anybody daring to touch that code will curse for the rest of his life

IFF we would add rabitmq to the publisher scripts, then of course the announcer would/could/should listen as well

jberry-suse:
/me remembers a proposal on this topic. oh well.

coolo:
> 
> Preferably this event would come from another place, like the "publisher holding off"-script for example, at the time when it does the switch - but anybody daring to touch that code will curse for the rest of his life
> 
now *that* explains @lnussel's behavior! :)

PR #1848, opened 2019-01-30
fix incompatible repo id tagging for openSUSE products

adrianschroeter:
please note that this is an incompatible change, but all our openSUSE product pool repo listings are currently anyway broken.


coolo:
The failure is unrelated:

```
$ flake8
./leaper.py:446:21: E117 over-indented
./build-fail-reminder.py:37:9: E117 over-indented
./deptool.py:304:29: E117 over-indented

PR #1851, opened 2019-02-05
Fix ignore list handling in repo_checker.pl

coolo:
Once write_package is called, it's in the meta data and there is no way to un-ignore it


coolo:
well, it does in my dry runs - when I run repo_checker.pl with the given ignore list, it works after my patches.

But from IRC discussion I get I should have used -w instead of -i

jberry-suse:
Yeah, you should see problems with multi-layered staging runs like adi.

PR #1852, opened 2019-02-05
[WIP] repo_checker: provide repo_checker-binary-whitelist-ring config option and remove explicit target project repository pair for staging projects

jberry-suse:
This should be what is desired by #1851. I have not run locally as I would have to warm up my caches, but I will if needs extensive debugging.


jberry-suse:
Need to consider effect on delete requests and handle filtering out "ports" via missing archs properly since not expanding entirely will not work for SLE.

PR #1854, opened 2019-02-06
Change repo checker to expect non-leaky rings

coolo:
This is based on a discussion with @jberry-suse. The way we currently rely on the target project is just too problematic, so we no longer - at the price of reporting more problems as ring problems.

Most requires should come from the staging project itself - if it's just too expensive, binaries to ignore can be put in repo_checker-binary-whitelist-ring in the target project's config.

I setup Factory this way, reported 9 bugs for SLE15 and created an extensive blacklist there - and started to do the same for leap (WIP).  

What I tested:
 - Factory letter stagings
 - Factory adi stagings
 - Leap letter stagings
 - SLE15 letter stagings

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -393,7 +393,13 @@ def repository_arch_state(apiurl, project, repository, arch):
     # just checking the mtimes of the repository's binaries
     url = makeurl(apiurl, ['build', project, repository, arch, '_repository'])
     from osclib.util import sha1_short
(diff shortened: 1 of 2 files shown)


coolo:
Superseding #1852 - and IMO ready

coolo:
Jimmy does not like the way I did this, but didn't veto either. So I'll deploy and continue watching

PR #1857, opened 2019-02-07
osclib/list_command: 24ae4ba broke formatting.

jberry-suse:
- 01d9504d90a403b8dddc44bf1f1208bb312f5f24:
    osclib/list_command: 24ae4ba81 broke formatting.
    
    ```
    The author was attempting to port for python3, but changed the meaning:
    
    - print ' ', line
    + print(' ' + line)
    
    The first prints "  $line", while the second:
                     " $line".
    
    This changes console output from:
    
    science
     666254 libArcus-lulzbot               -> delete       (delete request)
             3 more users
    
    back to:
    
    science
      666254 libArcus-lulzbot               -> delete       (delete request)
             3 more users
    
    (ignore lines up and standard double space indent used)
    ```

### DIFF
--- a/osclib/list_command.py
+++ b/osclib/list_command.py
@@ -59,7 +59,7 @@ class ListCommand:
                 if message:
                     line += '\n' + Fore.WHITE + message + Fore.RESET
 
-                print(' ' + line)
+                print(' ', line)
 
         if len(splitter.other):
             non_ring_packages = []


osclib/list_command.py
```diff
@@ -59,7 +59,7 @@ def perform(self, supersede=False):
                 if message:
                     line += '\n' + Fore.WHITE + message + Fore.RESET
 
-                print(' ' + line)
+                print('  ' + line)
```
aplanas: `print(' ', line)` is more natural here IMHO (two spaces and not very visible)

jberry-suse: Except that the output is then:

```
(' ', '$line')
```

coolo: not in python3 nor with the future:

```
>>> from __future__ import print_function
>>> line='hallo'
>>> print(' ', line)
  hallo
```

(which if you checked, the list_comand.py already imports)

aplanas: Indeed, what @coolo says.

jberry-suse: Doesn't bother me either way.

PR #1860, opened 2019-02-11
repo_checker: Simplify cycle check

coolo:
No longer compare against the target project's cycle, but just against a configured list of package names. This way we're not bound to refreezing stagings if we reduced cycles and it's clearer to the operator what happens and how to react to it.


repo_checker.py
```diff
@@ -278,52 +277,27 @@ def install_check_parse(self, output):
         if section:
             yield InstallSection(section, text)
 
-    @memoize(ttl=60, session=True)
-    def cycle_check_skip(self, project):
```
jberry-suse: When new cycles are introduced intentionally how will someone indicate that to repo-checker?

coolo: by changing the config.

nilxam:
Looks ok to me. We should just give a try on production, having r-c to accept cycle if it was in 'repo_checker-allowed-in-cycles'  at least we can work out bootstrap staging blocking.

coolo:
I have fixed the config for sle15, factory and 15.1 - and verified that maintenance incidents are not affected (unless we start to release 2 bootstrap packages in one incident, which may happen in the future)

PR #1862, opened 2019-02-11
Add kubic-kured-image and kubic-pause-image to container_products

Vogtinator:
Mostly untested - not sure how that could be done.

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -770,7 +770,9 @@ class ToTestFactory(ToTestBase):
                        ImageProduct('livecd-tumbleweed-gnome', ['i586', 'x86_64']),
                        ImageProduct('livecd-tumbleweed-x11', ['i586', 'x86_64'])]
 
-    container_products = [ImageProduct('opensuse-tumbleweed-image:docker', ['i586', 'x86_64'])]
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -802,7 +804,9 @@ class ToTestFactoryPowerPC(ToTestBase):
 
     image_products = [ImageProduct('opensuse-tumbleweed-image:lxc', ['ppc64le'])]
 
-    container_products = [ImageProduct('opensuse-tumbleweed-image:docker', ['ppc64le'])]
+    container_products = [ImageProduct('opensuse-tumbleweed-image:docker', ['ppc64le']),
+                          ImageProduct('kubic-kured-image', ['ppc64le']),
```
DimStar77: https://build.opensuse.org/package/show/openSUSE:Factory:PowerPC/kubic-kured-image

At least kured-image for PowerPC is in state excluded; so this can't work

Vogtinator: That's because `:PowerPC/kubic-kured-image` is not a link yet and images has no linkedbuild.

DimStar77: good catch; fixed for :ARM and :PowerPC - kured should thus be buiding

PR #1870, opened 2019-02-18
ttm: Add openSUSE:Leap:15.1:ARM:Images

Vogtinator:
- Contains JeOS and opensuse-leap-images

### DIFF
--- a/totest-manager.py
+++ b/totest-manager.py
@@ -160,7 +160,7 @@ class ToTestBase(object):
         if not base:
             base = self.project_base
         for binary in self.binaries_of_product(project, tree, repo=repo, arch=arch):
-            result = re.match(r'.*-(?:Build|Snapshot)([0-9.]+)(?:-Media.*\.iso|\.docker\.tar\.xz)', binary)
+            result = re.match(r'.*-(?:Build|Snapshot)([0-9.]+)(?:-Media.*\.iso|\.docker\.tar\.xz|\.tar\.xz)', binary)
             if result:
                 return result.group(1)
         raise NotFoundException("can't find %s iso version" % project)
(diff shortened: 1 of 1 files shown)


totest-manager.py
```diff
@@ -1016,6 +1016,20 @@ class ToTest151Images(ToTest150Images):
     def openqa_group(self):
         return 'openSUSE Leap 15.1 Images'
 
+class ToTest151ARMImages(ToTest150Images):
```
dirkmueller: Should inheriting ToTest151Images

totest-manager.py
```diff
@@ -1016,6 +1016,20 @@ class ToTest151Images(ToTest150Images):
...
+    image_products = [ImageProduct('JeOS', ['armv7l', 'aarch64'])]
+
+    container_products = [ImageProduct('opensuse-leap-image:docker', ['x86_64'])]
```
dirkmueller: Huh?

totest-manager.py
Vogtinator: Could, but it's also not necessary.

totest-manager.py
Vogtinator: What's not clear about this?

totest-manager.py
dirkmueller: I think it would better reflect the hierarchy: ARM 15.1 Images is a derivative of the 15.1 Images, not of 15.0 Images. you're correct that because you're currently overwriting all implementations of 15.1 Images it doesn't make a lot of difference, but it will be more futureproof when 15.0 Image baseclass is going to be faded out.

Vogtinator:
It looks like `zypper` and travis don't like each other - it reproducably segfaults when building the repo cache. Maybe not enough memory?

jberry-suse:
Should have [7.5G according to documentation](https://docs.travis-ci.com/user/reference/overview/). It seems something changed in last 3-4 days. I can reproduce locally with 32G of RAM.

PR #1876, opened 2019-02-25
Repo checker: work on staging projects as a whole

coolo:
Implement a new command that takes a target project and one of its
staging projects and will run the checks against that project (including
delete requests) and if all is fine approves all reviews selected in
that staging.

This is to be used as part of the staging pipeline (and has been tested
with Factory's staging projects)

### DIFF
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -464,6 +464,7 @@ fi
 %{_bindir}/osrt-status
 %{_bindir}/osrt-sync-rebuild
 %{_bindir}/osrt-unmaintained
+%{_bindir}/osrt-staging-installcheck
 %{_datadir}/%{source_dir}
 %exclude %{_datadir}/%{source_dir}/abichecker
 %exclude %{_datadir}/%{source_dir}/%{announcer_filename}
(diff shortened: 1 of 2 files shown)


coolo:
This should really be a new tool - as it's clearly not a reviewbot and I had to copy a lot of code as I need the data slightly different.

coolo:
But for now we need this *and* the review bot as adi projects don't really have any marker yet that we ran repo check. But this requires an OBS feature first.

repo_checker.py
```diff
@@ -70,6 +70,129 @@ def project_only(self, project, post_comments=False):
...
+                                          {'view': 'fileinfo_ext'})
+                            reverse_fileinfo = ET.parse(osc.core.http_GET(url)).getroot()
+                            for require in reverse_fileinfo.findall('requires_ext'):
+                                # extract >= and the like here too
+                                dep = require.get('dep').split(' ')[0]
+                                if dep != provide:
+                                    continue
+                                for provided_by in require.findall('providedby'):
+                                    if provided_by.get('name') in built_binaries:
+                                        continue
```
okurz: wow, 41 spaces :)

coolo: yeah, I was trying to squeeze it all in one function :)

jberry-suse:
Clearly a mess...

coolo:
I separated (and duplicated in parts) the code from repo checker

PR #1890, opened 2019-03-07
Add all checks from staging-required-checks-adi to main repo of adi p…

coolo:
…rojects

For Factory I want to switch away from repo checker reviews, but a required
check on project level. So first step: make it required for adi projects.


coolo:
tested against openSUSE:Factory (with config) and openSUSE:Leap:15.1 (without config)

PR #1893, opened 2019-03-08
No longer add repo checker as reviewer for Factory

coolo:


coolo:
having the config in code has quite severe drawbacks on changes :(

PR #1906, opened 2019-03-19
Replace deprecated (and not advised) yaml.load() with yaml.safe_load().

jberry-suse:
The following warning is output:

```
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
```


coolo:
hmm, the failure started 5 days ago. Sounds like some dep broke again :(

PR #1912, opened 2019-03-21
origin-manager: expose osc-plugin-origin commands via operator and provide initial supplementary interface.

jberry-suse:
- ca687d4089bcbc2f82e8a4c50f2719eb4a4ce72f:
    userscript/origin: provide initial supplementary interface.

- 0734bf756c55df56fe61e2c2c3b6544b2e217dbb:
    obs_operator: expose osc-plugin-origin commands via GET.

- 2c45688f2f2a6f6f0dc829ce913b7317a6a0fc06:
    obs_operator: extract osc environment setup as with object.

- 3293a24a894fdf4350d35a749fac8c79f13988d4:
    osc-origin: osrt_origin_lookup(): print time since generated.

- d5dff4d15c01513b67568be31f9664c3f84bdbf3:
    osc-origin: osrt_origin_lookup_file(): include project in cache file name.

No frills supplementary interface that makes use of operator server. Can expand OBS overlay JS to provide links to overall origin list and report, etc.

image

image

The "UI" as referred to will use the same basic format of querying the operator sever origin data paths, but will require additional information to be exposed.

### DIFF
--- a/userscript/origin.js
+++ b/userscript/origin.js
@@ -1,0 +1,26 @@
+// ==UserScript==
+// @name         OSRT Origin
+// @namespace    openSUSE/openSUSE-release-tools
+// @version      0.1.0
+// @description  Supplement OBS interface with origin information.
+// @author       Jimmy Berry
+// @match        */package/show/*
+// @require      https://code.jquery.com/jquery-3.3.1.min.js
+// @grant        none
+// ==/UserScript==
+
+(function()
+{
+    var pathParts = window.location.pathname.split('/');
+    var project = pathParts[pathParts.length - 2];
+    var package = pathParts[pathParts.length - 1];
+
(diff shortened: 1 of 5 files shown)


userscript/origin.js
```diff
@@ -0,0 +1,26 @@
...
+// ==/UserScript==
+
+(function()
+{
+    var pathParts = window.location.pathname.split('/');
+    var project = pathParts[pathParts.length - 2];
+    var package = pathParts[pathParts.length - 1];
+
+    var domain_parent = window.location.hostname.split('.').splice(1).join('.');
+    var subdomain = domain_parent.endsWith('suse.de') ? 'tortuga' : 'operator';
```
coolo: please don't abuse tortuga for that. Tortuga is the openqa installation used by 3+ developers 

You have 2 openstack installations available to you

PR #1919, opened 2019-03-25
Totest manager refactoring

coolo:


coolo:
@Vogtinator @DimStar77  if that's ok with you so far, I would move the configs into a project attribute and remove config from the code.

ttm/manager.py
```diff
@@ -0,0 +1,659 @@
...
+        root = ET.parse(f).getroot()
+        for binary in root.findall('binary'):
+            binary = binary.get('filename', '')
+            result = re.match(r'.*-([^-]*)-[^-]*.src.rpm', binary)
+            if result:
+                return result.group(1)
+
+        raise NotFoundException("can't find %s version" % self.project)
+
+    def current_sources(self):
```
Vogtinator: I don't like the "source" naming here - it should at least contain the word "version" in one way or another.

ttm/manager.py
```diff
@@ -0,0 +1,659 @@
...
+        if self.project.take_source_from_product is None:
+            raise Exception('No idea where to take the source version from')
+
+        if self.project.take_source_from_product:
+            if self.project.is_image_product:
```
Vogtinator: This could be avoided by just checking whether `main_products` is empty.

ttm/manager.py
coolo: I found there were too many versions - so I wanted something different than 'current_version'. Meanwhile I'd go with version_from_project

ttm/manager.py
coolo: I can do that.

coolo:
The dry runs were all fine, so I'd move forward.

ttm/manager.py
```diff
@@ -0,0 +1,660 @@
...
+        raise NotFoundException("can't find %s version" % self.project)
+
+    def version_from_project(self):
+        if not self.project.take_source_from_product:
+            return self.release_version()
+
+        if len(self.project.main_products):
+            return self.iso_build_version(self.project.name, self.project.main_products[0])
+
+        #assert(self.project.is_image_product)
```
Vogtinator: Should be removed

coolo: it's not a leftover debug - it's a comment

Vogtinator: It's a commented out leftover debug statement then - this assert can't possibly be useful in production.

coolo: Right, like most comments only help the reader to understand the logic - not the production

coolo: You might be confused by this comment being pseudo code, but that's only because it's easier to express this way

coolo: I reworked that whole image_product thingy now

Vogtinator:
Looking good - I just don't like this being implicit:

```
        if not len(self.project.main_products):
            version_file = version_file + '_images'
```

Simply by not having an empty `main_products` array it picks an entirely different version file. I don't have a better idea though.

Vogtinator:
Actually, it might be possible to delete that now - the `:Images` subprjs have their own `00Meta/version_*` now.

coolo:
let's keep more refactorings for another round. This was just a precondition for the real split

PR #1920, opened 2019-03-26
userscript/origin: use cross-domain request with credentials.

jberry-suse:
Otherwise, requests are denied with not running operator server in development mode. This is the same behavior used by drag-n-drop selects, just overlooked.


jberry-suse:
With this and deployment to heroes kubernetes one should be able to use.

PR #1930, opened 2019-03-28
Fix crash on images projects (openSUSE:Leap:15.0:Images)

coolo:
And reduce debug output - it's hard to see the actual debug output


coolo:
That's the problem with --dry disabling half the code :)

https://botmaster.suse.de/go/tab/build/detail/TTM.Leap_15.0_Images/24/Run/1/Run

Vogtinator:
ftp_products and main_products could be combined actually, by adding a `should_set_release` marker to those in the config (or `if "ftp" in product`...).

PR #1931, opened 2019-03-28
Fix one more crash in totest-manager on actual release

coolo:


coolo:
I found one more even, but now I went through all the usages.

PR #1935, opened 2019-03-28
Move some more pipelines from citest to official repo

coolo:


gocd/staging.bot.gocd.yaml
```diff
@@ -0,0 +1,146 @@
...
+        jobs:
+          Run:
+            resources:
+            - repo-checker
+            tasks:
+            - script: |-
+                cd scripts
+                ./staging-installcheck.py -A https://api.opensuse.org -p Leap:15.1
+    SBot.Regular.Leap.15.1:
+      group: Factory.pkglistgen
```
nilxam: I thought it should be Leap.pkglistgen group?

gocd/staging.bot.gocd.yaml
```diff
@@ -0,0 +1,146 @@
...
+      - Run:
+          approval:
+            type: manual
+          jobs:
+            Run:
+              timeout: 0
+              resources:
+              - staging-bot
+              tasks:
+              - script: osc -A https://api.opensuse.org staging -p Leap:15.1 rebuild
```
nilxam: '-p' parameter won't working with simplified 'Leap:15.1' IIRC, it works with '-p Factory' though.

coolo: yeah, that's a bug in the erb

gocd/staging.bot.gocd.yaml
coolo: that's true

coolo:
I rather move the others than adding an if here :)

PR #1941, opened 2019-03-28
Move gocd/ to dist/gocd/.

jberry-suse:
Just like kubernetes. Would likely make sense to move systemd, but need to handle subdirectories and Makefile. Otherwise, avoid more root directories.


coolo:
how do you define dist/? because 
  a) gocd references that git subdir in the config directly
  b) the gocd subdir is referenced in the gocd subdir itself
  c) we don't distribute it

As such: I would say, not a good idea

jberry-suse:
`dist/` being the home for scripts/aids pertaining to the various places the tools are distributed rather than one of the tools. The `ci` directory is clearly not a "distribution", nor the `kubernetes` directory, but rather aids for use in those environments or deploying the tools there.

a and b simply need to be resolved in the commit then and are not reasons.

coolo:
ok. in gocd/ are also tools used on gocd. So really not a good idea

jberry-suse:
> ok. in gocd/ are also tools used on gocd.

Exactly like `dist/ci` and my description.

> So really not a good idea

Does not follow the logic, but I remember why I stopped trying.

PR #1942, opened 2019-03-28
Remove the special logic for 000product

coolo:
If we release 000product with set_release the FTP tree is mis-named


Vogtinator:
Leap should be covered already by the correct config - the other TTBNew differences are covered.

PR #1946, opened 2019-04-03
WIP: Test my obs fix

coolo:


Vogtinator:
FYI, you can simply create a draft pull request in the GH web UI - it blocks merging even for repo owners.

coolo:
I've seen others doing it, but never bothered where the button is. Now I checked - next time :)

PR #1953, opened 2019-04-15
Fix problems found during testing Factory's TTM

coolo:


ttm/publisher.py
```diff
@@ -160,13 +160,19 @@ def send_amqp_event(self, current_snapshot, current_result):
...
             current_snapshot = self.version_from_totest_project()
         except NotFoundException as e:
             # nothing in test project (yet)
             self.logger.warn(e)
             current_snapshot = None
 
+        if not self.get_status('testing'):
+            # migrating. Normally the releaser maintains that snapshot, but to differentiate
+            # bootstrap of new project and migration we explicitly set 'NONE' if there is Nothing
+            self.update_status('testing', current_snapshot if current_snapshot else 'NONE')
```
Vogtinator: AFAICT the releaser's behavior is the same whether it's set to `NONE` or not set at all, so why set it to NONE explicitly?

coolo: See the code I added to releaser.py in this PR

coolo: Perhaps we can require bootstrap to use --force ?

Vogtinator: That sounds like a good idea anyway - I mostly care about getting rid of 'NONE' though, it's ugly and non-obvious what it does

coolo: I changed that now

PR #1954, opened 2019-04-15
TTM: add fixup command to update attribute

coolo:
It may be necessary to fix the TTM Status to be aligned with reality, so
provide a command line to do so easily (osc is unfortunately useless for
multiline attributes)

### DIFF
--- a/ttm/cli.py
+++ b/ttm/cli.py
@@ -15,7 +15,7 @@ import logging
 import ToolBase
 import cmdln
 
-from ttm.manager import QAResult
+from ttm.manager import ToTestManager, QAResult
 from ttm.releaser import ToTestReleaser
 from ttm.publisher import ToTestPublisher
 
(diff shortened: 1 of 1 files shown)


ttm/cli.py
```diff
@@ -68,3 +68,15 @@ def do_run(self, subcmd, opts, project):
         if ToTestPublisher(self.tool).publish(project) == QAResult.passed:
             ToTestPublisher(self.tool).wait_for_published(project)
         ToTestReleaser(self.tool).release(project)
+
+    @cmdln.option('--status', choices=['testing', 'publishing'], help='The status to fix')
```
Vogtinator: It's something that should be used manually only, so might as well allow the other choices as well here?

coolo: Yeah, but the others don't really have any mechanism value, so I don't consider them important enough. If I listed more here they would run out of sync just too quickly.

But I can drop the choices and make it a string argument - it's for experts only anyway

coolo: I moved the choices to help text now

PR #1955, opened 2019-04-15
Rework handling of packages with multiple specs

coolo:


coolo:
Note that the tests are broken as the obs.py testsuite is just not obs. And I won't touch them, but rather invest a little in a proper solution

PR #1957, opened 2019-04-16
Fix release of Leap

coolo:


Vogtinator:
IMO wrong approach, the migration code needs to take care of setting `published` as well, like https://github.com/openSUSE/openSUSE-release-tools/blob/master/ttm/publisher.py#L173

coolo:
but that risks to release a new snapshot while we haven't published. And the migration code already set testing by now, so your approach won't run. Unless of course you expect other customers of this code somewhere.

Vogtinator:
Right, I missed the `!` in line 183.

PR #1958, opened 2019-04-16
Add program for cleaning up maintenance_release projects for containers

Vogtinator:
Currently every single container released to `openSUSE:Containers:Tumbleweed` is kept forever, that needs to change.

Only those containers providing binaries to the latest five version for each
archs are kept, all others are deleted.

### DIFF
--- a/gocd/container-cleaner.gocd.yaml
+++ b/gocd/container-cleaner.gocd.yaml
@@ -7,7 +7,7 @@ pipelines:
       OSC_CONFIG: /home/go/config/oscrc-totest-manager
     materials:
       script:
-        git: https://github.com/openSUSE/osc-plugin-factory.git
+        git: https://github.com/openSUSE/openSUSE-release-tools.git
         branch: master
         destination: scripts
     timer:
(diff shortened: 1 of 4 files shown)


gocd/container-cleaner.gocd.yaml
```diff
@@ -0,0 +1,23 @@
+format_version: 3
+pipelines:
+  ContainerCleaner.Tumbleweed:
+    group: openSUSE.Checkers
+    lock_behavior: unlockWhenFinished
+    environment_variables:
+      OSC_CONFIG: /home/go/config/oscrc-totest-manager
+    materials:
+      script:
+        git: https://github.com/openSUSE/osc-plugin-factory.git
```
coolo: better use the proper name of the repo

Vogtinator:
Not sure why, but apparently the used timer spec `0 0-23/2 * ? * *` results in it running every 2nd minute.

coolo:
and you intentended hours? Then move it down one more ;)

https://docs.gocd.org/current/configuration/configuration_reference.html#timer

Vogtinator:
Apparently I read the wrong documentation. Fixed.

PR #1962, opened 2019-04-20
More porting to python3

coolo:
Travis changed default python to 3.6, which has impact on the way flake8 checks. So make sure we run fine under flake8 of python 3


lnussel:
the changes don't change the #!/usr/bin/python line so for example factory-package-news.py now no longer works with python2.

coolo:
It was never intentend to run on python3 - just to be compatible with it. I assume your osc is python3 now?

PR #1963, opened 2019-04-21
Replacing OBS.py with VCR tapes of OBS

coolo:
This is a PoC and the idea is to record responses of a real OBS instance into [VCR Cassettes](https://vcrpy.readthedocs.io/en/latest/usage.html), so that we can replay them while regression testing.

The docker-compose files to start the instance are bundled, the Dockerfile for the image as well - but it's meant to build on OBS.

The docker-compose start would only happen while developing or refreshing cassettes, normally they are not required.

This requires a lot of extra data in git, but IMO it's worth it in this context.


coolo:
https://travis-ci.com/coolo/osc-plugin-factory/jobs/194656359 for CI, I'm starting to like it.

tests/test.oscrc
```diff
@@ -0,0 +1,11 @@
+[general]
+apiurl = http://Admin:opensuse@localhost:3737
+
+[http://localhost:3737]
+user=Admin
+pass=opensuse
+
+[http://Admin:opensuse@localhost:3737]
```
coolo: @jberry-suse this was an attempt to force osc to always send basic auth headers - but this failed. osc always sends a reques that ends with 401 to then retry from exception handler with auth.

Do you have an idea how to force direct use of auth? This would cut the size of the cassettes in half :(

dist/ci/obs-ic/api.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
+
+su - frontend -c "RAILS_ENV=production rails r \"Project.find_or_create_by(remoteurl: 'https://api.opensuse.org/public', name: 'openSUSE.org').store\""
```
coolo: I guess I don't even need this by default - and one less .sh file to worry about.

jberry-suse:
> This is a PoC and the idea is to record responses of a real OBS instance into [VCR Cassettes](https://vcrpy.readthedocs.io/en/latest/usage.html), so that we can replay them while regression testing.

This has zero benefit over a real instance which we already have working. Not to mention a real instance allows one to develop against it and handle changes down the road without huge dumps. This is merely "automated mocks."

One should ask what are we lacking form the real setup I already made and how hard is it to add what is needed. The answer is very little.

coolo:
No, the answer is: running the test suite during package builds in OBS. But thanks for your feedback.

jberry-suse:
> No, the answer is: running the test suite during package builds in OBS. But thanks for your feedback.

This should not be prevented by a testing setup. If so that is a bigger problem. Huge static dumps are always a bad idea.

coolo:
online tests are always a bad idea as well - and your cloning relies on online instance.

jberry-suse:
Building from scratch via commands would be the best, but of course we do not even have the for the real projects as they are half manually created. There in lies the true problem.

coolo:
Well, if I want reproducible tests, I need to create the projects as part of the test. And that's what I get from using an empty OBS - and vcr speeds it up.

I don't like the dumps in git either, but I find them preferred over the alternatives. I find the manual mocking unacceptable by now and only testing on travis is tough too. Because whatever you do with a real instance: the data in it affects the results.

coolo:
And automatic mocking isn't such a bad term for it - and I find it sounds like a big win :)

coolo:
To summarize: in the new tests, every test creates a staging scenario with varying amount of data, works on it and removes all of it again. The OBS running against it is plain, but the interactions with it are recorded by the developer for faster tests.

lnussel:
still if you modify something that touches the logic how obs replies one needs to do a new recording right? Doesn't sound like too much fun.

Wrt amount of data, how about moving that to a submodule?

coolo:
> still if you modify something that touches the logic how obs replies one needs to do a new recording right? Doesn't sound like too much fun.

Well, there are 2 parts to this PR. One is avoiding manual httpretty as e.g. in obs.py or 
a590787e3368fb360c4bc085e2d5eb20215d5779 by creating test scenarios from scratch in a real OBS.

The other is recording the HTTP traffic to a) speed it up and b) make it offline available. Rerecording the traffic is *much* easier than adoption obs.py or manual mocks. And we can run travis without the records and the package builds with it, so we know that OBS or our logic changed.

And the recording is atm 3.8MB. If the goal of the submodule is to avoid large checkouts for users, I suggest we move all of tests/ into it.

coolo:
The submodule idea won't work - it would be basically impossible for contributors to create a pull request worth it. Because you need to commit to the submodule and afterwards create a PR moving the submodule. We already have a lot of fun with openqa tests and its needles - and there we just use 2 git checkouts and not a submodule.

coolo:
I measure 190 seconds to regenerate the VCR, 38 seconds replaying them. As more tests would use this mechanism, the distance would go grow - but I guess if you guys don't share my goal of running the test suite offline, the speedup is not worth the stop energy.

The VCR comes with its own set of problems (e.g. the need to replay also the time), so I can ditch it meanwhile with more happiness than in the beginning :)

lnussel:
I can't really judge whether this method is good or not in practice, didn't develop any tests for a long time. I agree that the manual mockup are not a good solution for sure though. IIRC last time I had to deal with them they didn't help either as OBS changed and broke the bots while the test in question was still green and it was rather hard to figure out what's going on.
So I guess with those VCR cassettes one would redo the recording and then look at the diff of the reference recording? is that easily understandable or does it require to review and understand thousands of lines of diff?

coolo:
Right now the diff is very noisy due to e.g. Date, X-Runtime or X-Request-ID headers that change. They can be filtered out - and I would really like to force auth to reduce all the pointless 401 requests (no matter if VCR or not), but @lethliel is on vacation to help me there.

PR #1974, opened 2019-05-02
PubSubConsumer: adopt to Pika 1.0 change

nilxam:


coolo:
this won't do, will append to it

PR #1975, opened 2019-05-02
Add CNAME for github pages serving.

jberry-suse:
Utilized for serving content of web directory.

Requires [poo#51002](https://progress.opensuse.org/issues/51002) or local entry.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+osrt.opensuse.org


coolo:
why not osr.too.ls ? :)

jberry-suse:
> why not osr.too.ls ? :)

Based operator and OBS server address on host domain (ie. `opensuse.org`). Otherwise either have to hard-code them or prompt the user in some fashion.

PR #1979, opened 2019-05-02
dist/package: handle python 2 vs python 3 dependencies.

jberry-suse:
- 0225a8e95537b976bbaf98d49248e6f53f96b6cb:
    dist/package: handle python 2 vs python 3 dependencies.
    
    Since osc was updated to require python 3 handle dependencies in
    same fashion as osc so as to get them right.

- be729c0bd143d273c7bb4a565a9ee281f8dbaccf:
    dist/package: require influxdb and telegraf for Tumbleweed.
    
    The packages have made there way into Tumbleweed and thus can now be
    required instead of suggested.

- 9214736f378e3d959ae5bcf625648e994432ccb8:
    dist/package: drop %check and related build requirements.
    
    Not feasible to run tests on OBS.

For Tumbleweed deployments, including kubernetes from container image, the pre-python 3 version of `osc` is packaged and used. If this PR is accepted that will break since it will depend on python 3 dependencies. If this state goes on for a while either this PR should not be accepted or a project level way of overriding to force python 2 will be needed to keep container building properly.

### DIFF
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -16,6 +16,17 @@
 #
 
 
+%if 0%{?suse_version} > 1500 || 0%{?sle_version} >= 150100
+%bcond_without python3
+%else
+%bcond_with    python3
+%endif
+%if %{with python3}
+%define use_python python3
+%else
+%define use_python python
+%endif
+
 %global __provides_exclude ^perl.*
 %define source_dir openSUSE-release-tools
 %define announcer_filename factory-package-news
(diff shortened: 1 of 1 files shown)


dist/package/openSUSE-release-tools.spec
```diff
@@ -16,6 +16,17 @@
 #
 
 
+%if 0%{?suse_version} > 1500 || 0%{?sle_version} >= 150100
```
coolo: I would prefer to get rid of python2 (and only test python3) - and require osc 0.165. But as an intermediate step, I'm fine with this.

jberry-suse: Sure, although keep in mind we package staging plugins and such for Leap 15.0 which has osc python 2.

coolo: Not in openSUSE:Tools:
https://build.opensuse.org/package/binary/openSUSE:Tools/osc/SLE_15/x86_64/osc-0.165.0-250.1.noarch.rpm

jberry-suse: Then the condition should be updated I suppose.

coolo: The problem is as soon as switch the interpreter of tools to /usr/bin/python3, the osclib requires both python version. And as I said: I rather get rid of python2 and say sorry for everyone with sle12/leap 42

jberry-suse: Not disagreeing with long-term plan, but the same applies to osc which is doing the split. Simply drop once all deployments are working on python3.

PR #1980, opened 2019-05-03
Add container for testing in leap 15.1

coolo:
The container will be built on OBS and be used on travis to avoid waiting
for the container to build (or to error because of download.opensuse.org madness)

### DIFF
--- a/dist/ci/testenv-leap151/osc-hotpatch.diff
+++ b/dist/ci/testenv-leap151/osc-hotpatch.diff
@@ -1,0 +1,11 @@
+--- /usr/lib/python3.6/site-packages/osc/core.py.bak    2019-05-02 18:14:01.088380882 +0000
++++ /usr/lib/python3.6/site-packages/osc/core.py        2019-05-02 18:14:19.204341426 +0000
+@@ -6318,7 +6318,7 @@
+         try:
+             comment = node.find('comment').text.encode(locale.getpreferredencoding(), 'replace')
+         except:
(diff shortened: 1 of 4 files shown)


dist/ci/testenv-leap151/run_as_tester
```diff
@@ -0,0 +1,7 @@
+#! /bin/sh
```
Vogtinator: Whitespace

coolo: It's perfectly fine with it and I find it easier to read - but I won't argue here

dist/ci/testenv-leap151/osc-hotpatch.diff
```diff
@@ -0,0 +1,11 @@
+--- /usr/lib/python3.6/site-packages/osc/core.py.bak    2019-05-02 18:14:01.088380882 +0000
++++ /usr/lib/python3.6/site-packages/osc/core.py        2019-05-02 18:14:19.204341426 +0000
+@@ -6318,7 +6318,7 @@
+         try:
+             comment = node.find('comment').text.encode(locale.getpreferredencoding(), 'replace')
+         except:
+-            comment = b'<no message>'
++            comment = '<no message>'
```
coolo: btw: the patch is most likely not correct, but it helps pass the test suite and that's all that counts.

coolo:
travis is busted - looks like bad mirror

PR #1981, opened 2019-05-03
Port to python3 and test in tumbleweed containers

coolo:
This is a rather drastic change (and contains a little VCR history I didn't want to squash).

There are several problems left to be tackled:
- document how to write/adopt tests
- streamline the multiple oscrcs we have now in tests
- @jberry-suse mentioned that some new functions in OBSLocal duplicate existing code, that needs to be refactored
- whatever I forgot

And then we need to decide how to move forward in general with python3. I would prefer if we switched all env lines to /usr/bin/python3 and require osc in python3 variant. osc on sle12 is still python2, but I think we can sacrifies that platform.


jberry-suse:
> (and contains a little VCR history I didn't want to squash).

That's what a separate branch is for locally. Committing history of bad/unused approaches is not a particularly sound thing to start doing. There would be ~10 times as many commits from me of an non-useful nature. As an utter minimum I assume it does not contain the enormous vcr dumps?

> Remove vcr usage - too controversial

Not sure that's the reason. The VCR approach is like testing a web application (perhaps OBS) and instead of bringing up a database I were to record all the query responses. Not only would there be massive data duplication, but if I simply re-ordered the fields selected in a query my dumps would have to be updated. The approach is extremely brittle and if I am recording them from an instance that one has to setup to be work properly with tests it only makes sense to just use it for testing directly.

jberry-suse:
At least going to rename the vcrhelpers and such before merging? It's one thing to have the history, another to wait until it works, but quite another to actually merge like that.

coolo:
I would have renamed it already - but I feel it should be merged with obslocal. So just shout your suggestions :)

And about the history squashing, I stopped squashing as I rewrote the tests while using vcr and I don't want to loose that. And no, the vcrs have always been in one git revision that was easy to drop.

coolo:
the coverage data is there now (even reported as only 25%, but as I had to realize that includes the tests) and is reported to coveralls - but then coveralls shows no files.

coolo:
https://coveralls.io/jobs/48083523 now contains coverage data - and still it's not reported here

coolo:
Leap 15.1 might not be released yet, but it seems too old for coveralls :(

coolo:
coverage reports work and vcrhelpers is history, I'm ready to merge, but I'm still unsure what to do with the shebangs. But I would prefer to do this in a followup - for the time being I can run the commands with python3 explicitly

PR #1982, opened 2019-05-03
userscript/origin.user.js: nice icons

lnussel:
spinner and such

### DIFF
--- a/userscript/origin.user.js
+++ b/userscript/origin.user.js
@@ -19,8 +19,16 @@
     var subdomain = domain_parent.endsWith('suse.de') ? 'tortuga' : 'operator';
     var url = 'https://' + subdomain + '.' + domain_parent + '/origin/package/' + project + '/' + package;
 
+    $('ul.clean_list, ul.list-unstyled').append('<li id="osrt-origin"><i class="fas fa-spinner fa-spin text-info"></i> loading origin...</li>');
     $.get({url: url, crossDomain: true, xhrFields: {withCredentials: true}, success: function(origin) {
-        if (origin.endsWith('failed')) return;
-        $('ul.clean_list, ul.list-unstyled').append('<li>Origin: <a href="/package/show/' + origin + '/' + package + '">' + origin + '</a>');
+        if (origin.endsWith('failed')) {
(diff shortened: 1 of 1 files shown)


userscript/origin.user.js
```diff
@@ -19,8 +19,16 @@
     var subdomain = domain_parent.endsWith('suse.de') ? 'tortuga' : 'operator';
     var url = 'https://' + subdomain + '.' + domain_parent + '/origin/package/' + project + '/' + package;
 
+    $('ul.clean_list, ul.list-unstyled').append('<li id="osrt-origin"><i class="fas fa-spinner fa-spin text-info"></i> loading origin...</li>');
```
jberry-suse: The icon does not work in old theme (which is the reason for the two selectors). Otherwise works properly. Given the old theme does not use an icon theme and uses cached icon URLs it is a bit cumbers to utilize from the outside like this. The new theme makes it trivial. I am fine with leaving like this, but just making you aware.

userscript/origin.user.js
```diff
@@ -19,8 +19,16 @@
...
     $.get({url: url, crossDomain: true, xhrFields: {withCredentials: true}, success: function(origin) {
-        if (origin.endsWith('failed')) return;
-        $('ul.clean_list, ul.list-unstyled').append('<li>Origin: <a href="/package/show/' + origin + '/' + package + '">' + origin + '</a>');
+        if (origin.endsWith('failed')) {
+            $('#osrt-origin').html('<i class="fas fa-bug text-warning"></i> failed to get origin info');
+        } else {
+            project = origin.trim();
+            if (project.endsWith('~')) {
+                project = project.slice(0, -1);
```
jberry-suse: This logic exists for the origin-manager interface as well. Once properly hosted my thought was to include the util.js (and expand) in userscripts with common stuff like this.

userscript/origin.user.js
jberry-suse: Would be an argument for adding back "origin: " prefix in the case if one cares about it. As otherwise it is rather context-less.

Old theme:
image

New theme:
image

jberry-suse:
A, presumably unintended, side-effect was that `failed to get origin info` now displays on every page even those not within origin manager configured projects. The error about no OriginConfig needs to be handled if going to display such a message.

PR #1995, opened 2019-05-10
travis: restore deployment functionality.

jberry-suse:
All other test suites were removed which makes the condition never met.


coolo:
The last auto deploy was after 1977 was merged, my changes hit later. I'll check - but triggering it from botmaster might be simple enough as well

jberry-suse:
I updated it manually as noted in the issue.

PR #1999, opened 2019-05-11
Remove more references to vcr

coolo:
I tried to replace the Mocking of functions with httpretty, but unfortunately httpretty replaces all other API calls with 404s then.

What we'd need is a library sniping out single requests - possibly by mocking the http_GET function, but that's tricky as well with caching already overwriting it.


coolo:
Merging so I can test the deployment - need a change in master for full test

PR #2000, opened 2019-05-11
Simplify travis setup even more: deploy in gocd

coolo:
-> https://build.opensuse.org/request/show/702118

I used a bot account for this, not to risk personal accounts.


jberry-suse:
Please use the throwaway account (after changing credentials since no longer needed) which is far more appropriate than mixing role with `staging-bot`. Turned a 3-4 line change into multiple hundreds and needlessly disconnected from the rest of our CI for absolutely no gain...

Alternatively, instead of `opensuse-releaseteam-travis` could make a new one (as I had thought about) named `opensuse-releaseteam`, but using `staging-bot` is just wrong.

coolo:
Feel free to create a new account. But I don't want to have a public password as maintainer in a devel project.

jberry-suse:
Read what I wrote, that is not what I suggested nor what existed before.

jberry-suse:
This problem existed somewhat before with the old repochecker user being overused and granted more and more privileges (it even superseded requests). I resolved it mostly by moving to more appropriate users, expect new pkglistgen just jump to new repo-checker user.

Now we have staging-bot being given maintainer of the package from which itself and tools are deployed which has nothing to do with staging process.

coolo:
Before it was using jberry user to deploy, not exactly perfect to me either.

jberry-suse:
Sure, but that user already needs and has access to push to package. Better than mixing. Frankly was tired of fighting with brain-dead OBS restrictions that prevent one from running `osc build` locally without an account.

If you were looking for less work running everything as is in the new container was a whole lot easier than reworking into gocd.

coolo:
Let's agree to disagree. Everything I can reproduce locally is a lot easier for me than everything I have to debug on travis-ci's infrastructure.

coolo:
and about the mulittude of accounts - I'm all for not mixing roles of bot accounts, there is already a lot of power to testtest-manager, repo-checker, factory-repo-checker and staging-bot.

PR #2002, opened 2019-05-11
Fix and test repository_path_expand on SLE12-SP5

coolo:
The way OBS collapses repositories is different to what we did

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from collections import OrderedDict
 from datetime import datetime
 from dateutil.parser import parse as date_parse
 import re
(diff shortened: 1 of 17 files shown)


osclib/core.py
```diff
@@ -359,11 +359,11 @@ def repository_path_expand(apiurl, project, repo, repos=None):
         # calls which effectively means the list grows even when new project.
         repos = []
 
-    if [project, repo] in repos:
-        # For some reason devel projects such as graphics include the same path
-        # twice for openSUSE:Factory/snapshot. Does not hurt anything, but
-        # cleaner not to include it twice.
-        return repos
+    for prev_project, prev_repo in repos:
+        # OBS only takes the first repository of a project into account, we should
```
coolo: @mlschroe is this what is done or did I guess wrongly?

PR #2009, opened 2019-05-13
suppkg_rebuild: Fix TypeError on python3

nilxam:
Convert all items to String in a List to fix TypeError when executing on python3.

File "suppkg_rebuild.py", line 42, in get_support_package_list
    if re.match('Preinstall', line) or re.match('VM[Ii]nstall', line) or re.match('Support', line):
  File "/usr/lib64/python3.6/re.py", line 172, in match
    return _compile(pattern, flags).match(string)
TypeError: cannot use a string pattern on a bytes-like object

### DIFF
--- a/suppkg_rebuild.py
+++ b/suppkg_rebuild.py
@@ -18,6 +18,7 @@ import osc.conf
 import osc.core
 
 from osc import oscerr
+from osc.util.helper import decode_list
 from osclib.conf import Config
 from osclib.stagingapi import StagingAPI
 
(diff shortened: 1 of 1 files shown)


suppkg_rebuild.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, project):
         self.api = StagingAPI(self.apiurl, self.project)
 
     def get_support_package_list(self, project, repository):
-        f = osc.core.get_buildconfig(self.apiurl, project, repository).splitlines()
+        f = [str(x) for x in osc.core.get_buildconfig(self.apiurl, project, repository).splitlines()]
```
coolo: no need to do this per line. Just split after converting to string. osc code itself wraps into ` from osc.util.helper import decode_it`

nilxam: or just using `decode_list`?

coolo: As you wish, I personally would go with `decode_it(get_buildconfig()).splitlines()` (pseudo code of course :)

PR #2013, opened 2019-05-13
Run staging-installcheck exclusively on python3

coolo:
It creates invalid reports with python2 and I see no point in debugging that


DimStar77:
:+1: Makes sense to move to the future

PR #2014, opened 2019-05-13
osclib/core: rework as package_list_kind_filtered().

jberry-suse:
- 9302407e33e86e516a02b78f583706e72d4dc2be:
    osclib/core: rework as package_list_kind_filtered().

- 0857bdfb9aaf6ec896b49f9b098119dc3aaceea9:
    osclib/core: provide package_kind().

Per [mailing list discussion](https://lists.opensuse.org/opensuse-buildservice/2019-05/msg00020.html) not going anywhere fast. As such better to have something to mostly works than mostly does not (for SLE and Leap :Update projects).

Still running a full set locally to see if anything crops up.

### DIFF
--- a/devel-project.py
+++ b/devel-project.py
@@ -16,7 +16,7 @@ from osclib.comments import CommentAPI
 from osclib.conf import Config
 from osclib.core import devel_project_fallback
 from osclib.core import entity_email
-from osclib.core import package_list_without_links
+from osclib.core import package_list_kind_filtered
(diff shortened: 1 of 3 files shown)


jberry-suse:
Needs to detect patchinfo packages.

jberry-suse:
Works well enough. Reduces SP5 package clutter from 19,095 "packages" to 3,965.

Issue #2022, opened 2019-05-14
staging accept python3 fail


lnussel:
File "/home/blah/.osc-plugins/osc-staging.py", line 518, in do_staging
    cmd.update_factory_version()
  File "/home/blah/.osc-plugins/osclib/accept_command.py", line 306, in update_factory_version
    new_product = re.sub(r'<version>\d{8}</version>', '<version>%s</version>' % curr_version, product_spec)
  File "/usr/lib64/python3.6/re.py", line 191, in sub
    return _compile(pattern, flags).sub(repl, string, count)
TypeError: cannot use a string pattern on a bytes-like object

PR #2023, opened 2019-05-14
Some TypeError fixes for python3

nilxam:
Fixes https://github.com/openSUSE/openSUSE-release-tools/issues/2022

Use `source_file_load()` to get product file. And make `source_file_load()` always return string.

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -301,8 +301,7 @@ class AcceptCommand(object):
         for product in products.findall('product'):
             product_name = product.get('name') + '.product'
             product_pkg = product.get('originpackage')
-            url = self.api.makeurl(['source', project, product_pkg,  product_name])
-            product_spec = http_GET(url).read()
+            product_spec = source_file_load(self.api.apiurl, project, product_pkg, product_name)
             new_product = re.sub(r'<version>\d{8}</version>', '<version>%s</version>' % curr_version, product_spec)
 
             if product_spec != new_product:
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -26,6 +26,7 @@ from osc.core import search
 from osc.core import show_package_meta
 from osc.core import show_project_meta
 from osc.core import show_results_meta
+from osc.util.helper import decode_it
 from osclib.conf import Config
 from osclib.memoize import memoize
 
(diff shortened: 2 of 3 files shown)


osclib/accept_command.py
```diff
@@ -254,7 +255,7 @@ def create_new_links(self, project, pkgname, oldspeclist):
 
         if len(filelist) > 1:
             # There is more than one .spec file in the package; link package containers as needed
-            origmeta = source_file_load(self.api.apiurl, project, pkgname, '_meta')
```
coolo: source_file_load is our own function - and I would say it always should return a `str`

nilxam: makes sense to me.

osclib/core.py
```diff
@@ -288,7 +289,7 @@ def project_pseudometa_file_load(apiurl, project, filename, revision=None):
     project, package = project_pseudometa_package(apiurl, project)
     source_file = source_file_load(apiurl, project, package, filename, revision)
     if source_file is not None:
-        source_file = source_file.decode('utf-8').rstrip()
+        source_file = source_file.encode('utf-8').decode('utf-8').rstrip()
```
jberry-suse: lol, python drunk?

nilxam:
travis is pending more than 12 hours?

osclib/core.py
nilxam: `str` in python3 is a Unicode object, `decode()` isn't an attribute of that object anyway, str->bytes->str is the way force it to converting the encoding. *However* in python3 the default encoding should be utf-8 if not defining `codeing` in the py file, so I think we should remove most of `decode()` cases in our code?

coolo: if source_file_load is now a str - then just don't do nothing, but rstrip() on it.

coolo: at least I can't follow the reasoning to have an encode() there

nilxam: because I'm not sure what is `decode('utf-8')` suppose to do before.

coolo: decode('utf-8') is converting byte to str - just the `decode_it` function does. It now breaks as you change the returned class of source_file_load, so removing it is the right thing to do

nilxam: yes, as I explained above. if `'utf-8'` is not a problem here.

nilxam: changed. if it has non-utf8 character breaks things we can revisit it here.

coolo: that's for decode_it to handle :)

nilxam: yeah, but my understanding was deocde_it isn't mandatory to return a utf8 decoding. :)

jberry-suse:
With the usage of `from osc.util.helper import decode_it` we are now unable to use pre-python3-switched osc :( Making a real mess of this transition.

PR #2024, opened 2019-05-15
check_tags_in_requests: issue state is deleted not removed.

### LINKED ISSUE #1991: changelog-checker, reject submission if we loose bug, fate or JIRA reference
As a Release Manager, I have to spend a lot of time on the submission (fixing dependency issues via _product changes, grouping on staging) before we figure out that changelog lost a bug or fate reference.

Autobuild prefers to reject these anyway, so I believe it's safe to have such rule in place

jberry-suse:
- 655bac280fe44e6c01cc28ef4227415c8997068b:
    check_tags_in_requests: escalate loss of issue references to a decline.

- 45e6b617e10bf627c191f56dc7edec7921ce5437:
    check_tags_in_requests: make issue singular since references is plural.

- a309e1c85b680a99a968987918ed5440c8775d61:
    check_tags_in_requests: issue state is deleted not removed.
    
    See #679, where it was originally added and provides example output.

Fixes #1991.

### DIFF
--- a/check_tags_in_requests.py
+++ b/check_tags_in_requests.py
@@ -84,7 +84,7 @@ by OBS on which this bot relies.
             self.logger.debug("reject: diff contains no tags")
             return False
         if deleted > 0:
-            self.review_messages['accepted'] = 'Warning: {} issues reference(s) deleted'.format(deleted)
+            self.review_messages['accepted'] = 'Warning: {} issue reference(s) deleted'.format(deleted)
             return True
         return True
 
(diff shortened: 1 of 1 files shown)


coolo:
Let's merge, but as I wrote in the issue I'm sceptical it works in general.

jberry-suse:
It did not work, note the change from `removed` to `deleted` and the example output from original issue that I verified on current request.

coolo:
right. So let's see how quickly the rms want this to be gone as it's too noisy - my expectations on reliable output from obs are low

PR #2032, opened 2019-05-16
origin-manager: port and switch to python3.

jberry-suse:
- 69f3fd13c8f48bbe74d635f868139e79737dbece:
    origin-manager: switch to python3.

- ceb4a9cea9598659a1970d57a367c9e4fd7b7d14:
    osclib/origin: cast filter() to list() to improve debug output [python3].

- 743b3a6cd63ce51b891a53ee39f6739b5d417282:
    osc-origin, osclib/origin: port first dict key/value access [python3].

- 7a1ff946cd2abf5cc4ca2873dd2553eacc739ac5:
    osclib/util: sha1_short(): encode data as utf-8 [python3].

### DIFF
--- a/osclib/util.py
+++ b/osclib/util.py
@@ -159,4 +159,4 @@ def sha1_short(data):
     if isinstance(data, list):
         data = '::'.join(data)
 
-    return hashlib.sha1(data).hexdigest()[:7]
+    return hashlib.sha1(data.encode('utf-8')).hexdigest()[:7]
--- a/origin-manager.py
+++ b/origin-manager.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
 from osclib.core import package_source_hash
 from osclib.origin import origin_annotation_dump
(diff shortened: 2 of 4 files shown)


jberry-suse:
The operator server, used for origin manager web interface, was always python 3. ;)

osclib/util.py
```diff
@@ -159,4 +159,4 @@ def sha1_short(data):
     if isinstance(data, list):
         data = '::'.join(data)
 
-    return hashlib.sha1(data).hexdigest()[:7]
+    return hashlib.sha1(data.encode('utf-8')).hexdigest()[:7]
```
coolo: this breaks all usages passing bytes into it :(

jberry-suse: hmm, really makes this function cumbersome to use as most of the data I want to feed into it originates from osc API call and selected via lxml xpath query.

```python
sha1_short(root.xpath('//foo/@attr'))
```

If there is a way to handle that without manually having to encode each entry that would be preferred. I find this whole game rather cumbersome considering it Just Worked (tm) in python 2.

coolo: my advice: get over it, python3 is reality.

But I fixed the function to work on both meanwhile

jberry-suse: > my advice: get over it, python3 is reality.

not questioned at all...fix looks good.

PR #2039, opened 2019-05-16
Fix check_source with python3

coolo:


check_source.py
```diff
@@ -149,9 +150,9 @@ def check_source_submission(self, source_project, source_package, source_revisio
         civs += 'LC_ALL=C perl %s _old %s 2>&1' % (source_checker, target_package)
         p = subprocess.Popen(civs, shell=True, stdout=subprocess.PIPE, close_fds=True)
         ret = os.waitpid(p.pid, 0)[1]
-        checked = p.stdout.readlines()
+        checked = decode_list(p.stdout.readlines())
 
-        output = '  '.join(checked).translate(None, '\033')
+        output = '  '.join(checked).replace('\033', '')
```
nilxam: personally I'd using static method `.translate(str.maketrans('','','\033'))` though, but changed to `replaced()` from `translate()` also ok. :)

check_source.py
```diff
@@ -149,9 +150,9 @@ def check_source_submission(self, source_project, source_package, source_revisio
         civs += 'LC_ALL=C perl %s _old %s 2>&1' % (source_checker, target_package)
         p = subprocess.Popen(civs, shell=True, stdout=subprocess.PIPE, close_fds=True)
         ret = os.waitpid(p.pid, 0)[1]
-        checked = p.stdout.readlines()
+        checked = decode_list(p.stdout.readlines())
```
nilxam: `decode_list` hasn't imported.

DimStar77: See line 16: ``` from osc.util.helper import decode_list```

check_source.py
coolo: I wondered what translate actually does - and I would even be more confused with the maketrans version. Because it's not translated, it's plainly removed. So I took the replace version :)

check_source.py
nilxam: Oh, I'm overlooked. My bad. :)

check_source.py
nilxam: yeah, `maketrans` is `(in, out, mapped to None)` :)

PR #2052, opened 2019-05-23
.gitignore: ignore osc symlink commonly used to develop against git osc.

jberry-suse:
While going through never merged branched and stash to get small origin-manager follow-ups I decided to add a bit for developing against git osc since it became especially relevant again with python 3 porting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,18 @@ If working on an `osc` plugin create symlinks for the plugin and `osclib` in eit
 
     ln -sr ./osc-staging.py ./osclib ~/.osc-plugins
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -42,6 +42,18 @@ If working on an `osc` plugin create symlinks for the plugin and `osclib` in eit
 
     ln -sr ./osc-staging.py ./osclib ~/.osc-plugins
 
+It can also be useful to work against a development copy of `osc` either to utilize new features or to debug/fix functionality. To do so one must place the development copy in the path to be loaded and utilize the wrapper script if working on `osc` plugins. One method to accomplish this is shown below.
+
+    # outside of openSUSE-release-tools checkout
+    git clone git clone https://github.com/openSUSE/osc.git
```
coolo: you only need gt clone once

PR #2054, opened 2019-05-24
accept_command: fix missing the url for uploading product file change

nilxam:
Regression from acc0ffc3852700cc996109f350f72dcb9ff1b3fe , we still need url for uploading product file change. That's why my testing on Leap works but now it's a problem on TW.

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -306,6 +306,7 @@ class AcceptCommand(object):
 
             if product_spec != new_product:
                 update_version_attr = True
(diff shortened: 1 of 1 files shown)


nilxam:
BTW DimStar reminds me maybe it's time to stop update version number to product file from now, just let pkglistgen to do that per the version attribute. Let's do that?

coolo:
I would have done it much earlier - your concern was the OBS project dashboard. But if we change the release process between Factory, ToTest and Tumbleweed, it will be broken anyway :(

PR #2070, opened 2019-05-28
Run buid fail reminder on botmaster

coolo:
The timer is stopped on packagelists and the data migrated. The service only ran for openSUSE:Factory atm (leap 15.1 was never enabled it seems)


okurz:
typo, s/buid/build/

PR #2079, opened 2019-05-29
Run suppkg_rebuild.py on gocd

coolo:


suppkg_rebuild.py
```diff
@@ -1,13 +1,9 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
nilxam: yes, this is on my TODO by the time we move it away from packagelists. Thanks.

PR #2081, opened 2019-05-29
Check for duplicated binaries as part of the staging projects

coolo:


coolo:
@DimStar77 set the current problems as whitelist into the config

PR #2093, opened 2019-05-31
Run factory announcer on gocd

coolo:
Ported the announcer to python3 and made the config parsing a little easier. The current state is no longer stored as filename, but in a config file


coolo:
I sent test mails to myself and they look good

coolo:
(and botmaster is prepared not to send 0527 again :)

PR #2098, opened 2019-06-01
Run origin manager reviews on gocd as well

coolo:


coolo:
By this packagelists migration is done

PR #2099, opened 2019-06-02
Fix inner-staging links to use cicount=copy

coolo:
We need to keep the cicount in sync for the kernel as kernel-syms requires kernel-default of kernel-source


coolo:
fixing sp5

PR #2107, opened 2019-06-05
Split repo_checker.py into 3 tools

coolo:
Split repo_checker.py into 3:

- staging-installcheck.py (so far sharing tons of code by means of copy&paste)
- maintenance-installcheck.py (Reviewbot)
- project-installcheck.py

The later will no longer comment into packages, for now it just updates the meta file - that is no longer read though. This is task for a followup PR


coolo:
clarified with @msmeissn that the maintenance incidents are declined (after disabling checks on 42.3/i586 per OSRT:Config)

clarified with @DimStar77 that adi projects no longer care about existing problems.

staging-installcheck.py
```diff
@@ -182,7 +153,7 @@ def staging(self, project, force=False):
             return True
 
         repository_pairs = repository_path_expand(api.apiurl, project, repository)
-        staging_pair = [project, repository]
+        [project, repository]
```
nilxam: what is this suppose to do?

coolo: that's what happens if you let a tool remove unused variables - it creates unused statements :)

I'll remove

PR #2110, opened 2019-06-12
Replace rebuildpacs.pl with project-installcheck functionality

coolo:


coolo:
https://transfer.sh/13Xvt8/state.yaml is currently a local file, but I will change it to store in OBS. I removed followup errors and mapped the binaries to the source containers (and stripped rebuild counters).

Also failed packages are excluded, so it's a good match to the build fail reminder. I plan to obsolete the project_only file with this file (I'm not really sure we need detailed history of it, it changes a lot)

DimStar77:
just FTR: rebuildpacs.pl was/is also used in various devel projects. I hope we won't lose this capability

coolo:
I'm aware - and it's the reason why I didn't put the state into staging dashboard right away. I guess I will put it in my favorite home:repochecker storage :)

project-installcheck.py
```diff
@@ -97,22 +125,193 @@ def project_repository(self, project):
...
+        for source in list(oldstate['check']):
+            if not source.startswith('{}/{}/{}/'.format(project, repository, arch)):
+                continue
+            if not os.path.basename(source) in succeeding:
+                continue
+            if source not in per_source:
+                self.logger.info("No known problem, erasing %s", source)
+                del oldstate['check'][source]
+
+        packages = ['branding-openSUSE', 'PackageKit-branding-openSUSE', 'xfce4-branding-openSUSE',
```
DimStar77: Should we make this into a configuration inside OBS/project as well? From the looks of it, I'd miss at least installation-images:MicroOS right now

coolo: Yes, we should.

coolo: Feel free to put it into osrt:config already for later pickup. But remember to do it for the ports as well

PR #2111, opened 2019-06-13
Add MicroOS announcer instance

sysrich:


gocd/announcer.gocd.yaml
```diff
@@ -36,3 +36,21 @@ pipelines:
...
+  Announcer.MicroOS:
+    group: openSUSE.Checkers
+    lock_behavior: unlockWhenFinished
+    environment_variables:
+      OSC_CONFIG: /home/go/config/oscrc-staging-bot
+    materials:
+      script:
+        git: https://github.com/openSUSE/openSUSE-release-tools.git
+    timer:
+      spec: 0 0 * ? * *
```
coolo: please run at a different time (also for kubic) - as the announcers access the same state file, we shouldn't risk too much overwriting the file. I can also copy away the state file 3 times to erase this problem.

sysrich: Done, what do you think?

PR #2113, opened 2019-07-04
Add manager_42 configuration for Package Hub 15 SP1

ismail:


coolo:
manager_42 is going to die soonish, but for now we can take it

CC @jberry-suse

jberry-suse:
What was the goal here? If I remember correctly manager_42 only submits updates and since 15.1 is locked there should be none. It does not do a diff and submit if a difference which perhaps is the intention.

ismail:
> What was the goal here? If I remember correctly manager_42 only submits updates and since 15.1 is locked there should be none. It does not do a diff and submit if a difference which perhaps is the intention.

This was for Package Hub only. Coolo mentioned manager_42 going away, what would be the replacement to keep 00Meta up to date?

jberry-suse:
The replacement will be a new update crawler as the origin manager code-base and workflow already handles the equivalent of `00Meta/lookup.yml`.

The question of course is that only one project is configured as a potential origin and that project is locked. Effectively, either matches or does not in either case manager_42 does nothing.

PR #2119, opened 2019-07-08
rebuildpacs: ignore issues of petsc:openmpi

DimStar77:
A further workaround for https://github.com/openSUSE/openSUSE-release-tools/issues/1222


jberry-suse:
It would seem this list should be in the remote config and not hard-coded.

coolo:
no, the whole script should be dropped - WIP.

jberry-suse:
5 months is a bit of time to leave it like this considering I already had things piped from remote config into perl portion. Hopefully, it's resolved.

PR #2121, opened 2019-07-09
gocd: add origin-manager to SLE checkers.

jberry-suse:


coolo:
I didn't spot it either, but if you touch the .erb files, you need to run generate.sh to 'render' them

coolo:
pushed to master

PR #2125, opened 2019-07-11
dist: package /etc/grafana with the same mode as grafana

DimStar77:


DimStar77:
This will address the file conflicts in TW:

```
found conflict of ceph-grafana-dashboards-14.2.1.468+g994fd9e0cc-1.2.noarch with openSUSE-release-tools-metrics-20190613.dd9c9d45-2.1.noarch
  /etc/grafana [mode mismatch: d755 root:root, d750 root:grafana]

found conflict of grafana-6.2.5-3.1.x86_64 with openSUSE-release-tools-metrics-20190613.dd9c9d45-2.1.noarch
  /etc/grafana [mode mismatch: d755 root:root, d750 root:grafana]
```

i.e. grafana and ceph-grafana-dashboards both bring /etc/grafana as 755 root:root - only we don't

PR #2127, opened 2019-07-12
fix SLE15 SP2 pkglist generator pipeline

fcrozat:


gocd/sp2.target.gocd.yaml
```diff
@@ -23,19 +23,16 @@ pipelines:
     group: SLE15.SP2.Target
     lock_behavior: unlockWhenFinished
     materials:
-      repos:
-        git: git://botmaster.suse.de/suse-repos.git
-        auto_update: true
-        whitelist:
-          - SUSE:SLE-15-SP2:GA_-_standard.yaml
```
coolo: you can't remove the suse-repos. That's the trigger of the whole pipeline. you need to have multiple materials - and if you do that you also need destination directories for both.

coolo: I'll push into your branch

PR #2132, opened 2019-07-18
gocd: update 15.1 jobs to 15.2.

jberry-suse:
More related to [poo#52094](https://progress.opensuse.org/issues/52094).


coolo:
And to your question: botmaster pulls the config directly from git. But Ludwig paused all the pipelines touching 15.1 - unless they were renamed by this (some contain 15.1 in the name and you changed them), they will stay paused

coolo:
but everyone with the openSUSE role can unpause

coolo:
unpaused all pipelines

PR #2134, opened 2019-07-19
rabbit-openqa: Catch exceptions as PUT to openqa might result in 404

coolo:


gocd/rabbit-openqa.py
```diff
@@ -209,7 +209,11 @@ def on_published_repo(self, payload):
         # notify openQA to sync the projects - the plugin will check itself it
         # the project is to be synced. For now we notify about every 'images' repo
         if payload['repo'] == 'images':
-            self.openqa.openqa_request('PUT', 'obs_rsync', str(payload['project']), 'runs')
+            try:
+                self.openqa.openqa_request('PUT', 'obs_rsync/{}/runs'.format(payload['project']), retries=0)
+            except RequestError as e:
+                self.logger.info("Got exception on syncing repository: {}".format(e))
+                pass
```
jberry-suse: `pass` is a noop in python and does not mean "pass on the exception" or somesuch. Unless raised in the except portion the exception is consumed.

coolo: I didn't mean anything but noop. I just left it there in case I want to remove the logger statement later

coolo: removed the pass

PR #2143, opened 2019-08-05
osclib/origin: always include project link when fetching source history.

jberry-suse:
This was debated as for SLE this would add some repetition when not
going to match a SLE origin anyway, but it is necessary to stabilize
maintenance origins.

When an :Update project is first created all sources are inherited which
means all the revisions from the top project's source container are
considered. Once an update is provided via a maintenance incident the
inherited sources are no longer presented and instead the only revision
seen is the maintenance update and future updates.

When a new version of Leap is created and all source containers are copied
from the previous version the origin will first be considered :Update of
the previous Leap version, but once a maintenance update is created it
will drop the the non-:Update prior version since the matching revision is
missing. If the update is submitted the origin will flip back to :Update.

In order to avoid this mess always utilize include_project_link=True
which will effective consider the maintenance update as the newest
revision while continuing to consider the inherited revisions.

For SLE, were all projects are stacked this means there will almost always
be 10 revisions to review after the first could releases.

Realistically this is yet another failing of the OBS source control model since
the original source revision available (via inheritance) in the :Update project is
lost after the first update.

An example, [request 720684](https://build.opensuse.org/request/show/720684).

### DIFF
--- a/osclib/origin.py
+++ b/osclib/origin.py
@@ -238,7 +238,8 @@ def origin_find(apiurl, target_project, package, source_hash=None, current=False
     return None
 
 def project_source_contain(apiurl, project, package, source_hash):
-    for source_hash_consider in package_source_hash_history(apiurl, project, package):
+    for source_hash_consider in package_source_hash_history(
+        apiurl, project, package, include_project_link=True):
         project_source_log('contain', project, source_hash_consider, source_hash)
         if source_hash_consider == source_hash:
             return True
@@ -285,7 +286,8 @@ def origin_find_fallback(apiurl, target_project, package, source_hash, user):
 
     # Attempt to find a revision of target package that matches an origin.
     first = True
-    for source_hash_consider in package_source_hash_history(apiurl, target_project, package):
+    for source_hash_consider in package_source_hash_history(
+        apiurl, target_project, package, include_project_link=True):
         if first:
             first = False
             continue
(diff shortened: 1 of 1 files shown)


jberry-suse:
# Before

```
$ osc origin -p openSUSE:Leap:15.2 package --debug yast2
[D] origin_find: openSUSE:Leap:15.2/yast2 with source 808ded7 (True, True, True)
[D] source_contain: SUSE:SLE-15-SP1:Update                   6ab9af9 == 808ded7
[D] source_contain: SUSE:SLE-15-SP1:Update                   0fa6eb2 == 808ded7
[D] source_contain: openSUSE:Leap:15.1:Update                6ab9af9 == 808ded7
[D] source_contain: openSUSE:Leap:15.1:Update                0fa6eb2 == 808ded7
[D] source_contain: openSUSE:Leap:15.1                       808ded7 == 808ded7 (match)
openSUSE:Leap:15.1
```

```
$ osc origin -p openSUSE:Leap:15.2 package --debug yast2-packager
[D] origin_find: openSUSE:Leap:15.2/yast2-packager with source 4e9ee55 (True, True, True)
[D] source_contain: SUSE:SLE-15-SP1:Update                   4e8da0f == 4e9ee55
[D] source_contain: openSUSE:Leap:15.1:Update                4e9ee55 == 4e9ee55 (match)
openSUSE:Leap:15.1:Update
```

# After

```
$ osc origin -p openSUSE:Leap:15.2 package --debug yast2
[D] origin_find: openSUSE:Leap:15.2/yast2 with source 808ded7 (True, True, True)
[D] source_contain: SUSE:SLE-15-SP1:Update                   6ab9af9 == 808ded7
[D] source_contain: SUSE:SLE-15-SP1:Update                   0fa6eb2 == 808ded7
[D] source_contain: SUSE:SLE-15-SP1:Update                   808ded7 == 808ded7 (match)
SUSE:SLE-15-SP1:Update
```

# Current

One can see the mix due to this as maintenance updates come out.

image

Once tracking a maintenance version it would remain on them which was less noticeable last cycle since it was late in the cycle.

nilxam:
Makes sense to me.

PR #2149, opened 2019-08-08
origin-manager: switch away from manager_42 and drop old configs

jberry-suse:
- cd6c00e4b6616cd3c2b0d3eb06982e4c34cc5239:
    config/manager_42: drop openSUSE:Leap:15.2 config.

- c8d7c6c8db40bbc41a98db5982858b242226071b:
    config/manager_42: drop SUSE:SLE-15-SP2:GA config.

- 29fb8fa29d79b42bd27221b2841830385e02d13b:
    gocd/checkers.suse: drop SLE-SP2.Manager42 job.

- 91b991e95906ded3b22337e6ecaed3153374dee5:
    gocd/checkers.suse: add S15.SP2.Origin.Manager.Update job.

Follow-up to #2145 which will utilize the update command for the two active products previously using manager_42 and update crawler. Additionally, remove manager_42 configs and gocd jobs.


jberry-suse:
Wait for #2150 to hopefully resolve Leap job in order to be hopeful about SLE job setup.

PR #2150, opened 2019-08-08
gocd/checkers.opensuse: wrap OS.Origin.Manager.Update with osc-origin install.

jberry-suse:
This is very ugly, but following the pattern laid forth.


jberry-suse:
Working: https://build.opensuse.org/request/show/721880

PR #2153, opened 2019-08-13
check_source: add a review for the security team if there are warnings about missing

jsegitz:
We want to turn some rpmlint errors into warnings and make them only mandatory in official project. We'll handle this via a review. Add logic to add reviews if warnings for missing whitelist entries are present in submits

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -18,6 +18,8 @@ from osclib.conf import Config
 from osclib.core import devel_project_get
 from osclib.core import devel_project_fallback
 from osclib.core import group_members
+from osclib.core import source_file_load
+from osclib.core import target_archs
 from urllib.error import HTTPError
 
 import ReviewBot
@@ -26,6 +28,8 @@ from osclib.conf import str2bool
 class CheckSource(ReviewBot.ReviewBot):
 
     SCRIPT_PATH = os.path.dirname(os.path.realpath(__file__))
+    AUDIT_BUG_URL = "https://en.opensuse.org/openSUSE:Package_security_guidelines#audit_bugs"
+    AUDIT_BUG_MESSAGE = "The package is submitted to an official product and it has warnings that indicate that it need to go through a security review. Those warnings can only be ignored in devel projects. For more information please read: {}.".format(AUDIT_BUG_URL)
 
     def __init__(self, *args, **kwargs):
         ReviewBot.ReviewBot.__init__(self, *args, **kwargs)
@@ -48,6 +52,8 @@ class CheckSource(ReviewBot.ReviewBot):
         self.repo_checker = config.get('repo-checker')
         self.devel_whitelist = config.get('devel-whitelist', '').split()
         self.skip_add_reviews = False
+        self.security_review_team = config.get('security-review-team', 'security-team')
+        self.bad_rpmlint_entries = config.get('bad-rpmlint-entries', '').split()
 
         if self.action.type == 'maintenance_incident':
             # The workflow effectively enforces the names to match and the
@@ -187,7 +193,50 @@ class CheckSource(ReviewBot.ReviewBot):
             elif self.repo_checker is not None:
                 self.add_review(self.request, by_user=self.repo_checker, msg='Please review build success')
 
+        if self.bad_rpmlint_entries:
+            if self.has_whitelist_warnings(source_project, source_package, target_project, target_package):
+                # if there are any add a review for the security team
+                # maybe add the found warnings to the message for the review
+                self.add_review(self.request, by_group=self.security_review_team, msg=CheckSource.AUDIT_BUG_MESSAGE)
+            if self.suppresses_whitelist_warnings( source_project, source_package ):
+                self.add_review(self.request, by_group=self.security_review_team, msg=CheckSource.AUDIT_BUG_MESSAGE)
+
         return True
+
+    def suppresses_whitelist_warnings( self, source_project, source_package):
+        # checks if there's a rpmlintrc that suppresses warnings that we check
+        found_entries = set()
+        contents = source_file_load(self.apiurl, source_project, source_package, source_package+'-rpmlintrc')
+        if contents:
+            matches = re.findall(r'addFilter\(["\']([^"\']+)["\']\)', contents)
+            for entry in self.bad_rpmlint_entries:
+                for match in matches:
+                    if match.startswith(entry):
+                        self.logger.info(f'found suppressed whitelist warning: {entry}')
+                        found_entries.add(entry)
+        return found_entries
+
+    def has_whitelist_warnings( self, source_project, source_package, target_project, target_package ):
+        # this checks if this is a submit to an product project and it has warnings for non-whitelisted permissions/files
+        found_entries = set()
+        url = osc.core.makeurl(self.apiurl, ['build', target_project])
+        xml = ET.parse(osc.core.http_GET(url)).getroot()
+        for f in xml.findall('entry'):
+            # we check all repos in the source project for errors that exist in the target project
+            repo = f.attrib['name']
+            query = { 'last' : 1, }
+            for arch in target_archs(self.apiurl, source_project, repo):
+                url = osc.core.makeurl(self.apiurl, ['build', source_project, repo, arch, source_package, '_log'], query = query)
+                try:
+                    result = osc.core.http_GET(url)
+                    contents = str(result.read())
+                    for entry in self.bad_rpmlint_entries:
+                        if (': W: '+entry in contents) and not (entry in found_entries):
+                            self.logger.info(f'found missing whitelist for warning: {entry}')
+                            found_entries.add(entry)
+                except HTTPError as e:
+                    self.logger.error('ERROR in URL %s [%s]' % (url, e))
+        return found_entries
 
     def is_devel_project(self, source_project, target_project):
         if source_project in self.devel_whitelist:


check_source.py
```diff
@@ -187,8 +193,60 @@ def check_source_submission(self, source_project, source_package, source_revisio
             elif self.repo_checker is not None:
                 self.add_review(self.request, by_user=self.repo_checker, msg='Please review build success')
 
+        if self.has_whitelist_warnings(source_project, source_package, target_project, target_package):
+            # if there are any add a review for the security team
+            # maybe add the found warnings to the message for the review
+            self.add_review(self.request, by_group='security-team', msg="The package is submitted to an official product and it has warnings that indicate that it need to go through a security review. Those warnings can only be ignored in devel projects. For more information please read: {}.".format(CheckSource.AUDIT_BUG_URL))
```
jberry-suse: `security-team` does not exist on IBS so this will cause crashes there and based on your use-case you likely want it disable on IBS. In general everything should be configurable via the configuration tooling already available. Add the security team review to `target_project_config()` following the convention there. Additionally, an option similar to `check-source-add-review-team` should be provided with the default set to false for Leap in `osclib/conf.py` (and we'll need to make sure both active SLE targets remote configs are update since they do not use defaults.

check_source.py
```diff
@@ -48,6 +49,11 @@ def target_project_config(self, project):
         self.repo_checker = config.get('repo-checker')
         self.devel_whitelist = config.get('devel-whitelist', '').split()
         self.skip_add_reviews = False
+        self.bad_rpmlint_entries = [
+            'polkit-unauthorized-',
+            'polkit-untracked-',
+            'permissions-',
+        ]
```
jberry-suse: If these are not configurable they should be a constant. It seems like they auto to just be configurable as Factory will be the only user and thus the whole list can be managed in Factory remote config (`OSRT:Config` attribute on `openSUSE:Factory`).

check_source.py
```diff
@@ -187,8 +193,60 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+        try:
+            result = osc.core.http_GET(url)
+            contents = str(result.read())
+            for entry in self.bad_rpmlint_entries:
+                if re.search('addFilter\(["\']'+entry,contents):
+                    self.logger.info(f'found suppressed whitelist warning {entry}')
+                    found_entries.append(entry)
+            return found_entries
+        except HTTPError as e:
+            # probably doesn't have a rpmlintrc
```
jberry-suse: No need for "probably" lets be sure.

```python
if e.code != 404:
    raise e
```

check_source.py
```diff
@@ -187,8 +193,60 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+            pass
+        return False
+
+    def has_whitelist_warnings( self, source_project, source_package, target_project, target_package ):
+        # this checks if this is a submit to an product project and it has warnings for non-whitelisted permissions/files
+        found_entries = []
+        # we only want this check in official projects
+        if target_project.startswith('openSUSE:') or target_project.startswith('SUSE:'):
```
jberry-suse: No need to hard-code prefixes. Based on feedback above if the option that controls adding the security review is enabled then go ahead and check otherwise do nothing.

check_source.py
```diff
@@ -187,8 +193,60 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+            url = osc.core.makeurl(self.apiurl, ['build', target_project])
+            xml = ET.parse(osc.core.http_GET(url)).getroot()
+            for f in xml.findall('entry'):
+                # we check all repos in the source project for error sthat exist in the target project
+                repo = f.attrib['name']
+                query = { 'last' : 1, }
+                # FIXME: Hardcoded architecture?
+                url = osc.core.makeurl(self.apiurl, ['build', source_project, repo, 'x86_64', source_package,'_log'], query = query)
```
jberry-suse: Could utilize `staging-archs` from config or `osclib.target_archs()` and pick or loop until a log found.

check_source.py
```diff
@@ -187,8 +193,60 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+                try:
+                    result = osc.core.http_GET(url)
+                    contents = str(result.read())
+                    for entry in self.bad_rpmlint_entries:
+                        # FIXME change that to W once we change it to a warning
+                        if ': E: '+entry in contents:
+                            self.logger.info(f'found missing whitelist warning {entry}')
+                            found_entries.append(entry)
+                except HTTPError as e:
+                    # the target might have repos that the source doesn't have, we'll ignore it
```
jberry-suse: This should be resolved by using `osclib.target_archs()` before calls against `source_project`.

check_source.py
```diff
@@ -187,8 +193,60 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+                # we check all repos in the source project for error sthat exist in the target project
+                repo = f.attrib['name']
+                query = { 'last' : 1, }
+                # FIXME: Hardcoded architecture?
+                url = osc.core.makeurl(self.apiurl, ['build', source_project, repo, 'x86_64', source_package,'_log'], query = query)
+                try:
+                    result = osc.core.http_GET(url)
+                    contents = str(result.read())
+                    for entry in self.bad_rpmlint_entries:
+                        # FIXME change that to W once we change it to a warning
```
jberry-suse: As in you will change to warning once this tool is deployed? If so then just change it in this code now and once merged it is auto-deployed and it can be changed in rpmlint.

jberry-suse:
Do you have an example request that can be used for local testing?

check_source.py
jberry-suse: Also as suggested by @lnussel this should probably also handle `.*` filters were possible.

jsegitz:
Thanks for the notes, I'll work through them.

I used sr#723033 for testing (but I modified the if condition that test that the sumbit goes to (open)SUSE to make this work

check_source.py
jsegitz: I had it in there to facilitate testing, will change it now. You would need to change it back if you want to test it with the submit mentioned below

check_source.py
jsegitz: No we want this in IBS and OBS. I was suprised to learn that we don't have the same group setup in IBS and will have this changed so we have the group in both instances

jsegitz:
I've added all the changes that were suggested, please have a look.

check_source.py
```diff
@@ -187,8 +191,59 @@ def check_source_submission(self, source_project, source_package, source_revisio
             elif self.repo_checker is not None:
                 self.add_review(self.request, by_user=self.repo_checker, msg='Please review build success')
 
+        if self.bad_rpmlint_entries:
+            # we only want this check in official projects
```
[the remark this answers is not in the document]
jsegitz: right, removed

check_source.py
```diff
@@ -187,8 +191,59 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+            if self.has_whitelist_warnings(source_project, source_package, target_project, target_package):
+                # if there are any add a review for the security team
+                # maybe add the found warnings to the message for the review
+                self.add_review(self.request, by_group=self.security_review_team, msg=CheckSource.AUDIT_BUG_MESSAGE)
+            if self.suppresses_whitelist_warnings( source_project, source_package ):
```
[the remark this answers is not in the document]
jsegitz: hm, I'm hesitant here. It should be rare that the review is added twice and I want to express that both are issues we want to be notified about. They don't really depend on each other, so I don't want to chain them together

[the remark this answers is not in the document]
jberry-suse: Fair enough, previously without the check stuff like this would cause periodic crashes, but that's fine.

check_source.py
```diff
@@ -187,8 +191,59 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+        url = osc.core.makeurl(self.apiurl, ['source', source_project, source_package, source_package+'-rpmlintrc'])
+        found_entries = []
+        try:
+            result = osc.core.http_GET(url)
+            contents = str(result.read(), encoding='utf-8', errors='replace')
+            if re.search(r'addFilter\(["\']\.[\+\*]', contents):
+                self.logger.info(f'suppresses all warnings/errors in rpmlintrc')
+                found_entries.append('suppresses all warnings/errors')
+            for entry in self.bad_rpmlint_entries:
+                if re.search(r'addFilter\(["\']'+entry, contents):
```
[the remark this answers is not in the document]
jberry-suse: As in matching all `addFilter` entries via a regex and then intersecting the set against the bad list or "starts with" anything in the bad list. That way no need to run multiple regex matches nor treat remote config (user input) as a regex.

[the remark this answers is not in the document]
jsegitz: Removing the REs is a good idea, but I want to keep the ability to match on prefixes. That's a bit longer, but it allows us to remove the special case for the 'matches everything rule' with
bad-rpmlint-entries: polkit-unauthorized- polkit-untracked- permissions- .* .+
and
```python
matches = re.findall(r'addFilter\(["\']([^"\']+)["\']\)', contents)
    for entry in self.bad_rpmlint_entries:
        for match in matches:
            if match.startswith(entry):
                self.logger.info(f'found suppressed whitelist warning: {entry}')
                found_entries.add(entry)
```

DimStar77:
> We want to turn some rpmlint errors into warnings and make them only mandatory in official project. We'll handle this via a review. Add logic to add reviews if warnings for missing whitelist entries are present in submits

Isn;t that the difference between rpmlint-Factory and rpmlint-Factory-strict? -strict is only used in openSUSE:* - one specific sample therein is the license check, which is a hard-fail on openSUSE:*, but only a warning in all other projects

jsegitz:
Seems like this is the only example. That would also have been a way but I like the review solution more because it allows us to approve submits even if rpmlint isn't happy with it.

Seems to me like travis is reviewing an old commit since the errors don't apply to the currenty commit. The force push caused a rerun, but it didn't seem to work on the current sources

jberry-suse:
> Seems to me like travis is reviewing an old commit

Not sure you pushed what you wanted based on the regex discussion as the code you commented was not pushed.

jsegitz:
yes, it really helps to push all of the changes and not only a changed commit message ;)

jsegitz:
So travis is happy. Would be great if we could get this done until end of this week since I'll be on FTO then and I would like to get this done before I leave

jberry-suse:
> Would be great if we could get this done until end of this week since I'll be on FTO

So someone else can handle the fallout? :)

I ran locally with your test request:

```
$ ./check_source.py --debug --dry id 723033
[I] checking 723033
[I] home:jsegitz:requestsource/libvirt@6f2b82a14a124e489db3a2f27a613e64 -> home:jsegitz:requesttarget/libvirt
[I] found suppressed whitelist warning: polkit-unauthorized-
[I] found suppressed whitelist warning: .*
[D] skipped adding duplicate review for security-team
[I] 723033 accepted: Check script succeeded
[D] 723033 review not changed
```

Based on the OSRT:Config that seems to be the appropriate result.

jberry-suse:
@jsegitz You will want to someone in the `factory-maintainers` group to make the appropriate changes to the `openSUSE:Factory` `OSRT:Config`.

jsegitz:
> So someone else can handle the fallout? :)

Exactly my plan ;)

> @jsegitz You will want to get someone in the factory-maintainers group to make the appropriate changes to the openSUSE:Factory OSRT:Config.

Will do. Thank you for the review

PR #2163, opened 2019-08-20
gocd: *.Origin.Manager.Update: shift timer to account for seconds.

jberry-suse:


coolo:
for the record: gocd only schedules every couple of minutes, so why they take seconds there is completely beyond me :)

PR #2177, opened 2019-08-29
 build-fail-notification: Mail to -factory after 4, dr after 6 weeks

coolo:


build-fail-reminder.py
```diff
@@ -157,37 +188,40 @@ def main(args):
...
+        # At least to report to the project for not building - send a mail to openSUSE-Factory
+        ProjectComplainList.sort()
+        to = 'openSUSE-Factory@opensuse.org'
+        fullname = "openSUSE Factory - Mailing List"
+        subject = "%(project)s - Build fail notification" % {'project': project}
+        text = MAIL_TEMPLATES[2] % {
+            'project': project,
+            'packagelist': ProjectComplainList,
+            'sender': sender,
+            }
```
DimStar77: packagelist and sender are not used in MAIL_TEMPLATES[2]

coolo: Dude, this is your code :)

DimStar77: I know - and this was from before I added the for pkg in blabla to list them nicely and not just in a weird way.

coolo: Having this in the same array is strange anyway. I'll extract it

coolo:
@DimStar77 technically I'm done here. But I would really like to add tests - as emails are a very public thing to go wrong :)

Give me some more time

DimStar77:
> Give me some more time

Sure, it's been down for three months already... a few days more won't make a huge difference

coolo:
I still struggle finding better ways to test this, but I guess we should get this rolling as it hopefully has an impact on TW

DimStar77:
let's - in a first instance it will likely send a couple hundred mails out to all the maintainers with failed packages (I am constantly on CC - also to know who really was addressed), It will be interesting in 2 weeks though, when things hit the factory ML :) (the script should not start with sending mails to -factory just yet, as it missed the maintainer messages. after all)

DimStar77:
```
gocd/build-fail-reminder.gocd.yaml;
1. Error parsing YAML. : Line 21, column 0: Expected a 'block end' but found: scalar : 
```

PR #2194, opened 2019-09-04
remove leap 42.3 and add 15.2

msmeissn:


Vogtinator:
42.3 is still in apimap.json, intentional?

msmeissn:
i also removed it from apimap.json

PR #2202, opened 2019-09-05
check_source: move checking of bad rpmlint entries to subprocess.

### LINKED ISSUE #2193: check_source: logic to prevent rpmlint entries that suppress all warnings triggers for valid entries
e.g. sr#728060
Package has
addFilter(".*W:.*permissions-symlink.*/bin/umount.*")

The .* at the beginning triggers this behavior

jsegitz:
We need to use user supplied REs to check this and this is a potential
DoS. This also fixes false positives for entries that start with
addFilter(".*

This fixes #2193

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -206,17 +206,27 @@ class CheckSource(ReviewBot.ReviewBot):
 
         return True
 
-    def suppresses_whitelist_warnings( self, source_project, source_package):
+    def suppresses_whitelist_warnings( self, source_project, source_package ):
         # checks if there's a rpmlintrc that suppresses warnings that we check
         found_entries = set()
         contents = source_file_load(self.apiurl, source_project, source_package, source_package+'-rpmlintrc')
         if contents:
+            contents = re.sub(r'(?m)^ *#.*\n?', '', contents)
             matches = re.findall(r'addFilter\(["\']([^"\']+)["\']\)', contents)
+            # this is a bit tricky. Since users can specify arbitrary regular expresions it's not easy
+            # to match bad_rpmlint_entries against what we found
             for entry in self.bad_rpmlint_entries:
                 for match in matches:
-                    if match.startswith(entry):
-                        self.logger.info(f'found suppressed whitelist warning: {entry}')
-                        found_entries.add(entry)
+                    # First we try to see if our entries appear verbatim in the rpmlint entries
+                    if entry in match:
+                        self.logger.info(f'found suppressed whitelist warning: {match}')
+                        found_entries.add(match)
+                    # if that's not the case then we check if one of the entries in the rpmlint file would match one
+                    # of our entries (e.g. addFilter(".*")
+                    elif re.search(match, entry) and match not in found_entries:
+                        self.logger.info(f'found rpmlint entry that suppresses an important warning: {match}')
+                        found_entries.add(match)
+
         return found_entries
 
     def has_whitelist_warnings( self, source_project, source_package, target_project, target_package ):


jberry-suse:
Also likely a good idea to write a simple loop that runs over all packages in factory and see how many trigger with your patterns to decide if they're what you want before changing again.

jsegitz:
osc cat home:jsegitz:requestsource/libvirt libvirt-rpmlintrc
e.g. 
addFilter(".*W:.*permissions-symlink.*/bin/umount.*")
triggered the check that tested for .* (matches all errors/warnings)

The separate process can be checked from the main process and be terminated if it takes too long. So DoS issues should be avoided. It's definitely quite a bit of complexity but AFAIU python we can't do it in the same process any other way.

I couldn't figure out a way to make sure that we reliable test the user provided REs against our list without using them. Users could provide patterns like
pol?kit-[ou]nautho.ized-foo
to get around polkit-unauthorized-foo. 

The Factory idea is great, I'll do that and get back here.

coolo:
In dubio: add too many reviews if the regexp matches 'auth' or has >X globs in it. But you're not the only reviewer, there is also an opensuse-review team that hopefully looks out for strange rpmlintrc additions.

jsegitz:
So I tested it against all rpmlint files in Factory and fixed one issue: The code didn't check if the warning was actually applied (pcp-rpmlintrc has a problematic entry but it's not active).
I think it's better to accept the added complexity to not have to resort to heuristics. Spawning a different process isn't that bad IMHO

jsegitz:
It shouldn't hide the problem since it'll add reviews for us if it runs into entries that trigger the problematic behavior. We can than have a look and ask the maintainer to remove it.

I don't think the timeout solution will work. From what I've read REs will acquire the GIL and therefor the example you linked to wouldn't work if you replace the sleep with an actual long running RE. And yes, there are other parts that can be DoS'd but IMHO that's not an argument to add more code that has this problem

coolo:
It's an argument to avoid overengineered code though, so please simplify the code. If we need to protect our tools against DoS from devel project maintainers, we'll have different problems.

jsegitz:
I don't agree but I changed the code. I didn't added a timeout because that will not work in the same process due to the GIL being held.

check_source.py
```diff
@@ -206,17 +206,28 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
-                    if match.startswith(entry):
-                        self.logger.info(f'found suppressed whitelist warning: {entry}')
-                        found_entries.add(entry)
+                    # First we try to see if our entries appear verbatim in the rpmlint entries
+                    if entry in match:
+                        self.logger.info(f'found suppressed whitelist warning: {match}')
+                        found_entries.add(match)
+                    # then we check if one of the entries in the rpmlint file would match one
+                    # of our entries (e.g. addFilter(".*")
+                    if re.search(match, entry) and match not in found_entries:
```
jberry-suse: It would seem the condition order should be reversed to allow for lazy execution otherwise all that is saved is log message.

check_source.py
```diff
@@ -206,17 +206,28 @@ def check_source_submission(self, source_project, source_package, source_revisio
 
         return True
 
-    def suppresses_whitelist_warnings( self, source_project, source_package):
+
```
jberry-suse: nit: extra newline

check_source.py
```diff
@@ -206,17 +206,28 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
                 for match in matches:
-                    if match.startswith(entry):
-                        self.logger.info(f'found suppressed whitelist warning: {entry}')
-                        found_entries.add(entry)
+                    # First we try to see if our entries appear verbatim in the rpmlint entries
+                    if entry in match:
+                        self.logger.info(f'found suppressed whitelist warning: {match}')
+                        found_entries.add(match)
+                    # then we check if one of the entries in the rpmlint file would match one
+                    # of our entries (e.g. addFilter(".*")
```
jberry-suse: nit: comment does not really make sense with the example as `.*` would match `.*` verbatim.

check_source.py
jberry-suse: Otherwise, it likely makes more sense to be an `elif`.

jberry-suse:
Also commit message should be changed to something more indicative of the change at this point (supporting regex whitelist entries).

check_source.py
jsegitz: elif makes sense, thanks

check_source.py
jsegitz: .* would match any entry in the list (with the new logic it doesn't contain entries for .* anymore). Not sure if I understand you correctly. The example is for an rpmlint entry that would cause the branch to be taken

jberry-suse: Yes, I see it now was thinking the re matched the other direction.

PR #2224, opened 2019-09-18
osclib/core: request_create_delete(): utilize pusedo osc.core function.

jberry-suse:
Current code causes an exception (seen by sentry) and does not properly work with non-default apiurl. Current code was a workaround for lack of proper function in osc.core so a pseduo function of what should be there is provided.

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -994,6 +994,36 @@ def request_create_delete(apiurl, target_project, target_package, message=None):
 
     return RequestFuture('delete {}/{}'.format(target_project, target_package), create_function)
 
+# Should exist within osc.core like create_submit_request(), but rather it was
+# duplicated in osc.commandline.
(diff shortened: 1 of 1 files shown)


osclib/core.py
```diff
@@ -986,14 +986,40 @@ def request_create_delete(apiurl, target_project, target_package, message=None):
...
 
     return RequestFuture('delete {}/{}'.format(target_project, target_package), create_function)
 
+# Should exist within osc.core like create_submit_request(), but rather it was
+# duplicated in osc.commandline.
+def create_delete_request(apiurl, target_project, target_package=None, message=None):
+    if not message:
+        message = message_suffix('created')
+
+    request = ETL.Element('request')
```
coolo: build-fail-reminder's version is a little easier:
```
                r = osc.core.Request()
                r.description = cgi.escape("[botdel] Package has failed to build for >= 6 weeks")
                r.create(apiurl)

Issue #2220, opened 2019-09-15
Threaded SUSE.OriginManagerUpdate gets stuck


coolo:
the job is dead for over 8 hours.

```
b'{"project":"Devel:Galaxy:Manager:3.2","package":"spacewalk-web","sender":"zypp-team"}'
[I] skipped irrelevant origin: Devel:Galaxy:Manager:3.2
[I] Still alive: 23:25:31.873459
[I] Closing channel (0): 'Normal shutdown' on <Channel number=1 OPEN conn=<SelectConnection OPEN transport=<pika.adapters.utils.io_services_utils._AsyncSSLTransport object at 0x7f9b3057e438> params=<ConnectionParameters host=rabbit.opensuse.org port=5671 virtual_host=/ ssl=True>>>
[I] Closing channel (0): 'Normal shutdown' on <Channel number=1 OPEN conn=<SelectConnection OPEN transport=<pika.adapters.utils.io_services_utils._AsyncSSLTransport object at 0x7f9b30f06668> params=<ConnectionParameters host=rabbit.suse.de port=5671 virtual_host=/ ssl=True>>>
[I] Received <Channel.CloseOk> on <Channel number=1 CLOSING conn=<SelectConnection OPEN transport=<pika.adapters.utils.io_services_utils._AsyncSSLTransport object at 0x7f9b3057e438> params=<ConnectionParameters host=rabbit.opensuse.org port=5671 virtual_host=/ ssl=True>>>
[I] Channel 1 was closed: (0, 'Normal shutdown')
[I] Closing connection (200): 'Normal shutdown'
[I] Connection.close is waiting for 1 channels to close: <SelectConnection CLOSING transport=<pika.adapters.utils.io_services_utils._AsyncSSLTransport object at 0x7f9b3057e438> params=<ConnectionParameters host=rabbit.opensuse.org port=5671 virtual_host=/ ssl=True>>
[I] Received <Channel.CloseOk> on <Channel number=1 CLOSING conn=<SelectConnection OPEN transport=<pika.adapters.utils.io_services_utils._AsyncSSLTransport object at 0x7f9b30f06668> params=<ConnectionParameters host=rabbit.suse.de port=5671 virtual_host=/ ssl=True>>>
[I] Channel 1 was closed: (0, 'Normal shutdown')
[I] Closing connection (200): 'Normal shutdown'
[...]

PR #2226, opened 2019-09-19
gocd/monitors: *.OriginManagerUpdate: set timeout to 10 minutes.

jberry-suse:
Would seem like this should be the default on all monitors since they run with same base class that spits out the still alive message (specifically for gocd). Will at least insure that a stuck state does not last a long time.

Related to #2220


coolo:
we never used timeouts so far - let's see if the jobs are killed cleanly

PR #2231, opened 2019-09-20
check_source: don't enforce name for kiwi images

lnussel:
The benefit of this check is unclear to me. It enforces the Name tag
inside he kiwi file to be the same as the package name in OBS while the
kiwi file name itself can be arbitrary.
So that means that kiwi package names end up encoding their distribution
and version (eg "openSUSE-Leap-15.2-Vagrant") and have to be renamed for
each release. Also means diffing between released versions isn't as
straight forward as it could be if the name was more flexibel.
If we wanted to enforce something then we should make sure the actual
kiwi file name is the same as the OBS package name. That is what the
spec file name check actually does after all (package hello needs to
have hello.spec ie srpm name hello).

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -145,7 +145,7 @@ class CheckSource(ReviewBot.ReviewBot):
         shutil.rmtree(os.path.join(target_package, '.osc'))
 
         new_info = self.package_source_parse(source_project, source_package, source_revision)
-        if new_info['name'] != target_package:
+        if not new_info['filename'].endswith('.kiwi') and new_info['name'] != target_package:
             shutil.rmtree(dir)
             self.review_messages['declined'] = "A package submitted as %s has to build as 'Name: %s' - found Name '%s'" % (target_package, target_package, new_info['name'])
             return False
@@ -293,6 +293,9 @@ class CheckSource(ReviewBot.ReviewBot):
 
         if xml.find('version') is not None:
             ret['version'] = xml.find('version').text
+
+        if xml.find('filename') is not None:
+            ret['filename'] = xml.find('filename').text
 
         return ret
 


Vogtinator:
> If we wanted to enforce something then we should make sure the actual
kiwi file name is the same as the OBS package name.

Sounds ok to me, but note that this would break quite a few images we have currently.

Also, OBS gets really confused if either the "name" or "version" in the .kiwi file are non-standard and screws up stuff like releasever tracking.

lnussel:
what is a non standard name?

Vogtinator:
> what is a non standard name?

All I know is that it totally broke when I tried to use `%OS_VERSION_ID%`. It was stuck at `1.1`.

coolo:
Can we have some test cases for this? So we can talk about facts?

Vogtinator:
> Can we have some test cases for this? So we can talk about facts?

Image was rebuilt manually once, yet both builds hat a relcnt of 1:

```
> osc buildhist home:favogt:branches:Virtualization:containers:images:openSUSE-Leap-15.2 -M docker opensuse-leap-image containers x86_64
time                  srcmd5                              vers-rel.bcnt    rev   duration
2019-09-23 08:39:27   95bbe26b15ac15a54916849c70bb7061    %OS_VERSION_ID%. 2        206
2019-09-23 08:42:18   95bbe26b15ac15a54916849c70bb7061    %OS_VERSION_ID%. 2        142
```

lnussel:
so you changed the version tag, not the name tag. this change is only about the name tag, no?

Vogtinator:
> so you changed the version tag, not the name tag. this change is only about the name tag, no?

Correct. I'll try the same with the name as well.

lnussel:
seems to work fine for me in home:lnussel:branches:openSUSE:Leap:15.2/kiwi-images-vagrant

PR #2234, opened 2019-09-24
origin-manager: skip package of type not (None or source) and similar for check_source.

### LINKED ISSUE #2229: origin-manager: ignore packages during review using same filter as list generation
Requests for [000package-groups](https://build.opensuse.org/request/show/731520) should be ignored by origin-manager during review the same way they are ignored when generating full cached lookup and updates.

jberry-suse:
- 160225b5e983057381f92a973054cf67219bac37:
    origin-manager: skip package of type not (None or source).
    
    Should skip review of meta packages and such. Although same logic would
    seem to apply to deletes origin-manager is used to enforce manual review
    of all deletes and as such even deleting of a meta package should be
    included.

- afd288b87d946df3464c4037b50eea41653ecec3:
    check_source: decline package of kind not (meta or source).

- 91aeaf1f55fd58bee0e0a1454c0a114fe9297a5e:
    check_source: replace meta check with package_kind() == meta.

- f63da476c21719a8634b7ab4427182ef68419f13:
    osclib/core: package_kind(): include underscore as meta prefix.
    
    Previously used by _product which is still used in SLE-12.

Fixes #2229.

An active example:

```
$ ./origin-manager.py --dry --debug id 732664
[I] checking 732664
[I] 732664 accepted: skipping meta package since not source
[D] 732664 review not changed
```

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -82,6 +82,9 @@ class CheckSource(ReviewBot.ReviewBot):
         if kind == 'meta':
             self.review_messages['accepted'] = 'Skipping all checks for meta packages'
             return True
+        elif kind != 'source':
+            self.review_messages['accepted'] = 'May not modify a non-source package of type {}'.format(kind)
+            return False
 
         inair_renamed = target_package != source_package
 
--- a/check_source.py
+++ b/check_source.py
@@ -18,6 +18,7 @@ from osclib.conf import Config
 from osclib.core import devel_project_get
 from osclib.core import devel_project_fallback
 from osclib.core import group_members
+from osclib.core import package_kind
 from osclib.core import source_file_load
 from osclib.core import target_archs
 from urllib.error import HTTPError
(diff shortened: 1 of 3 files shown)


check_source.py
```diff
@@ -77,9 +78,13 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
             return False
 
-        if target_package.startswith('00') or target_package.startswith('_'):
-            self.review_messages['accepted'] = 'Skipping all checks for product related packages'
+        kind = package_kind(self.apiurl, target_project, target_package)
+        if kind == 'meta':
+            self.review_messages['accepted'] = 'Skipping all checks for meta packages'
             return True
+        elif kind != 'source':
+            self.review_messages['accepted'] = 'May not modify a non-source package of type {}'.format(kind)
```
jberry-suse: @DimStar77 @nilxam @lnussel How often is a multi-spec sub-package overridden with a full source package? The override should still be usable in this case. Or would you rather this check not be added? For example, this would prevent a submit request against `kernel-debug` since it is a multi-spec subpackage of `kernel-source`.

nilxam: We should always decline the *multispec_subpackage* submission in the first place. The only exception here is user want to switch sub-package to be the main-package *+* package update at the same time, but that case happened very rarely, the override for that is good enough.

nilxam:
Jimmy, FYI https://github.com/openSUSE/openSUSE-release-tools/pull/2237

jberry-suse:
Yeah, thanks. Go figure included the `None` check in origin-manager chunk.

PR #2247, opened 2019-09-26
osclib/origin_listener: change from package.update to package.commit event.

### LINKED ISSUE #2239: origin-manager: synced SLE sources do not generate events
Based on watching the requests generated by the OBS listener and reviewing the logs the only time an update request from SLE is generated is when the back is update in Leap:15.1:Update or similar which triggers an update check which notices the SLE sources were changed. The logs I reviewed never contain an update message for a SLE package even when the revision history on OBS shows they were updated within the timespan covered by the logs.

Presumably, there is nothing on our that can be done about this short of reverting back to running the full crawler more often which I think should be done until OBS properly generates events. I am assuming the lack of events is due to some sort of hack used to create the sync since commits from IBS appear on OBS which is not possible with standard osc tooling. Another thing that would work out-of-the-box with git (pushing entire tree).

jberry-suse:
In theory fixes the problem of not picking up SLE events as it is claimed they generate `.commit` events while not generating `.update` events (unlike all other packages). In theory no harm from this (although I remember debating between the two events and feel like there was a reason for sticking to update).

Fixes #2239.

### DIFF
--- a/osclib/origin_listener.py
+++ b/osclib/origin_listener.py
@@ -36,7 +36,7 @@ class OriginSourceChangeListener(PubSubConsumer):
 
     def routing_keys(self):
         return [self._prefix + k for k in [
-            '.obs.package.update',
+            '.obs.package.commit',
             '.obs.request.create',
         ]]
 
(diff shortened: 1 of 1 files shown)


coolo:
that commit send a .update event is basically a bug. It happens because every commit 'updates' the last_updated column in the database (for 'latest commits' news feed).

PR #2249, opened 2019-09-27
Enable images repo once 000product was updated

coolo:
If installcheck or packagelists check fails, there is no point (normally) to build the media - so let the release manager disable the images repo (which also hides uninteresting fails) while the checks are failing and have gocd enable the images repo afterwards to start the build (and the openqa sync)


coolo:
We did that so far only for factory, now also for sp2 and leap. So as soon as sp5 is done, I would like to have select/unselect commands disable images repo automatically.

PR #2250, opened 2019-09-27
check_source: avoid crash if filename is not set

lnussel:


lnussel:
fixes the crash of #2250

PR #2254, opened 2019-10-04
BuildFailReminder: fix inversion of to and fullname

DimStar77:


DimStar77:
formataddr expects fullname, address in order to format the address - we had this inverted
and also inverted filling to and fullname when reading the user info of the maintainers (which is why this worked)

but we did not invert this when sending digest mails to opensuse-factory - those mails never passed

DimStar77:
```
 email.utils.formataddr(pair, charset='utf-8')

    The inverse of parseaddr(), this takes a 2-tuple of the form (realname, email_address) and returns the string value suitable for a To or Cc header
```

coolo:
rebased

PR #2257, opened 2019-10-09
Convert factory-package-news-web.py to python3

ggardet:


coolo:
If that's your intention, then also change the shebang

ggardet:
> If that's your intention, then also change the shebang

Fixed.

coolo:
@lnussel @DimStar77 I have no idea how and where this is deployed - so I shy away from merging

coolo:
@lnussel @DimStar77 any help?

lnussel:
runs on openqa and doesnt auto deploy. no impact.

PR #2260, opened 2019-10-15
pkglistgen: fix wrong parsing way on pkglistgen-locale

nilxam:
Since the data is from project attribute, pkglistgen should not parsing a string.
This change also renamed pkglistgen-local to pkglistgen-locale.


nilxam:
I guess no one tried pkglistgen-local before :)

PR #2262, opened 2019-10-16
Added Vagrant box in 15-SP2 release tool

gyr:
Added new SLED-15-SP2-Vagrant and SLES-15-SP2-Vagrant in release to TEST
setup for SLE-15-SP2

### DIFF
--- a/gocd/sp2.target.gocd.yaml
+++ b/gocd/sp2.target.gocd.yaml
@@ -82,6 +82,7 @@ pipelines:
         tasks:
         - script: |-
             for product in kiwi-templates-JeOS \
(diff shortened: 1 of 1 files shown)


gyr:
> Please be aware that vagrant is using multibuild you'll need to specify individual multibuild targets
> Please see SLE-12-SP5 https://gitlab.suse.de/sle-prjmgr/release-management-tools/commit/4f8cc15a4fa39eab87a2d993302f922c4581f594

seems that it does it out-of-the-box, e.g. kiwi-templates-JeOS has also multibuild and it is not required to list all of them. Could anyone confirm that?

coolo:
if you provide one flavor, you need to provide the others as well - if you provide no flavor, all flavors are released

PR #2263, opened 2019-10-16
build-fail-reminder: update api path to projects/:project/status

nilxam:
Recently the status api path changed from `project/status/:project` to `projects/:project/status`, update the api path in build-fail-reminder. This should also fixes travis failure.


jberry-suse:
The argument used to block useful features was always not to break API compatibility and yet this is blatant? Amusing.

Thanks for fixing.

coolo:
Yeah, the problem is that this is not using 'the api' but code I explictly added to be useful for scripts :)

PR #2264, opened 2019-10-16
Fix build fail reminder to use new URL for project status

coolo:


Vogtinator:
Dup of https://github.com/openSUSE/openSUSE-release-tools/pull/2263/files?

coolo:
Fun that github doesn't care it's already in master :)

PR #2270, opened 2019-10-21
Fix test plan URL for 15.2

lnussel:


coolo:
I take your word on that :)

PR #2272, opened 2019-10-22
Fix check_source on public cloud images

coolo:
The check for bcntsynctag can be very misleading - just because we align the build counters between 2 packages doesn't make them invalid submission targets. Better rely on the link check which is already implemented as fallback

And set a verbose decline reason for this case


mbgg:
Just as a heads up [SR 74102](https://build.opensuse.org/request/show/741702) is affected by this.

PR #2277, opened 2019-10-25
requestfinder: ConfigParser has been renamed to configparser in python3

nilxam:


nilxam:
just a question: should we keep the code python2 compatible? or just go ahead to python3 only?

coolo:
we're python3 only - the package buildrequires python3-osc

requestfinder.py
```diff
@@ -1,29 +1,25 @@
...
+#!/usr/bin/python3
 
-from ConfigParser import ConfigParser
+from configparser import ConfigParser
 from xdg.BaseDirectory import load_first_config
 from xml.etree import cElementTree as ET
+
 import sys
 import cmdln
 import logging
```
mimi1vx: can be remove too

requestfinder.py
```diff
@@ -1,29 +1,25 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 
-from ConfigParser import ConfigParser
+from configparser import ConfigParser
 from xdg.BaseDirectory import load_first_config
 from xml.etree import cElementTree as ET
```
mimi1vx: if is python3 only "his module will use a fast implementation whenever available. The xml.etree.cElementTree module is deprecated." from python 3.3

nilxam: It still exists there for backward compatible, that's why you can found cElementTree everywhere  in this repo, maybe at some point we should change it by another PR.

coolo:
please rebase

coolo:
that rebases? wow :)

nilxam:
rebase what? it's a commit only...

coolo:
I just wanted a green travis - I guess you're confusing it with squashing. Normally I git rebase origin/master and push -f, but your way worked too, as travis tests your commit against master

PR #2280, opened 2019-10-29
Use python3 in shebang to not pull in python2 dependency

scarabeusiv:


coolo:
our CI is busted and I failed to notice so far :(

coolo:
could you please use a personal fork for future PRs? thanks

scarabeusiv:
@coolo it was not on purpose, i just used the web edit so I don't have to clone it. I guess it took it as i have privs on openSUSE I don't need a fork....

PR #2285, opened 2019-11-04
Make the CI green again

coolo:
(it's mainly osc workaround)


coolo:
we discussed this in IRC today and it seems local overwrite this way isn't used - so we may as well remove this feature and rely on pure combination of remote config and conf.py changes

jberry-suse:
Short of development I would expect that to be the case. Before my introduction of remote config it was more useful for deployments as well.

PR #2295, opened 2019-11-06
New staging api for workflow

coolo:
Playground: https://build-test.opensuse.org/staging_workflows/1


osc-staging.py
```diff
@@ -146,35 +141,6 @@ def do_staging(self, subcmd, opts, *args):
...
-
-        All of the above may be restricted to a set of stagings.
-
-        The staging configuration is automatically cleared anytime staging
-        psuedometa is cleared (accept, or unstage all requests).
-
-        The keys that may be set in staging configuration are:
-
-        - repo_checker-binary-whitelist[-arch]: appended to target project list
-        - todo: text to be printed after staging is accepted
```
jberry-suse: Not only losing config command, but requested todo functionality. :(

coolo: Right. And if you read the commit log of 0dc2769f443735ca0904c8cef5c15b2dea4d3587 I checked with staging managers that they want it differently anyway.

jberry-suse: The other use-case was "lost" during `git mv ...`. If the functionality is no longer desired should just be dropped as not relevant to "new staging api".

coolo: You mean to cherry pick the drop of feature in another PR? Could be done - this is a draft

coolo:
Moving to openSUSE repo

PR #2296, opened 2019-11-06
osclib/origin: provide automatic update mode controls.

### LINKED ISSUE #942: update-crawler: mark packages to no longer receive version updates
A number of python packages are to no longer receive updates, but the burden is on the reviewer to remember that or look around for previous declined request. It would be better to mark, perhaps in origin file, the package to no longer receive automatic update requests via the crawler or at least blacklist say Factory updates.

Thoughts?

### LINKED ISSUE #2182: origin-manager: provide update frequency control (or ignore altogether)
Provide a method, likely an OBS attribute, of controlling the frequency of update submissions. This will primarily be useful for Factory which tracks many packages updated by CI, but is also useful for Leap. Along with this provide a configurable project level default.

Along with a time based frequency should be the ability to never create a request when an open one exists for the package.

Finally, a complete ignore option to halt updates.

jberry-suse:
- ba103f3609f8b56583f721bd156763678f5ef1ec:
    osclib/origin: provide automatic update mode controls.

- 35dab9824a1067ba6962e167968689cfda194ecb:
    osclib/core: request_create_submit(): provide frequency option.

- 5087480d7bff406244e8d09a951e028d19238e72:
    osclib/core: request_create_submit(): provide supersede flag.

- fbeccb34ecbeecbf68612ee2aea92c3cb19cbb88:
    osclib/core: request_action_simple_list(): include full history.
    
    Without this the requests are not nearly as useful when processing.

- 6bdd1a014d122e4eaa90e1c923f0f234b327ae8c:
    osclib/core: provide package_source_{changed,age}() functions.

- 34e21ad23034416aece231d09b0f1088c7ff856f:
    osclib/core: provide attribute_value_delete().

- b894ceae509461bf82c83bde39155d52aa974894:
    osclib/core: support package in attribute_value_{load,save}() functions.

Based on requests from devel package maintainers for Leap, likely want to default to a delay and frequency cap for `<devel>` origin. @lnussel thoughts? (for now I configured a one hour delay and daily frequency for both Leap and Factory devel projects)

Fixes #2182.
Fixes #942.

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -968,7 +968,7 @@ def request_action_simple_list(apiurl, project, package, states, request_type):
     # Disable including source project in get_request_list() query.
     before = conf.config['include_request_from_project']
     conf.config['include_request_from_project'] = False
-    requests = get_request_list(apiurl, project, package, None, states, request_type)
+    requests = get_request_list(apiurl, project, package, None, states, request_type, withfullhistory=True)
     conf.config['include_request_from_project'] = before
 
     for request in requests:
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -361,8 +361,9 @@ def package_list_kind_filtered(apiurl, project, kinds_allowed=['source']):
 
         yield package
 
-def attribute_value_load(apiurl, project, name, namespace='OSRT'):
-    url = makeurl(apiurl, ['source', project, '_attribute', namespace + ':' + name])
+def attribute_value_load(apiurl, project, name, namespace='OSRT', package=None):
+    path = list(filter(None, ['source', project, package, '_attribute', namespace + ':' + name]))
+    url = makeurl(apiurl, path)
 
     try:
         root = ETL.parse(http_GET(url)).getroot()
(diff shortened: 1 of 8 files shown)


jberry-suse:
Test hung twice and was terminated by travis. I do not see anything relevant to failing test and cannot reproduce locally.

jberry-suse:
Consistently fails on travis and only for this PR. My guess would be some sort of memory change or something. I'll have to investigate.

jberry-suse:
Updated my containers locally and consistently passes, but consistently hangs on travis...

osclib/core.py
```diff
@@ -520,6 +526,14 @@ def project_meta_revision(apiurl, project):
         apiurl, project, '_project', None, format='xml', meta=True))
     return int(root.find('logentry').get('revision'))
 
+def package_source_changed(apiurl, project, package):
+    root = ETL.fromstringlist(get_commitlog(
```
coolo: if you only need the latest history entry, calling get_commitlog is very expensive as it fetches the full history, parses it and creates custom XML out of it.

And _history has a limit parameter, but the osc function doesn't seem to support it. So I would instead fetch `/source/Kernel:stable/kernel-source/_history?limit=1` and take the time. This is what osc does to it:
```
        t = time.gmtime(int(node.find('time').text))
        t = time.strftime('%Y-%m-%d %H:%M:%S', t)
```

So just doing the time.gmtime will do, then you can also skip the date_parse

coolo:
my guess: you use a lot of random strings and travs-ci is short on entropy

osclib/core.py
jberry-suse: Only called when actually updating a package, not on every package during check. Also just trying to use the `osc.core` API which is rather lacking as usual.

jberry-suse: It's a bit more fun to get the date objects to play nice without using date_parse() as the timezones are wrong, then one with and one without.

jberry-suse:
> my guess: you use a lot of random strings and travs-ci is short on entropy

Interestingly it fails consistently on `tests/select_tests.py`. Which passes locally and I see no relevant changes.

PR #2316, opened 2019-11-12
Enable announcer for ARM for Tumbleweed, Kubic and MicroOS

ggardet:


config/announcer/openSUSE:Factory:ARM.yml
```diff
@@ -0,0 +1,31 @@
...
+  Please note that this mail was generated by a script.
+
+  The described changes are computed based on the aarch64 DVD.
+
+  The full online repo contains too many changes to be listed here.
+
+
+  Please check the known defects of this snapshot before upgrading:
+
+  https://openqa.opensuse.org/tests/overview?distri=opensuse&groupid=3&version=Tumbleweed&build={version}
```
coolo: I have the feeling that the openqa url is wrong

ggardet: No, it is fine, just try: https://openqa.opensuse.org/tests/overview?distri=opensuse&groupid=3&version=Tumbleweed&build=20191109

ggardet: groupid `1` is for `x86_64`, `3` is for `aarch64`

PR #2319, opened 2019-11-19
Fix rebuild command

coolo:


coolo:
the test suite is plagued by https://github.com/openSUSE/open-build-service/issues/8729

PR #2320, opened 2019-11-19
Fix for reviews (OBS issue is fixed)

coolo:


coolo:
damn, the test suite has another problem

PR #2325, opened 2019-11-20
suppkg_rebuild: Move to new staging api

nilxam:


suppkg_rebuild.py
```diff
@@ -105,8 +105,9 @@ def crawl(self):
         staging_projects = ["%s:%s"%(self.api.cstaging, p) for p in self.api.get_staging_projects_short()]
         cand_sources = defaultdict(list)
         for stg in staging_projects:
-            prj_meta = self.api.get_prj_pseudometa(stg)
-            prj_staged_packages = [req['package'] for req in prj_meta['requests']]
+            url = makeurl(self.apiurl, ['staging', self.project, 'staging_projects', stg, 'staged_requests'], { 'requests': 1 })
```
coolo: `staged_requests = self.api.project_status(stg, status=False)` is enough

PR #2331, opened 2019-11-21
Adopt to latest API changes

coolo:
we decided to drop the `number` as it's `id` everywhere else


coolo:
@nilxam 

Note that our test container already uses the new API, but it's not yet deployed. But we will have it deployed soonish

PR #2332, opened 2019-11-21
More fixes unignore

nilxam:


osclib/stagingapi.py
```diff
@@ -933,6 +933,11 @@ def rq_to_prj(self, request_id, project):
             msg = msg.format(request_id)
             raise oscerr.WrongArgs(msg)
 
+        # unignore a request selected to a project
+        requests_ignored = self.get_ignored_requests()
```
coolo: I think this should be an option and not be done by default. And for this option, there is an API parameter to select

coolo: https://github.com/openSUSE/open-build-service/issues/8525

coolo: So I would say for now just not handling ignored requests in select is the easist - and leave the direct move in the backlog.

PR #2337, opened 2019-11-24
Reuse function of accept to check local links

coolo:
There is no error reported yet, just output


coolo:
Both Factory and 15.2 are without report

PR #2341, opened 2019-11-26
stagingapi: Supersedes request is from the same project

nilxam:
If the request is from the same project then just supersedes the old ones. The request is from the different project then put it to excluded list as the right behavior for Leap.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -508,6 +508,11 @@ class StagingAPI(object):
                     self.do_change_review_state(request_id, 'declined',
                                                 by_group=self.cstaging_group, message=message)
                 else:
+                    # Supersedes request is from the same project
+                    if request_new.find('./action/source').get('project') == request_old.find('./action/source').get('project'):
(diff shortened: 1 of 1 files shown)


coolo:
At one point test cases for this would be really beneficial :(

osclib/stagingapi.py
```diff
@@ -508,6 +508,11 @@ def superseded_request(self, request, target_requests=None):
                     self.do_change_review_state(request_id, 'declined',
                                                 by_group=self.cstaging_group, message=message)
                 else:
+                    # Supersedes request is from the same project
+                    if request_new.find('./action/source').get('project') == request_old.find('./action/source').get('project'):
+                        message = 'sr#{} has newer source and is from the same project'.format(request_new.get('id'))
```
coolo: I wouldn't lie - we just don't know if it's newer source. But it will do

nilxam: Yes, or we can re-use old message `Replaced by sr#XXX`

PR #2355, opened 2019-12-04
Fix setprio command to use current API

### LINKED ISSUE #2353: staging setprio: 'Staging Project "FOO" does not exist
I just tried to bump up the prio (mainly for legal - even though nobody looks at it) for Staging G, but this results in:

> osc staging setprio G
Staging Project "G" does not exist.

from osc -d, it becomes apparent that this is stilkl using the routes through the old dashboard:

```
GET https://api.opensuse.org/staging/openSUSE:Factory/staging_projects/G?requests=1
```

(so does osc staging check)

coolo:
Fixes #2353

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -66,7 +66,7 @@ def _full_project_name(self, project):
 
 def lock_needed(cmd, opts):
     return not(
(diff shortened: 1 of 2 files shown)


coolo:
I removed the option to prio all staging projects. If someone means to increase all reviews at once to important, he may script this easily. But this is a too dangerous user error waiting to happen :)

PR #2360, opened 2019-12-05
Test select and accept of _multibuild packages

coolo:


tests/accept_tests.py
```diff
@@ -49,3 +50,69 @@ def test_accept_final_comment(self):
...
+
+        ret = SelectCommand(wf.api, staging.name).perform(['gcc9'])
+        ac = AcceptCommand(wf.api)
+        self.assertEqual(True, ac.accept_all(['A'], True))
+
+        # no stale links
+        self.assertEqual([], package_list(wf.apiurl, staging.name))
+        self.assertEqual(['gcc9', 'gcc9-tests', 'wine'], package_list(wf.apiurl, wf.project))
+
+    def test_accept_switch_to_multibuild_package(self):
```
DimStar77: In theory a former multibuild/multispec could also switch to pure multispec again

coolo: I thought about it and considered it too academic taking the trend

osclib/stagingapi.py
```diff
@@ -936,12 +936,16 @@ def get_sub_packages(self, package, project):
         # guarantee the sub-pacakges are created according to the
         # specfiles of main package. Therefore, main package must be
         # created before through get_sub_packages().
-        filelist = self.get_filelist_for_package(pkgname=package, project=project, expand='1', extension='spec')
+        filelist = self.get_filelist_for_package(pkgname=package, project=project, expand='1')
+        if '_multibuild' in filelist:
```
nilxam: I suppose `_multibuild` file should be filtered out in `get_sub_packages()` instead of return it back?

PR #2361, opened 2019-12-06
Delete adi projects whenever we'd build disable them

coolo:


osclib/stagingapi.py
```diff
@@ -1484,3 +1495,16 @@ def is_staging_bootstrapped(self, project):
...
+        if not self.is_adi_project(project):
+            return
+
+        if not self.is_staging_manager:
+            return
+
+        try:
+            delete_project(self.apiurl, project, force=True)
+        except HTTPError as e:
+            print(e)
```
nilxam: Why shouldn't it just `raise e`?

coolo: I just moved this - but I still think it's a good idea to continue here. If the adi project is gone or not doesn't matter much, but if we exit the accept command, we don't cleanup the rest

nilxam: ok

PR #2369, opened 2019-12-10
Do not return after adi project is accepted

### LINKED ISSUE #2368: osc staging accept does not 'complete its job'
Yesterday evening, I did a checkin using 'osc staging accept' (all green stagings - that was only a list of adi at that time, no letter staging was ready due to the certificate issues still)

The checkin went pretty ok, with a few things observed:
* The 'waiting for staging project 'xxx' to be accepted happened only for one adi (all the others were 'just being accepted after osc staging accept already long finished (not an issue per se, but weird)
* The 'trigger rebuild loop' for the diff between openSUSE:Factory and openSUSE:Factory:Rebuild did not happen
* Product Version in the openSUSE:Factory attribute was not updated (I bumped it this morning manually)

There were no errors in the accept output, the log was (non-debug log)

```
Accepting all acceptable projects
delete links to packages pending deletion...
triggering staging accepts...
Accepting request 755046: SDL2
waiting for staging project openSUSE:Factory:Staging:adi:57 to be accepted
```

coolo:
Fixes #2368

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -115,7 +115,7 @@ class AcceptCommand(object):
             print(f"Accepting request {req['id']}: {req['package']}")
             change_request_state(self.api.apiurl, str(req['id']), 'accepted', message='Accept to %s' % self.api.project)
 
-        for project in staging_packages.keys():
+        for project in sorted(staging_packages.keys()):
             print(f'waiting for staging project {project} to be accepted')
 
             while True:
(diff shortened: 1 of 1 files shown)


osclib/accept_command.py
```diff
@@ -128,9 +128,8 @@ def accept_all(self, projects, force=False, cleanup=True):
             self.api.accept_status_comment(project, staging_packages[project])
             if self.api.is_adi_project(project):
                 self.api.delete_empty_adi_project(project)
-                return
-
-            self.api.staging_deactivate(project)
+            else:
+                self.api.staging_deactivate(project)
 
             self.reset_rebuild_data(prj)
```
coolo: wait, there is more - this reset_rebuild_data shouldn't be done for adis

PR #2372, opened 2019-12-10
More rebuild script fixes for ARM

dirkmueller:
tested manually via a copy of the original project


coolo:
Can you please seperate the flake8 cleanups in PRs on their own? Thanks

dirkmueller:
@coolo there are separate commits, you can browse them one after the other (they even have a commit message :-) )

Issue #2350, opened 2019-12-02
installcheck: need a way to 'exclude' the .crt from the conflict list (or do more what rpm does)


DimStar77:
https://build.opensuse.org/package/view_file/home:repo-checker/reports/openSUSE:Factory:Staging:D

found conflict of kernel-debug-5.4.0-2.2.x86_64 with kernel-default-5.4.0-2.2.x86_64
  /etc/uefi/certs/F0D46AF3.crt

In theory, yes, this is correct - but kernel is a multiversion package in SUSE distro anyway so if this conflict were a problem we could also not install multiple versions of the kernel without a file conflict.

for installcheck, the new conflict is seen as modules are newly signed for all flavors (which was a bug in the past, only -default had them signed)

So, we either need a way to ignore this file conflict or be more flexible, as RPM is: since the packaged file is actually the same (same checksum), this is not really a conflict for RPM (you are allowed to own the SAME file in multiple packages, but not different files with the same path/name)

coolo:
https://jira.suse.com/browse/PM-355 lists several real problems with this 'theoretical' problem.

DimStar77:
you mean like https://bugzilla.suse.com/show_bug.cgi?id=1124431 where everybody heats up and no real solution is found?

In essence, RPM is 'fine' as long as the files are identical. Of course, when the certificate is renewed, the ID stays and the content changes and things will blow up.

In case of the kernel though the issue is even more fun: the file seen in staging does not exist in the real package shipped to users (a different certicicate is used in stagings compared to openSUSE:Factory) - so in fact we have, essentially, anyway no clue what happens in the end

DimStar77:
> https://jira.suse.com/browse/PM-355 lists several real problems with this 'theoretical' problem.

What was the final conclusion on SLE? Seems most bugs have been closed and I assume SLE still received kernel updates?

Since the filename of the certificate is based on its sha1sum (albeit shortened to 8 chars) any different certificate would have a different file name (minus fingerprint clashes)

DimStar77:
Ok, found how it was 'fixed' for SLE15SP2

```
# subpackages of ring packages with unacceptable requires (don't abuse)
# there are some bugs pending that would result in cleanup:
# 1124431	SUSE Linux Enterprise Server 15 SPx	Kernel	kernel-maintainers@forge.pr...	NEW	---	File conflict about /etc/uefi/certs/91A3B0B5.crt	09:52:44
[...]

PR #2376, opened 2019-12-13
findfileconflicts: Ignore kernel installed certificates in /etc/uefi/…

DimStar77:
…certs

The file names are generated out of the sha1sum of the file content. Thus we
know if the file has the same name, it has the same content as well. Multiple
packages are allowed to own the same file with the same content. This is already
proven functional by the kernel being flavor:multiversion - the kernel can
be installed in multiple versions already (which the checked never would have
caught file conflicts for: we only check against current packages in the distro)

Fixes issue #2350

### DIFF
--- a/findfileconflicts
+++ b/findfileconflicts
@@ -355,7 +355,7 @@ for my $tc (sort keys %tocheck) {
 	}
 	# got one!
 	$f =~ /^(\d+)\/(.*)/;
-	push @con, "$dirs[$1]$2$info";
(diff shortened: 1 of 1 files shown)


fcrozat:
LGTM, even if we can't really verify the premise (filename generated from checksum).

nilxam:
I'm fine to have this if SLE release manager and Tumbleweed release manager both are agree to have it, the only risk is once the fingerprint(filename) changed.

PR #2377, opened 2019-12-18
abichecker fixes

lnussel:


abichecker/abichecker.py
```diff
@@ -198,6 +215,15 @@ def check_source_submission(self, src_project, src_package, src_rev, dst_project
         # there were problems.
         ret = True
 
+        if self.has_staging(dst_project):
+            # what a wonderufl hack
```
coolo: neat word

abichecker/abichecker.py
```diff
@@ -514,6 +539,12 @@ def check_one_request(self, req):
 
         self.dblogger.request_id = req.reqid
 
+        self.staged_in = None
+        for r in req.reviews:
```
coolo: I'm not happy with this one. We should reduce not increase our dependency on this review - and make 'staging project' a property of a request. The problem is that there is little I have to offer as alternative.

`staging/$TARGET/staging_project?requests=1` will list all requests, but to request it only makes sense if you can do it for multiple requests at once (or rely on internal caching of osclib)

abichecker/abichecker.py
lnussel: just as wonderufl as the hack itself

abichecker/abichecker.py
lnussel: the bot is long running so it would have to prune the cache on each iteration. I could move the staged_in property into ReviewBot so it's at least in a central place and not hidden in abichecker. The implementation can then be changed once OBS has better APIs

coolo: sounds like a plan

PR #2382, opened 2020-01-10
Staging accept: build-disable the images repository

DimStar77:
With the current pipelines, we enable build of the images repository
once everything in /standard is build-succeeded and repo-checker passed.

Let's go the last mile and disable images when we check in the staging.

This should save some OBS build power (not building ISO images before the
stagings are actually ready) and also some openQA (not testing ISO images
before stagings settle)

I can foresee that sometimes an RM would still want to have preliminary
testing done - in which case the images repo can be manually build enabled
(or the pipeline be triggered to go beyond repo-checker, as it likely also
needs a recalculation of the content)

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -110,6 +110,7 @@ class AcceptCommand(object):
         for project in staging_packages.keys():
             u = self.api.makeurl(['staging', self.api.project, 'staging_projects', project, 'accept'], opts)
             http_POST(u)
+            self.api.switch_flag_in_prj(project, flag='build', state='disable', repository='images')
 
         for req in other_new:
             print(f"Accepting request {req['id']}: {req['package']}")


DimStar77:
did some checkin rounds with this already and it does what I expected

coolo:
I'm afraid the OBS staging dashboard makes staging projects acceptable too quick.

coolo:
See https://github.com/openSUSE/open-build-service/issues/7359#issuecomment-537016476

DimStar77:
I sort of what afraid that this case could happen for a short time. If I get it right, the gap would be:

* /standard has completed
* pipeline accepted the two installcheck: tests (build-success and repo-checker)

While the pkglistgen runs, /images is still disabled

Could we add another expected: test (like the installchecks) that gets confirmed by 'images enabled' in the pipeline?

DimStar77:
> Could we add another expected: test (like the installchecks) that gets confirmed by 'images enabled' in the pipeline?

That would actually also help in case pkglistgen fails and images is not being enabled at all

coolo:
That check would need to be on standard repository though.

coolo:
I did that now #2757 - the check is to be set to required once it's merged

PR #2384, opened 2020-01-14
Update Announcer for Leap 15.2

lkocman:
* Lubos Kocman is Release Manager for Leap since 1.1.2020

### DIFF
--- a/config/announcer/openSUSE:Leap:15.2.yml
+++ b/config/announcer/openSUSE:Leap:15.2.yml
@@ -30,4 +30,4 @@ iso: openSUSE-Leap-15.2-DVD-x86_64-Current.iso
 name: openSUSE:Leap:15.2
 subject: Leap 15.2 Build {version} released!
 url: http://download.opensuse.org/distribution/leap/15.2/iso/
-sender: Ludwig Nussel <ludwig.nussel@suse.de>
+sender: openSUSE release team <opensuse-releaseteam@opensuse.org>


lkocman:
@nilxam could you please review this one? Thank you

DimStar77:
:+1: 

But a few concerns:

* Are you subscribed to opensuse-factory so you can send mails to that list?
* Before you switch the RM sending out emails, I'd think it would be really nice if that new RM announced the switch on the mailing list, so that people are aware of a new person being in charge and not just happen to 'find out' by a different address sending out release announcement mails.

nilxam:
> But a few concerns:
> 
> * Are you subscribed to opensuse-factory so you can send mails to that list?
> * Before you switch the RM sending out emails, I'd think it would be really nice if that new RM announced the switch on the mailing list, so that people are aware of a new person being in charge and not just happen to 'find out' by a different address sending out release announcement mails.

ah, yes, these concerns sounds reasonable.

lkocman:
Let's postpone until there is official announcement set.  Thank you

coolo:
you better also pause the mail pipeline until this was done - to avoid mixed communication

lkocman:
I've discussed this topic and Ludwig had a nice idea of using the opensuse release team email. In this way we can visualize it as a team effort rather than individual.

coolo:
You still need to subscribe it - and please make sure it doesn't get email :)

coolo:
are  we done here? :)

lnussel:
the address is subscribed, go ahead

PR #2397, opened 2020-01-31
Staging accept: No longer set the version in .product files on accept

DimStar77:
In the past, we had a dashboard reading the version of the current source from the .product file.
Not immediately updating this version after a checkin would have been confusing as the dashboard would
have shown incorrect/outdated information.

the new dashboard (https://factory-dashboard.opensuse.org) reads the source version directly from the
attribute, so this mis-match is eliminated.

The .product files will be updated by pkglistgen in due time.

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -237,29 +237,13 @@ class AcceptCommand(object):
 
         project = self.api.project
         curr_version = date.today().strftime('%Y%m%d')
-        update_version_attr = False
-        url = self.api.makeurl(['source', project], {'view': 'productlist'})
 
-        products = ET.parse(http_GET(url)).getroot()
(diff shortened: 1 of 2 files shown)


DimStar77:
Meh - the test was simply ignoring that before - as there was not even a product to update (as far as I could find)

nilxam:
I just aware this change was updating ProductVersion attrib on Leap, `OSRT:ProductVersion` has changed to date time instead of 15.2, *luckily* that in pkglistgen it only going to replace version number if [the version option is undefined](https://github.com/openSUSE/openSUSE-release-tools/blob/f8ecad1d4612bf054d8d2cfa489679a0d543c24f/pkglistgen/tool.py#L563) where Leap has `<version>15.2</version>`

Issue #2402, opened 2020-02-11
ttm: Ignore "parallel_failed" fails in openQA


okurz:
## Motivation

label carry over in openQA works mainly on "failed" tests, not "parallel_failed" but ttm expects labels on "parallel_failed" openQA jobs as well meaning manual labelling work is required for repeated failures in a job that is part of a cluster with parallel jobs.

## Suggestion

As "parallel_failed" jobs do not have much helpful information by themselves (the important part is in the parallel job) I suggest to simply ignore jobs with result "parallel_failed" in ttm.

## Further details
For example dimstar needed to add a label manually here: https://openqa.opensuse.org/tests/1169725#comments

PR #2405, opened 2020-02-14
ttm: ignore parallel_failed jobs, but still list as failed

DimStar77:
Fixes https://github.com/openSUSE/openSUSE-release-tools/issues/2402

### DIFF
--- a/ttm/publisher.py
+++ b/ttm/publisher.py
@@ -78,7 +78,7 @@ class ToTestPublisher(ToTestManager):
                         # update reference
                         self.issues_to_ignore[ref] = build_nr
 
-                if ignored:
+                if ignored or job['result'] == 'parallel_failed':
                     self.failed_ignored_jobs.append(job['id'])
                     if labeled:
                         text = 'Ignored issue' if len(refs) > 0 else 'Ignored failure'


ttm/publisher.py
```diff
@@ -78,7 +78,7 @@ def overall_result(self, snapshot):
                         # update reference
                         self.issues_to_ignore[ref] = build_nr
 
-                if ignored:
+                if ignored or job['result'] == 'parallel_failed':
```
okurz: I wonder where other "uninteresting" jobs are ignored, e.g. "cancelled", etc.

But looks like it gets the job done :)

DimStar77: a canceled one is not uninteresting

okurz: Hm, so all others would count as "important" for TTM, including obsolete which should only happen when there is actually a new build so TTM would look that that one. That makes sense, yes.

DimStar77: Obsolete is only on 'replaced snapshots' - since we only rnu ttm against 'latest', it's not realistically happening.

In any case:
https://github.com/openSUSE/openSUSE-release-tools/blob/5728167f14fe3e72009e7f764a8455b3c8f5a6a9/ttm/publisher.py#L52

There we check for potential failed ones to handle.

An alternative option was to just remove parallel_failed from that list - but that in turn would mean it is not listed as 'failed' anymore in the ttm output; as I still want to see it (as ignored) I opted for the change as submitted

nilxam:
looks reasonable to me.

PR #2406, opened 2020-02-17
Release FTP repos before ISOs

coolo:
It takes a long time to sync the repos but ISOs are very quick. This lead
to the problem that openQA already started syncing the build while the
repo was synced up. As the repository is set to unpublished in between
the calls, openQA has no chance to see that more is to come.

So if we release the slow products first, openqa has a better chance to
catch up.


coolo:
CC @andrii-suse

PR #2419, opened 2020-03-16
No longer trigger obsrsync directly from rabbitmq messages

coolo:
Rely on rabbit-repoid git instead - only the first step


coolo:
It's obviously a regression, but I need to verify this step before the real fix

PR #2420, opened 2020-03-16
Sync script for obs_rsync

coolo:
This will be triggered on gocd if there are changes in the finished
repositories, checks the important repositories and notifies openQA


coolo:
CC @andrii-suse

PR #2423, opened 2020-03-18
Work around OBS issue 8994

coolo:
Instead of declining the by_project review, unselect the request and
decline the group review. This way we avoid the OBS problem


nilxam:
the 2nd rm_from_prj() https://github.com/openSUSE/openSUSE-release-tools/blob/master/osclib/stagingapi.py#L551 will print out 'no package or no request_id' is confusing @DimStar77 , harmless though

PR #2424, opened 2020-03-19
Support 'required' field for packages

### LINKED ISSUE #1844: pkglistgen: Support 'required' package role
For Factory stagings we want uninstallable "important" patterns to stop building Test DVDs. For Leap this is the default by mean of OSRT:Config's pkglistgen-ignore-unresolvable. Leaving it to openQA to detect KDE being uninstallable is a waste of resource

coolo:
Fixes #1844

### DIFF
--- a/pkglistgen.md
+++ b/pkglistgen.md
@@ -43,6 +43,8 @@ You can also adapt the solving on a package level by putting a hash into the pac
  Do not put the package into this group. Used to *force* certain packages into other modules
  * silent
  Use this package for dependency solving of groups "on top", but do not output the package for this group. Mainly to mark the product to use by adding release packages. Use with care, this breaks dependency chains!
+ * required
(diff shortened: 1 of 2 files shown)


DimStar77:
It seems not to work reliably as intended.

```[ 786s] Mar-24 03:56:06 <2> : [W]    patterns-containers-kubic_admin not found on i586```
(but 000package-groups has ```- patterns-containers-kubic_admin: [required,aarch64,x86_64]```

coolo:
hmm, where is that? I never said a combination with archs works :)

DimStar77:
Had this in Staging:D - where your 000package-groups submission was being staged. After identifying the fail, I unselected it to make the rest progress (can add to :O - the 'catch-all-failure Staging)

coolo:
no need

coolo:
the request was more to raise your interest and to see what other packages would be required

Issue #2428, opened 2020-04-20
Multibuild package spread across rings


DimStar77:
The openSUSE Release Tooling needs to support us with a new concept to split multibuild packages across rings (currently not yet possible - OBS support in the works)

OBS will introduce an ```excludebuild``` statement that can be used in prjconf to disable a specific build flavor in ring0 (e.g python3 to be disabled, python3:base enabled). But that will mean we will end up with openSUSE:Factory/python3 to be linkged to Ring0 AND ring1 (both having specific ecludebuild statements)

At the moment, this is 'invalid setup' for the staging tool - it would complain about a package found in multiple rings

adrianschroeter:
Just for reference: https://github.com/openSUSE/open-build-service/pull/9420

nilxam:
Another question in my mind is how would it looks like when the same package has inherited from different project, eg. letter staging(ring0 + ring1), maybe we need a solution for `staging freeze` command also.

PR #2432, opened 2020-04-29
allow multibuild package spreads across ring

nilxam:
* Allow multibuild package spreads across ring
* Remove dead code
* Do not add multibuild flavor package to the ring package list(frozenlinks)

Perfectly we should verify `excludebuild` macro on each ring, I'd take it as enhancement task for the future, for the current requirement this change should works well enough.

Fix https://github.com/openSUSE/openSUSE-release-tools/issues/2428


nilxam:
lets start fun with multibuild python3

PR #2434, opened 2020-04-30
miniobs: Fix database setup for rails 6

coolo:


coolo:
ok, this is 2 steps closer to a working CI again - the other failure I would fix in another PR

PR #2436, opened 2020-05-04
Update some ttm documentation

lnussel:


docs/totest.asciidoc
```diff
@@ -26,27 +26,85 @@ of the project openSUSE:Factory:ToTest, if the version is different
...
+      - x86_64
+take_source_from_product: true
+need_same_build_number: true
+set_snapshot_number: true
+test_subproject: ToTest
+EOF
+$ osc meta attribute MyProject -c -a OSRT:ToTestManagerConfig -F myattribute
+-------------------------------------------------------------------------------
+
+Create a 00Meta package in the main project for ttm to store some metadata:
```
coolo, line 91: I don't think ttm stores anything in 00Meta

lnussel: it did for me, otherwise I would not have created 00Meta:

`ttm/manager.py:        self.api.pseudometa_file_ensure(self.version_file(target), version,
ttm/manager.py:        return self.api.pseudometa_file_load(self.version_file('totest'))
ttm/publisher.py:            self.update_status('publishing', self.api.pseudometa_file_load(self.version_file('snapshot')))`

coolo: I see - that's `openSUSE:%(project)s:Staging/dashboard` otherwise

lnussel: It's the `pseudometa_package` setting in the config which is set to `:Staging/dashboard` for the main projects but defaults to `00Meta` https://github.com/openSUSE/openSUSE-release-tools/blob/027727ccf3d1d0d4dc16ce3f3e4954cb725e17a5/osclib/core.py#L322

PR #2438, opened 2020-05-11
Trigger rebuilds on zSystem as needed

dirkmueller:
I don't have access there to recover the build, which is in a pretty
bad state atm. lets see if the bot would have enough privileges to be
able to do so.


Vogtinator:
Fails when mirroring the `x86_64` arch, presumably because it's empty:

```
  File "./project-installcheck.py", line 154, in check_pra
    target_packages = catalog.get(directories[0], [])
AttributeError: 'NoneType' object has no attribute 'get'
[script-executor] Script completed with exit code: 1.
```

PR #2441, opened 2020-05-15
Fix crash in project-installcheck if there are no packages in the pra

Vogtinator:
Fix crash of the run added by https://github.com/openSUSE/openSUSE-release-tools/pull/2438


Vogtinator:
@coolo Did the OBS integration break? There's no project for this PR on OBS

coolo:
not sure what you're asking

Vogtinator:
The "OBS Package Build" status is stuck.

PR #2442, opened 2020-05-15
ignore packages that just have _channel inside

msmeissn:
this avoids lingering legal reviews of _channel like we see for kernel and livepatch submissions

(tag: business continuity 
tag: reduce marcus knowledge
)

### DIFF
--- a/legal-auto.py
+++ b/legal-auto.py
@@ -214,6 +214,9 @@ class LegalAuto(ReviewBot.ReviewBot):
                 continue
             if package == 'patchinfo' or package.startswith('patchinfo.'):
                 continue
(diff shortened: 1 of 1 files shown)


msmeissn:
hmm, this might track the package name, not cpontent

legal-auto.py
```diff
@@ -214,6 +214,9 @@ def _query_sources(self, project):
                 continue
             if package == 'patchinfo' or package.startswith('patchinfo.'):
                 continue
+	    # skip packages that have _channel inside
```
coolo, line 217: there appears to be an indenting problem

PR #2444, opened 2020-05-28
devel-project: "list" is the default command, Python 3.3 fix

lkocman:
Workaround for https://bugs.python.org/issue16308

kocman@localhost:~/Workspace/opensuse/openSUSE-release-tools> python3 devel-project.py 
  File "devel-project.py", line 300, in <module>
    main()
  File "devel-project.py", line 297, in main
    sys.exit(args.func(args))
AttributeError: 'Namespace' object has no attribute 'func'

### DIFF
--- a/devel-project.py
+++ b/devel-project.py
@@ -262,11 +262,10 @@ def common_args_add(parser):
     parser.add_argument('--repeat-age', type=int, default=7, metavar='DAYS', help='age after which a new reminder will be sent')
     parser.add_argument('--remind', action='store_true', help='remind maintainers to review')
 
-
-if __name__ == '__main__':
+def main():
     parser = argparse.ArgumentParser(description='Operate on devel projects for a given project.')
     subparsers = parser.add_subparsers(title='subcommands')
-
+    parser.set_defaults(func=list)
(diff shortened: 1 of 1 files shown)


lkocman:
CI Fails at   File "/usr/lib/python3.8/site-packages/osc/core.py", line 4238, in create_submit_request
    cgi.escape(message))
AttributeError: module 'cgi' has no attribute 'escape'

This seems to be not related to my change at all

coolo:
Can you rebase?

mvidner:
I have rebased this, but there is still a similar problem and I am not familiar enough with the APIs to fix it:

```console
$ python3 ./devel-project.py
...
utilities
vdr
windows:mingw:win32
windows:mingw:win64
zypp:Head
zypp:plugins
  File "./devel-project.py", line 300, in <module>
    main()
  File "./devel-project.py", line 297, in main
    sys.exit(args.func(args))
  File "./devel-project.py", line 74, in list
    if args.write:
AttributeError: 'Namespace' object has no attribute 'write'
```

It works fine with `python3 ./devel-project.py list`

mvidner (approved):
Having fixed this, I approve :)

PR #2448, opened 2020-06-09
factory-auto: add opensuse-review-team on submits to NonFree

DimStar77:
Packages submitted to openSUSE:*:NonFree did not end up on the queue
of the opensuse-review-team


DimStar77:
travis failure is unrelated - another PR already has the fix for that

Issue #2449, opened 2020-06-12
JUMP: supply pre-integration with project-wide execution of xcdhck tests


lkocman:
about 4k packages, essentially the media - KDE will now come from SLE. These packages will have binaries and sources synced 1:1 from IBS -> OBS on each build event.

This was also part of discussion with autobuild team https://build.suse.de/project/requests/SUSE:SLE-15-SP2:Update

Next step: Rudi would call our pipeline via a rest call on the IBS -> OBS sync event http://api.go.cd/current/#scheduling-pipelines

lkocman:
open

PR #2450, opened 2020-06-12
Initial version of xcdchk-testsuite for Jump

lkocman:
* Resolves Issue #2449

DO NOT MERGE

TODO 

- [x] make sure that it can be executed locally (I'm still getting some missing attr errors on python3).
- [ ] ask Thorsten if it's okay to share his tests on github (currently in gitlab)
- [ ] wrap kukuk/xcdchk
- [ ] handle dependencies on rpmdev-tools, libsolv-tools (installcheck)
- [ ] do not use /data as a data storage for xcdchk

Tool itself was moved here https://github.com/openSUSE/xcdchk

### DIFF
--- a/gocd/xcdchk.jump.opensuse.yaml
+++ b/gocd/xcdchk.jump.opensuse.yaml
@@ -1,7 +1,6 @@
 format_version: 3
-pipelines:
-  openSUSE.QA.Maint:
-    group: openSUSE.Checkers
+  XCDCHK.openSUSE_Jump:
+    group: openSUSE.XCDCHK
     lock_behavior: unlockWhenFinished
     environment_variables:
       OSC_CONFIG: /home/go/config/oscrc-xcdchk
--- a/gocd/xcdchk.jump.opensuse.yaml
+++ b/gocd/xcdchk.jump.opensuse.yaml
@@ -18,4 +18,4 @@ pipelines:
           - staging-bot
         tasks:
           - script: |-
-             ./xcdchk-testsuite.py
+             ./xcdchk.py
--- a/gocd/xcdchk.jump.opensuse.yaml
+++ b/gocd/xcdchk.jump.opensuse.yaml
@@ -8,9 +8,6 @@ pipelines:
     materials:
       script:
         git: https://github.com/openSUSE/openSUSE-release-tools.git
-    timer:
-      spec: 0 */10 * ? * *
-      only_on_changes: false
     stages:
     - Run:
         approval: manual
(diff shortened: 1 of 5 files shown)


gocd/xcdchk.jump.opensuse.yaml
```diff
@@ -0,0 +1,21 @@
...
+      script:
+        git: https://github.com/openSUSE/openSUSE-release-tools.git
+    timer:
+      spec: 0 */10 * ? * *
+      only_on_changes: false
+    stages:
+    - Run:
+        approval: manual
+        resources:
+          - staging-bot
```
coolo, line 18: staging-bot is for quick jobs. this sounds like an expensive job and as such should be in the repo-checker resource

gocd/xcdchk.jump.opensuse.yaml
```diff
@@ -0,0 +1,21 @@
+format_version: 3
+pipelines:
+  openSUSE.QA.Maint:
```
coolo, line 3: this pipeline is something else

xcdchk-testsuite.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/python2
```
coolo, line 1: there is for sure no python2 in gocd

gocd/xcdchk.jump.opensuse.yaml
```diff
@@ -0,0 +1,21 @@
...
+    group: openSUSE.Checkers
+    lock_behavior: unlockWhenFinished
+    environment_variables:
+      OSC_CONFIG: /home/go/config/oscrc-xcdchk
+    materials:
+      script:
+        git: https://github.com/openSUSE/openSUSE-release-tools.git
+    timer:
+      spec: 0 */10 * ? * *
```
coolo, line 12: it doesn't make sense to have this script run every 10 minutes - you want this as part of the release pipeline, not as a checker

xcdchk-testsuite.py
lkocman: fixed

gocd/xcdchk.jump.opensuse.yaml
lkocman: Fixed

gocd/xcdchk.jump.opensuse.yaml
lkocman: I did use a separate pipeline

  XCDCHK.openSUSE_Jump:
    group: openSUSE.XCDCHK

Does the group look like a good match?

coolo:
is this supposed to be continoued?

coolo:
I guess not

PR #2456, opened 2020-07-03
Sle15 sp3

gyr:
Adding 15-SP3 setup, I will make another PR to remove 15-SP2 setup as soon as I have a go from SLE RM's


coolo:
I'm not so happy with duplicating this, but I understand the timing issue

PR #2458, opened 2020-07-06
metrics/grafana/openqa: Add openSUSE Leap 15.1 and 15.2

okurz:
Not sure if there are further changes necessary, e.g. internal changes
to influxdb or whatever data source should provide the data.


coolo:
I seriously have no idea how it's deployed, but I can merge.

okurz:
Sure, maybe this will just bring dead source code up-to-date but even that shouldn't harm ;)

PR #2459, opened 2020-07-06
Update version attribute also in o:F:RISCV

andreas-schwab:


pkglistgen/update_repo_handler.py
```diff
@@ -199,6 +199,8 @@ def update_project(apiurl, project):
         key = str(key)
         if not opts['url'].endswith('/'):
             opts['url'] += '/'
+        if 'downloadcontent.opensuse.org' in opts['url']:
+            opts['url'].replace('downloadcontent.opensuse.org', 'download.opensuse.org')
```
DimStar77, line 203: We intentionally picked downloadcontent.o.o to not being redirected wildly across the internet to grab downloads for our update stacks. we know we have the stuff on our own mirror

andreas-schwab:
Sorry pushed the wrong branch.

PR #2470, opened 2020-08-31
pkglistgen: Add a way to merge files from 000update-repos

coolo:


coolo:
this is unmerged on purpose as I wanted to see how it moves out and then automate it further

coolo:
But for those who need to run this manually, the PR might be useful

coolo:
Automated now, ready to merge

PR #2476, opened 2020-09-04
TTM: Verify that all successfully built products are mentioned in the config

Vogtinator:
This is to avoid that added container images or appliance flavors are never
released due to the missing config addition.

This actually fails for `openSUSE:Factory` currently:

```
ERROR:ttm.manager:ceph-grafana-image is built for x86_64, but not mentioned as product
ERROR:ttm.manager:kubic-velero-image is built for x86_64, but not mentioned as product
ERROR:ttm.manager:openSUSE-MicroOS:ContainerHost-SelfInstall is built for x86_64, but not mentioned as product
ERROR:ttm.manager:openSUSE-Tumbleweed-Yomi:livecd is built for x86_64, but not mentioned as product
ERROR:ttm.manager:openSUSE-Tumbleweed-Yomi:pxeboot is built for x86_64, but not mentioned as product
ERROR:ttm.manager:opensuse-openldap-image is built for x86_64, but not mentioned as product
```

Same for `openSUSE:Jump:15.2`:
```
ERROR:ttm.manager:kiwi-templates-JeOS:MS-HyperV is built for x86_64, but not mentioned as product
ERROR:ttm.manager:kiwi-templates-JeOS:OpenStack-Cloud is built for x86_64, but not mentioned as product
ERROR:ttm.manager:kiwi-templates-JeOS:VMware is built for x86_64, but not mentioned as product
ERROR:ttm.manager:kiwi-templates-JeOS:kvm-and-xen is built for x86_64, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:docker is built for x86_64, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:lxc is built for x86_64, but not mentioned as product
ERROR:ttm.manager:kiwi-templates-JeOS:RaspberryPi is built for aarch64, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:docker is built for aarch64, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:lxc is built for aarch64, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:docker is built for ppc64le, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:lxc is built for ppc64le, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:docker is built for s390x, but not mentioned as product
ERROR:ttm.manager:opensuse-leap-image:lxc is built for s390x, but not mentioned as product
DEBUG:ttm.manager:config incomplete
```

The test returns successfully on `openSUSE:Factory:ARM`, `openSUSE:Factory:PowerPC`, `openSUSE:Factory:zSystems`, `openSUSE:Factory:WSL`, `openSUSE:Leap:15.2:Images`, `openSUSE:Leap:15.1:Images` and `openSUSE:Leap:15.2`.

### DIFF
--- a/ttm/releaser.py
+++ b/ttm/releaser.py
@@ -47,6 +47,10 @@ class ToTestReleaser(ToTestManager):
                 self.logger.debug('not snapshotable')
                 return QAResult.failed
 
+            if not self.all_built_products_in_config():
+                self.logger.debug('config incomplete')
+                return QAResult.failed
+
         self.update_totest(new_snapshot)
         self.update_status('testing', new_snapshot)
         self.update_status('failed', '')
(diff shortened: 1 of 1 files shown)


DimStar77:
IF already going that route, we should just eliminate the need to list them in TTM config at all.

DimStar77:
Having this done automatically would also eliminate the problem that updating the ttm config while a snapshot is being tested is a bad idea, as then publishing said snapshot does not work. (missing images on publish)

Vogtinator:
> IF already going that route, we should just eliminate the need to list them in TTM config at all.

There are some topics like:
* Is it an appliance or a container?
* Release the multibuild container or the flavors separately? In the `JeOS` case, that's actually not that simple...
* That means that *any* failure in `product_repo` is fatal

> Having this done automatically would also eliminate the problem that updating the ttm config while a snapshot is being tested is a bad idea, as then publishing said snapshot does not work. (missing images on publish)

That could actually be fixed by just ignoring nonexisting images in `:ToTest`

DimStar77:
Ok, let's take this as a good starting point to not miss images being published. definitively better than what we have now.

Vogtinator:
> > IF already going that route, we should just eliminate the need to list them in TTM config at all.

I thought about that some more and it's both a bit easier and a bit harder.

> There are some topics like:
> 
> * Is it an appliance or a container?

Would need one API call to check for a `.containerinfo` file in the build artifacts, so not too bad.

For publishing, ttm has to release all container products again from `:ToTest`, but knowing which products are containers requires either:
- Going through the build project again and checking there, which means the product repo has to be green for publishing (which is not desirable)
- Going through all binaries in `:ToTest` and releasing everything which looks like a container. To avoid releasing deleted/stale containers, it would have to delete all stale containers in `:ToTest` before releasing, which means that simply build disabling something in the build project effectively deletes it. As build failures are fatal, build disabling is the only way to work around that, but with that mechanism that's not working anymore.
- Saving the list of containers during release and loading it on publish (new attribute?)

> * Release the multibuild container or the flavors separately? In the `JeOS` case, that's actually not that simple...

The design of binary releasing in OBS actually forces a decision here:
- Releasing the multibuild container itself releases the sources and all binaries of all flavors
- Releasing a particular flavor releases the sources and the binaries for that flavor

So if the multibuild container itself is not `excluded` or `disabled`, it has to be released in its entirety anyway.
If not, the flavors can be handled like individual packages.

> * That means that _any_ failure in `product_repo` is fatal

This is probably a blocker for this. Especially for `openSUSE:Factory:ARM`, one of the dozens of `JeOS` images fails to build for whatever reason occasionally and currently it just uses the old binaries again.

nilxam (approved):
Looks a reasonable change to me, let's try it then.

PR #2483, opened 2020-10-06
Check git errors in rabbit-repoid

andrii-suse:


okurz (approved):

coolo:
unfortunately we rely on this command to check if we changed something, but git commit exists 1 if there is nothing to do - and so the pipeline stops

coolo:
So I guess changing to `git add . && if git commit -m "{}"; then git push; fi` ?

andrii-suse:
#2484

PR #2488, opened 2020-10-20
Add back setup for 15SP1 quarterly updates

gyr:
Maintenance asked to add back 15SP1 quarterly updates because they have a request to update it

### DIFF
--- a/gocd/checkers.suse.gocd.yaml.erb
+++ b/gocd/checkers.suse.gocd.yaml.erb
@@ -281,6 +281,35 @@ pipelines:
             - staging-bot
             tasks:
             - script: ./legal-auto.py -A https://api.suse.de --debug --legaldb http://legaldb.suse.de project $(cat /home/go/config/legal-auto-projects-suse)
+  SLE15.SP1.QR:
+    group: SLE.Checkers
+    lock_behavior: unlockWhenFinished
+    environment_variables:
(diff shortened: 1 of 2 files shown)


gocd/checkers.suse.gocd.yaml.erb
```diff
@@ -281,6 +281,35 @@ pipelines:
             - staging-bot
             tasks:
             - script: ./legal-auto.py -A https://api.suse.de --debug --legaldb http://legaldb.suse.de project $(cat /home/go/config/legal-auto-projects-suse)
```
coolo, line 283: please program this as loop then

coolo (changes requested):

gocd/checkers.suse.gocd.yaml.erb
gyr: sorry, but did you mean by "loop"

coolo: Well, replace the copy&pasted code with a loop over SP1 and SP2. this is ruby

coolo: Take inspiration from 

https://github.com/openSUSE/openSUSE-release-tools/blob/6956f3b65580b41069282186c99ae254aa83bfe7/gocd/announcer.gocd.yaml.erb#L3

coolo (approved):

PR #2494, opened 2020-11-18
Support for Dockerfile builds

Vogtinator:
Draft because:

- [x] Needs to be tested
- [x] Repo name wasn't discussed yet

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -186,7 +186,8 @@ class CheckSource(ReviewBot.ReviewBot):
         shutil.rmtree(os.path.join(target_package, '.osc'))
 
         new_info = self.package_source_parse(source_project, source_package, source_revision, target_package)
-        if not new_info.get('filename', '').endswith('.kiwi') and new_info['name'] != target_package:
+        filename = new_info.get('filename', '')
+        if not (filename.endswith('.kiwi') or filename == 'Dockerfile') and new_info['name'] != target_package:
             shutil.rmtree(dir)
             self.review_messages['declined'] = "A package submitted as %s has to build as 'Name: %s' - found Name '%s'" % (target_package, target_package, new_info['name'])
             return False
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1384,18 +1384,29 @@ class StagingAPI(object):
             linkproject = ''
             repository = '<repository name="standard">'
 
+        # Add "images" and "containerfile" repos if the main project has them.
+        # If both are present, they build against each other and the parent project.
+
         images_repo = ''
         if self.project_has_repo('images'):
+            containerfile_path = ''
+            if self.project_has_repo('containerfile'):
+                containerfile_path = f'<path project="{name}" repository="containerfile"/>'
+
             images_repo = f"""
                <repository name="images">
                  <path project="{name}" repository="standard"/>
+                 {containerfile_path}
                  <path project="{self.project}" repository="images"/>
                  <arch>x86_64</arch>
               </repository>"""
 
         containerfile_repo = ''
         if self.project_has_repo('containerfile'):
-            images_path = f'<path project="{name}" repository="images"/>' if images_repo else ''
+            images_path = ''
+            if self.project_has_repo('images'):
+                images_path = f'<path project="{name}" repository="images"/>'
+
             containerfile_repo = f"""
                <repository name="containerfile">
                  <path project="{name}" repository="standard"/>
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -142,17 +142,16 @@ class StagingAPI(object):
         raise Exception("setting packages_staged is not allowed")
 
     @property
-    def use_images(self):
-        # Determine if staging is bootstrapped.
-        meta = self.get_prj_meta(self.project)
-        xpath = 'repository[@name="images"]'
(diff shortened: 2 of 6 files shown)


dirkmueller (commented):
LGTM

dirkmueller:
personally if I may start the bikeshedding, I would prefer repository name "containers" over "docker". (for some reason we picked plural for images). and whether its its docker build or podman or buildah or whatever imho shouldn't matter, it is specific to "Dockerfile".

Vogtinator:
> personally if I may start the bikeshedding, I would prefer repository name "containers" over "docker". (for some reason we picked plural for images)

I'm against `containers` for the simple reason that there are other repo types which also build containers, like `images` with kiwi.

> and whether its its docker build or podman or buildah or whatever imho shouldn't matter, it is specific to "Dockerfile".

What matters most for a repo name is what it builds. For `standard` it's `Type: rpm`, for `images` it's `Type: kiwi` and for `docker` it's `Type: docker`. `dockerfile` would also work.

dirkmueller:
we didn't call repository "standard" as "rpm" either, so I'm not sure there should be an alignment between repository name and repository build type (there isn't one in what you quote as examples). 

I'd simply avoid the use of a trademarked name, hence some other word should be used instead.

Vogtinator:
Ping @DimStar77.

DimStar77:
'containers' seems a bit too broad, especially as we already build lxc and .docker containers (based on kiwi) in 'images' (where we also  have appliances - something TTM is splitting up for us again)

So, I somewhat don't like 'containers' as a repo name on oS:F -having it more relevant to docker makes more sense, docker, docks, docker_containers

Vogtinator:
Staging parts work: https://build.opensuse.org/staging_workflows/home:favogt:releasetest

`docker` builds can be based on `images` and vice versa.

TTM testing is next.

Vogtinator:
> Staging parts work: https://build.opensuse.org/staging_workflows/home:favogt:releasetest
> 
> `docker` builds can be based on `images` and vice versa.
> 
> TTM testing is next.

Both release and publish worked. An incomplete list `docker` products triggers an error as expected and without the list the behaviour is unchanged.

Vogtinator:
Forgot to mention, the `check_source.py` change works as well.

dirkmueller:
> 'containers' seems a bit too broad, especially as we already build lxc and .docker containers (based on kiwi) in 'images' (where we also have appliances - something TTM is splitting up for us again)

@DimStar77 well, I need a generic name that doesn't refer to "docker" because thats a trademark, and that says its a Dockerfile build (which could be docker, it could also be built with buildah etc).

"dockerfile" as repository would also work for me, but thats weird (we don't all "standard" repository as "spec" either, so its inconsistent). 

last suggestion "oci_images" ?

Vogtinator:
> > 'containers' seems a bit too broad, especially as we already build lxc and .docker containers (based on kiwi) in 'images' (where we also have appliances - something TTM is splitting up for us again)
> 
> @DimStar77 well, I need a generic name that doesn't refer to "docker" because thats a trademark, and that says its a Dockerfile build (which could be docker, it could also be built with buildah etc).

OBS requires `Type: docker` for those, which is unavoidable. Feel free to file an issue at https://github.com/openSUSE/open-build-service/ to get that renamed if you think this is an issue.

> "dockerfile" as repository would also work for me, but thats weird (we don't all "standard" repository as "spec" either, so its inconsistent).
>
> last suggestion "oci_images" ?

That's actually wrong in two ways: docker tarballs aren't OCI images and kiwi (`images`) can also build OCI images.

dirkmueller:
the "Type: docker" is more hidden I think, so less of an issue imho. I'd be fine with "dockerfile" , you didn't comment on that?

Vogtinator:
> the "Type: docker" is more hidden I think, so less of an issue imho.

Repo names are also not that visible, so I don't really see much difference there.

> I'd be fine with "dockerfile" , you didn't comment on that?

I suggested that myself in https://github.com/openSUSE/openSUSE-release-tools/pull/2494#issuecomment-729916396.

DimStar77:
> > I'd be fine with "dockerfile" , you didn't comment on that?
> 
> I suggested that myself in [#2494 (comment)](https://github.com/openSUSE/openSUSE-release-tools/pull/2494#issuecomment-729916396).

So that seems like 'dockerfile' is the winner everybody can live with?

Vogtinator:
It should be noted that at least podman/buildah is going in the direction of preferring `Containerfile` meanwhile: https://github.com/containers/buildah/pull/1860

OBS doesn't support that so far, but the repo name could be future proof...

dirkmueller:
I'm okay with "containerfile", it solves the "docker" issue which is the part I was mostly concerned about.

aginies:
Any progress? sounds like there is an agreement for "containerfile", we need to move forward, this is currently blocking us. thx.

Vogtinator:
> I'm okay with "containerfile", it solves the "docker" issue which is the part I was mostly concerned about.

For some reason I didn't get a notification for this comment...

I changed the repo name to `containerfile` everywhere and retested adi stagings and ttm, still working fine.

Vogtinator:
[14:00] <DimStar> fvogt: let's break things :)

So: Merge time!

PR #2498, opened 2020-11-26
QR: add stage to accept request create in IBS

gyr:


gocd/checkers.suse.gocd.yaml.erb
```diff
@@ -311,4 +311,11 @@ pipelines:
                 sed -e 's,home:maintenance-installcheck:15-<%= servicepack %>:QR,SUSE:SLE-15-<%= servicepack %>:Update:QR,' -i *
                 osc ci -m 'Fix repoid'
                 osc sr --yes SUSE:SLE-15-<%= servicepack %>:Update:QR -m 'Update Package List'
+    - Accept:
+        resources:
+        - repo-checker
```
coolo, line 316: you don't need a repo checker resource to call osc request.

But what's the point of this? I don't think the maintenance-installcheck user should have maintainer access to SUSE:SLE-15

admust: The idea behind that, to have a task to approve the request create via the job if the job finished successfully to merge changes to :QR project automatically

coolo: If that was goal you wouldn't need to create a SR at all - but have random scripts break your builds at random times. The goal here is to have a bot doing the work and then a human to review and accept it. And the bots shall not have write permissions to SUSE products.

admust: I'm not aware of what the script does, but from using it I found it make some changes related to packages has a different version or update kernel version released on group files that couldn't figure without a huge manual work.

I'm not sure which cases could crash the product build but from my experience, I didn't receive any issue for missing package in QU ISOs 

if there something we need to test or cover it please let me know

coolo:
I close this, it just doesn't make sense to me

PR #2499, opened 2020-11-30
another try to filter SUSE:Channels submits

msmeissn:
my last attempts did not seem to work ...


coolo:
you're aware you can dry run this on requests?

msmeissn:
i will debug this first

PR #2501, opened 2020-12-02
Don't produce directory conflicts of ghost files

lnussel:


lnussel:
https://build.opensuse.org/request/show/851301

coolo:
I'm torn. If package A got /var/log/messages as %ghost because it's generated and removed in %post(un) and then another package B adds it as %file, removing B will remove it also for A.

coolo:
@mlschroe what do you think?

lnussel:
The exception here is for directories only. Still, even for the example the bug would be the scriptlet then, not the file list per se, right?

The reason this came up in the first place is https://bugzilla.opensuse.org/show_bug.cgi?id=1179097 ie aaa_base owning legacy files in /etc/init.d. Those are not created nor removed in scriptlets, quite the contrary, they must not be removed. There's a historic mistake to flag them as %ghost rather than %ghost %config(noreplace). So we need to keep the latter around for a while. At the same time aaa_base is not suppose to even create /etc/init.d to avoid suggesting to use that dir for anything. That's why I flagged the dir also as %ghost.

lnussel:
ping?

coolo:
what's the point of a %config(noreplace) on a directory? I guess I don't get your explanation. But I guess I don't get why you would %ghost the directory anyway - if you don't create it in scriptlets.

lnussel:
The directory is just %ghost, not %config. The motivation is to get rid of remaining sysvinit bits in the system, ie not create anything related to sysv anymore. That includes references to /etc/init.d/{boot,after}.local. The entries from aaa_base were not marked %config before though. So if we'd get rid of those entries in the filelist, rpm would delete admin created files. So we either need to keep them as %config for a while, or use some scripting tricks to back the files up and move them back. Scripts do not know about file lists before and after, so would likely have to apply the trick on every single aaa_base update, or leave some trigger file etc. No beauty in that.
So the request at hand goes the %config(norplace) route for those .local files. Looks good, but now the bots and checks maze run  starts:

```
$ osc diff
Index: aaa_base.spec
===================================================================
--- aaa_base.spec       (revision 715)
+++ aaa_base.spec       (working copy)
@@ -201,7 +201,6 @@
 /etc/profile.d/ls.zsh
 %config /etc/shells
 %config /etc/ttytype
-%ghost %dir /etc/init.d
 %ghost %config(noreplace) /etc/init.d/boot.local
 %ghost %config(noreplace) /etc/init.d/after.local
 %ghost %config /etc/inittab
% osc build ..
[...]
[   82s] ... checking filelist
[   82s] aaa_base-84.87+git20201123.4f16b16-0.x86_64.rpm: directories not owned by a package:
[   82s]  - /etc/init.d
[   83s] 
[   83s] tanana failed "build aaa_base.spec" at Fri Dec 11 08:16:13 UTC 2020.
```

Leaving that in makes  `findfileconflicts` complain. Catch-22.

The file conflict check isn't a real one though. Both packages it complains about contain the directory with the same mode. The difference is in the rpm file flags only, ie one is %ghost, the other isn't. So the patch I posted here just zaps the ghost flag as it just doesn't lead to a conflict in reality.

coolo:
Looks like @mlschroe has no oppinion either

lnussel:
gracias

PR #2507, opened 2020-12-10
check_source.pl: One .changes file per package is enough

Vogtinator:
It's actually fine to have a single .changes file for multiple .spec files
in the same package, OBS picks that automatically. Relax the check to only
require a single .changes file for the package name.

Many multi-spec packages currently just duplicate the changes files anyway,
which can be avoided.

### DIFF
--- a/check_source.pl
+++ b/check_source.pl
@@ -75,17 +75,17 @@ my $changes_updated = 0;
 for my $spec (@specs) {
     $changes = $spec;
     $changes =~ s/\.spec$/.changes/;
-    if (!-f "$dir/$changes") {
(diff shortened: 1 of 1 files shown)


DimStar77:
obs-source_validator should use the same rule then - currently that one nags (warns) about missing changes files too

Vogtinator:
> obs-source_validator should use the same rule then - currently that one nags (warns) about missing changes files too

https://github.com/openSUSE/obs-service-source_validator/pull/88

Vogtinator:
> > obs-source_validator should use the same rule then - currently that one nags (warns) about missing changes files too
> 
> [openSUSE/obs-service-source_validator#88](https://github.com/openSUSE/obs-service-source_validator/pull/88)

^ got merged already!

PR #2508, opened 2020-12-12
Update aggregate.php

coogor:
Prepare for Leap 15.3


coolo:
For the record: I have no idea if and how this is deployed

PR #2509, opened 2020-12-15
Fix staging-installcheck.py being ignored if i586 is present

Vogtinator:
For openSUSE:Factory:Staging:adi:*, i586 is only enabled when necessary and so
the check is not set as required. Currently x86_64 is handled first (due to
target_archs), which means that for oS:F, i586 comes last and so the ugly hack
results in x86_64 getting a fake success and the actual status is set for i586
only, which gets ignored.

### DIFF
--- a/staging-installcheck.py
+++ b/staging-installcheck.py
@@ -236,8 +236,7 @@ class InstallChecker(object):
     def report_state(self, state, report_url, project, repository, buildids):
         architectures = self.target_archs(project, repository)
         for arch in architectures:
-            self.report_pipeline(state, report_url, project, repository, arch,
-                                 buildids[arch], arch == architectures[-1])
+            self.report_pipeline(state, report_url, project, repository, arch, buildids[arch])
 
     def gocd_url(self):
         if not os.environ.get('GO_SERVER_URL'):
@@ -262,15 +261,9 @@ class InstallChecker(object):
         return self.api.makeurl(['status_reports', 'built', project,
                                  repository, architecture, 'reports', buildid])
 
-    def report_pipeline(self, state, report_url, project, repository, architecture, buildid, is_last):
+    def report_pipeline(self, state, report_url, project, repository, architecture, buildid):
         url = self.report_url(project, repository, architecture, buildid)
         name = 'installcheck'
-        # this is a little bit ugly, but we don't need 2 failures. So save a success for the
-        # other archs to mark them as visited - pending we put in both
-        if not is_last:
-            if state == 'failure':
-                state = 'success'
-
         xml = self.check_xml(report_url, state, name)
         try:
             osc.core.http_POST(url, data=xml)


Vogtinator:
Seems to work: https://build.opensuse.org/staging_workflows/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:adi:29

DimStar77 (approved):
ok from my end.

coolo:
if i586 is required or not doesn't really matter - if it fails, it fails. So is in your case i586 succeeding? 

What the code is trying to do is avoiding a double installcheck fail in the dashboard. And I still think that's valuable - even though the OBS team made sure the visual impression of the dashboard no longer matters.

coolo:
that adi:29 succeeds on i586 makes sense - as it's only about -32bit packages.

Vogtinator:
> if i586 is required or not doesn't really matter - if it fails, it fails. So is in your case i586 succeeding?

No, it's marked as failed, but the staging doesn't care. There's only one checkmark for `installcheck` and it's green

coolo:
That does sound like a severe bug of the dashboard code then. Because the checks are per arch - and so their status shouldn't be "merged".

coolo:
Oh, I know - because the i586 repository doesn't have *any* required checks, it's status isn't even read. That's a rather suprising performance optimisation.

So basically: no matter if we put something into i586 or not, it will always be ignored - unless we also require it.

Vogtinator:
> Oh, I know - because the i586 repository doesn't have any required checks, it's status isn't even read. That's a rather suprising performance optimisation.
>
> So basically: no matter if we put something into i586 or not, it will always be ignored - unless we also require it.

Yep, that's my theory as well. Do you have a link to the code? I'd argue it's a bug there.

IMO the optimization here is too fragile and ugly, just to avoid having two red marks. So I'd prefer to drop it in any case.

coolo:
it still won't matter for i586 failures - and I do care about the two installchecks. this is not just adi

coolo:
but I agree that the last_arch hack is a little fragile - it should at least check that the other archs are failed as well.

Vogtinator:
> it still won't matter for i586 failures - and I do care about the two installchecks. this is not just adi

With this PR, installcheck will set all archs to failed if any one fails, so it would be sufficient to just mark one installcheck as required. That way it would only show up once, like in adis.

coolo:
The problem is that the repository is per arch - this is not to be confused with published repositories. So there is only one installcheck - but it exists in 2 repositories architectures.

Vogtinator:
> The problem is that the repository is per arch - this is not to be confused with published repositories. So there is only one installcheck - but it exists in 2 repositories architectures.

Which effectively means that the result for all repo arches is the same, and so only one of them is necessary. The ugly hack removed here made all but one of them report "success" to work around the duplicate result display, but that could be achieved simply by hiding the others on the OBS side, by configuring them as not required.

coolo:
it's not only about success/failure - it's also about pending. We want to know if installcheck for i586 didn't yet run. But yes, only if it's required it's seen (which is the case for !adi)

coolo:
and we agree that i586 in adi is an exception atm

Vogtinator:
IMO correct results are much more important than cosmetics, like having multiple installcheck error links.

For now I'd be fine with a workaround though, like dropping the hack in `target_archs` and sorting the other way around, so that `x86_64` comes last.

coolo:
I get that - but if you want multiple installchecks, you need to set it as required

Vogtinator:
What I care about is that this hack makes installcheck lie. It writes that for a specific architecture it passes, although it failed.

The only issue with the current status + this PR would be that there are to links to installchecks in the staging overview page, which is IMO a tiny issue compared to the other options here.

It would be possible to enable builds for all main project architectures in all adi stagings, which fulfills the assumptions made by this check again. That doesn't make this hack any better though.

coolo:
I'm not saying the hack is correct - I'm just saying removing it is fixing it :)

coolo (approved):
we should redo the "eliminate all but one failure" feature differently

PR #2515, opened 2021-01-04
gocd: enable staging bot and staging installcheck on SLE15 SP3 Backports

nilxam:


nilxam:
tested with https://build.opensuse.org/staging_workflows/openSUSE:Backports:SLE-15-SP3

nilxam:
let's do it!

PR #2519, opened 2021-01-18
Accept: Don't crash on actions that aren't submit/delete

### LINKED ISSUE #2381: staging accept: fails if there is a chgdev in state 'new'
```
Accepting all acceptable projects
  File "/usr/bin/osc", line 41, in <module>
    r = babysitter.run(osccli)
  File "/usr/lib/python3.7/site-packages/osc/babysitter.py", line 62, in run
    return prg.main(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 344, in main
    return self.cmd(args)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 367, in cmd
    retval = self.onecmd(argv)
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 501, in onecmd
  File "/usr/lib/python3.7/site-packages/osc/cmdln.py", line 1232, in _dispatch_cmd
  File "/home/dimstar_suse/.osc-plugins/osc-staging.py", line 426, in do_staging
    cmd.accept_all(args[1:], opts.force, not opts.no_cleanup)
  File "/home/dimstar_suse/.osc-plugins/osclib/accept_command.py", line 100, in accept_all
    self.requests[req['type']].append(req['package'])
KeyError: 'change_devel'
```

coolo:
If a change_devel request is grouped into a staging, it shouldn't crash
on it - the API will accept it, and we have no specific action for this
type

Fixes #2381

### DIFF
--- a/osclib/accept_command.py
+++ b/osclib/accept_command.py
@@ -92,12 +92,15 @@ class AcceptCommand(object):
 
             staging_packages[project] = []
             for request in status.findall('staged_requests/request'):
-                self.requests[request.get('type')].append(request.get('package'))
+                type = request.get('type')
+                if type in self.requests:
+                    self.requests[type].append(request.get('package'))
                 staging_packages[project].append(request.get('package'))
 
         other_new = self.find_new_requests(self.api.project)
         for req in other_new:
-            self.requests[req['type']].append(req['package'])
+            if req['type'] in self.requests:
+                self.requests[req['type']].append(req['package'])
 
         print('delete links to packages pending deletion...')
         self.delete_linked()


DimStar77:
Staging does not really make sense on change_devel imho.

My workflow there usually is:

* Staging list tells me there are change_devel requests (good, otherwise they tend to be 'lost')
* I approve the 'change' from a RelMgr PoV - i.e. accept the Staging review for factory-staging
* If no other reviews are pending,I am generally 'accepting to Factory' right away (causes no rebuilds, only metadata)
The bad part is if the 'devel project' has a pending review (i.e a non-maintainer initiated the devel change to a better fitting, better maintained project)- Then I can't (without overriding) accept therequest right away and generally want to give the maintainers to make a 'peaceful transition of power'  for the package. So waiting for ack/nak from the current devel prj sounds fair.

But in this case, when the devel prj maintainer clicks ok, we are now in state new - which, stagnig accept does not like (the pass through staging would probably allow to circumvent this - but this adds in normal cases too much delay imho)

coolo:
It still shouldn't crash

DimStar77:
> It still shouldn't crash

Indeed - preferably it would accept the ones in 'state=new'  (without crash)

coolo:
Hmm, I gently ignored change_devel requests here. If you say, those should still be accepted, I need to change it

PR #2520, opened 2021-01-19
pkglistgen: Check the return code of Popen calls

### LINKED ISSUE #2517: pkglistgen: no error handling for bs_mirrorfull
https://botmaster.suse.de/go/tab/build/detail/Pkglistgen.openSUSE_Factory/9245/pkglistgen/1/openSUSE_Factory_rings

    DEBUG:pkglistgen.tool:updating /home/go/.cache/openSUSE-release-tools/repository-meta/openSUSE:Factory:Rings:1-MinimalX/standard/x86_64
    connect to api.opensuse.org:443: Connection timed out

Afterwards generates a mostly empty solv file.
Looks like there is no error handling in https://github.com/openSUSE/openSUSE-release-tools/blob/deab50eb50ee7658ee3ff29a166833c7d6eb9039/pkglistgen/tool.py#L354

coolo:
Not updating the product in case of failure is preferred over creating bogus products

Fixes #2517

### DIFF
--- a/pkglistgen/tool.py
+++ b/pkglistgen/tool.py
@@ -351,9 +351,11 @@ class PkgListGen(ToolBase.ToolBase):
                 args.append('--nodebug')
                 args.append('{}/public/build/{}/{}/{}'.format(self.apiurl, project, repo, arch))
                 args.append(d)
-                p = subprocess.Popen(args, stdout=subprocess.PIPE)
-                for line in p.stdout:
-                    self.logger.info(line.decode('utf-8').rstrip())
+                with subprocess.Popen(args, stdout=subprocess.PIPE) as p:
+                    for line in p.stdout:
+                        self.logger.info(line.decode('utf-8').rstrip())
+                    if p.wait() != 0:
+                        raise Exception("Mirroring repository failed")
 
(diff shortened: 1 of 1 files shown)


coolo:
Tested locally that it catches if I kill -SEGV the perl process

DimStar77:
I think I'd like to see a 'broken/failed' product build instead of something that might accidentally still build based on old definitions, just because we could not update the product

coolo:
The question is how often we see'd temporary OBS burp causing a reset of the product file

coolo:
Taking that the pipeline runs every hour, the temporarly broken DVD would only be there for a short time. But if we reset it, we could very well cancel a running FTP build just because OBS was busy the 2nd hour after the update. I don't think you want that.

DimStar77:
hm.. good point - so if the DVD is tempo miscalculated due to such a failure, we should generally see a 2nd timeslot still in the active time while the FTP tree is still building - that one would hopefully be ok then.

coolo:
Yeah, having the pipeline status on the factory dashboard might be a good way around - so you see it easily if you cared. But that's an enhancement ticket and I don't do those :)

PR #2530, opened 2021-02-18
accept_command: use the value of attribute to set ProductVersion if always_set_productversion_to option presents

nilxam:


osclib/accept_command.py
```diff
@@ -239,7 +241,11 @@ def update_factory_version(self):
         # missleading.
 
         project = self.api.project
-        curr_version = date.today().strftime('%Y%m%d')
+        pred_productversion = self.config.get('always_set_productversion_to', '')
```
coolo, line 244: as noone will be able to discover this, how about you add it to osclib/conf.py's DEFAULT?

And happy new year!

coolo:
So this wasn't relevant after all?

nilxam:
sry, I forgot this since was concentrating on 15.3, PR updated.

Issue #2440, opened 2020-05-14
Stage non-ring delete requests in adi


DimStar77:
So far, we stage all delete requests (ring and non-ring) in letter stagings. This is not really useful (often they need more coordination, and tend to be moved around between stagings manu times)

I've started tomanually move delete requests for non-ring packages into adi stagings - which makes them easier to handle, and not blocking letters.

We should adjust our tooling to allow to automatically put delete requests into adi stagings, including splitting them out from existing  adi groups into new groups (currently, the tool only allows to select a del req into an existing adi staging)

There is a minor caveat: installcheck does not get a new repo id and does not ever recheck the adi if it fails on first attempt. So far, I work around this by linking in openSUSE-releases into the adi - which results in a daily rebuild of the staging and allows the staging to be retested. The final solution for that part would be a time-based recheck, which requires https://github.com/openSUSE/open-build-service/issues/9295 to be implemented

PR #2531, opened 2021-02-23
adi: stage Non-Ring delete requests in adi projects

DimStar77:
This is now acceptable as the installcheck bot verifies that we
cannot remove binaries we still care for.

Fixes https://github.com/openSUSE/openSUSE-release-tools/issues/2440


nilxam (approved):
We tried to stage DR to adi staging before and haven't seen any critical problem, I'm fine with it.

PR #2532, opened 2021-02-24
Add TTM config for Leap 15.3 WSL

lkocman:
Missing ttm configuration for WSL/15.3


coolo (changes requested):
That file is generated from ttm.gocd.yaml.erb

coolo:
Reopen if you have news

PR #2535, opened 2021-03-11
 build-fail-reminder: don't use cgi.escape, but rather html escape the fixed string directly

DimStar77:


build-fail-reminder.py
```diff
@@ -188,7 +188,7 @@ def main(args):
                 # Package failed to build for 6 weeks - file a delete request
                 r = osc.core.Request()
                 r.add_action('delete', tgt_project=project, tgt_package=package)
-                r.description = cgi.escape("[botdel] Package has failed to build for >= 6 weeks")
+                r.description = escape("[botdel] Package has failed to build for >= 6 weeks")
```
coolo, line 191: just using `&gt;= 6 weeks` is out of style? :)

DimStar77: sure; is an option - I don't mind changing it and not using escape (was just a 1 for 1 replacement without brain muscle)

DimStar77:
after we merge / deploy this, we should set the counter of everything that is >= 6 to 5 - to make sure the botdel action fires (in the cache file that is, of course)

coolo (approved):

coolo:
So tell me about that cache file

PR #2536, opened 2021-03-15
findfileconflicts: handle usrmerge

lnussel:
When a filesystem package with usrmerge is detected (/bin, /sbin, /lib
or /lib64 link to /usr), pretend all files in the affected dirs are
actually in /usr.

### DIFF
--- a/findfileconflicts
+++ b/findfileconflicts
@@ -3,14 +3,32 @@
 $| = 1;
 use strict;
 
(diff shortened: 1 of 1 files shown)


coolo:
I don't know that code, sorry - nor do I have the time to understand what Ludwig is trying to achieve here.

lnussel:
augen zu und durch? :)

PR #2537, opened 2021-03-17
Add Leap 15.3 Maintenance configuration

foursixnine:
Following: https://progress.opensuse.org/issues/72202

For now I don't know how to test it, so directions would be great :)

### DIFF
--- a/data/repos.json
+++ b/data/repos.json
@@ -28,5 +28,19 @@
             "http://download.opensuse.org/update/leap/15.2/non-oss/"
          ]
       }
-   }
+   },
+   "openSUSE:Leap:15.3:Update" : {
+         "settings" : {
+            "OS_TEST_ISSUES" : "",
+            "FLAVOR" : "DVD-Updates",
+            "DISTRI" : "opensuse",
+            "VERSION" : "15.3",
+            "ARCH" : "x86_64"
+         },
+         "test" : "kde",
+         "repos" : [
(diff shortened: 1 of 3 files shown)


foursixnine:
@msmeissn is there something else needed? /cc @okurz

msmeissn:
its not as easy.

you need 2 more repos:

at least:

http://download.opensuse.org/update/leap/15.3/backports/
this is needed definitely.

and also the sle import repos, but I am not sure if these are covered by:

http://download.opensuse.org/update/leap/15.3/sle/

these seem not yet be filled.

okurz:
> its not as easy.
> 
> you need 2 more repos:

Although it would be nice to see if tests would reveal these as missing :)

okurz:
So what I suggest is to have this merged and see what is missing keeping the above note in mind. Note that I don't have permissions to merge, others can

data/repos.json
```diff
@@ -28,5 +28,21 @@
             "http://download.opensuse.org/update/leap/15.2/non-oss/"
          ]
       }
-   }
+   },
+   "openSUSE:Leap:15.3:Update" : {
+         "settings" : {
+            "OS_TEST_ISSUES" : "",
```
foursixnine, line 34: I wonder if this is actually needed

foursixnine:
@mimi1vx Since Marcus is on Parental leave, do you know if this is all that's needed to enable maintenance repos for 15.3? (if so, can you merge? :P)

PR #2539, opened 2021-03-29
pkglistgen: ignore bootstrap_copy repo

lnussel:
Never use the bootstrap_copy repo for package solving. Staging content
might be very different so don't confuse the solver.

### DIFF
--- a/pkglistgen/tool.py
+++ b/pkglistgen/tool.py
@@ -578,7 +578,8 @@ class PkgListGen(ToolBase.ToolBase):
                                 project, scope, force, no_checkout,
                                 only_release_packages, stop_after_solve):
         self.all_architectures = target_config.get('pkglistgen-archs').split(' ')
-        self.repos = self.expand_repos(project, main_repo)
+        ignore_repos = set(target_config.get('pkglistgen-ignore_repos').split(' '))
+        self.repos = [ r for r in self.expand_repos(project, main_repo) if r[0] != project or r[1] not in ignore_repos ]
         print('[{}] {}/{}: update and solve'.format(scope, project, main_repo))
 
         group = target_config.get('pkglistgen-group', '000package-groups')
--- a/osclib/conf.py
+++ b/osclib/conf.py
@@ -42,6 +42,7 @@ DEFAULT = {
         'repo_checker-package-comment-devel': 'True',
         'pkglistgen-product-family-include': 'openSUSE:Leap:N',
         'pkglistgen-locales-from': 'openSUSE.product.in',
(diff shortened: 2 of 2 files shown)


coolo:
Add that default to osclib/conf.py instead so it's discoverable

coolo:
https://botmaster.suse.de/go/tab/build/detail/Pkglistgen.openSUSE_Factory/11219/pkglistgen/1/openSUSE_Factory_ARM - your default is not working

suntorytimed:
It's also broken for SLE: https://botmaster.suse.de/go/tab/build/detail/SLE15.SP3.RelPkgs/6435/Create.Release.Packages/1/Create.Release.Packages

pkglistgen/tool.py
```diff
@@ -578,7 +578,8 @@ def update_and_solve_target(self, api, target_project, target_config, main_repo,
                                 project, scope, force, no_checkout,
                                 only_release_packages, stop_after_solve):
         self.all_architectures = target_config.get('pkglistgen-archs').split(' ')
-        self.repos = self.expand_repos(project, main_repo)
+        ignore_repos = set(target_config.get('pkglistgen-ignore_repos').split(' '))
+        self.repos = [ r for r in self.expand_repos(project, main_repo) if r[0] != project or r[1] not in ignore_repos ]
```
Vogtinator, line 582: Is it intentional that the `ignore_repos` list doesn't influence recursive expansion?

If you have e.g. `projectA/standard -> projectB/bootstrap_copy -> projectC/standard`, then it would solve against `projectA/standard -> projectC/standard`

Vogtinator:
> https://botmaster.suse.de/go/tab/build/detail/Pkglistgen.openSUSE_Factory/11219/pkglistgen/1/openSUSE_Factory_ARM - your default is not working

This PR actually defines defaults for `openSUSE:Factory`, `openSUSE:Leap` and `openSUSE:Jump` only. For others, it works for `pkglistgen-archs` because that's explicitly defined in the `OSRT:Config` attribute there.

So either this needs to define defaults for the other projects as well or it's necessary to configure it explicitly everywhere. I'm in favor of the former, because most projects don't need this setting.

lnussel:
obviously the method is discoverable

pkglistgen/tool.py
lnussel: project C doesn't contain any packages in the factory case so doesn't matter.

sysrich:
This also seems to have broken all ring1 stagings in Factory, which can no longer find libc.so.6 and anything else from `bootstrap_copy`

PR #2547, opened 2021-04-08
Revert pkglistgen: ignore bootstrap_copy repo

sysrich:
#2539 broke quite a bit, including blocking all Factory Ring 1 stagings, so I think it's best we just revert it now before revisiting whati t was trying to fix


coolo:
Yeah, I guess you need to look if the project is actually linking against ring0 before ignoring it.

PR #2552, opened 2021-04-15
gocd: Drop obsolete projects, fix whitespace

Vogtinator:
- Drop Leap:15.1:Images and Leap:15.2:MicroOS
- Some .yaml files were changed manually, adjust the .erb source
- Delete some unnecessary/duplicate whitespace

### DIFF
--- a/gocd/microos-stagings.gocd.yaml
+++ b/gocd/microos-stagings.gocd.yaml
@@ -94,7 +94,6 @@ pipelines:
          - script: |-
              osc -A $STAGING_API api -X POST "/source/$STAGING_PROJECT?cmd=remove_flag&repository=images&flag=build"
 
-  
   MicroOS.Staging.B:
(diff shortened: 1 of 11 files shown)


Vogtinator:
@nilxam Apparently you hand-edited `staging.bot.gocd.yaml` after generating, because the `.erb` didn't match.

For backports, there are no `rebuild` and `select` calls, but for factory there are. For now I just removed them from factory as well, because the bot is disabled there anyway.

nilxam (approved):
When I clean up 15.2 configs last time, Ludwig told me he want to keep 15.2 MicroOS, if 15.2:MicroOS is obsoleted then sure thing we can drop it, other than the change looks fine to me.

Vogtinator:
Yep, Leap MicroOS was cancelled (for now)

PR #2558, opened 2021-05-07
Do rebuilds for Step

dirkmueller:


Vogtinator:
Released SLE is ABI compatible and such building only packages with changes sources should be enough. So what's the benefit of this?

coolo:
I trusted Dirk to know what he's doing - and your comment popped up the moment I pressed merge. We can revert of course if this is not the right thing to do

Vogtinator:
Right... I asked him on IRC what this PR is for. I didn't get an answer and only commented here later.

This should effectively be a noop, so just waste some time and space on botmaster.

dirkmueller:
@Vogtinator as I answered an hour ago on irc, Step currently builds with linkedbuild on all the sources including updates. if the maintenance team releases rebuild binaries to compensate for a incompatibility, those might not be  source change and hence need a rebuild then on the step side. 

we can remove the rebuilds on non-Update soon, but I think the :Update rebuilds will have to remain. we'll see. we haven't really decided yet how updates will be done for Leap going forward.

PR #2564, opened 2021-05-21
Support Leap:15.3:ARM:Images as well

dirkmueller:
We need this to run TTM against :images and properly determine the
pseudometa package.


ggardet:
CI failed.

dirkmueller:
> CI failed.

thanks, fixed

Vogtinator:
You probably want to enable this in gocd as well?

ggardet:
> You probably want to enable this in gocd as well?

I guess `gocd/totestmanager.gocd.yaml` should be updated as well

ggardet (approved):

PR #2565, opened 2021-05-26
check_source.py: Decline if source_revision not set

Vogtinator:
Accepts https://build.opensuse.org/request/show/895535, but declines https://build.opensuse.org/request/show/867727.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -114,6 +114,10 @@ class CheckSource(ReviewBot.ReviewBot):
             self.review_messages['declined'] = 'Only one action per request allowed'
             return False
 
+        if source_revision is None:
+            self.review_messages['declined'] = 'Submission not from a pinned source revision'
+            return False
+
         kind = package_kind(self.apiurl, target_project, target_package)
         if kind == 'meta':
             self.review_messages['accepted'] = 'Skipping all checks for meta packages'


DimStar77 (approved):

lnussel (approved):

coolo:
Well ... I can pin Base:System/brp-check-suse to revision 1, which is a link to home:coolo/brp-check-suse, submit that revision 1 and it would get reviewed. Then I change home:coolo and you would have the same problem you're fighting here (and don't say it doesn't happen, that's how texlive packages are submitted).

So having a kind of revision is just slightly more useful unless you verify expanded sources are submitted. And this really should be OBS's job. But I guess I'm preaching to the choir here.

Vogtinator:
Yes, there was quite a bit of discussion in IRC and on MLs about that.

lnussel:
https://github.com/openSUSE/open-build-service/pull/10992 is supposed to fix it on server side but does not move.
So meanwhile I guess Factory would have to decline source packages with links to somewhere outside Factory itself.

coolo:
that still doesn't solve the submit+delete+revive. we have to require hashes and those hashes to point to expanded sources. It would require a developer project maintainer with ill intentions for Factory to become victim of such attacks, devel projects are more likely to become it, but they rarely have large time windows between reviews and accepts.

Before we accept @Vogtinator's check_source patch, I rather fix OBS :)

PR #2567, opened 2021-05-27
Add Backports to the updates being tested

a-kpappas:
This commit contains the necessary configuration and code changes. The products
referenced by them have also been created on openqa.opensuse.org


Vogtinator (approved):
Bit strange that backports has its own separate DVD, but LGTM

PR #2569, opened 2021-05-31
GitHub actions

dirkmueller:


coolo:
do we need to enable something on the repo to trigger it?

okurz:
You should not need additional settings. But I suggest to first introduce github actions, merge it, check that it works and only then remove the travis CI config as both GHA and travis can run in parallel.

coolo:
travis won't no longer return, so it removed itself

coolo:
Added the github actions file now (just copied Dirk's)

okurz:
> travis won't no longer return, so it removed itself

We are running projects with travis-ci.com (not travis-ci.org) just fine

coolo:
possibly, but if I need to change the setup anyway, I can change to the real thing :)

PR #2570, opened 2021-06-04
check_source.py: Validate content of _links in submissions

Vogtinator:
Recursively expand links until the end or reaching a "safe" project.
When a link isn't static, fail the review.

Draft because:
- [ ] Breaks submissions to devel projects, because branches to them aren't allowed
- [ ] Allowing links to different packages in `openSUSE:` might also be bad
- [ ] Needs to check for project links as well?
- [ ] The `_link` parsing needs special care e.g. `package="../evil/package"` (`osc.core.makeurl` just does `'/'.join(path)`!)

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -106,6 +106,40 @@ class CheckSource(ReviewBot.ReviewBot):
 
         return ret
 
+    def check_static_source(self, project, package, revision):
+        # Allow links to openSUSE:, but not Staging
+        if project.startswith('openSUSE:') and ':Staging' not in project:
+            return True
+
+        # The revision has to be a hash
+        if revision is None:
+            self.review_messages['declined'] = 'Revision not set'
+            return False
+
+        if len(revision) != 32:
+            self.review_messages['declined'] = 'Revision not a hash'
+            return False
+
+        url = osc.core.makeurl(self.apiurl, ['source', project, package, '_link'], {'expand': 0, 'rev': revision})
+        try:
+            linkxml = ET.parse(osc.core.http_GET(url)).getroot()
+        except HTTPError as e:
+            if e.code == 404:
+                # Not a link
+                return True
+
+            raise
+
+        # Check links recursively
+        lprj = linkxml.get('project', project)
+        lpkg = linkxml.get('package', package)
+        lrev = linkxml.get('revision', None)
+        if not self.check_static_source(lprj, lpkg, lrev):
+            self.review_messages['declined'] += f' (from {project}/{package}/_link@{revision})'
+            return False
+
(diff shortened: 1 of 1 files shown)


check_source.py
```diff
@@ -114,8 +148,8 @@ def check_source_submission(self, source_project, source_package, source_revisio
             self.review_messages['declined'] = 'Only one action per request allowed'
             return False
 
-        if source_revision is None:
-            self.review_messages['declined'] = 'Submission not from a pinned source revision'
+        # The SR itself can use the commit number safely, because it's declined on package deletion
```
DimStar77, line 151: but after recreating the package, the SR can be 'reopened'

Vogtinator: IIRC I tried this once but got `is in a final state and can't be changed`

Vogtinator: If we have to enforce hashes in SRs, then pretty much every SR to `oS:F` currently will be declined...

Vogtinator: Reopening should also restart the reviews. But if it happens quickly enough, then the caches might make this a false positive.

(This rabbit hole goes really deep...)

DimStar77: https://build.opensuse.org/request/show/897561

DimStar77: declining non-hash SRs can't be enforced without the backend transparently changing the rev - a simple `osc sr DEVELPRJ PKG openSUSE:Factory` creates an sr with a numeric rev

Vogtinator: For a transition period, the bot could decline the sr and create a new one with hash instead

DimStar77: giving the users a decline message on everything and then not having the users as creator makes it hard to enter into dialogs through the SR UI

Should rather be the backend rewriting it from 'rev x' to hash-rev (it has all the information after all)

coolo:
Why so complicated? You only want to verify it's expanded sources on a pinned revision, no? If you have a revision, compare the srcmd5 of /source/$project/$package?rev=$rev with /source/$project/$package?rev=$rev&expand=1

I don't see why it's important where the sources point to - as long as we accept what we reviewed.

check_source.py
coolo: The above SR just proves that we shouldn't try to enforce this client side. This really should be enforced on the source revision system

Vogtinator: Yes, so far there's no progress or even acknowledgement on the OBS side.

coolo: Yeah, let me use my levers... :)

coolo:
Moving to https://github.com/openSUSE/open-build-service/pull/11263

PR #2577, opened 2021-06-21
gocd/totestmanager.gocd.yaml: Update list of projects

Vogtinator:
- Leap:15.3 is frozen, drop it
- Add 15.3:WSL and :ARM:Images


nilxam:
👍

Issue #1947, opened 2019-04-03
freeze: rename to rebase


lnussel:
the name of the freeze command is rather unintuitive. It should be rename to rebased to better reflect common naming practice.

lkocman:
Ack, I was also initially confused by the feature functionality. Please keep the freeze for compat reasons as well.

ancorgs:
Isn't this solved by #2584?

coolo:
I guess the openSUSE prefix broke the auto close

PR #2584, opened 2021-06-30
Staging: add 'rebase' as an alias to the 'freeze' command

DimStar77:
Fixes issue #1947


coolo:
Better also update/refresh docs/staging.asciidoc

coolo (approved):

PR #2585, opened 2021-07-01
Ci small tunning

jreidinger:
What is done in this pr:

- update badge in readme to reflect change to github actions
- move linter to own job to speed up a build for hopefully around 30 seconds

Also badge for codecov needs to update, but I do not have permissions to see proper link in settings.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -1,4 +1,4 @@
-name: tests
+name: Checks
 
 on:
   push:
@@ -9,7 +9,7 @@ on:
     - master
 
 jobs:
-  docker:
+  tests:
     timeout-minutes: 40
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-test.yml
```diff
@@ -35,3 +32,21 @@ jobs:
       if: always()
       run: docker-compose -f dist/ci/docker-compose.yml down
 
+  linters:
+    runs-on: ubuntu-latest
+
```
coolo, line 37: that's a a lot of copy&paste for 20s (from >20min). I just checked and looks like github doesn't support yaml anchors.

jreidinger: yep. As it is just single docker image it can maybe use container directly, but it is still something we are checking...another idea is to add there also `perl -Wc` call for linters, but it will probably need adaptation of docker image to contain also required perl dependencies.

PR #2586, opened 2021-07-01
Perl linter

jreidinger:
Is that check_source.pl used at all? Some bugs in it looks bad enough, especially that regexp.

### DIFF
--- a/check_source.pl
+++ b/check_source.pl
@@ -132,7 +132,7 @@ for my $test (glob("/usr/lib/obs/service/source_validators/*")) {
 
 my $odir = getcwd();
 
-chdir($dir) || "chdir $dir failed";
+chdir($dir) || die "chdir $dir failed";
 for my $patch (glob("*.diff *.patch *.dif")) {
     $patches{$patch} = 'current';
 }
@@ -213,7 +213,7 @@ if (-d "$old") {
                 $diff = diff "$old/$changes", "$dir/$changes";
             }
             for my $line (split(/\n/, $diff)) {
-                next unless $line =~ m/^+/;
+                next unless $line =~ m/^\+/;
                 $line =~ s/^\+//;
                 for my $patch (keys %patches) {
                     if (index($line, $patch) != -1) {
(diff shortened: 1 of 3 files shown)


jreidinger:
Too make this work it needs to adapt test docker container to contain also all required perl libraries.

coolo:
Yes, check_source.pl is used by check_source.py (just to add to the fun of it :)

And that regexp bug isn't as bad as it first looks - it just means we also look for patches mentioned in other parts of the diff, not just the + part. I guess this leads to 0 problems in reality - but good catch anyway.

coolo:
And the CI container is defined in `dist/ci/testenv-tumbleweed/Dockerfile` - but I just learned it's broken:
https://build.opensuse.org/package/show/openSUSE:Tools:Images/osrt-testenv-tumbleweed

jreidinger:
@coolo thanks, I will check it and try fix that container.

coolo:
Fixed the container in OBS - the problem was an inconsistent file provides between Factory and openSUSE:Tools:Images

.perl_checker.sh
```diff
@@ -0,0 +1,4 @@
+#! /usr/bin/sh
+for i in `find . | grep 'p[lm]$'`; do
```
coolo, line 2: please no hidden files as commands. We have dist/ci for that

coolo (changes requested):

coolo (approved):

DimStar77:
> Yes, check_source.pl is used by check_source.py (just to add to the fun of it :)
> 
> And that regexp bug isn't as bad as it first looks - it just means we also look for patches mentioned in other parts of the diff, not just the + part. I guess this leads to 0 problems in reality - but good catch anyway.

The SLE submissions actually relied on this bug

i.e:
* a patch is added in SLE-SP-(n-1), referenced in the changelog, as is needed.
* in SLE-SP-n, the package is taken from Factory, which never had the patch. The diff now shows the patch being 'lost', which was considered ok (the changelog never mentions the patch being added, so we also do not drop it in the changelog)
* So, the fix is technically seen correct, but causes more issues for SLE now - as patches can't formally be 'lost' anymore

DimStar77:
The above note is also a problem for TW-proper: if I have to submit a 'revert submission SR', then patches are being dropped the same way (i.e. never added) - the bot would now decline those too

coolo:
Well, then this should be properly implemented not rely on a bug

DimStar77:
> Well, then this should be properly implemented not rely on a bug

Of course - I wasn't trying to imply we should put the bug back in place. Just that the bug was unnoticed for a long time and 'served us well'

suntorytimed:
Agreeing with DimStar77 and coolo, we should not rely on a bug for this. Nevertheless we require a proper implementation sooner than later, as this will hit us with a couple of submissions to SLE. I have opened a bug and linked it to this PR.

Issue #2524, opened 2021-02-03
Migrate from nose to pytest


s-t-e-v-e-n-k:
The test runner nosetests has been unmaintained for years, so investigating moving to pytest or another test runner would be awesome.

PR #2587, opened 2021-07-01
do not print coverage to stdout

jreidinger:
First step to make test output a bit more readable.


jreidinger:
still WIP. I need to check why option is not recognized from https://nose.readthedocs.io/en/latest/plugins/cover.html

coolo:
https://github.com/nose-devs/nose/commit/7c26ad1e6b7d308cafa328ad34736d34028c122a - unreleased since 2016 :)

coolo:
There is a reason https://github.com/openSUSE/openSUSE-release-tools/issues/2524 exists

coolo:
let's close this one

PR #2590, opened 2021-07-02
Update exclude section in .flake8

ancorgs:
As far as I can see, there is no longer any reason to exclude `openqa-maintenance.py` and `openqa` is not longer there.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
-exclude = .open-build-service/, abichecker, openqa, openqa-maintenance.py
+exclude = .open-build-service/, abichecker
 max-line-length = 100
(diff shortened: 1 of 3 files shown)


.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
-exclude = .open-build-service/, abichecker, openqa, openqa-maintenance.py
+exclude = .open-build-service/, abichecker
```
coolo, line 2: .open-build-service isn't here either. Basically all of 495878ff0f460b7f44dded36b8600ba8fdd8a794 can be reverted

ancorgs: Done. Added a new commit for that.

coolo:
the openSUSE infrastructure had a bad weekend :(

coolo (approved):

ancorgs:
Since we are not in a hurry, I will wait until the openSUSE infrastructure is awake to re-run the tests and merge in green.

PR #2595, opened 2021-07-08
use pytest instead of nosetests

jreidinger:
experimental switch to run tests with pytest

### DIFF
--- a/dist/ci/docker-compose-test.sh
+++ b/dist/ci/docker-compose-test.sh
@@ -15,10 +15,13 @@ done
 cd /code
 for file in tests/*_tests.py; do
   if ! test -f /code/.without-coverage; then
-    COVER_ARGS="--with-coverage --cover-xml --cover-package=. --cover-inclusive"
+    COVER_ARGS="--cov=. --cov-append --cov-report=xml"
+  else
+    COVER_ARGS="--no-cov"
   fi
-  echo "running tests from $file..."
-  run_as_tester nosetests $COVER_ARGS -c .noserc -s $file
+  # TODO: Review bot test failed without log-level set to debug
+  # TODO: due to memoize tests cannot be run together, otherwise it start failing
+  run_as_tester pytest $COVER_ARGS --log-level=DEBUG $file
 done
 
 set -x
--- a/dist/ci/testenv-tumbleweed/Dockerfile
+++ b/dist/ci/testenv-tumbleweed/Dockerfile
@@ -5,9 +5,9 @@ FROM opensuse/tumbleweed
 RUN zypper -n ar http://download.opensuse.org/repositories/openSUSE:/Tools/openSUSE_Factory/ openSUSE:Tools
 RUN zypper --gpg-auto-import-keys ref
 
-RUN zypper in -y osc python3-pytest python3-nose python3-httpretty python3-pyxdg python3-PyYAML \
+RUN zypper in -y osc python3-pytest python3-httpretty python3-pyxdg python3-PyYAML \
    python3-pika python3-mock python3-cmdln python3-lxml python3-python-dateutil python3-colorama \
-   python3-influxdb python3-coverage python3-pytest-cov python3-coveralls libxml2-tools curl python3-flake8 \
+   python3-influxdb python3-pytest-cov python3-coveralls libxml2-tools curl python3-flake8 \
    shadow vim vim-data strace git sudo patch openSUSE-release openSUSE-release-ftp \
    perl-Net-SSLeay perl-Text-Diff perl-XML-Simple perl-XML-Parser build \
    obs-service-download_files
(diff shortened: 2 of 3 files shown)


dist/ci/testenv-tumbleweed/Dockerfile
```diff
@@ -5,9 +5,9 @@ FROM opensuse/tumbleweed
 RUN zypper -n ar http://download.opensuse.org/repositories/openSUSE:/Tools/openSUSE_Factory/ openSUSE:Tools
 RUN zypper --gpg-auto-import-keys ref
 
-RUN zypper in -y osc python3-nose python3-httpretty python3-pyxdg python3-PyYAML \
+RUN zypper in -y osc python3-pytest python3-httpretty python3-pyxdg python3-PyYAML \
```
coolo, line 8: you better create another PR to add both. You can still remove nose in this.

jreidinger: well, it is still WIP. I plan to temporary just for testing install it here. Locally it failed after several tests for unavailable obs API, but it is probably caused by my local machine computation power.

coolo: that's my point: you can't test pytest without it being in the OBS container

jreidinger: well, I plan to install it with zypper for now as temporary solution for testing that all tests can pass.

jreidinger: trick we use often in yast where we have shared base image for yast and if additional library is needed for one specific module we install it with zypper as part of run to not grow too much generic image.

jreidinger:
OK, so it runs faster...but failing. I will need to debug what exactly and why failed.

dist/ci/docker-compose-test.sh
```diff
@@ -13,13 +16,8 @@ until curl http://api:3000/about 2>/dev/null ; do
...
 cd /code
-for file in tests/*_tests.py; do
-  if ! test -f /code/.without-coverage; then
-    COVER_ARGS="--with-coverage --cover-xml --cover-package=. --cover-inclusive"
-  fi
-  echo "running tests from $file..."
-  run_as_tester nosetests $COVER_ARGS -c .noserc -s $file
-done
+# TODO: Review bot test failed without log-level set to debug
+run_as_tester pytest --log-level=DEBUG --cov=. tests/*_tests.py
```
jreidinger, line 20: this log level is strange...maybe nosetests has it by default and it can be reason why pytest failed in that review bot test.

dist/ci/docker-compose-test.sh
```diff
@@ -13,13 +16,13 @@ until curl http://api:3000/about 2>/dev/null ; do
...
 cd /code
-for file in tests/*_tests.py; do
-  if ! test -f /code/.without-coverage; then
-    COVER_ARGS="--with-coverage --cover-xml --cover-package=. --cover-inclusive"
-  fi
-  echo "running tests from $file..."
-  run_as_tester nosetests $COVER_ARGS -c .noserc -s $file
+for i in tests/*_tests.py; do
+  # TODO: Review bot test failed without log-level set to debug
+  # FIXME: JR: for not yet known reason running all tests together failed
```
coolo, line 21: there is global state kept in osc library and in memoize

coolo:
At least the speed is on par now :smile:

jreidinger:
@coolo yeah, right. Thanks for explain why running all tests together does not work. Maybe reseting that memoize between tests can help as now annoying part is that it fail in first failed test and you do not see all failures at once.

coolo:
I tried for quite a while - but memoize is not the only global state. So I gave up in favor of having test coverage at all :(

dist/ci/docker-compose-test.sh
```diff
@@ -13,12 +13,10 @@ until curl http://api:3000/about 2>/dev/null ; do
 done
 
 cd /code
-for file in tests/*_tests.py; do
-  if ! test -f /code/.without-coverage; then
-    COVER_ARGS="--with-coverage --cover-xml --cover-package=. --cover-inclusive"
-  fi
-  echo "running tests from $file..."
-  run_as_tester nosetests $COVER_ARGS -c .noserc -s $file
+for i in tests/*_tests.py; do
```
coolo, line 16: hmm, why did you rename my variable `file` now `i`? 

Why did you make it imposible to run without coverage now? it's a waste of time while developing locally

coolo (changes requested):

dist/ci/docker-compose-test.sh
jreidinger: ok, will revert it back.

coolo (approved):

PR #2599, opened 2021-07-12
Preparation for switch from nosetests to pytest

jreidinger:
The preparation is needed so no workaround is needed when validating
that all tests passed when switched from nosetests to pytest. Nosetests
will be removed as part of final switch change.

final pr: https://github.com/openSUSE/openSUSE-release-tools/pull/2595


jreidinger:
thanks

PR #2600, opened 2021-07-13
pkglistgen: initialize pkglistgen for Leap 15.4

nilxam:


gocd/pkglistgen.opensuse.gocd.yaml.erb
```diff
@@ -52,4 +52,56 @@ pipelines:
             - repo-checker
             tasks:
              - script: python3 ./pkglistgen.py --apiurl https://api.opensuse.org handle_update_repos <%= project %>
+<% end -%>
+  Pkglistgen.openSUSE_Leap_15.4:
```
coolo, line 56: why the copy paste?

nilxam: fixed format error anyway

PR #2601, opened 2021-07-13
Check for maintainers in the source project

### LINKED ISSUE #2429: Ensure 'factory-maintainers' are devel prj maintainers
This is an angle on https://github.com/openSUSE/open-build-service/issues/9310 - trying to at least prevent to run into the permission errors OBS could throw on accept

When a submission is incoming, we shuld ensure the sourceproject (devel prject) hast factory-maintainers (any configurable group - based on OSRT:Config I'd say) defined as maintainer.

During accept this is nescessay to convert the source package into a branch of the target package  (OBS:InitializeDevelProject)

We envision something like:
* factory-auto does not accept (or decline) the review it has open for the package if factory-maintainers is not maintainer in the source project/devel project
* Comment about this fact on the request (to make sure everybody looking at it know about the issue)
* Send an Addrole request to the source prject, to have the issue corrected

imobachgs:
* It introduces a new configuration parameter `required-source-maintainer`.
* If defined, it is expected to be a maintainer of the devel project.
* If that's not the case, the request is declined and an `add_role` request is created.

Fixes #2429

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -21,6 +21,8 @@ from osclib.core import group_members
 from osclib.core import package_kind
 from osclib.core import source_file_load
 from osclib.core import target_archs
+from osclib.core import create_add_role_request
+from osclib.core import maintainers_get
 from urllib.error import HTTPError
 
 import ReviewBot
@@ -52,6 +54,7 @@ class CheckSource(ReviewBot.ReviewBot):
         self.mail_release_list = config.get('mail-release-list')
         self.staging_group = config.get('staging-group')
         self.repo_checker = config.get('repo-checker')
+        self.required_maintainer = config.get('required-source-maintainer', '')
         self.devel_whitelist = config.get('devel-whitelist', '').split()
         self.skip_add_reviews = False
         self.security_review_team = config.get('security-review-team', 'security-team')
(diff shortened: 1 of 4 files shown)


osclib/core.py
```diff
@@ -1098,6 +1098,21 @@ def create_change_devel_request(apiurl, source_project, source_package,
             tgt_project=target_project, tgt_package=target_package)
     return create_request(apiurl, action, message)
 
+def create_add_role_request(apiurl, target_project, user, role, target_package=None, message=None):
+    """Create an add_role request
+
+    user -- user or group name. If it is a group, it should start with 'group:'.
```
jreidinger, line 1104: I would use "had to start" instead of "should" as it sounds less strict.

jreidinger:
now it looks good for me, but I am not domain expert

tests/check_source_tests.py
```diff
@@ -70,6 +79,30 @@ def test_devel_project(self):
         self.assertReview(req_id, by_user=(self.bot_user, 'accepted'))
         self.assertReview(req_id, by_group=(REVIEW_TEAM, 'new'))
 
+    def test_source_maintainer(self):
+        """Declines the request when the 'required_maintainer' is not maintainer of the source project"""
```
coolo, line 83: I'm missing a positive test. I have my suspicion....

coolo (changes requested):

tests/check_source_tests.py
imobachgs: Good catch. It looks like `maintainers_get` only returns users, but not groups.

imobachgs: Well, my assertion is not accurate: in case a group is a maintainer of the project, it returns the users within the group, not the group itself.

coolo (approved):

check_source.py
```diff
@@ -11,6 +11,8 @@
 except ImportError:
     import cElementTree as ET
 
+from lxml import etree as ETL
```
DimStar77, line 14: do we really need two XML libraries in here now? we have xml.etree(or cElementTree) and lxml both imported (works, of course, but…)

coolo: actually cElementTree is supposed to die, but we can do that refactoring later IMO

DimStar77: sure - no problem in going with what we have now

PR #2608, opened 2021-07-22
Document and extend OBSLocal

jreidinger:
- Documentation of OBSLocal extended and adapted to Python's standard format (reStructured Text).
- Added new methods to `OBSLocal.Project` to extend the current management of meta information
- Bonus: small fixes in `check_source_test.py`

### DIFF
--- a/tests/OBSLocal.py
+++ b/tests/OBSLocal.py
@@ -380,13 +380,13 @@ class StagingWorkflow(object):
                                       project_links=project_links)
         return self.projects[name]
 
-    def submit_package(self, package=None, project=None):
+    def submit_package(self, package, project=None):
         """Creates submit request from package to target project.
 
-        Both have to exist, otherwise looks at create_submit_request method.
+        Both have to exist (Use :func:`create_submit_request` otherwise).
 
-        :param package: package to submit TODO: what means None here?
-        :type package: Package or None
+        :param package: package to submit
+        :type package: Package
         :param project: project where to send submit request, None means use the default.
         :type project: Project or str or None
         :return: created request.
--- a/tests/OBSLocal.py
+++ b/tests/OBSLocal.py
@@ -168,10 +168,10 @@ class TestCase(unittest.TestCase):
 class StagingWorkflow(object):
     def __init__(self, project=PROJECT):
         """
-        Initialize the configuration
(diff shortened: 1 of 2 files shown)


tests/OBSLocal.py
```diff
@@ -405,6 +422,18 @@ def remove(self):
 
 class Project(object):
     def __init__(self, name, reviewer={}, maintainer={}, project_links=[], create=True, with_repo=False):
+        """
+          Represents Project in OBS.
+
+              Parameters:
+                  name(str): Project name
+                  reviewer: TODO
+                  maintainerr: TODO
```
teclator, line 431:
```suggestion
                  maintainer: TODO
```

ancorgs:
This doesn't seem to follow any of the four more extended docstring formats (see https://realpython.com/documenting-python-code/#docstring-formats, I have found the same list in several other places).

Other parts of this repository (like `osclib/stagingapi.py` or  `osclib/comments.py`) seem to be using reStructured Text. So I would suggest to stick to that instead of introducing our own (unofficial) variant.

tests/OBSLocal.py
```diff
@@ -331,6 +381,17 @@ def create_project(self, name, reviewer={}, maintainer={}, project_links=[]):
         return self.projects[name]
 
     def submit_package(self, package=None, project=None):
+        """Creates submit request from package to target project.
+
+        Both have to exist, otherwise looks at create_submit_request method.
+
+        :param package: package to submit TODO: what means None here?
```
teclator, line 388: Looks like mandatory

mvidner: yes, I have committed a fix to that effect

mvidner (changes requested):
Thanks @jreidinger and @ancorgs for a thorough documentation!

So we are documenting (and extending) some OBS API. Surely there is some other python binding of the same HTTP API in `osc`? We should link to it so that we don't reinvent the wheel now or in the future.

mvidner:
... which is in https://github.com/openSUSE/osc/blob/master/osc/core.py

mvidner (approved):
Having added the links, approving :)

mvidner:
@ancorgs excellent! thanks! still LGTM.

PR #2610, opened 2021-07-24
do not add factory-auto to backports reviews anymore

msmeissn:
similar to 15.3 and other leap, backports updates has no strict factory requirement... currently just ignore it for 15-sp3,
but probably all of 15 are the same


coolo:
First of all: this means the script is not obsolete, right? It seems to act to all release incident requests to openSUSE:Maintenance and adds maintainers.

https://build.opensuse.org/request/show/906899 indicates you ignore factory-source reviews if added - so possibly we should kill that review all together? That bot implements a policy, if that policy has been dropped, we should drop that part as well.

PR #2611, opened 2021-07-26
Avoid creation of duplicate add_role requests (fixes #2609)

### LINKED ISSUE #2609: Do not DoS devel projects with addrole requests
As we now enabled the factory-auto bot to ensure devel projects have factory-maintainers assigned as maintainer (to perform actions on accept), we have seen some sub projects that never caused problem but that do not have factory-maintainers assigned (hidden inheritance model of permissions in OBS, e.g. maintainers of devel:languages:python have access in devel:languages:python:azure, without being listed as maintainers)

So far, all is good - and I do not mind the group being listed at all (makes even more sense than the hidden inheritance model)

yet, we should probably not spam a project with 'one addrole request per declined submission'

* Only create a addrole request for factory-maintainers if there is not currently one pending already

Let's avoid things like
Screenshot from 2021-07-23 11-51-46
which would only add confusion and frustration

jreidinger:


check_source.py
```diff
@@ -171,9 +172,14 @@ def check_source_submission(self, source_project, source_package, source_revisio
             )
 
             try:
-                add_role_msg = 'Created automatically from request %s' % self.request.reqid
-                add_role_reqid = create_add_role_request(self.apiurl, source_project, self.required_maintainer,
-                                                         'maintainer', message=add_role_msg)
+                add_roles = get_request_list(self.apiurl, source_project, req_state=['new'], req_type='add_role')
+                add_roles = list(filter(self.select_add_role, add_roles))
```
DimStar77, line 176: Doesn't that mean any random open addrole request will be used? There is s chance that somebody else is already requesting managership, without it being 'our' request

coolo: That's the job of the filter, so I don't think so.

But the request may be in `review` if the devel project has reviewers set.

jreidinger: good point, I was not sure about review state. So I think it makes sense to add it.
@DimStar77 maybe better filter name can be used. I just do not have a good name in my mind that explain it selects add role that has maintainer for user or group set to required user.

jreidinger:
@ancorgs I think it is ready now including new test for inherited roles.

check_source.py
```diff
@@ -171,9 +172,15 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
-                                                         'maintainer', message=add_role_msg)
+                add_roles = get_request_list(self.apiurl, source_project,
+                    req_state=['new', 'review'], req_type='add_role')
+                add_roles = list(filter(self.select_add_role, add_roles))
+                if len(add_roles) > 0:
+                    add_role_reqid = add_roles[0].reqid
+                else:
+                    add_role_msg = 'Created automatically from request %s' % self.request.reqid
+                    add_role_reqid = create_add_role_request(self.apiurl, source_project, self.required_maintainer,
+                                                            'maintainer', message=add_role_msg)
```
ancorgs, line 183: What about extracting all this to some function like `find_or_create_add_role_req`, `ensure_add_role_req` or whatever name you like? The body of `check_source_submission` is huge already.

check_source.py
```diff
@@ -341,7 +348,8 @@ def source_has_correct_maintainers(self, source_project):
         """Checks whether the source project has the required maintainer
 
         If a 'required-source-maintainer' is set, it checks whether it is a
-        maintainer for the source project.
+        maintainer for the source project. Inherited maintainership is
+        intentionally ignored to have explicit maitainer set.
```
ancorgs, line 352: typo: maintainer

check_source.py
```diff
@@ -356,6 +364,15 @@ def source_has_correct_maintainers(self, source_project):
 
         return self.required_maintainer in maintainers
 
+    def select_add_role(self, request):
```
ancorgs, line 367: This filter has nothing to do with the request being an add_role one, but with the person assigned to the action. So I would change the name to something like `select_required_maintainer`.

check_source.py
ancorgs: I made a suggestion below about the name

check_source.py
ancorgs: Please, write a sentence to describe what the function returns

ancorgs (changes requested):
Review finished. Some requests added about code organization

check_source.py
```diff
@@ -166,28 +166,7 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
-                    add_role_reqid = create_add_role_request(self.apiurl, source_project, self.required_maintainer,
-                                                            'maintainer', message=add_role_msg)
-                declined_msg += ' Created the add_role request %s for addressing this problem.' % add_role_reqid
-            except HTTPError as e:
-                self.logger.error(
-                    'Cannot create the corresponding add_role request for %s: %s' % (self.request.reqid, e)
-                )
-
-            self.review_messages['declined'] = declined_msg
+            self.__ensure_add_role_request(source_project)
```
ancorgs, line 169: Not exactly what I meant. :wink:

Now the `ensure_add_role_request` function does two different things, the one suggested by its name and also declining the request. Declining the request inside a function call is inconsistent with the rest of `check_source_submission` (in which that is done explicitly in the body of the main function).

I was thinking on something like:

```python
if not self.source_has_correct_maintainers(source_project):
    declined_msg = (
        'This request cannot be accepted unless %s is a maintainer of %s.' %
        (self.required_maintainer, source_project)
    )

    add_role_req = self._ensure_add_role_request(source_project)
    if add_role_req:
        declined_msg += ' Created the add_role request %s for addressing this problem.' % add_role_req.reqid
    self.review_messages['declined'] = declined_msg
```

check_source.py
```diff
@@ -364,7 +343,33 @@ def source_has_correct_maintainers(self, source_project):
...
+            declined_msg += ' Created the add_role request %s for addressing this problem.' % add_role_reqid
+        except HTTPError as e:
+            self.logger.error(
+                'Cannot create the corresponding add_role request for %s: %s' % (self.request.reqid, e)
+            )
+
+        self.review_messages['declined'] = declined_msg
+
+    def __select_required_maintainer(self, request):
+        """Selects add role requests that adds required maintainer user or group"""
```
ancorgs, line 372: Nitpicking: this actually selects nothing (that's done by the filter function). This returns true if the action for the given request points to required_maintainer.

check_source.py
jreidinger: well, I try to shorten up more, but if this is more readable for you, I can adapt it.

ancorgs (approved):
LGTM

Issue #2612, opened 2021-07-27
Shut down useless bots and drop obsolete code


ancorgs:
There are several tools marked as obsolete in [CONTENTS.md](https://github.com/openSUSE/openSUSE-release-tools/blob/master/CONTENTS.md). So we plan to drop those tools by:

- Shutting them down from botmaster (gocd)
- Removing them from the spec file used for [the package at OBS](https://build.opensuse.org/package/show/openSUSE:Tools/openSUSE-release-tools)
- Dropping the code from the repository (the tools will always live in our hearts and in git's history)

This is the list of tools and the rationale for dropping them:

- check_maintenance_incidents (alias maintbot) - Looks like a [very outdated script](https://github.com/openSUSE/openSUSE-release-tools/pull/2604#discussion_r674063873) obsoleted by origin-manager. **BEWARE** maybe is not obsolete (see https://github.com/openSUSE/openSUSE-release-tools/pull/2610)
- leaper - Obsoleted by origin-manager. Currently deployed in comment-only mode for the maintenance workflow, [but it's also useless there](https://github.com/openSUSE/openSUSE-release-tools/pull/2604#discussion_r674031669)
- unmaintained - [Not longer used](https://github.com/openSUSE/openSUSE-release-tools/pull/2604#discussion_r673719609), based on data created by leaper
- sync-rebuild - The code was [merged to the 'accept' command](https://github.com/openSUSE/openSUSE-release-tools/pull/2604) 
- update_crawler - [Obsoleted](https://github.com/openSUSE/openSUSE-release-tools/pull/2604/files#r669910587) by origin-manager.
- check_source_in_factory - Outdated, it was [decided to drop it](https://github.com/openSUSE/openSUSE-release-tools/pull/2604#discussion_r674046338) bit it never happened
- manager_42.py - Apparently obsoleted by origin-manager. Currently not deployed.
- status - Apparently obsolete since it [checks for bots that are already retired](https://github.com/openSUSE/openSUSE-release-tools/pull/2604/commits/5ade13a48f21e4244ff1aad1a866936e6af2384d#r674885134)
- obs-clone - Was [used by Jimmy to create tests](https://github.com/openSUSE/openSUSE-release-tools/pull/2604/commits/5ade13a48f21e4244ff1aad1a866936e6af2384d#r674871919), but we use the OBS docker-compose approach now.
- obs-operator - Nobody seems to be using it nowadays
[...]

PR #2613, opened 2021-07-27
Drop obsolete tools

ancorgs:
This implements the easy (ie. less controversial) part of #2612.

Marked as draft because we agreed to wait at least one week before dropping anything, so people can provide feedback.

This pull request includes removal of the items marked below. According to that issue, we were supposed to drop the following:

- [ ] check_maintenance_incidents (alias maintbot) - Not dropped (yet) because maybe is not obsolete (see https://github.com/openSUSE/openSUSE-release-tools/pull/2610)
- [ ] obs-operator - Not so simple. Check #2614 
- [X] check_source_in_factory - **Not fully dropped** because it is still being called from [check_tags_in_requests](https://github.com/openSUSE/openSUSE-release-tools/blob/3990fecd037e4a/check_tags_in_requests.py#L91).
- [X] leaper
- [X] unmaintained
- [X] sync-rebuild
- [X] update_crawler
- [X] manager_42.py
- [X] status
- [x] obs_clone
- [x] scan_baselibs


ancorgs:
@lkocman After we merge this, you may want to check for references to the dropped tools at https://github.com/openSUSE/openSUSE-release-process

nilxam (approved):
No problem with this. I may revive *update_crawler* soon or later as an simple updater for Leap/Backports with a re-wrote code.

coolo:
7 days and still no dead kittens filling the streets!

PR #2615, opened 2021-07-29
Update mini obs

jreidinger:
It is still WIP as due to build depedencies 15.3 mini obs failing ( probably need to wait till obs finish publish of it )

obs commit: https://github.com/openSUSE/open-build-service/commit/fbfede9e51e155469dbfd3923f1807af2cb93fe0

### DIFF
--- a/dist/ci/miniobs/Dockerfile
+++ b/dist/ci/miniobs/Dockerfile
@@ -1,7 +1,7 @@
 #!BuildTag: osrt_miniobs
-FROM opensuse/leap:15.1
+FROM opensuse/leap:15.3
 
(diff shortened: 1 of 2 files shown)


dist/ci/miniobs/Dockerfile
```diff
@@ -1,7 +1,7 @@
 #!BuildTag: osrt_miniobs
-FROM opensuse/leap:15.1
+FROM opensuse/leap:15.3
```
coolo, line 2: this part is merged now - as 15.1 repo was removed

jreidinger: good, thanks. So I think it is time to close this PR as having worker that do real build is still too long way which we in the end do not take.

PR #2616, opened 2021-08-04
Add pipeline to publish the vagrant boxes to Vagrant Cloud

dcermak:
This is a draft for migrating the publishing pipeline of the opensuse vagrant boxes from our internal jenkins instance to botmaster.

TODO:
- [ ] somehow get the vagrant cloud API token into gocd
- [ ] verify that this can actually work


Vogtinator:
I'd like to do the same for the dockerhub container upload script, but so far didn't have time to discuss the secret data setup...

The code for the uploading should be in the same repo as the gocd pipeline configuration IMO.

IMO it should also use distro packages and not `bundle install`

dcermak:
> I'd like to do the same for the dockerhub container upload script, but so far didn't have time to discuss the secret data setup...

@coolo Added the secret variable for me to gocd. It should now be there, albeit I am unsure how it can be actually accessed.

> The code for the uploading should be in the same repo as the gocd pipeline configuration IMO.

I have moved the script into this repository.

> IMO it should also use distro packages and not `bundle install`

I hope that this can be done by adding a subpackage to `openSUSE-release-tools`.

coolo:
The subpackage is only useful if you need it locally in your cmdline. For anything else you need to update the CI container at https://build.opensuse.org/package/view_file/openSUSE:Tools:Images/gocd-agent-release-tools/Dockerfile?expand=1

dashboard.generator.gocd.yaml has the only other use of a secret variable right now

dcermak:
> The subpackage is only useful if you need it locally in your cmdline. 

Well it would certainly be convenient to have it available on the commandline, then I could use it in jenkins as well.

> For anything else you need to update the CI container at https://build.opensuse.org/package/view_file/openSUSE:Tools:Images/gocd-agent-release-tools/Dockerfile?expand=1

Is that file managed only via OBS or does it live somewhere in git?

> dashboard.generator.gocd.yaml has the only other use of a secret variable right now

Yes, I found that one already. What are the values for `secret_config_id` and `secret_key` that you configured in gocd?

coolo:
> > The subpackage is only useful if you need it locally in your cmdline.
> 
> Well it would certainly be convenient to have it available on the commandline, then I could use it in jenkins as well.
In jenkins you can also run it from git I suppose?

> 
> > For anything else you need to update the CI container at https://build.opensuse.org/package/view_file/openSUSE:Tools:Images/gocd-agent-release-tools/Dockerfile?expand=1
> 
> Is that file managed only via OBS or does it live somewhere in git?
Just in OBS.
> 
> > dashboard.generator.gocd.yaml has the only other use of a secret variable right now
> 
> Yes, I found that one already. What are the values for `secret_config_id` and `secret_key` that you configured in gocd?
That's the same for all pipelines - and as I reused the same json, it's RSYNC_FOR_COOLO

dcermak:
Stephan Kulow ***@***.***> writes:

>> > The subpackage is only useful if you need it locally in your cmdline.
>> 
>> Well it would certainly be convenient to have it available on the commandline, then I could use it in jenkins as well.
> In jenkins you can also run it from git I suppose?

Yes, I could.

coolo:
Do as you like. Just don't add the package because you believe it fixes the container :)

dcermak:
> Do as you like. Just don't add the package because you believe it fixes the container :)

I've moved it out of the package and will send a submitrequest to the dockerfile once my dependencies are in Factory.

dcermak:
I have tested the uploader locally using the container with the changes from https://build.opensuse.org/request/show/914976 and it appears to work.

Makefile
```diff
@@ -2,7 +2,7 @@ SUBDIRS = factory-package-news abichecker
 
 include Makefile.common
 
-pkgdata_BINS = $(shell find * -maxdepth 0 -executable -type f)
+pkgdata_BINS = $(shell find * -maxdepth 0 -executable -type f|sed '/.*rb$$/d')
```
Vogtinator, line 5: If you want to exclude something, just drop the executable bit or if that's not possible, add a `! -name`

gocd/vagrant-publisher.gocd.yaml
```diff
@@ -0,0 +1,22 @@
...
+      spec: 0 0 0 ? * *
+      only_on_changes: false
+    stages:
+    - Run:
+        approval: manual
+        resources:
+        - obs-to-vagrantcloud
+        tasks:
+        - script: |
+            ruby.ruby3.0 obs-to-vagrantcloud.rb --url https://download.opensuse.org/tumbleweed/appliances/boxes/Tumbleweed.x86_64.json --organization opensuse --new-box-name Tumbleweed.x86_64 --provider libvirt
```
Vogtinator, line 21: Is it necessary to hardcode the version here? This way it needs to be updated on every container update with newer minor version of ruby...

gocd/vagrant-publisher.gocd.yaml
```diff
@@ -0,0 +1,22 @@
+format_version: 3
+pipelines:
+  Vagrant.Publish:
```
Vogtinator, line 3: This only deals with openSUSE, maybe mention that? Most pipelines mention either openSUSE, Factory or Leap in their name

obs-to-vagrantcloud.rb
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env ruby
```
Vogtinator, line 1: The shebang doesn't match how it's called in by gocd

Vogtinator (changes requested):

gocd/vagrant-publisher.gocd.yaml
dcermak: Sadly, yes :-(

The dependencies of obs-to-vagrantcloud only exist for Ruby 3.0, but `/usr/bin/ruby` is ruby 2.7 in Tumbleweed and the ruby interpreter is not managed by alternatives…

gocd/vagrant-publisher.gocd.yaml
dcermak: I have renamed it to `vagrant-publisher.opensuse.gocd.yaml`.

gocd/vagrant-publisher.opensuse.gocd.yaml
```diff
@@ -0,0 +1,22 @@
+format_version: 3
+pipelines:
+  Vagrant.Publish:
```
Vogtinator, line 3: I was actually referring to the pipeline name, which should mention `openSUSE`.

The filename can stay as it was, it might contain also SLE pipelines in the future.

Vogtinator:
Looks good except for the pipeline name

gocd/vagrant-publisher.opensuse.gocd.yaml
```diff
@@ -0,0 +1,22 @@
...
+    group: Vagrant
```
coolo, line 4: If you introduce a new group, you won't see the pipeline. The access rights are per group - so please don't

gocd/vagrant-publisher.opensuse.gocd.yaml
dcermak: Like this?

gocd/vagrant-publisher.opensuse.gocd.yaml
dcermak: Is the new group name ok? I am not sure how gocd handles sub-groups.

coolo: I don't think there is something like a sub-group. the group we have is called openSUSE.checkers - and openSUSE users can operate it - and all Users can see it. By default new groups are only visible to admins - I don't think that's what you want.

dcermak: Hm, so what would you suggest that I use as the group?

Vogtinator: Pretty much everything uses `openSUSE.Checkers`

dcermak: `openSUSE.Checkers` it is then.

okurz:
when-i-deploy-to-production-i-do-it-on-friday-52721060


you are right. Better wait some hours for the afternoon to click the button ;)

PR #2620, opened 2021-08-12
pkglistgen: Handle loops in project path expansion

dirkmueller:
For some setups that have recursive paths we run into an endless
loop. Adding a visited project tracking to break the loop of expansion.

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -419,8 +419,10 @@ def attribute_value_delete(apiurl, project, name, namespace='OSRT', package=None
         apiurl, list(filter(None, ['source', project, package, '_attribute', namespace + ':' + name]))))
 
 @memoize(session=True)
-def repository_path_expand(apiurl, project, repo):
+def repository_path_expand(apiurl, project, repo, visited_repos=None):
     """Recursively list underlying projects."""
+    if visited_repos is None:
+        visited_repos = set()
     repos = [[project, repo]]
     meta = ET.fromstringlist(show_project_meta(apiurl, project))
     paths = meta.findall('.//repository[@name="{}"]/path'.format(repo))
@@ -431,7 +433,11 @@ def repository_path_expand(apiurl, project, repo):
 
     # ...which is expanded recursively
     if len(paths) > 0:
-        repos += repository_path_expand(apiurl, paths[-1].get('project', project), paths[-1].get('repository'))
+        p_project = paths[-1].get('project', project)
+        p_repository = paths[-1].get('repository')
+        if (p_project, p_repository) not in visited_repos:
+            visited_repos.add((p_project, p_repository))
+            repos += repository_path_expand(apiurl, p_project, p_repository, visited_repos)
     return repos
 
 @memoize(session=True)


Vogtinator:
What's an example for such a weirdly configured project?

osclib/core.py
```diff
@@ -414,26 +414,43 @@ def attribute_value_save(apiurl, project, name, value, namespace='OSRT', package
...
+
 def attribute_value_delete(apiurl, project, name, namespace='OSRT', package=None):
     http_DELETE(makeurl(
         apiurl, list(filter(None, ['source', project, package, '_attribute', namespace + ':' + name]))))
 
-@memoize(session=True)
+
 def repository_path_expand(apiurl, project, repo):
     """Recursively list underlying projects."""
+    visited_repos = set(f"{project}:{repo}")
```
Vogtinator, line 425: That's already done by `_recursive_repository_path_expand`, FWICT the split into separate functions isn't actually needed?

osclib/core.py
```diff
@@ -414,26 +414,43 @@ def attribute_value_save(apiurl, project, name, value, namespace='OSRT', package
...
+        repos += [[p_project, p_repository]]
 
     # ...which is expanded recursively
     if len(paths) > 0:
-        repos += repository_path_expand(apiurl, paths[-1].get('project', project), paths[-1].get('repository'))
+        p_project = paths[-1].get('project', project)
+        p_repository = paths[-1].get('repository')
+        if f"{p_project}:{p_repository}" not in visited_repos:
+            visited_repos.add(f"{p_project}:{p_repository}")
+            repos += _recursive_repository_path_expand(apiurl, p_project, p_repository, visited_repos)
```
Vogtinator, line 450: There's already `repos` which could be passed instead of `visited_repos`

Vogtinator (changes requested):
I thought a bit about this and I think this algorithm breaks for setups like:

```
devel:prj/tw
hack:repo/tw
devel:prj/tw
```

where it would no longer recurse into the bottom path. So the recursion check should only look at the last entry. Not sure how OBS behaves in such cases.

osclib/core.py
dirkmueller: no, we want to add the one we recursed on.

osclib/core.py
dirkmueller: the split isn't needed, I can do it with a default argument as well. changed.

dirkmueller:
> where it would no longer recurse into the bottom path. So the recursion check should only look at the last entry. Not sure how OBS behaves in such cases.

it actually does handle that now (visited_repos is only updated with the parent project/repository that we recurse into)

osclib/core.py
```diff
@@ -431,7 +431,11 @@ def repository_path_expand(apiurl, project, repo):
 
     # ...which is expanded recursively
     if len(paths) > 0:
-        repos += repository_path_expand(apiurl, paths[-1].get('project', project), paths[-1].get('repository'))
+        p_project = paths[-1].get('project', project)
+        p_repository = paths[-1].get('repository')
+        if (p_project, p_repository) not in visited_repos:
+            visited_repos.add((p_project, p_repository))
```
Vogtinator, line 437: I think you're falling into a nasty python trap here: defaulted arguments are persistent and mutable!
The next time this is called, the defaulted `repository_path_expand` has the content from last time...

```
def a(b, c=set()):
        c.add(b)
        return c

print(a("foo"))
print(a("bar"))
```

```
{'foo'}
{'foo', 'bar'}
```

Vogtinator:
Code looks good now, except for the trap you fell into

osclib/core.py
dirkmueller: I'm using the =None way of handling defaulted arguments.

osclib/core.py
```diff
@@ -419,8 +419,10 @@ def attribute_value_delete(apiurl, project, name, namespace='OSRT', package=None
         apiurl, list(filter(None, ['source', project, package, '_attribute', namespace + ':' + name]))))
 
 @memoize(session=True)
-def repository_path_expand(apiurl, project, repo):
+def repository_path_expand(apiurl, project, repo, visited_repos=None):
     """Recursively list underlying projects."""
+    if visited_repos is None:
+        visited_repos = set()
     repos = [[project, repo]]
     meta = ET.fromstringlist(show_project_meta(apiurl, project))
```
Vogtinator, line 427: To also avoid visiting the toplevel project, this should be added here already, and can then be removed in line 439:

```
    if visited_repos is None:
        visited_repos = set()
    visited_repos.add((project, repo))
    repos = [[project, repo]]
    meta = ET.fromstringlist(show_project_meta(apiurl, project))
```

Vogtinator:
It works fine here, and with the suggested change it does exactly what I'd intuitively expect it to do, at least for my test project:

```
  <repository name="recurse">
    <path project="home:favogt:recurtest" repository="path1"/>
    <path project="openSUSE.org:openSUSE:Factory" repository="standard"/>
    <path project="home:favogt:recurtest" repository="path2"/>
    <arch>x86_64</arch>
  </repository>
  <repository name="path2">
    <path project="home:favogt:recurtest" repository="path1"/>
    <path project="home:favogt:recurtest" repository="recurse"/>
  </repository>
  <repository name="path1">
    <path project="home:favogt:recurtest" repository="notthis"/>
  </repository>
  <repository name="notthis"/>
```

Results in:

``´
['home:favogt:recurtest', 'recurse'],
['home:favogt:recurtest', 'path1'],
['openSUSE.org:openSUSE:Factory', 'standard'],
['home:favogt:recurtest', 'path2'],
['home:favogt:recurtest', 'path1']
```

`osc buildinfo -d` shows:

```
  <path project="home:favogt:recurtest" repository="recurse" url="http://download.suse.de/ibs/home:/favogt:/recurtest/recurse/"/>
  <path project="home:favogt:recurtest" repository="path1" url="http://download.suse.de/ibs/home:/favogt:/recurtest/path1/"/>
  <path project="openSUSE.org:openSUSE:Factory" repository="standard"/>
  <path project="home:favogt:recurtest" repository="path2" url="http://download.suse.de/ibs/home:/favogt:/recurtest/path2/"/>
```

So there's the difference that OBS avoids the redundant revisiting of `path1`. Not ideal, but I can't think of a way that could break in practice.

osclib/core.py
Vogtinator: Actually, OBS doesn't allow this, which is understandable. So this doesn't really make any difference.

PR #2621, opened 2021-08-12
pkglistgen: Delete sources after creating the final solv files

Vogtinator:
When a project is switched from `refresh: true` to `refresh: false`, the
`15.1:update_123XXX.packages.xz` files are kept alongside the new "final"
`15.1:update.packages.xz`. However, the `update_and_solve` command looks at all
of those, which there can be a lot of, and takes a long time to look at
effectively redundant data.

With this commit, the old files will be deleted for projects with
`refresh: false` once the final solv file is created. This is also done for
already "finalized" projects in a separate block.

This is the first time I'm looking at this code, so I might be totally wrong here.
It looks like this is done manually for `openSUSE:Factory/000update-repos`
though.

### DIFF
--- a/pkglistgen/update_repo_handler.py
+++ b/pkglistgen/update_repo_handler.py
@@ -259,6 +259,13 @@ def update_project(apiurl, project):
 
         if os.path.exists(packages_file + '.xz'):
             print(path, 'already exists')
+
+            # Some old versions didn't do cleanup, do it now. Can be removed once everything was cleaned up
+            if not opts.get('refresh', False):
+                for file in glob.glob(os.path.join(repo_dir, '{}_*.packages.xz'.format(key))):
+                    os.unlink(file)
+                    package.delete_file(os.path.basename(file))
+
             continue
 
         solv_file = packages_file + '.solv'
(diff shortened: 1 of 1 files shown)


coolo:
Yes, the switch to false is done manually - I merge all existing solv files then. I don't think this would happen if you just switch to false and your code removing the input afterwards.

I wonder if it matters though - the only case I can think of is a package getting introduced during 15.4 development that is dropped in the final release.

Vogtinator:
> Yes, the switch to false is done manually - I merge all existing solv files then. I don't think this would happen if you just switch to false and your code removing the input afterwards.

Initially I thought it would merge the files, like the `_and_before` path, but it doesn't. So it would indeed lose information about temporarily added packages.

> I wonder if it matters though - the only case I can think of is a package getting introduced during 15.4 development that is dropped in the final release.

Also for maintenance, when e.g. 15.2 goes EOL and the updates repo is changed to `refresh: false`.

If this matters, it would be possible to merge those files, like you did manually.

I don't think this cleanup has that much of an impact anymore, it was just very visible on 15.4 update repos, with >900 solv files: https://build.opensuse.org/package/show/openSUSE:Leap:15.4/000update-repos?rev=27

That was probably fixed by b452cfba699ee305e31b1ce12c71423017b4a7df, but never cleaned up until now.

coolo:
So I don't think we should merge this as is. Switching a repo to False should be a one time action - and as such we shouldn't try to be smart after the fact.

We can create a subcommand to switch a repo and do what's needed instead.

PR #2623, opened 2021-08-19
Rebuildibility poc

jreidinger:
draft to have a tool to test a various approaches for detecting rebuildability issues.

Still WIP, but feel free to comment approach.


osclib/remote_project.py
```diff
@@ -0,0 +1,130 @@
...
+            return self._merge_inherited(res, self.metadata.linked_projects(recursive=True))
+        else:
+            return res
+
+    def _merge_inherited(self, res, linked_projects):
+        merge_dict = { pkg.name: pkg for pkg in res }
+        cleaning_regex = re.compile(r'\.[0-9]+$')
+        for project in linked_projects:
+            for pkg in project.get_packages(inherited = False):
+                # TODO: is there better way then regex?
```
coolo, line 65: Note that ruby2.5 also matches your regexp but is not maintenance specific, so relying on regexp is basically a no go. `/source/SUSE:SLE-15-SP2:Update/zypper.15590/_meta` will tell you `<releasename>zypper</releasename>`.

osclib/remote_package.py
```diff
@@ -0,0 +1,43 @@
...
+
+        :raises:
+          PackageNotFound: if the package is not found in the given project
+        """
+        url = osc.core.make_meta_url('pkg', (project_name, package_name), osc.conf.config['apiurl'])
+        try:
+            return cls.from_xml(osc.core.http_GET(url))
+        except HTTPError as e:
+            if e.code == 404:
+                raise PackageNotFound('Package %s not found' % (package_name))
```
coolo, line 43: you need to reraise in else - or you will get very strange results

rebuildability_check.py
```diff
@@ -0,0 +1,64 @@
...
+import osc.core
+import osc.conf
+
+import osclib.remote_project
+from osclib.remote_project import RemoteProject
+
+class RebuildabilityChecker(object):
+    def __init__(self, project_str):
+        self.logger = logging.getLogger('RebuildibilityChecker')
+        self.project = RemoteProject.find(project_str) # apiurl should be read from osc.conf.config['apiurl'], osclib config class looks like has different goal?
```
coolo, line 16: that might be just me but could you limit your line length to a value that makes reading the comments possible without scrolling left and right on github page? That would be awesome

osclib/remote_package.py
```diff
@@ -0,0 +1,43 @@
...
+
+class RemotePackage(object):
+    """ This class represents a package on the build service side """
+    def __init__(self, name, project=None):
+        self.name = name
+        self.project_name = project
+
+    def copy(self, target_project_name, expand=False):
+        apiurl = osc.conf.config['apiurl']
+        osc.core.copy_pac(apiurl, self.project_name, self.name, apiurl, target_project_name, self.name, expand=expand)
```
coolo, line 21: note that copy_pac is greatly expensive operating. For such use cases a revisioned _link is much clearly preferred.

rebuildability_check.py
```diff
@@ -0,0 +1,64 @@
...
+    def result(self):
+        packages = self.project.get_packages(inherited = True) # package should include also project from which it comes. Recursive means include also inherited packages so reading meta and linked projects are needed
+        self.logger.debug("Packages %s" % [pkg.project_name + " -> " + pkg.name for pkg in packages])
+
+        title = "Testing rebuild of whole parent project"
+        description = "Temporary project including expanded copy of parent to verify if everything can be rebuild from scratch"
+        rebuild_project = self.project.create_subproject("Rebuild", title, description) # how to handle if it exists? Clean it and use? What if we do not have permission to create subproject?
+
+        testing_packages = [package.copy(rebuild_project.name, expand=True) for package in packages]
+        while not all([p.builds.is_finished for p in testing_packages]): # builds is object for handling building and is_finished means that all builds are finished ( or disabled )
```
coolo, line 27: I assume this is supposed to read as pseudo code? Because you don't need to query all packages' build status just to find out that one is not finished (taking that `is_finished` is going to be an external API call).

rebuildability_check.py
```diff
@@ -0,0 +1,64 @@
...
+            # write some progress about number of pass, failures, what is waiting, etc. using p.builds query ( ensure it is not memoized )
+            print("Working like a crazy monk")
+
+        # TODO: delete rebuild project or keep it for inspection?
```
coolo, line 31: in principle: all succeeded packages are normally uninteresting. The failed ones you might need to show around. Not sure the yast team ever accepted a "it failed, fix it - that's all I have!" :)

But depending on how often you plan to run this, you surely do not want to have a lot of these projects around as they fill up several hundered GB easily. So reusing them can be key.

osclib/remote_project.py
jreidinger: Good catch with that ruby...in general I a bit worry about performance as when I check SLE15-SP4 there was a lot of updates in update channel, but lets not preoptimalize in advance.

rebuildability_check.py
jreidinger: goal of this comments is mainly discussion about code where we cooperate. It should not be there before we merge it.
But thanks for catch

osclib/remote_package.py
jreidinger: yeah, I expect it is expensive. I just want to avoid any diff issue. Revisioned _link can be solution as I expect in SLE15 it won't change much. If we want to use it on staging it can be more tricky as staging change more frequently and link can be broken. ( question is how big issue it is as it means that it is removed from staging )

rebuildability_check.py
jreidinger: yeah, we decide to start with pseudo code how the API can look like and adapt it as we implement it. Maybe we can even find some project specific query to find unfinished build.

rebuildability_check.py
jreidinger: Maybe it can be even parameter specific e.g. for SLE-15-SP4 I expect we want to keep it and just reuse project when we want another rebuild test. For some other usage like maintenance update, different behavior can make sense.

osclib/remote_package.py
coolo: if you plan to rebuild all of SLE15 history against a moving target you want to go back to the drawing board anyway.

rebuildability_check.py
coolo: Before you discuss such things, learn how long it takes to create and recompile a single project. So before you start to worry about reports and cleanup, just create one of these beasts

jreidinger: yep, that is a goal of this tool, to start playing with it and as side-effect create a bit of high level API that helps with whatever we decided to do.

rebuildability_check.py
```diff
@@ -23,7 +23,7 @@ def result(self):
         description = "Temporary project including expanded copy of parent to verify if everything can be rebuild from scratch"
         rebuild_project = self.project.create_subproject("Rebuild", title, description) # how to handle if it exists? Clean it and use? What if we do not have permission to create subproject?
 
-        testing_packages = [package.copy(rebuild_project.name, expand=True) for package in packages]
+        testing_packages = [package.link(rebuild_project.name) for package in packages]
```
coolo, line 26: note that only works for truely static targets. In case of update projects, you really want to have a revision to your link. Like osc linkpac --current does

jreidinger:
OK, so first run finished. It uses code after commit https://github.com/openSUSE/openSUSE-release-tools/pull/2623/commits/ede0f67ade8f6b19775642638445bfc37be0b98c
repo is at https://build.suse.de/project/show/home:jreidinger:SP4-rebuild-testing:Rebuild
runtime for creating  such repo is
```
real	231m10.171s
user	11m26.858s
sys	1m36.694s
```

building of such repo started 23.08. at 22:00

coolo:
Don't hold your breath :)

image

jreidinger:
@coolo nice graph (where I can find it?), so we have starting point where we can just improve. At least it already catches several dependency and build failures, so it proof it is kind of useful ( and we can compare how many catches other approaches ).

coolo:
On the project page, click on the repo name, ignore the disgusting amount of build cyles and look for "Check the rebuild time for x86_64"

jreidinger:
thanks

imobachgs:
> What we did in the past: Query /source/?view=info and then compare `<filename>` with package name

Thanks! I found out exactly that query last night checking [the OBS code](https://github.com/openSUSE/open-build-service/blob/master/src/api/app/controllers/public_controller.rb#L43-L69) :smile: because I felt that things were too complicated.

coolo:
Bootstrap is indeed interesting for embedded but for a fraction of the package. But it's impossible if the package sources aren't rebuildable to begin with - and that's what the issue asks for IMO.

jreidinger:
OK, I fix issue with getting proper source project name and now even more interesting issue appear. Build info for update project is empty. No clue why. Will ask on monday on build service chat - https://api.suse.de/build/SUSE:SLE-15-SP3:Update/standard/x86_64/_builddepinfo?view=revpkgnames

coolo:
https://chat.suse.de/channel/help-obs?msg=MqjFS46pCdKJd7LHm

coolo:
Closing the PR - leaving the branch

PR #2625, opened 2021-08-24
skippkg-finder: overwrites NON_FTP_PACKAGES.group for obsoleted/unneeded package from SLE or openSUSE

nilxam:
`NON_FTP_PACKAGES.group` maintaining a list of package(RPM) to be skipped in ftp-tree build, after CtLG, any binary built from SLE15 scope product including incidents from update has appeared in *pool* repository, there are many outdated and obsoleted binary still presented in case ftp-tree's build method is `wildcard(*)` - selecting anything does exist, therefore `skippkg-fidner` helps to find those *obsoleted* package from SLE, and since it will overwrite the skip-list entirely, `skippkg-finder` is handling *unneeded* SLE/openSUSE specific package also. This script should runs before `pkglistgen` since `pkglistgen` would parse `NON_FTP_PACKAGES.group` while creating ftp kiwi.

Related: https://github.com/openSUSE/openSUSE-release-process/issues/75 and https://github.com/openSUSE/openSUSE-release-process/issues/74
Test run result: https://build.opensuse.org/package/show/openSUSE:Leap:15.4/000package-groups

### DIFF
--- a/skippkg-finder.py
+++ b/skippkg-finder.py
@@ -1,0 +1,372 @@
+#!/usr/bin/python3
+
+import argparse
+import logging
+import sys
+
+from urllib.error import HTTPError
+
+import re
+from lxml import etree as ET
+from collections import namedtuple
+
+import osc.conf
+import osc.core
+from osc.core import http_GET
+from osc.core import makeurl
+from osc import oscerr
+import osclib
+from osclib.core import source_file_ensure
+
+SUPPORTED_ARCHS = ['x86_64', 'i586', 'aarch64', 'ppc64le', 's390x']
+DEFAULT_REPOSITORY = 'standard'
+
+META_PACKAGE = '000package-groups'
+
+class SkippkgFinder(object):
+    def __init__(self, opensuse_project, sle_project, print_only, verbose):
+        self.opensuse_project = opensuse_project
+        self.sle_project = sle_project
+        self.print_only = print_only
+        self.verbose = verbose
+        self.apiurl = osc.conf.config['apiurl']
+        self.debug = osc.conf.config['debug']
+
+    def is_sle_specific(self, package):
+        """
+        Return True if package is provided for SLE only or a SLE forking.
+        Add new condition here if you do not want package being added to
+        selected_binarylist[].
+        """
+        pkg = package.lower()
+        prefixes = (
+                'desktop-data',
+                'libyui-bindings',
+                'libyui-doc',
+                'libyui-ncurses',
+                'libyui-qt',
+                'libyui-rest',
+                'lifecycle-data-sle',
+                'kernel-livepatch',
+                'kiwi-template',
+                'mgr-',
+                'migrate',
+                'patterns',
+                'release-notes',
+                'sap',
+                'sca-',
+                'skelcd',
+                'sle-',
+                'sle_',
+                'sle15',
+                'sles15',
+                'spacewalk',
+                'supportutils-plugin',
+                'suse-migration',
+                'susemanager-',
+                'yast2-hana'
+                )
+        suffixes = (
+                '-caasp',
+                '-sle',
+                'bootstrap'
+                )
+        matches = (
+                'gtk-vnc2',
+                'ibus-googlepinyin',
+                'infiniband-diags',
+                'llvm',
+                'lua51-luajit',
+                'lvm2-clvm',
+                'osad',
+                'rhncfg',
+                'python-ibus',
+                'python-pymemcache',
+                'suse-build-key',
+                'suse-hpc',
+                'txt2tags',
+                'zypp-plugin-spacewalk',
+                'zypper-search-packages-plugin'
+                )
+        if pkg.startswith(prefixes) or pkg.endswith(suffixes) or pkg in matches:
+            return True
+        if 'sles' in pkg or\
+                'sled' in pkg or\
+                'sap-' in pkg or\
+                '-sap' in pkg or\
+                'eula' in pkg or\
+                'branding' in pkg:
+            return True
+        return False
+
+    def get_packagelist(self, project, by_project=True):
+        """
+        Return the list of package's info of a project.
+        If the latest package is from an incident then returns incident
+        package.
+        """
+
+        pkglist = {}
+        packageinfo = {}
+        query = {'expand': 1}
+        root = ET.parse(http_GET(makeurl(self.apiurl, ['source', project],
+                                 query=query))).getroot()
+        for i in root.findall('entry'):
+            pkgname = i.get('name')
+            orig_project = i.get('originproject')
+            is_incidentpkg = False
+            # Metapackage should not be selected
+            if pkgname.startswith('000') or\
+                    pkgname.startswith('_') or\
+                    pkgname.startswith('patchinfo.') or\
+                    pkgname.startswith('skelcd-') or\
+                    pkgname.startswith('installation-images') or\
+                    pkgname.endswith('-mini'):
+                continue
+            # Ugly hack for package has dot in source package name
+            # eg. go1.x incidents as the name would be go1.x.xxx
+            if '.' in pkgname and re.match(r'[0-9]+$', pkgname.split('.')[-1]) and \
+                    orig_project.startswith('SUSE:') and orig_project.endswith(':Update'):
+                is_incidentpkg = True
+                if pkgname.startswith('go1') or\
+                        pkgname.startswith('bazel0') or\
+                        pkgname.startswith('dotnet') or\
+                        pkgname.startswith('rust1') or\
+                        pkgname.startswith('ruby2'):
+                    if not (pkgname.count('.') > 1):
+                        is_incidentpkg = False
+
+            # If an incident found then update the package origin info
+            if is_incidentpkg:
+                orig_name = re.sub(r'\.[0-9]+$', '', pkgname)
+                incident_number = int(pkgname.split('.')[-1])
+                if orig_name in pkglist and pkglist[orig_name]['Project'] == orig_project:
+                    if re.match(r'[0-9]+$', pkglist[orig_name]['Package'].split('.')[-1]):
+                        old_incident_number = int(pkglist[orig_name]['Package'].split('.')[-1])
+                        if incident_number > old_incident_number:
+                            pkglist[orig_name]['Package'] = pkgname
+                    else:
+                        pkglist[orig_name]['Package'] = pkgname
+            else:
+                pkglist[pkgname] = {'Project': orig_project, 'Package': pkgname}
+
+        if by_project:
+            for pkg in pkglist.keys():
+                if pkglist[pkg]['Project'].startswith('SUSE:') and self.is_sle_specific(pkg):
+                    continue
+                if pkglist[pkg]['Project'] not in packageinfo:
+                    packageinfo[pkglist[pkg]['Project']] = []
+                if pkglist[pkg]['Package'] not in packageinfo[pkglist[pkg]['Project']]:
+                    packageinfo[pkglist[pkg]['Project']].append(pkglist[pkg]['Package'])
+            return packageinfo
+
+        return pkglist
+
+    def get_project_binary_list(self, project, repository, arch, package_binaries={}):
+        """
+        Returns binarylist of a project
+        """
+
+        # Use pool repository for SUSE namespace project.
+        # Because RPMs were injected to pool repository on OBS rather than
+        # standard repository.
+        if project.startswith('SUSE:'):
+            repository = 'pool'
+
+        path = ['build', project, repository, arch]
+        url = makeurl(self.apiurl, path, {'view': 'binaryversions'})
+        root = ET.parse(http_GET(url)).getroot()
+
+        for binary_list in root:
+            package = binary_list.get('package')
+            package = package.split(':', 1)[0]
+            index = project + "_" + package
+
+            if index not in package_binaries:
+                package_binaries[index] = []
+            for binary in binary_list:
+                filename = binary.get('name')
+                result = re.match(osclib.core.RPM_REGEX, filename)
+                if not result:
+                    continue
+
+                if result.group('arch') == 'src' or result.group('arch') == 'nosrc':
+                    continue
+                if result.group('name').endswith('-debuginfo') or result.group('name').endswith('-debuginfo-32bit'):
+                    continue
+                if result.group('name').endswith('-debugsource'):
+                    continue
+
+                if result.group('name') not in package_binaries[index]:
+                    package_binaries[index].append(result.group('name'))
+
+        return package_binaries
+
+    def exception_package(self, package):
+        """
+        Do not skip the package if matches the condition.
+        package parameter is source package name.
+        """
+
+        if '-bootstrap' in package or\
+                'Tumbleweed' in package or\
+                'metis' in package:
+            return True
+        # These packages must have a good reason not to be single-speced
+        # from one source.
+        if package.startswith('python2-') or\
+                package.startswith('python3'):
+            return True
+        return False
+
+    def exception_binary(self, package):
+        """
+        Do not skip the binary if matches the condition
+        package parameter is RPM filename.
+        """
+
+        if package == 'openSUSE-release' or\
+                package == 'openSUSE-release-ftp' or\
+                package == 'openSUSE-Addon-NonOss-release':
+            return True
+        return False
+
+    def crawl(self):
+        """Main method"""
+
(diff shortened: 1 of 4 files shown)


skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+DEFAULT_REPOSITORY = 'standard'
+BINARY_REGEX = r'(?:.*::)?(?P<filename>(?P<name>.*)-(?P<version>[^-]+)-(?P<release>[^-]+)\.(?P<arch>[^-\.]+))'
+RPM_REGEX = BINARY_REGEX + r'\.rpm'
+
+META_PACKAGE = '000package-groups'
+
+makeurl = osc.core.makeurl
+http_GET = osc.core.http_GET
+http_POST = osc.core.http_POST
+http_PUT = osc.core.http_PUT
```
jreidinger, line 31: why not `from osc.core import http_PUT` and same for others?

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+from urllib.error import HTTPError
+
+import re
+from lxml import etree as ET
+from collections import namedtuple
+
+import osc.conf
+import osc.core
+from osc import oscerr
+from osc.util.helper import decode_it
```
jreidinger, line 16: where it is used?

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+from osclib.core import source_file_ensure
+
+OPENSUSE = 'openSUSE:Leap:15.4'
+SLE = 'SUSE:SLE-15-SP4:GA'
+SUPPORTED_ARCHS = ['x86_64', 'i586', 'aarch64', 'ppc64le', 's390x']
+DEFAULT_REPOSITORY = 'standard'
+BINARY_REGEX = r'(?:.*::)?(?P<filename>(?P<name>.*)-(?P<version>[^-]+)-(?P<release>[^-]+)\.(?P<arch>[^-\.]+))'
```
jreidinger, line 23: as they are regexp why not `BINARY_REGEX = re.compile(r'...` and some comment explaining regexp would be nice as regexps are kind of tricky.

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument('-A', '--apiurl', metavar='URL', help='API URL')
+    parser.add_argument('-d', '--debug', action='store_true',
+                        help='print info useful for debuging')
+    parser.add_argument('-o', '--opensuse-project', dest='opensuse_project', metavar='OPENSUSE_PROJECT',
+                        help='openSUSE project on buildservice (default: %s)' % OPENSUSE,
+                        default=OPENSUSE)
+    parser.add_argument('-s', '--sle-project', dest='sle_project', metavar='SLE_PROJECT',
+                        help='SLE project on buildservice (default: %s)' % SLE,
+                        default=SLE)
```
jreidinger, line 322: I am not sure if those defaults are good idea as they become obsolete in next release, so we either break compatibility with updated default or has obsolete default.

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+
+    uc = SkippkgFinder(args.opensuse_project, args.sle_project, args.print_only, args.verbose)
+    uc.crawl()
+
+
+if __name__ == '__main__':
+    description = 'Overwrites NON_FTP_PACKAGES.group according to the latest sources. '\
+                  'This tool only works for Leap after CtLG implemented.'
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument('-A', '--apiurl', metavar='URL', help='API URL')
```
jreidinger, line 314: here on other way I expect default to api.opensuse.org

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+        if pkg.endswith('bootstrap') or pkg.endswith('-caasp') or\
+                pkg.endswith('-sle'):
+            return True
+        if pkg == 'suse-build-key' or pkg == 'suse-hpc' or\
+                pkg == 'llvm' or\
+                pkg == 'txt2tags' or\
+                pkg == 'zypper-search-packages-plugin' or\
+                pkg == 'python-pymemcache' or\
+                pkg == 'python-ibus':
+            return True
```
jreidinger, line 79: there is a lot of hardcoded stuff and I kind of miss rationally for bunch of them. Consider that you are promoted and someone like me need to adapt script for next version. So at least document how to obtain such list of SLE specific stuff.
Also e.g. parts with startswith can be quite error prone. Any software starting with patterns or migrate is matched.
Also that generic sap matcher can be error prone e.g. even `pinesap-grow`

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+                    pkgname.endswith('-mini'):
+                continue
+            # ugly hack for go1.x incidents as the name would be go1.x.xxx
+            if '.' in pkgname and re.match(r'[0-9]+$', pkgname.split('.')[-1]) and \
+                    orig_project.startswith('SUSE:') and orig_project.endswith(':Update'):
+                is_incidentpkg = True
+                if pkgname.startswith('go1') or pkgname.startswith('bazel0') or \
+                        pkgname.startswith('dotnet') or pkgname.startswith('ruby2'):
+                    if not (pkgname.count('.') > 1):
+                        is_incidentpkg = False
```
jreidinger, line 112: this can be error prone. We are facing something similar in https://github.com/openSUSE/openSUSE-release-tools/pull/2623 and use releasename - https://github.com/openSUSE/openSUSE-release-tools/pull/2623/files#diff-c1d94e5890752eab31a911fc75e1bcada9d337a7315c97eb8fe8c7046e9f7d30R85

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+            for pkg in pkglist.keys():
+                if pkglist[pkg]['Project'].startswith('SUSE:') and self.is_sle_specific(pkg):
+                    continue
+                if pkglist[pkg]['Project'] not in packageinfo:
+                    packageinfo[pkglist[pkg]['Project']] = []
+                if pkglist[pkg]['Package'] not in packageinfo[pkglist[pkg]['Project']]:
+                    packageinfo[pkglist[pkg]['Project']].append(pkglist[pkg]['Package'])
+            return packageinfo
+
+        return pkglist
```
jreidinger, line 138: looks like you need similar functionality like https://github.com/openSUSE/openSUSE-release-tools/pull/2623/files#diff-22a4ac30776b144d4e04816a4e9ccfc2f8999d985f675e160c7a21bf88d2a775R52

maybe we can coordinate and submit already that RemotePackage and RemoteProject objects?

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+
+    def get_project_binary_list(self, project, repository, arch, package_binaries={}):
+        """
+        Returns binarylist of a project
+        """
+
+        if project.startswith('SUSE:'):
+            repository = 'pool'
```
jreidinger, line 146: why this overwrite is needed?

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+        Return true if the given project or package exists
+        """
+
+        if package:
+            url = makeurl(self.apiurl, ['source', project, package, '_meta'])
+        else:
+            url = makeurl(self.apiurl, ['source', project, '_meta'])
+        try:
+            http_GET(url)
+        except HTTPError:
```
jreidinger, line 188: this is error prone as it does not need to be 404 but e.g. also temporary 500. We also have similar method https://github.com/openSUSE/openSUSE-release-tools/pull/2623/files#diff-22a4ac30776b144d4e04816a4e9ccfc2f8999d985f675e160c7a21bf88d2a775R87

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+            return False
+        return True
+
+    def origin_metadata_get(self, project, package):
+        """
+        Returns origin infomration
```
jreidinger, line 194: typo infomration

skippkg-finder.py
```diff
@@ -0,0 +1,333 @@
...
+        Do not skip the package if matches the condition
+        """
+
+        if package.startswith('python2') or package.startswith('python3') or \
+                package.startswith('preinstallimage-'):
+            return True
+        if package.endswith('-bootstrap'):
+            return True
+        if 'Tumbleweed' in package or 'metis' in package:
+            return True
```
jreidinger, line 228: I expect there rationale for non obvious exceptions like metis or python

skippkg-finder.py
nilxam: That is a pain for openSUSE release engineer to consider what SLE package should not pick to openSUSE, *none can tell*, those hardcoded stuff here can be 1) it has been agreed to fork in openSUSE which means we have different version between Leap and SLE , 2) really should not take to openSUSE, like desktop-data-SLE and such.

Yes, we do rebuild patterns in Leap, so any patterns from SLE should not take to openSUSE, migration stuff also, and of course it's true for sap stuff. We don't have `pinesap-grow` exist in SLE.

skippkg-finder.py
nilxam: I know, but the reason is I don't want to file an GET event on every package including every incidents, otherwise the script can not finish in 1 day or 2 even, suppose that all of packages from SLE15:GA, SLE15:Update, SP1:GA, SP1:Update ... SP4:GA + Backports + Leap... code looks ugly indeed for those special case like `go1.11.xxxxx` or others, but it saves time.

skippkg-finder.py
nilxam: maybe, but those special handling in my code have to be there, and I can not skip `kernel-livepatch` stuff, then the script can find them and skips them afterward when added to NON_FTP_PACKAGES.group

skippkg-finder.py
nilxam: That is part of CtLG design, SLE RPMs has injected to pool repository on OBS, not standard repository, so you will not find RPMs in standard repo if SLE: namespace project.

skippkg-finder.py
nilxam: there are some python pacakge wasn't single-spec'ed, eg. python-abc is python3 only, and there is python2-abc for the compatibility reason. metis is a special case since it was conflicting to scotch however we can not really delete a package in SLE, see https://bugzilla.suse.com/show_bug.cgi?id=1182799

jreidinger: thanks for explanation. It would be nice to have in code so we do not forget in future when the need to touch script appear.

skippkg-finder.py
jreidinger: maybe that special name skips can be parametrized, so we all benefit from that code?

skippkg-finder.py
jreidinger: @imobachgs I think working on making it more efficient, so hopefully in future we will have faster detection of such  cases.

skippkg-finder.py
nilxam: I didn't look into RemoteProject and RemotePackage closely, the major 'packagelist` in my code stores package's binarylist which would be selected to ftp-tree, eg. `yast2`
if you do `/source/openSUSE:Leap:15.4?view=info` - inherited ones merged, `yast2` is from `SUSE:SLE-15-SP3:Update`, and it wasn't the right place that RPMs from, it is `yast2.19962`, so `SUSE:SLE-15-SP3:Update/yast2.19962` RPMs have to be selected to ftp-tree, the binarylist of `yast2.19962` has stored to that major `packagelist`.

Any other RPM from other `yast2.xxxxx` updates will not be selected in case it does not exist in `SUSE:SLE-15-SP3:Update/yast2.19962`'s binarylist, these `others` will add to `skiplist`.

jreidinger: interesting, thanks for explanation

skippkg-finder.py
nilxam: the default is from osc config, isn't it?

skippkg-finder.py
nilxam: to me, if `/source/openSUSE:Leap:15.4?expand=1` can have 1) a `originpacakge` attrib to each entry, and 2) a `releasename` attrib to each entry, then my code can be simplified a lot :)

maybe like

    <entry name="yast2" originproject="SUSE:SLE-15-SP3:Update" originpackage="yast2.19962"/>
    <entry name="yast2.19962" originproject="SUSE:SLE-15-SP3:Update" originpackage="yast2.19962" releasename="yast2"/>

skippkg-finder.py
nilxam: I didn't bother to create another one, just copy it from https://github.com/openSUSE/openSUSE-release-tools/blob/master/osclib/core.py#L36 , I need `<filename>` only though.

skippkg-finder.py
```diff
@@ -0,0 +1,331 @@
...
+                if pkgname.startswith('go1') or pkgname.startswith('bazel0') or \
+                        pkgname.startswith('dotnet') or pkgname.startswith('ruby2'):
+                    if not (pkgname.count('.') > 1):
+                        is_incidentpkg = False
+
+            # if an incident then update the package origin info
+            if is_incidentpkg:
+                orig_name = re.sub(r'\.[0-9]+$', '', pkgname)
+                incident_number = int(pkgname.split('.')[-1])
+                if orig_name in pkglist and pkglist[orig_name]['Project'] == orig_project:
```
nilxam, line 111: BTW the code here is comparing incident number of the same source package from the matched `originproject`, the newest one has picked, avoiding spend time to poke package's meta for incident package name.

skippkg-finder.py
jreidinger: yes, it is. Just I am not sure if it should depends on it here. What if it is run by someone who has as default api.suse.de? but it is not critical note. It just surprise me.

skippkg-finder.py
jreidinger: yeah, for us too. Maybe we can ask build service guys what they think about it

skippkg-finder.py
jreidinger: well, if it is copy, maybe you can use it directly. either call `osclib.core.BINARY_REGEX` or if you want to import it then `from osclib.core import BINARY_REGEX` works in a same way and as bonus if regex change to match newer binaries, your code start using it.

skippkg-finder.py
imobachgs: Using the `/source/openSUSE:Leap:15.4?view=info` you can have the origin, but the `releasename` is still missing:

```xml
<sourceinfo package="yast2" rev="2" vrev="3.9" srcmd5="62841f713933618908b699c060d1d747" lsrcmd5="3e4e02a6e681a21402659868e83b6c2a" verifymd5="7189b727ca2be435c4a45fcc4241e7dc">
<filename>yast2.spec</filename>
<originproject>SUSE:SLE-15-SP3:Update</originproject>
<linked project="openSUSE:Backports:SLE-15-SP4" package="yast2"/>
<linked project="SUSE:SLE-15-SP4:GA" package="yast2"/>
<linked project="SUSE:SLE-15-SP3:Update" package="yast2"/>
<linked project="openSUSE:Leap:15.4" package="yast2.19962"/>
<linked project="openSUSE:Backports:SLE-15-SP4" package="yast2.19962"/>
<linked project="SUSE:SLE-15-SP4:GA" package="yast2.19962"/>
<linked project="SUSE:SLE-15-SP3:Update" package="yast2.19962"/>
</sourceinfo>
```

And, as @nilxam commented, an additional GET is not an option. I will ask the build service team for some help.

nilxam: the url `/source/openSUSE:Leap:15.4?expand=1` in `skippkg_finder` gives `originproject` already, just need to find the proper linked incident package in case it pointing to an incident, same for `?view=info` though

skippkg-finder.py
```diff
@@ -0,0 +1,330 @@
...
+        if pkg.startswith('skelcd') or pkg.startswith('release-notes') or\
+                pkg.startswith('sle-') or pkg.startswith('sle_'):
+            return True
+        if 'sles' in pkg or 'sled' in pkg:
+            return True
+        if 'sap-' in pkg or '-sap' in pkg or pkg.startswith('sap'):
+            return True
+        if 'eula' in pkg or 'branding' in pkg:
+            return True
+        if pkg.startswith('sle15') or\
```
imobachgs, line 49: You could simplify this condition by just passing a list of prefixes:

```python
PREFIXES=('sle15', 'sles15', 'suse-migration') # just a subset
if pkg.startswith(PREFIXES):
  # whatever
```

PR #2626, opened 2021-08-25
Small OBSLocal refactoring

ancorgs:
While preparing al alternative to the work done by @jreidinger in the `build_test` branch, I found pretty inconvenient how `OBSLocal.StagingWorkflow` is coupled to the Factory case. Simulating an SLE-like scenario with it is kind of inconvenient (you basically ignore all the projects created by the constructor and start from scratch).

This pull request tries to make OBSLocal a bit more useful by:

- Making it a bit more generic (not so coupled to simulate the Factory scenario)
- Improving the documentation in several parts (and tweaking some function names to better state what they do)
- Adding some functionality regarding the execution of several review bots in the same test (extracted from `TestFactorySubmitRequest`).

It's probably much easier to review this commit by commit.


joseivanlopez (approved):
LGTM

PR #2631, opened 2021-09-03
Add pipeline to setup QR home project

gyr:


gocd/checkers.suse.gocd.yaml.erb
```diff
@@ -312,3 +312,31 @@ pipelines:
...
                 osc sr --yes SUSE:SLE-15-<%= servicepack %>:Update:QR -m 'Update Package List'
 <% end -%>
+<% ['SP3'].each do |servicepack| -%>
+  SLE15.<%= servicepack %>.QR.Setup:
+    group: SLE.Checkers
+    lock_behavior: unlockWhenFinished
+    environment_variables:
+      OSC_CONFIG: /home/go/config/oscrc-maintenance-installcheck
+    timer:
+      spec: 0 0 0 ? * THU
```
coolo, line 322: I think a manual pipeline doesn't need a timer :)

gyr: Fixed. Copied/pasted pipeline setup from SLE15.SP3.QR without double checking it :(

PR #2632, opened 2021-09-06
skippkg-finder: add two packages to the extra list for zypper dup reason

nilxam:


jreidinger (approved):

nilxam:
Note: for those extras, I will move it to OBS attribute or metapackage if I need to tweak it so often.

PR #2633, opened 2021-09-07
Remove attributes in tests

joseivanlopez:
## Problem

The *factory_submit_request_test* requires the attribute *ProducVersion* to exist. This is because accepting a submit request requires a *ProductVersion* to properly work. But the *factory_submit_request_test* does not fail even though it does not set a *ProductVersion* at all. This is because there are dependencies between tests.

The attributes like the *ProductVersion* are not deleted when a test finishes. So, *factory_submit_request_test* will work if a previous test creates a *ProductVersion* attribute. For example, if *accept_test* is executed before, then *factory_submit_tests* passes. Otherwise, it would fail.   

## Solution

Always ensure that a *ProductVersion* attribute is created in order to make possible other actions like accepting a submit request. And clean up all the attributes when a Staging Workflow is destroyed. This avoids dependencies between tests.

### DIFF
--- a/tests/OBSLocal.py
+++ b/tests/OBSLocal.py
@@ -195,6 +195,7 @@ class StagingWorkflow(object):
         self.requests = []
         self.groups = []
         self.users = []
+        self.attributes = {}
         logging.basicConfig()
 
         # clear cache from other tests - otherwise the VCR is replayed depending
@@ -230,6 +231,10 @@ class StagingWorkflow(object):
         self.config = Config(APIURL, project)
 
     def create_attribute_type(self, namespace, name, values=None):
+        if not namespace in self.attributes: self.attributes[namespace] = []
+
+        if not name in self.attributes[namespace]: self.attributes[namespace].append(name)
+
         meta = """
         <namespace name='{}'>
             <modifiable_by user='Admin'/>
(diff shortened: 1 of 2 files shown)


imobachgs (approved):

jreidinger (approved):

tests/OBSLocal.py
```diff
@@ -217,6 +218,8 @@ def __init__(self, project=PROJECT):
         self.setup_remote_config()
         self.load_config()
         self.api = StagingAPI(APIURL, project)
+        # The ProductVersion is required for some actions, for example, when a request is accepted
+        self.create_attribute_type('OSRT', 'ProductVersion', 1)
```
ancorgs, line 222: Doesn't this belong to `setup_remote_config`? Just like the call to `self.create_attribute_type('OSRT', 'Config', 1)`. Isn't it?

ancorgs: Moreover, I think that doing this after the call to `load_config` is specially bad, since it guarantees `self.config` is outdated in comparison to what the OBS instance holds. Or am I lost?

ancorgs: Hmmm, not really. `load_config` loads the value of the attribute `Config` associated to the target project. That has nothing to do with the existence of an attribute type called `ProductVersion`.

As far as I understand, attributes are associated to concrete projects and attribute types are a global thing that defines which attributes can exist. So there is no direct relationship between `self.config' and the attribute types beyond `Config` itself.

So maybe my fault, but calling `self.attributes` to the attribute holding the list of attribute types (not to be confused with real attributes) didn't make it any easier to understand.

ancorgs: I tried to address that confusion as part of https://github.com/openSUSE/openSUSE-release-tools/pull/2626

PR #2634, opened 2021-09-07
skippkg-finder: move extra packages to the OSRT attribute

nilxam:
The extra packages are packages should be skipped in ftp-tree but we won't for some reasons, packagelist is not fixed depend on package status in SLE side, move the list to OSRT attribute avoiding modify this list so often.

### DIFF
--- a/skippkg-finder.py
+++ b/skippkg-finder.py
@@ -17,6 +17,7 @@ from osc.core import makeurl
 from osc import oscerr
 import osclib
 from osclib.core import source_file_ensure
+from osclib.conf import Config
 
 SUPPORTED_ARCHS = ['x86_64', 'i586', 'aarch64', 'ppc64le', 's390x']
 DEFAULT_REPOSITORY = 'standard'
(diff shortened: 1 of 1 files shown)


skippkg-finder.py
```diff
@@ -32,6 +33,12 @@ def __init__(self, opensuse_project, sle_project, print_only, verbose):
         self.apiurl = osc.conf.config['apiurl']
         self.debug = osc.conf.config['debug']
 
+        config = Config.get(self.apiurl, self.opensuse_project)
+        # binary rpms of packages from `skippkg-finder-skiplist-ignores`
+        # be found in the `package_binaries` thus format must to be like
+        # SUSE:SLE-15:Update_libcdio.12032, PROJECT-NAME_PACKAGE-NAME
+        self.skiplist_ignored = set(config.get('skippkg-finder-skiplist-ignores', '').split(' '))
```
jreidinger, line 40: is somewhere documentation of all attributes? Maybe it would be worth to document format ( space separated ) and intention for that attribute.

nilxam: The default attribute for config thingy is `OSRT:Config`, see https://github.com/openSUSE/openSUSE-release-tools/blob/master/osclib/core.py#L375 and https://github.com/openSUSE/openSUSE-release-tools/blob/master/osclib/conf.py#L273 .

There is no strict format rule for `values`, like OSRT:Config at https://build.opensuse.org/attribs/openSUSE:Factory some uses comma and some uses space for splitting. And unfortunately I don't remember we have documented every `value`, perhaps some has comment out in code.

nilxam:
I'm eager to have a new build because some *extras* have to be added, I'll merge this now.

PR #2636, opened 2021-09-15
Switch to pytest from nose

s-t-e-v-e-n-k:
The nosetests test runner has not been touched upstream for years, and
is at best unmaintained. Switch from using it to using pytest. Also
change the docs to specify Python 3 modules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,8 @@ install:
 check: test
 
 test:
-	# to see more add -v -d -s --nologcapture
-	$(wildcard /usr/bin/nosetests-2.*) -c .noserc
+	# to see more add -vv
+	/usr/bin/pytest
 
 package:
 	touch dist/package/$(package_name).changes
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -89,7 +89,7 @@ BuildArch:      noarch
 Requires:       libxml2-tools
 Requires:       python3-httpretty
 Requires:       python3-mock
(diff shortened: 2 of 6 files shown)


dist/package/openSUSE-release-tools.spec
```diff
@@ -89,7 +89,7 @@ BuildArch:      noarch
 Requires:       libxml2-tools
 Requires:       python3-httpretty
 Requires:       python3-mock
-Requires:       python3-nose
+Requires:       python3-pytest
```
coolo, line 92: I guess this subpackage should just die. Developers run tests in the container

Makefile
```diff
@@ -34,8 +34,8 @@ install:
 check: test
 
 test:
-	# to see more add -v -d -s --nologcapture
-	$(wildcard /usr/bin/nosetests-2.*) -c .noserc
+	# to see more add -vv
+	/usr/bin/pytest
```
coolo, line 38: I guess the same goes here - the tests won't work with a make test as they depend on running in a container. So just removing this should help

coolo:
I was suprised to see your PR as @jreidinger already converted to pytest some time ago. But it seems you touched mostly documentation and dead code.

s-t-e-v-e-n-k:
My whole plan was to remove the nose Requires on the -devel package, and ended with git grep nose

s-t-e-v-e-n-k:
But okay, I shall close this PR, and open a new one to remove the -devel package entirely, and the make target.

PR #2638, opened 2021-09-16
Test to showcase a submit request to SLE

ancorgs:
This is the SLE equivalent to #2617 

As an important "bonus track", this include the class `SLEWorkflow(StagingWorkflow)` that allows to easily setup the "onion" of projects used to implement package inheritance in SLE. It should be convenient to easily write future SLE-focused tests. From that point of view, this is the cherry on top for #2626.


jreidinger (approved):

joseivanlopez (approved):
LGTM. Quite well explained! Thanks a lot, it helps so much to understand the SLE workflow.

PR #2642, opened 2021-09-21
Fix weakremovers

coolo:


pkglistgen/tool.py
```diff
@@ -431,15 +431,17 @@ def create_weakremovers(self, target, target_config, directory, output):
 
                 pool.createwhatprovides()
 
+                accepted_archs = set(self.all_architectures)
+                accepted_archs.add('noarch')
+
                 for s in oldsysrepo.solvables_iter():
-                    if s.arch == 'src':
+                    if not s.arch in accepted_archs:
```
Vogtinator, line 438: `not in`

pkglistgen/tool.py
```diff
@@ -431,15 +431,17 @@ def create_weakremovers(self, target, target_config, directory, output):
 
                 pool.createwhatprovides()
 
+                accepted_archs = set(self.all_architectures)
```
Vogtinator, line 434: Can this possibly have issues with `i686` vs `i586` (and similar flavors)?

coolo: we can swap the arch replacement with the check

pkglistgen/tool.py
coolo: There are 127 of such uses - that's E713 btw

Vogtinator: Is it at least consistently wrong?

Vogtinator:
Should work as designed, just can't tell for sure whether the design is correct...

PR #2643, opened 2021-09-21
Remove E265 from flake ignore list

coolo:


coolo:
@Vogtinator made me look there!

Vogtinator (approved):

abichecker/abichecker.py
```diff
@@ -546,7 +546,7 @@ def check_one_request(self, req):
         else:
             # we probably don't want abichecker to spam here
             # FIXME don't delete comment in this case
-            #if state is None and not self.text_summary:
+            # if state is None and not self.text_summary:
             #    self.text_summary = 'abichecker will take a look later'
```
Vogtinator, line 550: Technically needs to be this, but who cares:
```suggestion
            #     self.text_summary = 'abichecker will take a look later'
```

coolo: abichecker is actually completely ignored in flake - it's just my autofixer that went there. So I can take out that diff - but who cares :)

PR #2647, opened 2021-09-22
Adds dependency for full check_source testing

jreidinger:


coolo (approved):

jreidinger:
Looks like it is not needed. see https://github.com/openSUSE/openSUSE-release-tools/pull/2648#issuecomment-925330101

Issue #2640, opened 2021-09-21
#2586 breaks detection of SLE-only patch removal


suntorytimed:
We have recently stumbled upon several declines of the factory-auto bot for SLE 15 SP4 submissions. It seems like PR #2586 is breaking a behaviour (or rather fixing a bug) we relied on in the SLE submit process. More specifically following change in line 216:

```
-    next unless $line =~ m/^+/;
+    next unless $line =~ m/^\+/;
```

We rely on this for cases in which a patch never got submitted to Factory but only in SLE (e. g. when Factory already has a newer version with the fix in it) and we want to sync the SLE version with the one from Factory again. This results in patches being dropped, which were never added to Factory. An example would be this submission: https://build.suse.de/request/show/252417

We have defined an exception for such cases in the Factory First Policy and openSUSE documentation, that it is enough to add those removed patches to the submit request message instead. (see https://en.opensuse.org/openSUSE:Creating_a_changes_file_(RPM)#Cross_Service-Pack_merges_for_SLE and “What if the changelog differs?” on https://wiki.suse.net/index.php/SUSE/Factory_First_Policy)

DimStar77:
The bug fix also made another thing impossible,namely 'revert submits' on TW:

Consider this:
* Package FOO is submitted from the devel prj to rev 2, it adds a patch
* the SR is accepted (passed all tests)
* As RelMgr I realize it causes way too much havoc, bugs found, fails, whatever, so I create an sr to revert to rev 1 (`osc sr openSUSE:Factory FOO -r 1 openSUSE:Factory -m "Revert" --nodevel`)

This revert will also make the patch vanish; due to the bug, this was accepted so far, newly this will fail.

The logic to limit to lines starting with `+` in the diff is this causing some trouble (we should also consider `-` lines)

DimStar77:
```
-    next unless $line =~ m/^\+/;
+    next unless $line =~ m/^[\+\-]/;
```

?

Vogtinator:
I'd try `next unless $line =~ m/^[-+]/;`

DimStar77:
> I'd try `next unless $line =~ m/^[-+]/;`

might work - I never know when one needs to escape + and -; the + not being escaped was the original bug, but then it was not inside []

coolo:
I don't think we want that either. we want to accept patches *removed* to appear in -, not added patches.

DimStar77:
[...]

PR #2648, opened 2021-09-22
check_source: Fix revert case for patch detection (fix#2640)

jreidinger:
depends on https://github.com/openSUSE/openSUSE-release-tools/pull/2647


check_source.pl
```diff
@@ -213,8 +212,10 @@ BEGIN
                 $diff = diff "$old/$changes", "$dir/$changes";
             }
             for my $line (split(/\n/, $diff)) {
-                next unless $line =~ m/^\+/;
-                $line =~ s/^\+//;
+                # Check also if changes do not revert additional of patch
```
jreidinger, line 215: typo additional instead of addition ( but feel free to correct my grammar )

tests/OBSLocal.py
```diff
@@ -557,12 +557,16 @@ def create_submit_request(self, project, package, text=None):
...
+               if package already exists
+        :type add_commit: bool
         :return: created request.
         :rtype: Request
         """
         project = self.create_project(project)
         package = Package(name=package, project=project)
-        package.create_commit(text=text)
+        if add_commit:
+            package.create_commit(text=text)
```
jreidinger, line 569: the condition is neeed as check source checker otherwise complain about not mentioned file README created in this method at https://github.com/openSUSE/openSUSE-release-tools/pull/2648/files#diff-f26c9a37e9fec1909032289e143103fe74d3ca070d31b9512a0e14a51c1ae985R1008
I keep default true to avoid break of other tests

tests/OBSLocal.py
```diff
@@ -1015,7 +1019,7 @@ def commit_files(self, path):
         :param path: path to a directory containing the files that must be added to the package
         """
         for filename in os.listdir(path):
-            with open(os.path.join(path, filename)) as f:
+            with open(os.path.join(path, filename), 'rb') as f:
```
jreidinger, line 1022: opening as binary is needed for gzipped tarball.

jreidinger:
Interesting that test passes even without https://github.com/openSUSE/openSUSE-release-tools/pull/2647 so it is probably just warning and confuse me, when I play with docker compose. So I expect it is ready now even without that format spec.

tests/OBSLocal.py
ancorgs: Why is that a comment here and not in the code?

tests/OBSLocal.py
```diff
@@ -557,12 +557,16 @@ def create_submit_request(self, project, package, text=None):
         :type package: str
         :param text: commit message for initial package creation
         :type text: str
+        :param add_commit: whenever add initial package commit. Useful to disable
```
ancorgs, line 560: I guess this is "whether", not "whenever"

tests/check_source_tests.py
```diff
@@ -69,7 +69,59 @@ def test_devel_project(self):
...
 
-        req_id = self.wf.create_submit_request(SRC_PROJECT, 'blowfish').reqid
+        req_id = self.wf.create_submit_request(SRC_PROJECT, 'blowfish', add_commit = False).reqid
+
+        self.assertReview(req_id, by_user=(self.bot_user, 'new'))
+
+        self.review_bot.set_request_ids([req_id])
+        self.review_bot.check_requests()
+
+        self.assertReview(req_id, by_user=(self.bot_user, 'accepted'))
```
ancorgs, line 79: JFYI: you could save some lines by using `assertReviewScript` (you first need to assign `self.script = 'check_source.py`) or `assertReviewBot` (you need to setup the bot using `setup_review_bot` instead of manually... which would also save you some lines in the tear up).

tests/check_source_tests.py
```diff
@@ -69,7 +69,59 @@ def test_devel_project(self):
...
+        )
+
+    def test_patch_in_changelog(self):
+        """Accepts a request if it adds patch and it is mentioned in changelog"""
+        self._setup_devel_project()
+
+        req_id = self.wf.create_submit_request(self.devel_package.project,
+            self.devel_package.name, add_commit = False).reqid
+
+        self.assertReview(req_id, by_user=(self.bot_user, 'new'))
```
ancorgs, line 108: Same again, you could save some lines if you wish. This basically duplicates the code of `assertReviewBot`

tests/OBSLocal.py
jreidinger: makes sense, will add it.

tests/check_source_tests.py
jreidinger: yeah, I duplicate existing code to prevent another source if mistakes, but cleaning makes sense.

check_source.pl
ancorgs: Apart from the grammar, I miss a comment with some explanation that could prevent this from happening again in the future. If not an explanation, at least a link to https://github.com/openSUSE/openSUSE-release-tools/issues/2640 or https://github.com/openSUSE/openSUSE-release-tools/pull/2586#issuecomment-923755993

ancorgs: What about something like this?

```python
# Check whether the line mentions a patch being added (starts with +) or removed (starts with -)
next unless $line =~ m/^[+-]/;
# If any of those cases, remove the patch from the list
$line =~ s/^[+-]//;
```

ancorgs (approved):
LGTM

jreidinger:
Lets merge it as it cause trouble in reverts and SLE workflow and fix for TW glibc/docker issue can take some time.

PR #2649, opened 2021-09-23
Enabling for openSUSE:Leap:15.3:Update:Respin

nilxam:
* Added an alternative_project option to skippkg-finder
* Added openSUSE:Leap:15.3:Update:Respin to ttm, pkglistgen and updaterepos


gocd/pkglistgen.opensuse.gocd.yaml.erb
```diff
@@ -84,6 +84,38 @@ pipelines:
...
+  if project.size > 1
+    options=" -s #{project[1]}"
+    name = name + "_#{project[1]}"
+  end
+  -%>
+          <%= name %>:
+            resources:
+            - repo-checker
+            tasks:
+            - script: python3 ./skippkg-finder.py -A https://api.opensuse.org -o openSUSE:Leap:15.3:Update:Respin -t openSUSE:Leap:15.3 -s SUSE:SLE-15-SP3:GA
```
jreidinger, line 115: Why is there SP3:GA and not SP3:Update?

nilxam: Good catch! I also changed command to check `openSUSE:Leap:15.3:Update` instead.

jreidinger (approved):

PR #2651, opened 2021-09-30
CI: Switch to 15.3 env (maintained in OBS until we can switch back)

coolo:


coolo:
let's go with priv?ledged TW

PR #2652, opened 2021-09-30
CI: Run Tumbleweed container as priveleged in docker

coolo:
This avoids the syscall filter breaking with new glibc

### DIFF
--- a/dist/ci/docker-compose.yml
+++ b/dist/ci/docker-compose.yml
@@ -7,6 +7,7 @@ x-testenv: &testenv
   image: registry.opensuse.org/opensuse/tools/images/containers_tumbleweed/osrt_testenv_tumbleweed:latest
   volumes:
     - "../..:/code"
+  privileged: true
 
 services:
   db:


okurz (commented):
/priveleged/privileged/

you put the right word in source and then mistyped in the commit message :D

ancorgs:
I'm failing to reproduce the failed test locally in my computer. So I guess I will open a separate pull request just for doing some trial-and-error on github actions.

ancorgs:
On the bright side, changing this the test suceeds. https://github.com/openSUSE/openSUSE-release-tools/pull/2653/commits/b78844333ff90db117c5223600adac551ea752d1

That proves the problem is not really in the licensedigger bot (which was heavily mocked and thus should not have failed). But in factory-auto that is not as easy to please as it was before (although it keeps working for me in local). That's in fact good news (to some degree, I would be happier if I could reproduce locally).

coolo:
> you put the right word in source and then mistyped in the commit message :D

Of course I copy&pasted into the source :)

ancorgs:
This alternative PR fixes the unveiled problem https://github.com/openSUSE/openSUSE-release-tools/pull/2654

coolo:
yours is without typo

PR #2654, opened 2021-10-01
CI: Run Tumbleweed containers as privileged including fixes

ancorgs:
This includes the fix from #2652 and also fixes for the errors it revealed in the tests (that are actually a consequence of e0d3ba3c95dfb9567ac52ecb6a7fb2372fea655b)


jreidinger (approved):

tests/factory_submit_request_tests.py
```diff
@@ -61,7 +61,7 @@ def setUp(self):
         FreezeCommand(self.wf.api).perform(STAGING_PROJECT_NAME)
 
         # Create the submit request
-        self.request = self.wf.create_submit_request(DEVEL_PROJECT, 'wine')
+        self.request = self.wf.create_submit_request(DEVEL_PROJECT, 'wine', add_commit=False)
```
jreidinger, line 64: funny part is that I use `add_commit=True` as default to be backwards compatible

PR #2656, opened 2021-10-11
try to skip channel files as submitted with kernels and livepatches, as

msmeissn:
they cannot be legal reviewed

i tried various approaches, but this should work hopefully.


coolo:
http://legaldb.suse.de/reviews/details/268568 - I see

PR #2665, opened 2021-11-01
Move remaining vagrant publishers from Jenkins to botmaster

dcermak:


coolo:
If one of them fails, the rest won't run. If you want to avoid that, make them individual tasks.

dcermak:
> If one of them fails, the rest won't run. If you want to avoid that, make them individual jobs of the same stage.

I have not considered that, thanks for the suggestion!

gocd/vagrant-publisher.gocd.yaml
```diff
@@ -18,5 +18,224 @@ pipelines:
         - staging-bot
         tasks:
         - script: |
-            ruby.ruby3.0 obs-to-vagrantcloud.rb --url https://download.opensuse.org/tumbleweed/appliances/boxes/Tumbleweed.x86_64.json --organization opensuse --new-box-name Tumbleweed.x86_64 --provider libvirt
-            ruby.ruby3.0 obs-to-vagrantcloud.rb --url https://download.opensuse.org/tumbleweed/appliances/boxes/Tumbleweed.x86_64.json --organization opensuse --new-box-name Tumbleweed.x86_64 --provider virtualbox
+            ruby.ruby3.0 obs-to-vagrantcloud.rb --url https://download.opensuse.org/tumbleweed/appliances/boxes/Tumbleweed.x86_64.json --organization opensuse --new-box-name Tumbleweed.x86_64
+
+  openSUSE.Tumbleweed.aarch64.Vagrant.Publish:
```
coolo, line 23: no no no, these are pipelines. I talked about different jobs in the same pipeline.

Compare gocd/pkglistgen.opensuse.gocd.yaml.erb (you might take advantage from .erb as well as you WET a lot here

dcermak: Oh, ok. Apologies, I'm not very familiar with gocd yet.

I have converted it to a erb template and that makes matters a lot shorter indeed, thanks for the suggestion!

PR #2668, opened 2021-11-03
add option to use newest package for resolving dependencies

dirkmueller:
In some cases we'd like to chose from all available repositories
for selecting the right package. in appliance builds and there
is an OBS-UseUnorderedRepos flag for enabling this behavior. Similarly
product-builds have a flag "use_newest_version" that can be set
to true. pkglistgen needs to know about these variant as well
to properly resolve package dependencies.

### DIFF
--- a/pkglistgen/tool.py
+++ b/pkglistgen/tool.py
@@ -225,7 +225,7 @@ class PkgListGen(ToolBase.ToolBase):
                     solvable.unset(solv.SOLVABLE_CONFLICTS)
                     solvable.unset(solv.SOLVABLE_OBSOLETES)
                 # only take the first solvable in the repo chain
-                if solvable.name in solvables:
+                if not self.use_newest_version and solvable.name in solvables:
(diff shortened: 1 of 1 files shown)


pkglistgen/tool.py
```diff
@@ -575,6 +575,7 @@ def update_and_solve_target(self, api, target_project, target_config, main_repo,
                                 project, scope, force, no_checkout,
                                 only_release_packages, stop_after_solve):
         self.all_architectures = target_config.get('pkglistgen-archs').split(' ')
+        self.use_newest_version = str2bool(target_config.get('pkglistgen-use-newest-version'))
```
coolo, line 578: I fear you're missing a default for all the other projects

dirkmueller: it simply defaults to false which matches normal behavior

https://github.com/openSUSE/openSUSE-release-tools/blob/8f4795d135733193b7350e7c92c6107149f10c81/osclib/conf.py#L196

dirkmueller: we can be more explicit, fixed.

coolo (approved):

PR #2669, opened 2021-11-18
Fix handling missing log files

witekest:
The logs before 2021-09-05 are not available anymore.
The change takes account of it and skips parsing older logs.
Additionally, the database is not dropped on every run.


lkocman:
Looks good

ddemaio:
@witekest and @lkocman metrics.opensuse.org seems to show an error "html/template: "index" is undefined" I'm not sure if this is from the merge. It seems like the error might have come before that. https://status.opensuse.org/ doesn't show this, so I'm not sure if it's the server causing the issue or the code.

PR #2671, opened 2021-11-22
 legal-auto: Move the "ignore channels" feature in the proper function

coolo:
The query_sources function is only used for product scan, which is not
relevant for request review. Fixing it to improve performance on live
patch reviews.


msmeissn:
looks good to me!

PR #2673, opened 2021-12-08
metrics: telegraf: Fix URL for openqa.opensuse.org after machines were moved to openSUSE internal network

okurz:
Related progress issue: https://progress.opensuse.org/issues/102975


ggardet:
Thanks @okurz !!

PR #2676, opened 2021-12-15
Add 15.4/ARM Images totestmanager

dirkmueller:


ggardet (approved):
LGTM.  
We need to have a ttmconfig in OBS prior to the merge, though.

PR #2688, opened 2022-02-02
Check whether the FTP trees contain OSRT:ProductVersion before releasing

Vogtinator:
Otherwise it might release a build which uses the wrong version.

### DIFF
--- a/ttm/releaser.py
+++ b/ttm/releaser.py
@@ -229,6 +229,17 @@ class ToTestReleaser(ToTestManager):
                                            self.project.product_repo, arch):
                         return False
 
+        # The FTP tree isn't released with setrelease, so it needs to contain
+        # the product version already.
+        product_version = self.get_product_version()
+        if product_version is not None:
+            for product in self.project.ftp_products:
+                for binary in self.binaries_of_product(self.project.name, product):
+                    # The NonOSS tree doesn't include the version...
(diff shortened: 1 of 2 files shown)


ttm/releaser.py
```diff
@@ -229,6 +229,17 @@ def is_snapshotable(self):
...
 
+        # The FTP tree isn't released with setrelease, so it needs to contain
+        # the product version already.
+        product_version = self.get_product_version()
+        if product_version is not None:
+            for product in self.project.ftp_products:
+               for binary in self.binaries_of_product(self.project.name, product):
+                   # The NonOSS tree doesn't include the version...
+                   if binary.endswith('.report') and 'NonOss' not in binary and product_version not in binary:
+                       self.logger.debug(f'{binary} in {product} does not include {product_version}')
```
DimStar77, line 240: "does not include" or "does not match" ?

Rest seems all right

Vogtinator: Really "does not include". For Leap, the `ProductVersion` is `15.4`, so fully matching that is not trivial without running into more edge cases.

DimStar77: for Leap, OSRT:ProductVersion is hard-set to 15.4; so the check is close to NOP for Leap, but should at least not break

DimStar77 (approved):

PR #2692, opened 2022-02-10
ToolBase: use --osc-debug for osc's config value, not the tools own debug parameter

DimStar77:


DimStar77:
@coolo Otherwise, ttm --debug spits out all the osc-debug too, and when using --debug --osc-debug, it's spit out twice

DimStar77:
The root of evil was faa760d61de5cf9a796b4535c0e6e1428e064997

PR #2696, opened 2022-02-15
Support to move one staging into another

### LINKED ISSUE #1105: osc staging: possibility to merge two staging projects
Every now and then, it would be handy to be able to 'merge' two staging projects (more commonly on adi, but sometimes also on letter stagings)

maybe any of our overly complex syntax options already does it - but I could not work out a set :)

Something like: osc staging merge adi:25 adi:65 -> move all packages from adi:25 to adi:65

(currently I do this using a shell function:

```
adimerge () 
{ 
    osc ls openSUSE:Factory:Staging:adi:$1 | xargs osc staging select --move openSUSE:Factory:Staging:adi:$2
}
```

)

coolo:
osc staging select --move adi:26 adi:14

If move is given, all staging projects given after the first will be seen as source

Fixes #1105

### DIFF
--- a/docs/staging.asciidoc
+++ b/docs/staging.asciidoc
@@ -159,12 +159,19 @@ selecting the request.  If we want to move a request from one staging
 project to another one, we need to use the parameter +--move+:
 
 --------------------------------------------------------------------------------
-osc staging select --move B 12345
+osc staging select B --move 12345
 --------------------------------------------------------------------------------
 
 This command will move #12345 from openSUSE:Factory:Staging:A to
 openSUSE:Factory:Staging:B
 
+You can also merge staging projects by giving other staging projects as arguments:
+
+--------------------------------------------------------------------------------
+osc staging select B --move A C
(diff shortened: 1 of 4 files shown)


docs/staging.asciidoc
```diff
@@ -165,6 +165,11 @@ osc staging select --move B 12345
 This command will move #12345 from openSUSE:Factory:Staging:A to
 openSUSE:Factory:Staging:B
 
+You can also merge staging projects by giving other staging projects as arguments:
+
+--------------------------------------------------------------------------------
+osc staging select --move B A C
```
DimStar77, line 171: I'd either write this as `osc staging select B --move A C` or add a further comment that this command merges A and C into B

coolo: As you wish, not that I think anyone ever read this file :)

DimStar77: like every doc; docs are just written so we can say RTFM

DimStar77 (approved):

PR #2697, opened 2022-02-15
Factory staging: report pkglistgen status

DimStar77:


coolo (approved):
try it.

DimStar77:
> try it.

saw some errors already - `PYTHONPATH` was not set to scripts, which means reporting would have failed...

coolo:
Remember to make it a required check if you want to block it accepting stagings

PR #2700, opened 2022-02-16
staging: consider first staging a target even without --move

### LINKED ISSUE #2699: 'osc staging select D $pkg' -> tries to select 'D' and finds a staging for it
Regression from dafa42d3dfafcee79137fe17a223e83b1924087f

```
> osc staging select D kdepim-runtime 
0#requests_61b97a9@all:
  bootstrap_required: false
  group: all
  requests:
    955387: kdepim-runtime
  staging: E
  strategy:
    args:
      requests:
      - D
      - kdepim-runtime
    name: requests

Accept proposal? [y/n] (y): n
```

so it understood 'D' as being the name of a package that it wanted to stage, and tried to find a staging for it (the first empty one being E at the moment)

DimStar77:
Fixes #2699

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -451,7 +451,7 @@ def do_staging(self, subcmd, opts, *args):
                 # also makes it consistent to distinguish them from request IDs.
                 #
                 # also support --move passing 2 or more staging projects to merge
-                if arg in existing_stagings and arg not in stagings and (len(stagings) == 0 and opts.move):
+                if arg in existing_stagings and arg not in stagings and not (len(stagings) > 0 and opts.move):
                     stagings.append(api.extract_staging_short(arg))
                 elif arg not in requests:
                     requests.append(arg)


DimStar77:
With this, I could do 'Staging select D rook':

```
> osc staging select D rook
(1/1) Adding request "955388" to project "openSUSE:Factory:Staging:D"
```

merging adi:

```
> osc staging select adi:15 --move adi:24
(1/1) Moving "955345" from "openSUSE:Factory:Staging:adi:24" to "openSUSE:Factory:Staging:adi:15"
```

If specifying a 2nd adi without move, it is interpreted as a package name:

```
> osc staging select adi:15 adi:24
No SR# found for: adi:24
```

which sounds acceptable

coolo:
The fix is wrong. Not sure when this was used for the last time, but the use case back then was "staging select D E" and then it would picked a strategy to shuffle between D and E

DimStar77:
> The fix is wrong. Not sure when this was used for the last time, but the use case back then was "staging select D E" and then it would picked a strategy to shuffle between D and E

oh - those ominous strategies... the problem is that the current code is even worse than the broken (unused) strategies..

What would be the correct fix in your opinion?

something like:

```
if arg in existing_stagings and arg not in stagings and ((len(stagings) == 0 and opts.move) or not opts.move):
       stagings.append(api.extract_staging_short(arg))
```

?

coolo:
yes, good thinking. Now if you convert that de morgan style you end up with `and not ((opts.move or len(stagings) > 0) and opts.move)`. Now opts.move is both and and or, so reduce it to and and you'll get `and not (opts.move and len(stagings) > 0)`

So: we want to skip adding staging if we have move and it's the 1+ staging. Easy, eh? :)

To be honest I had the feeling the condition is wrong, but in the examples I tried it worked, so I though my feeling is wrong.

osc-staging.py
```diff
@@ -451,7 +451,7 @@ def do_staging(self, subcmd, opts, *args):
                 # also makes it consistent to distinguish them from request IDs.
                 #
                 # also support --move passing 2 or more staging projects to merge
-                if arg in existing_stagings and arg not in stagings and (len(stagings) == 0 and opts.move):
+                if arg in existing_stagings and arg not in stagings and (len(stagings) == 0):
```
coolo, line 454:
```suggestion
               if arg in existing_stagings and arg not in stagings and not (len(stagings) > 0 and opts.move)
```

coolo (changes requested):

coolo (approved):

PR #2702, opened 2022-02-17
Remove --by-develproject for adi command

### LINKED ISSUE #1366: adi: remove --by-develproject flag in favor of project config
At this point everything was changed to work with the flag always set as part of `staging-bot` service. As such it seems like either just dropping the flag or setting based on project config would be best since it is not something the user should be specifying (ie. it should either always be included or always not) and as such is just cumbersome. The code should be reviewed to see if the flag can simply be removed given the changes made to how it works.

coolo:
It's pointless by now as we have no project that uses devel projects
and would accept submissions not coming from there

Fixes #1366 (kindof)

### DIFF
--- a/gocd/checkers.suse.gocd.yaml.erb
+++ b/gocd/checkers.suse.gocd.yaml.erb
@@ -131,7 +131,6 @@ pipelines:
 
                osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA rebuild
                osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA list --supersede
-               osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA adi --by-develproject
                osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA select --non-interactive --merge --try-strategies
                osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA unselect --cleanup
                osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA repair --cleanup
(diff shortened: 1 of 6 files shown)


gocd/checkers.suse.gocd.yaml.erb
```diff
@@ -131,7 +131,6 @@ pipelines:
 
                osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA rebuild
                osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA list --supersede
-               osc -A https://api.suse.de staging -p SUSE:SLE-15-SP4:GA adi --by-develproject
```
DimStar77, line 134: no need to run 'adi' at all? Even without `--by-develproject`?

coolo: SP4 and MicroOS don't have rings ...

osc-staging.py
```diff
@@ -451,7 +449,7 @@ def do_staging(self, subcmd, opts, *args):
                 # also makes it consistent to distinguish them from request IDs.
                 #
                 # also support --move passing 2 or more staging projects to merge
-                if arg in existing_stagings and arg not in stagings and (len(stagings) == 0 and opts.move):
+                if arg in existing_stagings and arg not in stagings and not (len(stagings) > 0 and opts.move):
```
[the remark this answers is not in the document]
coolo: that was most likely still in my editor from the experiments I did yesterday. Will clean this

DimStar77 (approved):

PR #2703, opened 2022-02-18
Harmonize imports on using lxml for XML parsing

coolo:


coolo:
Every usage of `osc.core` is potentially harmful as some of the functions return `xml.etree` and e.g. ET.tostring will fail then. But let's hope our tools are either covered or easy to repair.

PR #2705, opened 2022-02-18
Replace variable name 'l' (x is fine)

coolo:
https://www.flake8rules.com/rules/E741.html


gocd/rabbit-openqa.py
```diff
@@ -2,6 +2,8 @@
 
 import argparse
 import logging
+
+from debugpy import listen
```
coolo, line 6: oops

PR #2708, opened 2022-02-18
Targeting E123 - Closing bracket does not match indentation

coolo:
https://www.flake8rules.com/rules/E123.html


.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 exclude = abichecker
 max-line-length = 100
-ignore = E501,F401,E128,E251,E201,E202,E302,E305,F841,E261,E712,E126,E711,E125,E123,E101,E124,E127,E701,E714,W504,E129,E741
+ignore = W503,W504,E501,F401,E128,E251,E201,E202,E302,E305,F841,E261,E712,E126,E711,E125,E124,E127,E701,E714,E129,E741
```
coolo, line 4: W503 is a warning and the warning says that it's actually wrong to warn about it and flake will soon warn about the opposite - so let's ignore it :)

ReviewBot.py
```diff
@@ -765,6 +765,15 @@ def comment_write(self, state='done', result=None, project=None, package=None,
 
         self.comment_handler_remove()
```
coolo, line 766: @DimStar77 as we learned this week my logic code can use a reviewer. So if you may...

PR #2714, opened 2022-02-21
Create codeql-analysis.yml

coolo:
Just testing what it brings


coolo:
I can't see any value in it

PR #2717, opened 2022-02-22
Remove the origin operator and its user scripts

coolo:
The service is down for 2 years and more and it was unused ever since.
The idea just never took off (also because it was pretty late in the
cycle and leap wasn't developed much longer the way it used to when
the design for this was created).

For the next SLE/Leap it's just very, very unclear how it will look
like and I'm 99% sure it's not going to go back to old Leap. So the
whole operator needs more work and maintenance than the current team
has, so keep it in the git history only.


Vogtinator:
+1 :-/

This also removes the code for staging DND, which is not necessarily tied together with the origin commands, but I suppose that could be implemented in OBS itself fairly easily nowadays as well...

Vogtinator (approved):

ancorgs (approved):
I'm all for every change that reduces the "noise" in the repo. No offense intended by the term "noise" - sure the operator was a great tool,  but having it in the master branch just brings confusion nowadays (it will always live in git history).

coolo:
The drag & drop uses the operator (and is relying on the old staging dashboard code)

PR #2718, opened 2022-02-22
skippkg_finder: fix unconditional matches if the first regex element is space

nilxam:
In case regex list has not been defined any value, the first element will be space only, this causes unconditional matched any binary name. Exit the for loop if the first element is space only.


nilxam:
merging... this is urgent for 15.3 Respin.

PR #2720, opened 2022-02-25
pkglistgen: Also support zstd for weakremovers.inc

coolo:


pkglistgen/update_repo_handler.py
```diff
@@ -236,7 +218,7 @@ def fixate_target(root, package, fixate):
                 return False
             del opts['refresh']
             oldfiles = target_files(package.dir, key)
-            newfile = os.path.join(package.dir, f'{fixate}_package')
+            newfile = os.path.join(package.dir, f'{fixate}_packages')
```
coolo, line 221: evil typo :(

PR #2722, opened 2022-02-25
freeze: Copy weakremovers.inc from target project to avoid bitrot

### LINKED ISSUE #1924: freeze command: copy weakremovers.inc
The weakremovers.inc in stagings get old - but we don't want to calculate them. So having freeze take the one from target project should be good enough

coolo:
Fixes #1924


coolo:
CC @DimStar77

DimStar77 (approved):
LGTM

PR #2731, opened 2022-03-04
pkglistgen: Deprecate scope 'all'

coolo:
Within gocd we run each job on its own worker and don't rely on builtin loop. So no need to care for a remote config


coolo:
@DimStar77 can you please remove it from OSRT:Config in openSUSE:Factory ? It's unused now already

Issue #1186, opened 2017-10-10
botmaster: employ log monitoring for various services to detect errors


jberry-suse:
Given the importance of these services at minimum some basic alerting should be setup to notify re-occurring errors. There have been a number of instances where bots were failing to do something over and over, but nothing is done until someone notices.

jberry-suse:
[IT Ticket #94035](https://infra.nue.suse.com/Ticket/Display.html?id=94035)

coolo:
I can't see that IT ticket, but assuming nothing has happened there. botmaster is mirrored for running server and running agents - but the pipelines themselves are not monitored. Problem is that some of them are bound to report an error - even if it's reported elsewhere.

coolo:
I wonder how we can do that. We definitely still notice too late if a service keeps failing - we need botmaster to monitor itself somehow.

coolo:
First step: I enabled "Breaks" notification on gocd for myself.

coolo:
Those emails turned out to be very useful to find seldom crashes. And everyone (from SUSE, sorry) can add notifications for pipelines of interest. There are more pipelines failing than what are problems (e.g. build failures blocking further stages), but those are easy to ignore.

PR #2733, opened 2022-03-04
pkglistgen: Don't exit 1 if we skip weakremovers due to restarted builds

coolo:
Since 30daff5613803667be7422056431116e84c07d59 we check if the repos are
still what we expected them to be when we started and stop calculating
weakremovers.inc if the build didn't stay still for the full runtime of
the pkglistgen. But the exception makes it impossible to spot real
failures - so make this a regular exit


coolo:
Related to https://github.com/openSUSE/openSUSE-release-tools/issues/1186

PR #2740, opened 2022-03-07
leapmicro: add leapmicro to pipeline

nilxam:


gocd/leapmicro.target.gocd.yaml
```diff
@@ -0,0 +1,39 @@
...
+      only_on_changes: false
+    materials:
+      git:
+        git: https://github.com/openSUSE/openSUSE-release-tools.git
+    stages:
+    - Update.000product:
+        resources:
+        - repo-checker
+        tasks:
+          - script: ./pkglistgen.py -A https://api.opensuse.org --debug update_and_solve -p openSUSE:Leap:Micro:5.2 -s target --force
```
coolo, line 38: why would you run this with force on time base?

PR #2741, opened 2022-03-07
installcheck: Report all delete requests

coolo:
It's just very annoying otherwise


coolo:
CC @DimStar77

DimStar77 (approved):

coolo:
Working on the timeout problem - the code isn't covered though, so merging

PR #2743, opened 2022-03-08
tests: Make use of fixtures to be able to test different workflows

coolo:
Check_source supports multiple different target projects only
different by their remote config. So put this config into a fixture,
so we can mark every test with what config it should run


ancorgs (approved):
LGTM. I didn't know about pytest fixtures. I like how they help to clarify the context for each test/functionality.

PR #2750, opened 2022-03-14
Staging installcheck: Do not overwrite package variable

coolo:
Fixing strange looking installcheck reports talking about the wrong
package


nilxam (approved):
:+1:

PR #2752, opened 2022-03-17
pkglisten: Generate a comment for staging diffs

coolo:
This is the next step towards having multiple submissions to SLE stagings


coolo:
OBS was updated to ruby3.1 and miniobs will need to follow :(

PR #2753, opened 2022-03-17
CI: adapt miniobs to use ruby 3.1

coolo:


coolo:
@vpereira gave me the tip to use rails helpers to avoid hardcoding system ruby

PR #2757, opened 2022-03-19
Disable images repo on accepting stagings

coolo:
Following up from @DimStar77's https://github.com/openSUSE/openSUSE-release-tools/pull/2382

Building and testing staging products that aren't ready yet is wasteful and confusing.


coolo:
Did some experiments and this doesn't cut it. We have to wait for the scheduler to reevaluate if the DVDs need to be rebuild. Right after enabling the repo, the old ISOs still match and it may look like it's all fine - before the scheduler finds out that it needs rebuilding. But that's easy to fix

DimStar77 (approved):

Issue #1833, opened 2019-01-15
Competing SRs for the packagelist


gyr:
Scenario:
Using gitlab-pipeline and having 2 or more staging projects falling due to Package List Generator reported a mismatch between the calculated package list and the reference.

In order to fix them SRs for 000package-groups must be submitted.

Problem:
The SRs can not be done in parallel due to OBS is not able to "merge" SRs that touch the same file (000package-groups). So, staging projects must be fixed sequentially, e.g. fix the first staging project then rebase (freeze) the second staging to get latest 000package-groups and then fix it, repeat this process for the subsequent falling staging projects.

gyr:
Possible solution:
Instead of require a SR to fix it, a bot will provide a comment with the 000package_list diff, leaving the decision to accept or not for the user. If so the diff will be applied within the pipeline (retrigger the pipeline and apply the diff on staging accepts).

nilxam:
An option can be:

Don't submit groups.yml change directly against 000package-groups but from another package.

Eg. having a `000package-groups-change` with the content like:

<pre>
sle_module_desktop_applications: includes - XXX
sle_module_desktop_applications: excludes - XXX
</pre>

then extend the pkglistgen to be can process this change and apply it to `000package-groups` in each staging if `000package-groups-change` exists. Additionally to that, `000package-group` must not be from frozenlinks, but an instantiated `000package-group` in each staging what linked to last revision in main project's one. The last step is copy `000package-groups` content to main project and decline `000package-groups-change` when doing *staging accept*. We can also *re-create* `000package-groups` package in every staging project during a *staging accept*(then pkglistgen will re-processing it per 000package-groups-change if it exists).

hmm... not sure whether this sounds over engineering though.

jberry-suse:
> The SRs can not be done in parallel due to OBS is not able to "merge" SRs that touch the same file (000package-groups).

Seems like the crux of the problem is a VCS one then. Assuming the lines are not overlapping this would be trivial with git as is done with pull requests even in this repository. We can over-engineer a workaround as the majority of the code in this repository although this one breaks a lot of assumptions.

coolo:
OBS can handle merges as well (it even has extra code to merge .changes files) - that's not the problem. The problem is that the expected output is part of 000package-groups and the expected output depends on other changes in single staging projects.

jberry-suse:
> OBS can handle merges as well (it even has extra code to merge .changes files) - that's not the problem.

[...]

PR #2761, opened 2022-03-21
Change 000package-groups on staging accept

coolo:
At least refactor it - this might need more testing before we go live with it


coolo:
That's the final step to "fix" #1833 - it obviously doesn't implement "Competing SRs to the packagelist", but it solves let's say 80% of the underlying issue.

PR #2762, opened 2022-03-21
Add osc-pcheck: support devel prj maintainers

DimStar77:


DimStar77:
For lack of a better osc-plugin repo, I think we can fit this in here.

It's used by a few devel project maintainers already to get an overview of what in a specific project has a diff that's not submitted, a submitted change, or also packages that are not links at all (potentially missing to be submitted to Factory).

Originally started in GNOME: namespace, but other maintainers are using it too and are asking for a package.

I don't think it makes sense to have it in an own repo

DimStar77:
so far this plugin lived (for a couple of years) in
https://github.com/DimStar77/openSUSE-helpers/blob/master/osc-plugins/osc-pcheck.py

but as more users seem to be using it (as I keep on mentioning it :) ) it makes sense to also at least provide it in packaged form

osc-pcheck.py
```diff
@@ -0,0 +1,123 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+from __future__ import print_function
```
coolo, line 17: the future is now

osc-pcheck.py
```diff
@@ -0,0 +1,123 @@
...
+            pmap.setdefault(key, []).append(pkg)
+            todo.setdefault(elm.get('project'), []).append(elm.get('package'))
+        md5s[pkg] = sinfo.get('verifymd5')
+    for prj, pkgs in todo.items():
+        sinfos = osc.core.get_project_sourceinfo(apiurl, prj, True, *pkgs)
+        for pkg, sinfo in sinfos.items():
+            key = '%s/%s' % (prj, pkg)
+            for p in pmap[key]:
+                vmd5 = md5s.pop(p)
+                if vmd5 != sinfo.get('verifymd5'):
```
coolo, line 66: that's a little heavy indented. Just negate the != here and continue, so you can move the following block one left. Or possibly use a function

osc-pcheck.py
```diff
@@ -0,0 +1,123 @@
...
+        query = "(state/@name='new' or state/@name='review') and " \
+                "(action/source/@project='{project}' or submit/source/@project='{project}') and " \
+                "(action/source/@package='{package}' or submit/source/@package='Packafe')".format(project=project, package=package)
+        result = osc.core.search(self.apiurl, request=query)
+        collection = result['request']
+        for root in collection.findall('request'):
+            return root.get('id')
+        return 0
+
+    def get_rev(self, project, package):
```
coolo, line 111: that function looks a little unused

osc-pcheck.py
DimStar77: That's whaI get for submitting 7 year old code... back then, the future was not yet (osc was still py2)

coolo:
One last thing: please add it to CONTENTS.md

PR #2764, opened 2022-03-22
dashboard: Add Leap 5.2 micro

ggardet:


coolo:
Can I close this? It's WIP for 3 weeks now

ggardet:
`source` and `published`  are not fetched properly. And I failed to find the reason so far. Any suggestion?

coolo:
I have no idea about Leap Micro

ggardet:
> `source` and `published` are not fetched properly. And I failed to find the reason so far. Any suggestion?
> 
> EDIT:
> 
> * for `published`, I guess this is because `OSRT:ToTestManagerStatus` does not set `published` value.
> * for `source`, it would be  `ProductVersion` which is missing.

`source` is now fixed thansk to updated attributes. `published` should be fixed with next ttm run.
So, it is ready to merge.

PR #2765, opened 2022-03-23
Remove rpmlint.log parsing feature from factory-auto

coolo:
It only works in a fraction of cases:
 - it checks only %{name}-rpmlintrc while rpmlint supports more such file names
 - it looks into the devel project's build log where we could see earlier or later builds
 - it searches for repositories that build against openSUSE:Factory while most devel projects build against openSUSE:Tumbleweed, in which case the whole check is skipped
 
And on top of that: it duplicates what rpmlint-config-Factory-strict does in stagings - block security relevant rpmlint warnings.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -16,8 +16,6 @@ from osclib.core import devel_project_get
 from osclib.core import devel_project_fallback
 from osclib.core import group_members
 from osclib.core import package_kind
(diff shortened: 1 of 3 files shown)


jsegitz (approved):
As this has not been working properly lets kill it

PR #2768, opened 2022-03-24
Fix staging accept approving package list changes

coolo:
The order is important, look at the comments *before* deleting them


coolo:
github is on strike again :(

PR #2774, opened 2022-03-28
Repo checker escalation 

### LINKED ISSUE #933: Create escalation system for build failures and repo_checker discoveries
As the last escalation step a bug should be created for packages with build failures.

coolo:
Fix #933


coolo:
@DimStar77 pointed out that the rebuild info is outdated for packages that have been deleted. So we need to get this fixed first

PR #2778, opened 2022-03-29
project-installcheck: Store when we first saw unresolvable and failed results

coolo:
@DimStar77 wants to have unresolvable packages be part of the build fail reminder, but for that we need to store the first time we saw the unresolvable.

(I guess we could do the same also for failures and work around issues in the project status route)

### DIFF
--- a/project-installcheck.py
+++ b/project-installcheck.py
@@ -223,6 +223,22 @@ class RepoChecker():
         for p in root.findall('.//status'):
             buildresult[p.get('package')] = p.get('code')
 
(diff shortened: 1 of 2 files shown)


DimStar77:
> (I guess we could do the same also for failures and work around issues in the project status route)

I'm in favor of doing the same route for all cases; mostly likely makes the code easier

coolo:
Jein - right now the failures are looked at in build-fail-reminder only

coolo:
Extended this now

PR #2779, opened 2022-03-29
Force i586 in adi projects

coolo:
This replaces #1304 with a more brute force approach to fix not only 32bit libs but all failures we see on i586. We waste so much energy retrying failed builds, it's really not worth it not trying i586 on adi.

### DIFF
--- a/osclib/stagingapi.py
+++ b/osclib/stagingapi.py
@@ -1431,12 +1431,16 @@ class StagingAPI(object):
           {repository}
             <path project="{self.cstaging}" repository="standard"/>
             <path project="{self.project}" repository="standard"/>
-            <arch>x86_64</arch>
           </repository>
           {images_repo}
           {containerfile_repo}
         </project>"""
 
+        root = ET.fromstring(meta)
+        repository = root.find('.//repository[@name="standard"]')
+        for arch in self.cstaging_archs:
+            a = ET.SubElement(repository, 'arch')
+            a.text = arch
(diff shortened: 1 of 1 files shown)


DimStar77 (approved):

nilxam:
this can be a trouble for Backports's staging since Backports only enabled a short list of package for i586...

DimStar77:
> this can be a trouble for Backports's staging since Backports only enabled a short list of package for i586...

They are enabled using 'includebuild', right? with all the others being excluded? This should inherit into the adi stagigns if setup properly

coolo:
but they are build enabled through project config

nilxam:
> > this can be a trouble for Backports's staging since Backports only enabled a short list of package for i586...
> 
> They are enabled using 'includebuild', right? with all the others being excluded? This should inherit into the adi stagigns if setup properly

limited to Backports project only `%if "%_project" == "openSUSE:Backports:SLE-15-SP4"`

coolo:
So we use all staging-archs and still check for baselibs for backports?

nilxam:
> So we use all staging-archs and still check for baselibs for backports?

1. `staging-archs` has x86_64 only for Backports. 
2. since Leap doesn't support i586, those i586 enabled package are for the special purpose if `-32bit` is required, like wine, steam, etc., they're not enabled because it has baselibs.

coolo:
I changed it now to use all configured staging archs - so no change for Leap

PR #2781, opened 2022-03-29
factory-source seems fully defunct, remove maintenance incident revie…

msmeissn:
…ws for it


coolo:
Then let's get rid of factory-source - it's already declared obsolete in CONTENTS.md anyway

coolo:
https://github.com/openSUSE/openSUSE-release-tools/pull/2783 is the full way

PR #2786, opened 2022-03-30
Fix adi projects having no archs

coolo:


DimStar77 (approved):

coolo:
hmm, the obs build doesn't really care about the updated spec file

PR #2795, opened 2022-04-05
check_bugowner: Decline new packages without bugowner entry

coolo:


coolo:
Another failure of codecov upload - tests passed

PR #2796, opened 2022-04-05
Accept_command: create set_bugowner requests

coolo:
CC @suntorytimed (untested in real life, but the test case should cover us)


suntorytimed:
I guess we can test it with the libadwaita submission we have in Staging.

coolo:
codecov upload again... I can't spot a pattern :(

PR #2798, opened 2022-04-06
project-installcheck: Take 'excluded' as gone package

### LINKED ISSUE #2797: build-fail-reminder sends mail for already fixed issues
Just got noticed by a user that he received a notification by build-fail-reminder for something that he had already fixed (and indeed looks like fixed before the mail went out):

```
> Following-up the reminder of one week ago, we have to inform you that
> 'obs-service-cargo_audit' is still showing a problem in openSUSE:Factory. See
> https://build.opensuse.org/package/show/openSUSE:Factory/obs-service-cargo_audit
> 
> Since Fri Mar 11 22:09:45 2022 we noticed the following problem:
>  Uninstallable: nothing provides cargo-audit needed by obs-service-cargo_audit-0.1.8~0.noarch
```

sr#965790 was merged on March 30th, which addressed the issue (so the user acted on the first notification). The package was changed insofar that it is no longer a noarch package, so the package which the mail talks about does not actually exist anymore

coolo:
Fixes #2797

### DIFF
--- a/project-installcheck.py
+++ b/project-installcheck.py
@@ -268,7 +268,7 @@ class RepoChecker():
             if not source.startswith('{}/{}/{}/'.format(project, repository, arch)):
                 continue
             code = buildresult.get(os.path.basename(source), 'gone')
-            if code == 'gone':
+            if code == 'gone' or code == 'excluded':
                 del oldstate['check'][source]
             if code != 'succeeded':
                 self.logger.debug(f"Skipping build result for {source} {code}")


coolo:
Testing this will take another hour or so ...

coolo:
https://build.opensuse.org/package/rdiff/home:coolo/test-dashboard?linkrev=base&rev=22 looks good to me.

DimStar77:
> https://build.opensuse.org/package/rdiff/home:coolo/test-dashboard?linkrev=base&rev=22 looks good to me.

looks good to me too - shows that the 'go-to solution for i586 build fails is `ExcludeArch`

PR #2806, opened 2022-04-22
Add support to containerfile repo to SLE15

gyr:


Vogtinator (changes requested):
`SUSE:SLE-15-SP4:GA:TEST/containerfile` got renamed to `containers`

For consistency, `sles15-image` should be released into `SUSE:SLE-15-SP4:GA:TEST/containers` as well, that needs to be done explicitly as the other images still go to the default release target.

Vogtinator (changes requested):
Looks like it's missing the part for `Release.Images.To.Publish` as well

gocd/sp4.target.gocd.yaml
```diff
@@ -83,10 +85,16 @@ pipelines:
         - script: |-
             set -e
             PRODUCTS=$(osc -A https://api.suse.de ls SUSE:SLE-15-SP4:GA | grep "^SLES15-SP4" | grep -v release)
-            for product in kiwi-templates-Minimal sles15-image 000product $PRODUCTS; do
+            CONTAINERS=$(osc -A https://api.suse.de results -b -s succeeded SUSE:SLE-15-SP4:GA -r containerfile | cut -f1 -d ' '| uniq)
```
Vogtinator, line 88: I'm a bit suspicious of the `-s succeeded`. Images without that would silently be skipped. There could be a race between the check above and now...

vasiliy-ul:
Hey, ping here. @Vogtinator , do you think this is now good to go?  Can we move it forward?

Vogtinator:
> Hey, ping here. @Vogtinator , do you think this is now good to go? Can we move it forward?

It's on my todo list, I'll have a look later

gocd/sp4.target.gocd.yaml
```diff
@@ -82,11 +84,18 @@ pipelines:
         tasks:
         - script: |-
             set -e
+            osc -A https://api.suse.de release SUSE:SLE-15-SP4:GA -r containerfile
+            osc -A https://api.suse.de release SUSE:SLE-15-SP4:GA sles15-image --target-project=containers
```
Vogtinator, line 88: I think this needs

```suggestion
            osc -A https://api.suse.de release --target-project=SUSE:SLE-15-SP4:GA --target-repository=containers -r images SUSE:SLE-15-SP4:GA sles15-image
```

gocd/sp4.target.gocd.yaml
```diff
@@ -152,8 +169,10 @@ pipelines:
             tasks:
               - script: |-
                   set -e
+                  osc -A https://api.suse.de release SUSE:SLE-15-SP4:GA -r containerfile
+                  osc -A https://api.suse.de release SUSE:SLE-15-SP4:GA sles15-image --target-project=containers
```
Vogtinator, line 173: Same as above

Vogtinator (changes requested):

gocd/sp4.target.gocd.yaml
gyr: Fixed

Vogtinator (approved):
Let's see whether it also works in practice

suntorytimed:
Looks like it fails. The to test is now releasing all packages to images. So all of standard is being released to TEST in images.

PR #2808, opened 2022-04-28
enable cdi- and kubevirt containers for publish

msmeissn:
also monitor the "containerfile" repo for readyness before doing the
publish


msmeissn:
dup of !2806

PR #2809, opened 2022-05-11
Fix packages released to containers repo

gyr:
Limited to only container packages to be released to :TEST. Previously
all packages were being released.


Vogtinator:
I thought OBS was a bit smarter than that and doesn't release excluded packages... Apparently I was a bit too optimistic.

coolo:
Remember release works on sources *and* binaries. So excluded has no meaning to half of it.

PR #2812, opened 2022-05-25
Remove all Kubic tooling

sysrich:
Kubic is being wound down, so we dont need things like the Announcer any more


coolo:
could you please fix the gocd config?

Vogtinator:
I ran gocd/generate.sh and pushed the result

Issue #2789, opened 2022-04-01
'freeze' build enables images repository


DimStar77:
As we now implemented the more complete pipelines with 'images' only being built after the /standard repo is completely done plus DVD recalculated, we build disable /images on staging accept.

staging freeze though does re-enable it again - and with freeze being a common action before selecting packages into an empty staging, in fact the pipeline is never used as intended

meta pre-freeze
```
  <build>
    <disable repository="bootstrap_copy"/>
    <disable repository="images"/>
    <disable/>
  </build>
```

meta during freeze
```
  <build>
    <enable repository="bootstrap_copy"/>
    <disable/>
  </build>
```

meta post freeze
```
  <build>
    <disable repository="bootstrap_copy"/>
    <disable/>
  </build>
```

PR #2818, opened 2022-06-17
freeze: fix crash when rebasing staging that was earlier enabled by stagingbot

DimStar77:
Fixes issue#2817

### DIFF
--- a/osclib/freeze_command.py
+++ b/osclib/freeze_command.py
@@ -37,14 +37,6 @@ class FreezeCommand(object):
 
         self.api.retried_PUT(url, ET.tostring(meta))
 
-    def is_images_disabled(self):
-        flag = self.api.get_flag_in_prj(self.prj, flag='build', repository='images')
-        if flag == 'disable':
-            return True
-        if flag == 'enable':
-            return False
-        return self.api.get_flag_in_prj(self.prj, flag='build') == 'disable'
-
     def create_bootstrap_aggregate(self):
         self.create_bootstrap_aggregate_meta()
         self.create_bootstrap_aggregate_file()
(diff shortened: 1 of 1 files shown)


DimStar77:
Regression was apparently introduced with 27c4ae0a0

coolo:
Let's discuss what we want first.
The bug happens if you freeze while the project is building and finished the pipelines before (the images disabled is gone). Leaving the images enabled on a freeze is not necessarily what we want (in #2789 you made a point that freeze is used before selecting new packages).

So what are the chances you want freeze to keep the images repo enabled? And if "always" is the answer, then I'd still favor not to set it to 'enabled' explicitly but rather keep do nothing (i.e. extend the if)

DimStar77:
> Let's discuss what we want first. The bug happens if you freeze while the project is building and finished the pipelines before (the images disabled is gone). Leaving the images enabled on a freeze is not necessarily what we want (in #2789 you made a point that freeze is used before selecting new packages).
> 
> So what are the chances you want freeze to keep the images repo enabled? And if "always" is the answer, then I'd still favor not to set it to 'enabled' explicitly but rather keep do nothing (i.e. extend the if)

I guess 'disable' on freeze is probably always correct, as the next step will be 'building' and then the pipelines re-enabling it.

the only 'error' I can see is if a freeze does not actually change (i.e. freeze, wait until enabled, freeze - but without actually having done any changes to :Factory yet); of course the freeze would be quite non-sense here, but can potentially happen (mostly on 'full rebase of all stagings after prjconf changes like needed for py3.10 switch)

If the 'freeze even if no changes were done' still manages to give us a new repoid and the pipelines passing, then disabling should be the safe bet in all cases.

coolo:
I would say we go with disabled always. If this turns out to be a problem, we can touch the 000release-package on freeze to be sure to get a new repoid.

coolo (approved):

Issue #2820, opened 2022-06-23
sourcechecker crashes on git managed packages


DimStar77:
pending SR: https://build.opensuse.org/request/show/984629

sourcechecker bot crashes with:

```
[I] target package does not exist openSUSE:Factory/obs-service-download_assets
Traceback (most recent call last):
  File "/godata/pipelines/openSUSE.Source.Check/ReviewBot.py", line 198, in check_requests
    good = self.check_one_request(req)
  File "/godata/pipelines/openSUSE.Source.Check/ReviewBot.py", line 442, in check_one_request
    ret = func(req, a)
  File "/godata/pipelines/openSUSE.Source.Check/ReviewBot.py", line 534, in check_action_submit
    return self.check_source_submission(a.src_project, a.src_package, a.src_rev, a.tgt_project, a.tgt_package)
  File "/godata/pipelines/openSUSE.Source.Check/./check_source.py", line 197, in check_source_submission
    CheckSource.checkout_package(self.apiurl, source_project, source_package, revision=source_revision,
  File "/godata/pipelines/openSUSE.Source.Check/./check_source.py", line 419, in checkout_package
    result = osc.core.checkout_package(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/osc/core.py", line 5184, in checkout_package
    raise oscerr.OscIOError(None, 'Install the obs-scm-bridge package to work on packages managed in scm (git)!')
osc.oscerr.OscIOError
[I] 984629 ignored
[script-executor] Script completed with exit code: 1.
```

The package metadata has, in the devel project:
```
<package name="obs-service-download_assets" project="openSUSE:Tools">
  <title/>
  <description/>
  <scmsync>https://github.com/openSUSE/obs-service-download_assets.git</scmsync>
</package>
```

(but this is not the first package we have which is git based - python-orangebox also does that already)

coolo:
Newly deployed - no longer crashing! :) 
https://build.opensuse.org/request/show/984629

PR #2821, opened 2022-06-23
package: require git-scm-bridge by osrt-check-source

DimStar77:
The bridge is required for git-managed sources being submitted
to Factory in order to check out the code

Fixes issue #2820


coolo:
Note though that our infrastructure isn't using these packages. So it's for your own entertainment only

DimStar77:
> Note though that our infrastructure isn't using these packages. So it's for your own entertainment only

right - so the fix in the package is 'the right thing to do' - but nobody will realize it.

DimStar77:
Should be better now - also added obs-scm-bridge to the dockerfile

coolo:
Note however that the dockerfile you edited is just used to run the testsuite, not in our infrastructure.

The gocd agent runs in https://build.opensuse.org/package/view_file/openSUSE:Tools:Images/gocd-agent-release-tools/Dockerfile?expand=1

DimStar77:
EVIL!

PR #2831, opened 2022-08-04
Fix cache handling with osc from git master

coolo:
osc master no longer wraps urlopen, so use the underlying function


coolo:
@dmach do I have another option?

PR #2832, opened 2022-08-04
Move publish_distro 817e024 into openSUSE-release-tools

lkocman:
Suggested in https://github.com/openSUSE/publish_distro/issues/13 to separate tool and configuration.

### DIFF
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -187,6 +187,19 @@ BuildArch:      noarch
 %description origin-manager
 Tools for managing the origin of package sources and keeping them in sync.
 
+%package publish-distro
+Summary:        Tool for publishing ftp-stage to ftp-prod
+Group:          Development/Tools/Other
+Requires:       rsync
+# Currently in openSUSE infra repo
+Requires:       python-mb
+Requires(pre):  shadow
+BuildArch:      noarch
+
+%description publish-distro
+publish_distro tool and related configuration in publish_distro to rsync
+content from OBS to ftp-stage/ftp-prod on pontifex host.
+
 %package repo-checker
 Summary:        Repository checker service
 Group:          Development/Tools/Other
@@ -474,6 +487,9 @@ exit 0
 %files origin-manager
 %{_bindir}/osrt-origin-manager
 %{_datadir}/%{source_dir}/origin-manager.py
+
+%files publish-distro
+%{_bindir}/publish-distro
 
 %files repo-checker
 %{_bindir}/osrt-project-installcheck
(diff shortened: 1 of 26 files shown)


lkocman:
https://github.com/openSUSE/publish_distro/pull/21 (removal)

ancorgs:
First of all, please add the description of this new tool at https://github.com/openSUSE/openSUSE-release-tools/blob/master/CONTENTS.md as part of this pull request.

The main problem with this repository is that it has been historical just a bunch of poorly documented scripts (which doesn't even describe their function when executed without arguments), thrown into the repository by different people via pull requests with one-liner descriptions.

The file CONTENTS.md is at least a poor attempt to fix that by providing some context so we can know what a script is used for (and by who), where is it deployed and whether we can remove it from the repository at a given point of time.

Talking about deployment. Is this expected to become part of https://build.opensuse.org/package/show/openSUSE:Tools/openSUSE-release-tools ?

coolo:
And shouldn't all those configs move too from that repo? If so, better move it to a subdirectory, add a README there explaining (and point to that from CONTENT.md as @ancorgs mentions) and if you plan to deploy it on pontifex, add it to the package as well.

DimStar77:
> @DimStar77 would such proposal be okay with you?

If we want to consolidate this into a single repo, then it does make sense to move it all together

Question is if we win anything over having it in a sep repo

lkocman:
The original bonus was that we consume tool in a more controlled way.

coolo:
I personally don't care either way - you can also package the other repo. But if you want it to be in this repo, it needs to follow the rules

dirkmueller:
@lkocman the underlying issue should be fixed now, can you please rebase?

coolo:
This still lacks packaging - I thought that was the plan to begin with

dist/package/openSUSE-release-tools.spec
```diff
@@ -187,6 +187,19 @@ BuildArch:      noarch
 %description origin-manager
 Tools for managing the origin of package sources and keeping them in sync.
 
+%package publish-distro
+Summary:        Tool for publishing ftp-stage to ftp-prod
+Group:          Development/Tools/Other
+Requires:       rsync
+# Currently in openSUSE infra repo
+Requires:       python-mb
```
coolo, line 195: python-? that sounds like python2. Is that true?

coolo: All of the mb calls seem to be commented

coolo (approved):
Good enough for me. I didn't look at the script itself - I did that mistake some years ago... :)

lkocman:
Hehe thank you @coolo :-) Merging

PR #2833, opened 2022-08-04
Dont use garbled patches for test fixture

dirkmueller:
obs-service-source_validator recently gained the feature to reject
empty or garbled patches, and hence this was failing. Use a validly
formatted patch instead.


lkocman:
Ah this is why the test in my PR was failing.

coolo (approved):

PR #2834, opened 2022-08-05
gocd: LEO is replaced with ALP

fcrozat:


gocd/generate.sh
```diff
@@ -13,7 +13,7 @@ done
 
 grep group: *.yaml | cut -d: -f3 | sort -u | while read group; do
    case $group in
-	   BCI|Factory|Leap|Admin|LEO|MicroOS|Monitors|openSUSE.Checkers|SLE15.Stagings|SLE15.Target|SLE.Checkers)
+	   BCI|Factory|Leap|Admin|ALP|MicroOS|Monitors|openSUSE.Checkers|SLE15.Stagings|SLE15.Target|SLE.Checkers)
```
coolo, line 16: Possibly you leave it in group LEO - if you rename the group you will loose access to it and ALP is still rather code name.

coolo (approved):

gocd/generate.sh
fcrozat: fixed

PR #2835, opened 2022-08-08
StagingBot: call staging rebuild first (support random fails) and handle Factory:NonFree

DimStar77:


coolo:
What changed between https://github.com/openSUSE/openSUSE-release-tools/pull/2552 and now? :)

DimStar77:
> What changed between #2552 and now? :)

oh, we already had `rebuild` as action and removed that

Feels a bit like only
`                osc -A https://api.opensuse.org staging -p openSUSE:<%= project %> select --non-interactive --merge --try-strategies`

should have been removed IMHO; 

> For backports, there are no rebuild and select calls, but for factory there are. For now I just removed them from factory as well, because the bot is disabled there anyway.

That was true back then - but my goal was to get back on the bot running on botmaster instead of running on my own box

coolo (approved):

PR #2837, opened 2022-08-09
Setup for SLE15 SP4 Quarterly Update

gyr:


gocd/checkers.suse.gocd.yaml
```diff
@@ -348,6 +348,34 @@ pipelines:
...
+        approval:
+          type: manual
+        jobs:
+          Run:
+            resources:
+            - repo-checker
+            tasks:
+            - script: |-
+                cd scripts
+                ./pkglistgen.py -A https://api.suse.de --debug update_and_solve -p home:maintenance-installcheck:15-SP4:QR -s target --force
```
coolo, line 373: I don't think these projects exist atm

PR #2843, opened 2022-08-22
Set timeout on the suse checkers

dirkmueller:
This prevents the short-living checkers from not making any progress
due to an endless hanging job


coolo:
Please pick a larger timeout for the legal import pipelines as those can run hours if we add a new code base.

coolo:
Force pushed to the branch as Dirk is afk

PR #2844, opened 2022-08-23
install and link publish_dist

dirkmueller:
Otherwise all other PRs fail with a build error on the spec.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ include Makefile.common
 
 pkgdata_BINS = $(shell find * -maxdepth 0 -executable -type f)
 pkgdata_SCRIPTS=$(wildcard *.py *.pl *.sh)
-pkgdata_SCRIPTS+=bs_mirrorfull findfileconflicts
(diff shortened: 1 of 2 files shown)


Vogtinator:
Fails:

```
[   26s] ERROR: link target doesn't exist (neither in build root nor in installed system):
[   26s]   /usr/bin/osrt-publish_distro -> /usr/share/openSUSE-release-tools/publish_distro
[   26s] Add the package providing the target to BuildRequires and Requires
```

lkocman (approved):
Fine by me

dirkmueller:
> Fails:

The CI is not testing what the PR does. thats how the previous PR could go in and then broke everything else. 

I don't know where the CI code is, but you can see that the _service file is checking out the spec file from master not from the PR brnach

Vogtinator:
> > Fails:
> 
> The CI is not testing what the PR does. thats how the previous PR could go in and then broke everything else.

From a quick look I assumed it would test this PR because it already calls it `/usr/bin/osrt-publish_distro`, which is only now in `%files`

PR #2846, opened 2022-08-29
Add test for group in bugowner line

coolo:


coolo:
No idea what was wrong with the original request

Issue #2850, opened 2022-08-31
factory dashboard no longer shows 'published' version


DimStar77:
For 'published', we only see n.a

![image](https://user-images.githubusercontent.com/532559/187703139-e2fedc33-84ef-458b-b5aa-9e4a6fc7efef.png)

DimStar77:
That info should be coming from the attribute `OSRT:ToTestManagerStatus`, but in openSUSE:Factory that currently is:

```
failed: '20220830'
testing: '20220830'
```

There is no published entry there.

coolo:
It disappeared very recently though: https://github.com/bmwiedemann/openSUSE/commit/d62b4273863d78725768bf8f62a00cd2ed699286#diff-5b2127f002175b7eb280966ee0b9c83296b60049560aeb3c13ca3f616a33928aL489

DimStar77:
That seems almost like a race between ttm and a check-in round (?)

coolo:
Hmm, looking at log -M openSUSE:Factory, totest-manager commits every 15 minutes

coolo:
Note that @bmwiedemann only updates the git twice a day (at 12:50 and 18:50 we see commits) - but there are ~100 commits a day

coolo:
73005 killed it:
```
r73005 | totest-manager | 2022-08-31 13:30:17 | 02d70af6c063685503759b2d26b07d55 | None | 

<no message>
----------------------------------------------------------------------------
r73004 | totest-manager | 2022-08-31 13:15:27 | a7f09ecf2ef99536c17b4f4e26fe6fe7 | None | 

<no message>
```

```
@@ -485,10 +485,7 @@
 set_snapshot_number: True</value>
   </attribute>
   <attribute namespace="OSRT" name="ToTestManagerStatus">
-    <value>failed: '20220830'
-published: '20220829'
-publishing: '20220829'
-testing: '20220830'
+    <value>testing: '20220830'
 </value>
   </attribute>```

**Comment by coolo** [comment]
So this should be this TTM run:

```
[script-executor] -------------------------
install -D /home/go/config/openqa-client.conf /home/go/.config/openqa/client.conf
scripts/totest-manager.py -A https://api.opensuse.org --debug run openSUSE:Factory
[script-executor] -------------------------
[script-executor] Script Type: bash
[script-executor] Script written into '/go/pipelines/TTM.Factory/fd81a302-5373-48af-9b41-06004d229cd5.sh'.
[...]

PR #2851, opened 2022-09-01
TTM: Add log message for changing status

coolo:


coolo:
I can't do anything for the actual issue #2850 but at least totest-manager won't spam the history any more and we get debug output what actually changed in the botmaster history next time.

PR #2853, opened 2022-09-14
Remove publish_distro dependency on python-mb

lkocman:
* non of the mirrorbrain is used any more
* sections using mb are already disabled


lkocman:
Thank you!

PR #2857, opened 2022-09-19
Handle more revisions to ignore (handling icewm)

coolo:


aplanas (approved):

coolo:
Before: 560 failed imports
After: 175 failed imports

Progress! :)

PR #2863, opened 2022-09-20
Ignore delete requests (that got restored)

coolo:
Test case: cpupower


aplanas (approved):
We should collect that the conditions that set `ignore` in is self processing step.

coolo:
Down to 26 failed imports

PR #2869, opened 2022-09-26
Fix import determinism

coolo:
Don't treat any random network glitch as broken revision (which leads to ignoring it, changing the whole tree). The VM the importer runs on has some network outage bursts unfortunately, so we need to be defensive

### DIFF
--- a/git-importer.py
+++ b/git-importer.py
@@ -709,30 +709,34 @@ class Revision:
         return f"[{self.__str__()}]"
 
     def check_link(self):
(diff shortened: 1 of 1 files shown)


coolo:
@aplanas And that's the reason I use Exception very exceptionally :)

coolo:
`urllib.error.URLError: <urlopen error [Errno 99] Cannot assign requested address> ` was the interim error

aplanas (approved):

PR #2870, opened 2022-09-26
Make OBS retry more aggressive

coolo:


coolo:
@aplanas if you have a good idea how to avoid the c&p, let me know

aplanas (approved):

PR #2871, opened 2022-09-27
Allow using SR to instantiate packages

gyr:
Due to SLSA constrains, directly instantiate packages is restricted. Check for "package-instantiation" on description to allow request without changes to be accepted and so to be used to instantiate package on target project


coolo:
Please don't. Just detect the situation yourself and accept it. If it's 100% the same sources, the situation is clear - because OBS wouldn't have allowed to create the SR to begin with if it wasn't for new code streams.

coolo:
As you closed the PR: When I said "please don't", I meant do not rely on the requestor's description.

PR #2873, opened 2022-10-05
Re-download the accepted revisions to be on the safe side

coolo:
Test case: texlive-specs-* - they point to a different package that we don't import, so the merges are random


aplanas (approved):

coolo:
https://gitea.opensuse.org/rpm/texlive-specs-p/commit/e08cb8e0ba6bcddea356385ab1f019e3cbffde22 is an example of these random sync up commits.

PR #2874, opened 2022-10-10
WIP: Save staging summary in attributes

SchoolGuy:
This PR has the aim to achieve two things:

- switch saving the staging summary inside an OSRT namespace attribute that is newly created for this purpose
- submit the changelog to a devel project of the `000package-groups` package

The reasoning behind this effort is that with [SLSA](https://slsa.dev/) the release manager has no write access to the GA image anymore. During a `osc staging accept` the changes should be saved though. To preserve the functionality that was lost with complying to SLSA and to reduce pressure on autobuild for reviewing these changes more often then needed this PR was created.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 .docker-tmp/
 .bash_history
 osc
+.venv
+.vscode
+.idea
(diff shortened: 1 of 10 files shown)


SchoolGuy:
It was agreed that we try something simpler for now. I will close this and if needed reopen it.

PR #2877, opened 2022-10-13
Add typing hints

SchoolGuy:
This PR does two things:

- Add typing hints and docstrings to some of the functionality of this repository
- Fix typos in the documentation

PR #2879, opened 2022-10-13
Fixes for osc-1.x

dmach:
These fixes will be needed when we distribute osc-1.x instead of the current 0.x.
Also, it might be good to test integration with openSUSE/osc's master more frequently and report any issues found.

### DIFF
--- a/tests/OBSLocal.py
+++ b/tests/OBSLocal.py
@@ -71,14 +71,18 @@ class TestCase(unittest.TestCase):
         with open(OSCRC, 'w+') as f:
             f.write('\n'.join([
                 '[general]',
-                'apiurl = http://api:3000',
+                'apiurl = {}'.format(APIURL),
                 'http_debug = false',
                 'debug = false',
                 'cookiejar = {}'.format(OSCCOOKIEJAR),
-                '[http://api:3000]',
+                '[{}]'.format(APIURL),
                 'user = {}'.format(userid),
                 'pass = opensuse',
                 'email = {}@example.com'.format(userid),
+                # allow plain http even if it is insecure; we're testing after all
+                'allow_http = 1',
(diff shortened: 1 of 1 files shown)


dmach:
> Our testing is supposed to happen on openSUSE:Tools

That if usually too late.
We make a new osc release, submit a request to openSUSE:Tools when the request gets accepted, it is usually forwarded to Factory. Testing on OBS:Server:Unstable (which contains osc built from git master) would help us to catch issues earlier.

coolo:
But that's clearly out of scope of this repo - and as you can see I even fail to monitor one container building, leave alone two.

coolo:
The failing comment_tests leads me to believe that your change in oscParse is not working with the osc we have.

coolo:
confirmed. if I readd the line, it works again in local tests

dirkmueller:
@dmach are you planning to work on this? Many osc 1.x.fixes have been merged already. Not sure if there's still something that needs to be done?

dmach:
> @dmach are you planning to work on this? Many osc 1.x.fixes have been merged already. Not sure if there's still something that needs to be done?

@dirkmueller I'm not sure, to be honest. Probably not.
If the current code works with osc 1.x, then this pull request may not be needed any more.

@SchoolGuy you've pushed many commits to this PR - how about you, do you want them in or is it something that got to the code in different pull requests already?

DimStar77:
There is at least https://github.com/openSUSE/openSUSE-release-tools/issues/2928 reported to fail with osc 1.x

SchoolGuy:
The commits that are part of this PR are cherry picked. For example https://github.com/openSUSE/openSUSE-release-tools/pull/2877 is part of this PR. I believe that after a rebase all of my commits should be gone. If not I would be confused and would need to look deeper @dmach ...

PR #2885, opened 2022-11-11
check_bugowner: Don't get tripped over trailing space

coolo:


coolo:
OBS is down. We have to wait for it to be back up to see CI results :(

PR #2887, opened 2022-11-16
gocd: allow to publish SLE BCI repository using a token

fcrozat:


gocd/bci.gocd.yaml
```diff
@@ -189,23 +191,25 @@ pipelines:
         - SLE
         environment_variables:
           OSC_CONFIG: /home/go/config/oscrc-totest-manager
+          BCI_TOKEN: '{{SECRET:[opensuse.secrets][BCI_TOKEN]}}'
```
coolo, line 194: IBS_BCI_RELEASE_TOKEN would be a little more obvious - but let's stay with that for now.

gocd/bci.gocd.yaml
```diff
@@ -189,23 +191,25 @@ pipelines:
...
         resources:
         - staging-bot
         tasks:
         - script: |-
+            SPRJ=SUSE:SLE-15-SP4:Update:BCI
```
coolo, line 199: The file is autogenerated - to avoid exactly what happened here: that you forget about SP5.

gocd/bci.gocd.yaml
```diff
@@ -106,23 +106,25 @@ pipelines:
...
         - script: |-
+            SPRJ=SUSE:SLE-15-SP3:Update:BCI
             for arch in aarch64 ppc64le s390x x86_64 ; do
-             SPRJ=SUSE:SLE-15-SP3:Update:BCI
-             PKG="000product:SLE_BCI-ftp-POOL-$arch"
-             PRJ="SUSE:Products:SLE-BCI:15-SP3:$arch"
-             osc -A https://api.suse.de api -X POST "/source/$SPRJ/$PKG?cmd=release&target_project=$PRJ&repository=images&target_repository=images"
+              PKG="000product:SLE_BCI-ftp-POOL-$arch"
+              PRJ="SUSE:Products:SLE-BCI:15-SP3:$arch"
+              curl -X POST -H "Authorization: Token $BCI_TOKEN" "https://api.suse.de/trigger/release?project=$SPRJ\&package=000product:SLE_BCI-ftp-POOL-$arch\&targetproject=$PRJ\&targetrepository=images\&filter_source_repository=\"images\""
```
coolo, line 118: Why not use $PKG here?

gocd/bci.gocd.yaml
```diff
@@ -106,23 +106,25 @@ pipelines:
...
+              #              osc -A https://api.suse.de/ token --trigger $BCI_TOKEN --operation release $SPRJ $PKG
```
coolo, line 119: please reference https://github.com/openSUSE/osc/issues/1194 and drop the commented code.

coolo (changes requested):

gocd/bci.gocd.yaml
fcrozat: ok, thanks

gocd/bci.gocd.yaml
fcrozat: each SP has its own release pipeline, SP5 is handled in lines 281+ (old version of the script, I'll update it)

coolo: The file is still autogenerated

fcrozat: sorry, it took me a while to understand: the entire yaml file is generated from erb file !! Will adapt my PR

coolo: That's why we have a CI in this repo - so you get email about failures :)

coolo (approved):

PR #2890, opened 2022-11-17
Update version of the config plugin

coolo:


coolo:
This is a little mean as we can't easily validate that the script executor will like the config within yaml

PR #2894, opened 2022-11-23
gocd/rabbit-openqa.py: Always include flavor, name and machine in the check

Vogtinator:
When including images in staging checks we get conflicting names as we have e.g. `microos@uefi-staging` for both `Staging-DVD` and `MicroOS-Image-ContainerHost`.

### DIFF
--- a/gocd/rabbit-openqa.py
+++ b/gocd/rabbit-openqa.py
@@ -79,11 +79,10 @@ class Project(object):
         # collect job infos to pick names
         for job in openqa:
             print(staging, iso, job['id'], job['state'], job['result'],
-                  job['settings']['MACHINE'], job['settings']['TEST'])
+                  job['settings']['FLAVOR'], job['settings']['TEST'], job['settings']['MACHINE'])
             openqa_infos[job['id']] = {'url': self.listener.test_url(job)}
             openqa_infos[job['id']]['state'] = self.map_openqa_result(job)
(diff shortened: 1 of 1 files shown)


Vogtinator:
The required checks in existing staging projects hardcode the simplified check names, so this would break them. Before merging this that needs to be adjusted.

For now I'll create a separate PR which keeps the simplified name in case there is no conflict.

coolo:
Not a fan. "openqa:MicroOS-Image-ContainerHost-microos@uefi-staging" is quite a beast of a check name. But as I don't have to look at the staging dashboard all day, I'll leave that to @DimStar77, @suntorytimed, @gyr and @sysrich and peers I forgot.

Vogtinator:
> Not a fan. "openqa:MicroOS-Image-ContainerHost-microos@uefi-staging" is quite a beast of a check name. But as I don't have to look at the staging dashboard all day, I'll leave that to @DimStar77, @suntorytimed, @gyr and @sysrich and peers I forgot.

FTR, for SLE the longest name would be `Online-Y-Staging-rescue_system_sle11sp4@64bit` - not that bad.

coolo:
plus the openqa: prefix. Simulating in Factory dashbord it would give 
image

I'm just asking questions!

coolo (approved):

Vogtinator:
@suntorytimed, @gyr

Your input is still missing. If there are no objections, I'd like to merge this next week.

PR #2899, opened 2022-11-24
gocd/rabbit-openqa.py: Improve/fix check name generation

Vogtinator:
The previous algorithm for conflict resolution was not flexible enough (for TW we need FLAVOR in addition) and had bugs in edge cases:
* taken_names was not updated for solved conflicts
* Depending on the order of jobs the result could differ

This implements a different way which also includes FLAVOR and has stable output.

### DIFF
--- a/gocd/rabbit-openqa.py
+++ b/gocd/rabbit-openqa.py
@@ -6,6 +6,7 @@ import logging
 import json
 import osc
 import re
+from collections import defaultdict
 from osc.core import http_GET, http_POST, makeurl
 from osclib.conf import Config
 from osclib.stagingapi import StagingAPI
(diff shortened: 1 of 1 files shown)


coolo:
Looks like the miniOBS container are broken :(

gocd/rabbit-openqa.py
```diff
@@ -100,12 +101,38 @@ def fetch_openqa_jobs(self, staging, iso, openqa_infos):
...
+                f"{openqa_info['flavor']}-{openqa_info['name']}@{openqa_info['machine']}"]
+
+    def make_unique_check_names(openqa_infos):
+        # Make a map of "possible name" -> [jobs]
+        map_name_jobs = defaultdict(list)
+        for id in openqa_infos:
+            for name in Project.check_name_combinations(openqa_infos[id]):
+                map_name_jobs[name].append(id)
+
+        # For each job, pick the shortest unique name
```
coolo, line 123: when you're talking about stable output, consider the situation that your event comes in after the first flavor was scheduled and then another event comes in after the second flavor. Will you still name the jobs the same?

Vogtinator: I also thought about that and it's basically what I'd like to fix with #2894.

coolo: Just to document what I discussed with Fabian in RL: I would suggest to have those names we know conflicting in its full spelling as required check. And if there is a required check matching a candidate name, pick it.

Vogtinator: To document my RL reply: I don't like that approach either, because it's not obvious why it's necessary, it's easy to forget to add a check on (seemingly simple) changes on the openQA side, and when (not if!) something goes wrong the failure mode is really unintuitive.

Vogtinator:
#2894 was merged instead

PR #2901, opened 2022-11-28
gocd: SUSE:ALP:ARM doesn't exist

fcrozat:


coolo:
I accepted the previous PR, but this one is just over the top. So I give you this: https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53

PR #2902, opened 2022-11-29
Adjust 15.5 appliance patterns for publish_distro

lkocman:
Old JeOS patterns prior to raname did not match the current image names and blocked publishing.
These new patterns seem to work on pontifex.


Vogtinator (changes requested):
All of those are built though.

lkocman:
Should be all set whitespaces were corrected as well.

Vogtinator (approved):

PR #2903, opened 2022-11-29
publish_distro add  --now to set PUBLISH_DELAY=0

lkocman:
* we have to set it to 0 avery single time when we're doing some debuging on pontifex and such approach leaves dirty checkout
* let's add users option to override it

### DIFF
--- a/publish_distro
+++ b/publish_distro
@@ -70,6 +70,10 @@ for arg do
         do_cleanup=1
         date="90 days ago"
         continue
+  elif [ "$arg" == '--now' ]; then
+	PUBLISH_DELAY=0
+        shift
+        continue
   else
         newargs+=("$arg")
   fi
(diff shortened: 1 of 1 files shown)


publish_distro
```diff
@@ -70,6 +70,10 @@ for arg do
         do_cleanup=1
         date="90 days ago"
         continue
+  elif [ "$arg" == '--now' ]; then
+	PUBLISH_DELAY=0
```
coolo, line 74: I get the feeling the indentation is broken - or you used tabs, which is just as bad

lkocman: Good catch! I fixed this particular line, otherwise I'll do a separate PR with retab of the rest of the file. Seems like it was mostly using tabs.

shundhammer (approved):
LGTM

PR #2906, opened 2022-12-12
Retab publish distro

lkocman:
* based on https://github.com/openSUSE/openSUSE-release-tools/pull/2903#discussion_r1035751458
* depends on https://github.com/openSUSE/openSUSE-release-tools/pull/2903
* retab with two spaces insetad of tabs
* supply vim modeline

### DIFF
--- a/publish_distro
+++ b/publish_distro
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+# vim:sw=4:ts=4:et:
 #
 # The script is executed periodically by cronjob
 # as part of ~mirror/bin/publish_factory_leap wrapper on pontifex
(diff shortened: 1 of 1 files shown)


mvidner (changes requested):
I'd like to point out that https://github.com/openSUSE/openSUSE-release-tools/blob/master/.editorconfig exists and says that `*` files (shell) use 2 spaces
Your choice of 4 spaces is inconsistent :)

lkocman:
> I'd like to point out that https://github.com/openSUSE/openSUSE-release-tools/blob/master/.editorconfig exists and says that `*` files (shell) use 2 spaces Your choice of 4 spaces is inconsistent :)

Force pushed with two tabs instead

publish_distro
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+# vim:sw=4:ts=4:et:
```
dirkmueller, line 3: That now needs updating or removal

dirkmueller (changes requested):
See inline comment

coolo:
Superseded by #2910 (fixing the editor config conformance for all)

PR #2910, opened 2023-01-02
Validate that the files match the editor config

coolo:
I only learned about its existance due to pr 2906, so let's enforce it


coolo:
Ignoring all of python here as we run other linters and ec complains about "Wrong amount of left-padding spaces" if arguments are split and indented based on the brackets.

PR #2916, opened 2023-01-23
Allow specifying a devel project that has "Factory" in its name

dirkmueller:
we should only expand with openSUSE in front if it is starting with "Factory". Also add ALP handling and hide the warning if a full project has been specified.

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -53,14 +53,15 @@ def _full_project_name(self, project):
     if project.startswith(('openSUSE', 'SUSE')):
         return project
 
-    if 'Factory' in project or 'openSUSE' in project:
+    if project.startswith('Factory'):
         return 'openSUSE:%s' % project
 
-    if 'SLE' in project:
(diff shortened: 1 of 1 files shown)


osc-staging.py
```diff
@@ -53,14 +53,15 @@ def _full_project_name(self, project):
     if project.startswith(('openSUSE', 'SUSE')):
         return project
```
coolo, line 54: So you used ALP? Because if had SUSE:ALP, it would have ended here, no?

dirkmueller: no, in my concrete usecase I used `devel:Factory:git-workflow` which has been rewritten into `openSUSE:devel:Factory:git-workflow` which does not exist. however the fix for `-p ALP` would now rewrite this into -p `SUSE:ALP`

coolo: I see. I suppose most users use a fully qualified project name - I was actually suprised seeing this code :)

coolo (approved):

PR #2918, opened 2023-01-24
Fixes for osc-1.x

dirkmueller:
Starting with the minimal changes from #2879 again

### DIFF
--- a/check_tags_in_requests.py
+++ b/check_tags_in_requests.py
@@ -4,6 +4,8 @@ import sys
 
 import osc.conf
 import osc.core
+
+from osclib.core import get_request_list_with_history
 from lxml import etree as ET
 from urllib.error import HTTPError, URLError
 
@@ -72,7 +74,7 @@ class FactorySourceChecker(ReviewBot.ReviewBot):
                         project = project[len(prefix):]
                         sr = srprefix
                         break
-            requests = osc.core.get_request_list(apiurl, project, package, None, ['new', 'review'], 'submit')
+            requests = get_request_list_with_history(apiurl, project, package, None, ['new', 'review'], 'submit')
         except (HTTPError, URLError):
             self.logger.error("caught exception while checking %s/%s", project, package)
             return None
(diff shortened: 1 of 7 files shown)


check_tags_in_requests.py
```diff
@@ -72,7 +74,7 @@ def _check_requests(self, project, package, rev):
                         project = project[len(prefix):]
                         sr = srprefix
                         break
-            requests = osc.core.get_request_list(apiurl, project, package, None, ['new', 'review'], 'submit')
+            requests = get_request_list_with_history(apiurl, project, package, None, ['new', 'review'], 'submit')
```
Vogtinator, line 77: This didn't use `withfullhistory`, so it should behave the same with the compatibility wrapper. Were you able to find the cause of the difference?

osclib/core.py
```diff
@@ -56,10 +56,52 @@ def groups_members(apiurl, groups):
 
     return members
 
-# osc uses xml.etree while we rely on lxml
+
+def get_request_list_with_history(
```
Vogtinator, line 60: Can you add a deprecation warning or at least a `# TODO` comment?

check_tags_in_requests.py
dirkmueller: Yes, the return values are different. I can search the details tomorrow. for now I'd like to get this in to have a green CI to iterate on improving the quality on this.

osclib/core.py
dirkmueller: there's one on the line below (63). I can make it a deprecation warning when everything else is green.

Vogtinator: AFAIK a warning wouldn't make it less green than it was previously, wouldn't it?

dirkmueller: It does not. just want to get to see that everything works before adding further code changes (that would restart the CI). do want to leave from work today at some point soon ;)

Vogtinator: Ah, so it's just about the push, I see. In that case we'll need a separate PR anyway.

> do want to leave from work today at some point soon ;)

Protip: You can do that while the CI is running - /me writes this reply from home :stuck_out_tongue:

PR #2920, opened 2023-01-24
Fix regex for access logs parsing

witekest:
The format of access log messages has changed on 2022-12-23.

Fixes https://progress.opensuse.org/issues/123133

### DIFF
--- a/metrics/access/ingest.php
+++ b/metrics/access/ingest.php
@@ -1,7 +1,7 @@
 #!/usr/bin/php
 <?php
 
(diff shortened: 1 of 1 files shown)


dirkmueller (approved):
Assuming it is tested, +1. 

is it necessary to make this regex matching both for old and new data?

witekest:
The new regex has been tested with the logs of the last 12 days.

Once aggregated data is stored in cache files, so actually after completing this step we don't need logs anymore and backward compatibility is not needed.
On the other hand if we decide to change the aggregation rules we'd have to rerun on all available logs, so here it would be useful.

I've reviewed and modified the regex for backwards compatibility.

dirkmueller (approved):

PR #2922, opened 2023-01-25
Use the 15.4 container

dirkmueller:


coolo:
merge with the osc fixes, so we have one good

PR #2925, opened 2023-01-30
Fix gocd config ident - listing a stage as a job

coolo:


coolo:
I force pushed yet more fixes. It's a bit unfortunate that the plugin has no further way to lint the config :(

PR #2929, opened 2023-02-02
skippkg-finder: add conditionals support for the product ftp-tree

nilxam:
In some cases we would not like to add package to ftp-tree on an specific arch, this can be done by multiple 'group' to be defined in NON_FTP_PACKAGES.group, however NON_FTP_PACKAGES.group is auto-generated by skippkg-finder according the buildresults and the project attributes, this commit will add conditionals support to skippkg-finder, in other words, to be able to define arch dependent package in NON_FTP_PACKAGES.group via the project attribute.

To enable it, 'use group' is also needed to be existsed in the product file, the available condition/platform can be found in meta pacakge.

### DIFF
--- a/skippkg-finder.py
+++ b/skippkg-finder.py
@@ -44,6 +44,8 @@ class SkippkgFinder(object):
         self.skiplist_supplement_regex = set(config.get('skippkg-finder-skiplist-supplement-regex', '').split(' '))
         # drops off RPM from a list of the supplement RPMs due to regex
         self.skiplist_supplement_ignores = set(config.get('skippkg-finder-skiplist-supplement-ignores', '').split(' '))
+        # conditional support scenario
+        self.skiplist_conditionals = set(config.get('skippkg-finder-conditional-scenarios', '').split(' '))
 
     def is_sle_specific(self, package):
         """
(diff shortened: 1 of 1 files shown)


skippkg-finder.py
```diff
@@ -325,21 +341,26 @@ def crawl(self):
...
+        for item in self.skiplist_conditionals:
+            # node[0] is the condition, node[1] is the package
+            # an example of the format: only_x86_64:glibc-32bit
+            node = item.split(':')
+            if node[0] not in cond_list:
+                cond_list[node[0]] = []
+            cond_list[node[0]].append(node[1])
+        for cond in cond_list.keys():
+            group = self.create_group('NON_FTP_PACKAGES_' + cond, cond, cond_list[cond])
+            skip_list += group
```
Vogtinator, line 357: You're appending multiple XML documents which is technically not valid. Can you do the appending on the ET level?

Vogtinator: FWICT you end up with a document like

```
<group>
   stuff...
</group>
<group>
   more stuff...
</group>
```

That is not a valid XML document and can't be used with `<xi:include/>`. If it works, I suppose OBS ignores the spec or implements it wrongly.

nilxam: yes, it's not a valid xml format, but it works with product-builder, multiple groups in a file is allowed, like `dvd.group` hjas multiple group tree as `dvd.x86_64`, `dvd.s390x`, etc,.

Vogtinator: Sigh. Even with `<?xml ...` in between?

nilxam: > Can you do the appending on the ET level?

Can't do it in the one file(NON_FTP_PACKAGES.group), a `group` tree can only have an `conditional` element, in case want to make every group file to be an standard xml format, then we need multiple NON_FTP_PACKAGES_$contidions.group file for each `group`.

Vogtinator: *shrug* this is all chaotic anyway

PR #2932, opened 2023-02-09
Create new pipeline to deploy sle-prjmgr-tools scripts

gyr:


Vogtinator:
Can you elaborate on the change? FWICT you added two new pipelines but their primary goal isn't deployment of scripts

suntorytimed:
The commit is splitting up the pipelines, as right now there are to release to test jobs in the pipeline, one which is still following the old API call way and another one using sle-prjmgr-tools. To use and test the sle-prjmgr-tools we require a separate pipeline though, which is created with this commit.

dirkmueller (approved):

PR #2935, opened 2023-03-02
Fix handling of preinstall images

dirkmueller:
These have special name handling in `_preinstallimage`. The package name has to be "preinstallimage-$name" where $name is what is given in the _preinstallimage description

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -221,10 +221,13 @@ class CheckSource(ReviewBot.ReviewBot):
 
         new_info = self.package_source_parse(source_project, source_package, source_revision, target_package)
         filename = new_info.get('filename', '')
-        if not (filename.endswith('.kiwi') or filename == 'Dockerfile') and new_info['name'] != target_package:
+        expected_name = target_package
+        if filename == '_preinstallimage':
+            expected_name = 'preinstallimage'
+        if not (filename.endswith('.kiwi') or filename == 'Dockerfile') and new_info['name'] != expected_name:
             shutil.rmtree(dir)
             self.review_messages['declined'] = "A package submitted as %s has to build as 'Name: %s' - found Name '%s'" % (
-                target_package, target_package, new_info['name'])
+                target_package, expected_name, new_info['name'])
             return False
 
         if not self.check_service_file(target_package):


Vogtinator:
Looks like `preinstallimage-minimal` already violates that - it has `Name: base`! I wonder whether the name is actually used anywhere, it doesn't appear to be in the preinstallimage info.

Vogtinator:
Looks like this is broken in more cases. If the `Name` does not really matter, can we just ignore it?

dirkmueller:
@Vogtinator all existing preinstallimages were checked in outside the bot, so the bot was never able to do any checking on it. this PR is improving over that situation. 

what we can do is enforce that "name" is set to "preinstallimage" always (which is the default if it isn't set). patch updated.

Vogtinator:
> @Vogtinator all existing preinstallimages were checked in outside the bot, so the bot was never able to do any checking on it. this PR is improving over that situation.
> 
> what we can do is enforce that "name" is set to "preinstallimage" always (which is the default if it isn't set). patch updated.

Is there any advantage of having the name set to `preinstallimage` or match the source container name outside of consistency?

dirkmueller:
@Vogtinator the easiest would be to not set the name, in which case the spec file parser sets the name to "preinstallimage". this is what this PR is doing (now, didn't do before).

dirkmueller:
> > @Vogtinator all existing preinstallimages were checked in outside the bot, so the bot was never able to do any checking on it. this PR is improving over that situation.
> > what we can do is enforce that "name" is set to "preinstallimage" always (which is the default if it isn't set). patch updated.
> 
> Is there any advantage of having the name set to `preinstallimage` or match the source container name outside of consistency?

currently the name is not used by anything. it used to be required to be set because some logic in the OBS crashed when no Name was available.  This has been fixed to set the name to "preinstallimage" if it isn't set.

Vogtinator:
Ok, so this is just a cosmetical change. Considering that all previous preinstallimage changes were done skipping the bot, this PR won't break anything...

Vogtinator (approved):

dirkmueller:
@Vogtinator its not cosmetical for SLE.

check_source.py
```diff
@@ -221,10 +221,13 @@ def check_source_submission(self, source_project, source_package, source_revisio
 
         new_info = self.package_source_parse(source_project, source_package, source_revision, target_package)
         filename = new_info.get('filename', '')
-        if not (filename.endswith('.kiwi') or filename == 'Dockerfile') and new_info['name'] != target_package:
+        expected_name = target_package
+        if target_package.startswith('preinstallimage-'):
```
Vogtinator, line 225: Actually, checking for `filename == "preinstallimage"` might be better

Vogtinator:
> @Vogtinator its not cosmetical for SLE. the sle release managers require the "bot" to pass and currently there is no way to pass it.

I'm referring to the Name of preinstallimages. Is that not just cosmetical in SLE?

check_source.py
dirkmueller: but then we're missing that the package name is starting with 'preinstallimage-'

dirkmueller:
> > @Vogtinator its not cosmetical for SLE. the sle release managers require the "bot" to pass and currently there is no way to pass it.
> I'm referring to the Name of preinstallimages. Is that not just cosmetical in SLE?

I think we can do withotu the "Name" in _preinstallimage. but we still need this PR to fix it

check_source.py
Vogtinator: Yep, that has to be done separately:

```
if it builds a preinstallimage:
    make sure the source package name starts with "preinstallimage-"
    make sure the "binary" name is unset resp. "preinstallimage"
```

Without that, it would accept submitting an RPM package to `preinstallimage-foo` which had `Name: preinstallimage`

PR #2936, opened 2023-03-03
update gocd pkglistgen calls for latest ALP flavor

fcrozat:


Vogtinator (approved):
LGTM, but should be squashed on merge

PR #2941, opened 2023-03-09
Drop flavors in publish_distro config for leap15.5 

lkocman:
* They're not really used for anything later anyway and only cause issue on renames, etc.


lkocman:
This is the current visual interpretation, so far I see only one aarch64 image. I'll investigate what's going on

Screenshot from 2023-03-09 14-46-59

Issue #2943, opened 2023-03-22
installcheck: delete requests confused by own multibuild flavors


DimStar77:
Very often we see delete requests for packages that have a _multibuild flle; often times, one of the flavor depends on the other flavors

installcheck (and also cleanup_rings) do not take into account that they are linked together.

A typical scenario: delete req of rust1.62; nothing in the distro uses this anymore, except rust1.62:test (it's own test suite)
deleting rust1.62 obviously makes rust1.62:test disappear as well

The test should be enhanced to take the entire set of packages to disappear into account (and for sure it should also check that all flavors binaries are not used by anything else in the distro - in case that is not done yet)

PR #2945, opened 2023-03-22
Handle deletion of packages with _multibuild (#2943)

Vogtinator:
Also consider its multibuild flavors for deletion.

### DIFF
--- a/staging-installcheck.py
+++ b/staging-installcheck.py
@@ -21,6 +21,7 @@ from osclib.core import (builddepinfo, depends_on, duplicated_binaries_in_repo,
 
 from osclib.repochecks import installcheck, mirror
 from osclib.stagingapi import StagingAPI
+from osclib.memoize import memoize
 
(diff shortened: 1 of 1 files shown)


Vogtinator:
With this, deletion of `rust1.65` is accepted (previously wasn't) while deletion of `webkit2gtk3` is not (previously was).

DimStar77:
> With this, deletion of `rust1.65` is accepted (previously wasn't) while deletion of `webkit2gtk3` is not (previously was).

That sounds like a very welcome and correct change (I assume it would have accepted wk del req because the "" flavor does nothing)

Vogtinator:
> > With this, deletion of `rust1.65` is accepted (previously wasn't) while deletion of `webkit2gtk3` is not (previously was).
> 
> That sounds like a very welcome and correct change (I assume it would have accepted wk del req because the "" flavor does nothing)

Yep!

DimStar77 (approved):
LGTM

PR #2946, opened 2023-03-23
Rewrite cleanup_rings.py

Vogtinator:
Previously it went through each package's provides->required_by chain.
This does not deal with cycles correctly and there was only a hack to deal
with single-package cycles.

Instead, recursively walk through all required packages, starting from the
build enabled images and take note of all needed source packages.
This also handles _multibuild flavors as separate packages.

Depends on #2944.


Vogtinator:
> [09:42] <DimStar> fvogt: STaging:O installcheck looks good with that new list

Vogtinator:
Staging :O with the proposed changes is happy

PR #2958, opened 2023-04-19
staging-installcheck.py: Detect some edge cases in check_required_by

Vogtinator:
It's possible that OBS says that A provides cap required by B, but for B it does not show that cap is required or that cap is even provided by A. Detect and handle that explicitly.


Vogtinator:
Let's test in production, WCGW

PR #2959, opened 2023-05-11
devel-project: limit request has review state only

nilxam:


devel-project.py
```diff
@@ -190,6 +190,10 @@ def reviews(args):
     for devel_project in devel_projects:
         requests = get_review_list(apiurl, byproject=devel_project)
         for request in requests:
+            # get_review_list() behavior has been changed in osc
+            # https://github.com/openSUSE/osc/commit/00decd25d1a2c775e455f8865359e0d21872a0a5
+            if not request.state.name == 'review':
```
Vogtinator, line 195: Why not

```suggestion
            if request.state.name != 'review':
```

nilxam: just like that way, I'll change it to use `!=`, I don't mind :)

Vogtinator:
Works as expected. Let's merge this as a safeguard until a proper fix like https://github.com/openSUSE/osc/pull/1323 is available.

PR #2960, opened 2023-05-12
Add ALP pipelines for IBS

gyr:


Vogtinator (approved):
A case of "Let's see what happens" I guess...

gyr:
That is true for the changes related to Staging, they should remain disable on botmaster until a Staging area is configured for the project on IBS. The reason for submit them is due to it was a request from ALP architects

nilxam:
Is this PR still valid?

dirkmueller:
has conflicts, so I guess it is obsolete, please let us know if you think otherwise and don't hesitate to reopen then.

PR #2966, opened 2023-06-05
Move factory-package-news-web.py script to gunicorn

asdil12:
This script is used eg. for the /snapshot-changes/opensuse/Tumbleweed
route and was using the WSGIScriptAlias apache command.
With the switch to nginx we need this more modern approach.

Ticket: https://progress.opensuse.org/issues/130312


asdil12:
This is already applied on o3.

Martchus (approved):

PR #2967, opened 2023-06-06
also monitor the SUSE:ALP repositories

jsrain:
reported by openQA, which is not monitoring ALP repos


mloviska:
Kudos to @andrii-suse

nilxam (approved):

PR #2968, opened 2023-06-13
project-installcheck.py: Honour installcheck-ignore-conflicts config + tiny cleanup

Vogtinator:
Now we have some packages in the distro which fail installcheck
semi-intentionally: libstdc++6-gccX (with X < system GCC) conflicts with rpm
indirectly. While this can be ignored in stagings already, this caused an issue
once it's in the distro: The installcheck failure contains the release number
and so it gets rebuilt all the time in the hope to make it installable again.


Vogtinator:
No objections so far - let's just see what breaks.

PR #2972, opened 2023-06-15
Switch to Leap 15.5

dirkmueller:


Vogtinator:
I set up the `containers155` repo with the container here, let's see

PR #2974, opened 2023-06-19
dist: Add python3-requests

Vogtinator:
Needed by some scripts and tests.


Vogtinator:
This PR is not technically wrong, but we need https://build.opensuse.org/request/show/1093753 as well which will pull in requests in the right version then again.

PR #2977, opened 2023-06-22
Don't decline SRs from non-devel project for scmsync packages

dcermak:


check_source.py
```diff
@@ -152,10 +152,15 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
-                    self.review_messages['declined'] = 'Expected submission from devel package %s/%s' % (
-                        devel_project, devel_package)
-                    return False
+                    # check if the devel project & package are using scmsync
+                    # => waive the devel project source submission requirement
+                    meta = ET.fromstringlist(show_package_meta(self.apiurl, devel_project, devel_package))
+                    scm_sync = meta.find('scmsync')
+                    if not scm_sync or (scm_sync and not scm_sync.text.startswith(f"https://src.opensuse.org/pool/{source_package}")):
+                        # Not from proper devel project/package and not self-submission.
+                        self.review_messages['declined'] = 'Expected submission from devel package %s/%s' % (
```
dirkmueller, line 161: maybe a check on the user being the scm-sync bot user? so that we don't allow submitrequests from *not* the bot user accidentally (that bypass the git workflow)

dcermak: That will require quite a bit of refactoring, as the information who submitted the request is not forwarded into the individual `check_action_*` functions.

dirkmueller: That sucks

dirkmueller: can we check that the submitting source project is coming from home:defolos or whatever the trusted prefix will be? maybe we use a reasonable prefix that isn't home: and that we can actually keep..

dcermak: Turns out it wasn't that much refactoring work after all, I've implemented checking for the user now as well.

dirkmueller: e.g. check for src_project == `devel:Factory:git-workflow:staging` and then we use that one for submission source?

check_source.py
```diff
@@ -33,6 +33,8 @@ class CheckSource(ReviewBot.ReviewBot):
 
     SCRIPT_PATH = os.path.dirname(os.path.realpath(__file__))
 
+    ALLOWED_SCM_SUBMITTERS = ("defolos", )
```
Vogtinator, line 36: That's something for the config, not hardcoded

check_source.py
```diff
@@ -152,10 +154,20 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
-                    return False
+                    # check if the devel project & package are using scmsync
+                    # => waive the devel project source submission requirement
+                    meta = ET.fromstringlist(show_package_meta(self.apiurl, devel_project, devel_package))
+                    scm_sync = meta.find('scmsync')
+                    if (
+                            not scm_sync or
+                            (
+                                scm_sync and
+                                not scm_sync.text.startswith(f"https://src.opensuse.org/pool/{source_package}")
```
Vogtinator, line 165: Using `startswith` here will return false positives for URLs like `https://src.opensuse.org/pool/{source_package}-subpkg` or `https://src.opensuse.org/pool/{source_package}/../../hacker/pwned.git`

Vogtinator (changes requested):

check_source.py
dirkmueller: if the devel project has a hacker/pwned reference, we have larger issues

dirkmueller:
> * What's the issue with the current code, which does not look at scmsync at all?

> * If the devel project isn't used for submission, why is it set at all in OBS?

Because the devel project also provides the functionality to build the same package for older distributions via additional repositories. that functionality (that many users care about as there are > 50% leap users) would otherwise get lost. 

> * What's special about the `https://src.opensuse.org/pool/{source_package}` path?

its the location of where the sources of a package are authoritatively living once git workflow has been activated.

dirkmueller:
> That's orthogonal, I'm just referring to the `<devel project="some:where" package="foo"/>` property in the `openSUSE:Factory/foo` package meta. If the package is from git, this could simply be omitted?

How do you know that its from git? "somewhere" that information needs to come from, until we have switched over completely. plus the information of devel projects is still useful for package maintainership rights handling. 

Eventually we can switch over user access control to git but its still a long way to get there. it would be good to be able to start somewhere. do you have a better suggestion?

Vogtinator:
> > That's orthogonal, I'm just referring to the `<devel project="some:where" package="foo"/>` property in the `openSUSE:Factory/foo` package meta. If the package is from git, this could simply be omitted?
> 
> How do you know that its from git? "somewhere" that information needs to come from, until we have switched over completely. plus the information of devel projects is still useful for package maintainership rights handling.

At some point the scmsync property is set on those packages in oS:F and the devel prop removed. Until then the submissions will be from the devel prj I imagine?

> Eventually we can switch over user access control to git but its still a long way to get there. it would be good to be able to start somewhere. do you have a better suggestion?

check_source.py
Vogtinator: Right, I see this only checks the scmsync tag of the devel pkg, not the checked request. That should probably be done as well somewhere?

dirkmueller:
> At some point the scmsync property is set on those packages in oS:F and the devel prop removed. Until then the submissions will be from the devel prj I imagine?

maybe, that will be a longer road though, because it means we need to reimplement staging outside buildservice. currently staging and all the bots operate on submitrequest, and submitrequests do not work with scmsyncs. so we can't set this right now. 

Setting it in the devel project instead still keeps the functionality of devel projects alive plus makes it obvious that this package is no longer maintained in the devel project but elsewhere (and srs to devel project will fail, which is a feature in that regard).

check_source.py
dcermak: The SR will be generated by a bot, thus I don't really see the need for that. It will not really make anything more secure, only hinder us in changing things later on.

dcermak:
I've changed the bot to now check the name of the source project and grab the allowed sources from OBS as well instead of hardcoding them.

check_source.py
```diff
@@ -62,6 +62,7 @@ def target_project_config(self, project):
         self.allow_valid_source_origin = str2bool(config.get('check-source-allow-valid-source-origin', 'False'))
         self.valid_source_origins = set(config.get('check-source-valid-source-origins', '').split(' '))
         self.add_devel_project_review = str2bool(config.get('check-source-add-devel-project-review', 'False'))
+        self.allowed_scm_submission_sources = set(config.get('allowed-scm-submission-sources', 'home:defolos').split(' '))
```
dirkmueller, line 65:
```suggestion
        self.allowed_scm_submission_sources = set(config.get('allowed-scm-submission-sources', 'devel:Factory:git-workflow:staging').split(' '))
```

dirkmueller (approved):

check_source.py
```diff
@@ -152,10 +153,21 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+                    # check if the devel project & package are using scmsync
+                    # => waive the devel project source submission requirement
+                    meta = ET.fromstringlist(show_package_meta(self.apiurl, devel_project, devel_package))
+                    scm_sync = meta.find('scmsync')
+                    if (
+                            not scm_sync or
+                            (
+                                scm_sync and
+                                not scm_sync.text.startswith(f"https://src.opensuse.org/pool/{source_package}")
+                            ) or source_project not in self.allowed_scm_submission_sources
```
dirkmueller, line 165: If you want create subprojects per SR, this needs to be a "prefix" check.

dcermak: Should be done now, but please double check it.

check_source.py
```diff
@@ -62,6 +62,8 @@ def target_project_config(self, project):
         self.allow_valid_source_origin = str2bool(config.get('check-source-allow-valid-source-origin', 'False'))
         self.valid_source_origins = set(config.get('check-source-valid-source-origins', '').split(' '))
         self.add_devel_project_review = str2bool(config.get('check-source-add-devel-project-review', 'False'))
+        self.allowed_scm_submission_sources = set(
+            config.get('allowed-scm-submission-sources', 'devel:Factory:git-workflow:staging').split(' '))
```
nilxam, line 66: why does it needs a hardcoded value? add an config-bit to OSRT:Config that implied you want to have a flexible configurations, `devel:Factory:git-workflow:staging` can be set on the target project instead of to hardcode it here.

dcermak: What you are asking for is what the code should be doing. `devel:Factory:git-workflow:staging` is just the default value if no config is set.

Vogtinator: IMO `devel:Factory` only makes sense for `openSUSE:Factory` so the default here should be empty

dcermak:
Max Lin ***@***.***> writes:

> @nilxam commented on this pull request.
>
>
>
>> @@ -62,6 +62,8 @@ def target_project_config(self, project):
>          self.allow_valid_source_origin = str2bool(config.get('check-source-allow-valid-source-origin', 'False'))
>          self.valid_source_origins = set(config.get('check-source-valid-source-origins', '').split(' '))
>          self.add_devel_project_review = str2bool(config.get('check-source-add-devel-project-review', 'False'))
> +        self.allowed_scm_submission_sources = set(
> +            config.get('allowed-scm-submission-sources', 'devel:Factory:git-workflow:staging').split(' '))
>
> Yes, the default value should be empty.

Done

nilxam (approved):

check_source.py
```diff
@@ -150,12 +151,25 @@ def check_source_submission(self, source_project, source_package, source_revisio
...
+                    # check if the devel project & package are using scmsync & match the allowed prj prefix
+                    # => waive the devel project source submission requirement
+                    meta = ET.fromstringlist(show_package_meta(self.apiurl, devel_project, devel_package))
+                    scm_sync = meta.find('scmsync')
+                    if (
+                            not scm_sync or
+                            (
+                                scm_sync and
+                                not scm_sync.text.startswith(f"https://src.opensuse.org/pool/{source_package}")
+                                and all(not source_project.startswith(allowed_src) for allowed_src in self.allowed_scm_submission_sources)
```
dirkmueller, line 166: isn't that supposed to be `any(` rather than `all(` ?

dcermak: No, the branch should only be taken if the source project name begins with something that is none of the allowed sources.

PR #2983, opened 2023-07-03
Add cleanup_32bit.py tool

Vogtinator:
This uses code from cleanup_rings to figure out which packages are needed
for a 32bit build and runtime environment for steam and wine.

Draft because the API in cleanup_rings had to be changed and that broke cleanup_rings.


cleanup_32bit.py
```diff
@@ -0,0 +1,66 @@
...
+from osclib.core import fileinfo_ext_all
+from osclib.conf import Config
+from osclib.stagingapi import StagingAPI
+from osclib.cleanup_rings import CleanupRings
+
+class Cleanup32bit(ToolBase.ToolBase):
+    def run(self, prj: str, arch: str):
+        Config(self.apiurl, prj)
+        cr = CleanupRings(StagingAPI(self.apiurl, prj))
+        cr.whitelist = set(["wine", "wine-nine-standalone", "wine:staging"])
```
dirkmueller, line 14: Needed or requiredlist perhaps? whitelist is pretty generic ( and non inclusive) terminology

cleanup_32bit.py
```diff
@@ -0,0 +1,66 @@
...
+        cr.fill_pkginfo(prj, "standard", arch)
+
+        # Make sure those pkgs are also installable
+        for wppra in [("openSUSE:Factory:NonFree", "steam", "standard", "x86_64")]:
+            (wprj, wpkg, wrepo, warch) = wppra
+            for fileinfo in fileinfo_ext_all(self.apiurl, wprj, wrepo, warch, wpkg):
+                for providedby in fileinfo.findall('requires_ext/providedby[@name]'):
+                    name = providedby.get('name')
+                    # Those are not built as i586
+                    if "libgcc" in name or "libstdc++" in name:
```
dirkmueller, line 28: Maybe 'startswith'?

cleanup_32bit.py
Vogtinator: It's actually a list in CleanupRings still, can be changed in a later refactor commit after I got this one working

cleanup_32bit.py
Vogtinator: Changed

cleanup_32bit.py
Vogtinator: It's a dict now with more info, renamed to `force_required`

Issue #2954, opened 2023-04-17
Replacing or dropping publish_distro


lkocman:
Hello team,

publish_distro seems to have a simple job of rsyncing one directory to another (ftp-stage to ftp-prod), with a delay of n-hours, and there is a cleanup mechanism to keep older copies of files for n-days. I guess the first comes handy to ensure that all data is written/synced to ftp-stage by obs publisher, and the latter comes handy so people don't get their repodata constantly invalidated during development cycle.

However, we do see weird anomalies when  appliances for released version of Leap 15.X are not refreshed for months and simply calling script with --force fixes it.

Configuration is doubtful (regexpes that don't have much use in the current logic) etc. And in overal the script tries to be too smarty-pants for what it's supposed to do.

[publish.log](https://github.com/openSUSE/openSUSE-release-tools/files/11249703/publish.log)

Could we have a simplier replacement and do we need it at all?

andrii-suse:
FYI, I am investigating options for combined solution covering/integrating publish_distro, reporusher and tumbleweed-snapshot. Maybe MirrorCache integration as well. Have some ideas atm, but not sure how it will look like. So in case of any ideas or status check - let me know.

PR #2989, opened 2023-07-14
Add test for publish_distro

andrii-suse:
I am investigating how to improve publish_distro script, so having some kind of tests with debug output will help to track what exactly changes.


Vogtinator:
I'm wondering to which extent tests on this level actually work in practice. If the design and mechanisms are changed like #2954, then this might not be the case...

andrii-suse:
> I'm wondering to which extent tests on this level actually work in practice. If the design and mechanisms are changed like #2954, then this might not be the case...

My understanding that it is OK if tests sometimes don't cover everything.
And It is good if they cover at least something and help to visually track changes in functionality.

t/02-publish_distro_leap.sh.log
```diff
@@ -24,4 +24,3 @@ rsync -avvhiH t/tmp/ftp-stage/pub/opensuse//distribution/leap/15.5/repo --link-d
 rsync -avhiH t/tmp/ftp-stage/pub/opensuse/source//distribution/leap/15.5/ --link-dest=t/tmp/ftp-stage/pub/opensuse/source//distribution/leap/15.5 t/tmp/ftp/pub/opensuse/source//distribution/leap/15.5 --delete-after
 rsync -avhiH t/tmp/ftp-stage/pub/opensuse/debug//distribution/leap/15.5/ --link-dest=t/tmp/ftp-stage/pub/opensuse/debug//distribution/leap/15.5 t/tmp/ftp/pub/opensuse/debug//distribution/leap/15.5 --delete-after
 curl -X POST -F version=111.11 https://openqa.opensuse.org/snapshot-changes/opensuse/15.5/current
-stage/pub/opensuse/debug//distribution/leap/15.5 t/tmp/ftp/pub/opensuse/debug//distribution/leap/15.5 --delete-after
```
andrii-suse: I have no idea why that line is truncated like this, but it has gone when stdout was replaced with stderr for logs when --dry flag is set

lkocman:
I mean I'm supportive of any test coverage, however, ... I was hoping we'd eventually replace what we have :-)

PR #2990, opened 2023-07-14
Fix sle15 literal import

dcermak:


dist/package/openSUSE-release-tools.spec
```diff
@@ -58,6 +58,10 @@ Requires:       python3-pycurl
 Requires:       python3-python-dateutil
 Requires:       python3-pyxdg
 Requires:       python3-requests
+# typing extensions are needed on SLE & Leap
+%if 0%{?suse_version} < 1600
```
Vogtinator, line 62: Is it in ALP? If yes, can be changed to `<= 1500`

dcermak: I hope that alp is at python >= 3.8! Then it wouldn't be required.

Vogtinator: Should be 3.11

dcermak: Fixed

.github/workflows/ci-test.yml
```diff
@@ -42,6 +42,47 @@ jobs:
...
+
+      - name: fix the file permissions of the repository
+        run: chown -R $(id -un):$(id -gn) .
+
+      - name: add the openSUSE:Tools repository for Tumbleweed
+        run: zypper -n ar https://download.opensuse.org/repositories/openSUSE:/Tools/openSUSE_Tumbleweed/openSUSE:Tools.repo
+        if: ${{ contains(matrix.image, 'tumbleweed') }}
+      - name: add the openSUSE:Tools repository for Leap
+        run: zypper -n ar https://download.opensuse.org/repositories/openSUSE:/Tools/15.5/openSUSE:Tools.repo
+        if: ${{ contains(matrix.image, 'leap') }}
```
Vogtinator, line 69:
```suggestion
        if: ${{ contains(matrix.image, '15.5') }}
```

.github/workflows/ci-test.yml
```diff
@@ -42,6 +42,47 @@ jobs:
...
+        run: zypper -n in git
+
+      - uses: actions/checkout@v3
+
+      - name: fix the file permissions of the repository
+        run: chown -R $(id -un):$(id -gn) .
+
+      - name: add the openSUSE:Tools repository for Tumbleweed
+        run: zypper -n ar https://download.opensuse.org/repositories/openSUSE:/Tools/openSUSE_Tumbleweed/openSUSE:Tools.repo
+        if: ${{ contains(matrix.image, 'tumbleweed') }}
```
Vogtinator, line 66: IMO `if` should be above `run`

.github/workflows/ci-test.yml
```diff
@@ -42,6 +42,47 @@ jobs:
...
+        if: ${{ contains(matrix.image, 'leap') }}
+
+      - name: install the dependencies of openSUSE-release-tool
+        run: |
+          zypper -n --gpg-auto-import-keys refresh
+          zypper -n source-install openSUSE-release-tools
+
+      - name: FIXME, install missing dependencies
+        run: |
+          zypper -n in python3-requests python3-typing_extensions
```
Vogtinator, line 78: The latter only on 15.5?

.github/workflows/ci-test.yml
```diff
@@ -42,6 +42,47 @@ jobs:
...
+
+      - name: run a simple smoke test whether --help actually works
+        run: |
+          for f in $(find . -maxdepth 1 -type f -executable -print); do
+              eval "$f --help"
```
Vogtinator, line 83:
```suggestion
              "$f" --help
```

.github/workflows/ci-test.yml
dcermak: It doesn't matter where you put the `if`.

.github/workflows/ci-test.yml
dcermak: requests is missing on both os versions

.github/workflows/ci-test.yml
Vogtinator: Not semantically, but usually conditions are written before the code that they guard, to match the flow of reading and thought.

.github/workflows/ci-test.yml
```diff
@@ -42,6 +42,47 @@ jobs:
...
+        run: zypper -n ar https://download.opensuse.org/repositories/openSUSE:/Tools/openSUSE_Tumbleweed/openSUSE:Tools.repo
+        if: ${{ contains(matrix.image, 'tumbleweed') }}
+      - name: add the openSUSE:Tools repository for Leap
+        run: zypper -n ar https://download.opensuse.org/repositories/openSUSE:/Tools/15.5/openSUSE:Tools.repo
+        if: ${{ contains(matrix.image, 'leap') }}
+
+      - name: install the dependencies of openSUSE-release-tool
+        run: |
+          zypper -n --gpg-auto-import-keys refresh
+          zypper -n source-install openSUSE-release-tools
```
Vogtinator, line 74:
```suggestion
          zypper -n source-install -d openSUSE-release-tools
```

for dependencies only.

But don't we actually need *runtime* requires here?

.github/workflows/ci-test.yml
dcermak: :shrug: done

.github/workflows/ci-test.yml
dcermak: We do need them, that's why requests was missing.

Vogtinator: So `zypper in openSUSE-release-tools` would be more correct, though one "generation" behind?

dcermak: yes, that's now in the job as well

dcermak:
The test failure is a failure uploading to codecov.

Vogtinator (approved):

PR #2992, opened 2023-07-20
Revert temporary fixups, these are working now

dirkmueller:


Vogtinator:
`./check_source.py:3:1: F401 'contextlib.suppress' imported but unused`

"seem to be working" == "are working"?

Vogtinator (approved):

PR #2998, opened 2023-07-26
bci_repo_publish.py: Return 1 if the last published build is over a week old

Vogtinator:
This way it's easy to spot whether a manual look is required.

I'm not entirely sure whether this is a good idea. Ideally someone looks directly when builds or tests fail and fixes this before a week expires, but as additional safe guard it's probably worth it.

### DIFF
--- a/gocd/bci_repo_publish.py
+++ b/gocd/bci_repo_publish.py
@@ -26,6 +26,7 @@ class BCIRepoPublisher(ToolBase.ToolBase):
         ToolBase.ToolBase.__init__(self)
         self.logger = logging.getLogger(__name__)
         self.openqa = OpenQA_Client(server='https://openqa.suse.de')
+        self.last_publish_mtime = 0
 
     def version_of_product(self, project, package, repo, arch):
         """Get the build version of the given product build, based on the binary name."""
@@ -43,7 +44,7 @@ class BCIRepoPublisher(ToolBase.ToolBase):
         url = makeurl(self.apiurl, ['build', project, repo, arch, package])
         root = ET.parse(http_GET(url)).getroot()
         mtime = root.xpath('/binarylist/binary[@filename = "_buildenv"]/@mtime')
-        return mtime[0]
+        return int(mtime[0])
 
     def openqa_jobs_for_product(self, arch, version, build):
         """Query openQA for all relevant jobs"""
@@ -76,12 +77,12 @@ class BCIRepoPublisher(ToolBase.ToolBase):
 
         return True
 
-    def run(self, version, token=None):
+    def publish_if_possible(self, version, token=None):
+        """If the OBS project is finished and the built binaries are newer than
+        the ones in the publish location, check openQA for results. If they pass,
+        use the given token to release the builds into the target project.
+        The mtime of the last published build is written to self.last_publish_mtime."""
         build_prj = f'SUSE:SLE-{version}:Update:BCI'
-
-        if not self.is_repo_published(build_prj, 'images'):
-            self.logger.info(f'{build_prj}/images not successfully built')
-            return
 
         # Build the list of packages with metainfo
         packages = []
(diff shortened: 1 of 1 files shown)


dirkmueller:
I think we should simply stop publishing when there are failed builds? I'm not sure what hte cutoff of 7 days is solving?

Vogtinator:
> I'm not sure what hte cutoff of 7 days is solving?

Currently the bot fails only if some API call actually fails, but not if a build failed or openQA test failed, as this is just expected behaviour. This means that if the repo is unresolvable for a month, the only sign is that there haven't been any openQA runs recently. Or if openQA fails for a month because nobody had a look, the jobs are still green.

With this change it's visible in botmaster as a pipeline failure if there hasn't been a new build published for a week.

gocd/bci_repo_publish.py
```diff
@@ -147,6 +156,7 @@ def run(self, version, token=None):
         # Trigger publishing
         if token is None:
             self.logger.warning('Would publish now, but no token specified')
+            self.last_publish_mtime = packages[0]['built_mtime']
```
dirkmueller, line 159: why are we setting last_publish_mtime in this error case?

Vogtinator: It's not really an error case, more of a dry run case

gocd/bci_repo_publish.py
```diff
@@ -95,6 +96,13 @@ def run(self, version, token=None):
                 'publish_prj': f'SUSE:Products:SLE-BCI:{version}:{arch}'
             })
 
+        if not self.is_repo_published(build_prj, 'images'):
+            self.logger.info(f'{build_prj}/images not successfully built')
+            pkg = packages[0]
+            self.last_publish_mtime = \
+                self.mtime_of_product(pkg['publish_prj'], pkg['name'], 'images', 'local')
```
dirkmueller, line 103: and here? we're not actually publishing, we're aborting because it wasn't successfully built?

Vogtinator: Yes, which means we have to fetch the last published mtime from OBS to know how long ago that was

gocd/bci_repo_publish.py
dirkmueller: sure, but there are plenty of other returns before hand that are not setting `last_publish_mtime`. so this is a bit inconsistent.

gocd/bci_repo_publish.py
dirkmueller: as a nit pick, people usually avoid escaping newlines in python. you can write this in a more pythonic way:

``` python

     self.last_publish_mtime = self.mtime_of_product(
           pkg['publish_prj'], pkg['name'], 'images', 'local')
```

dirkmueller:
so if I understand this correctly. the only goal of the whole logic is to make the checkbox "red" rather than "green" in the botmaster listing/graph. 

how about we change the method to return True/False on whether it actually did a publish or not and then simply convert that to an exit code of 0 / 1?

then it would be red if it had something to publish but didn't publish (because tests failed or no matching run found or publish failed), and otherwise green? 

does the UI support tripple state? then we could encode the "still waiting" states with another color..

gocd/bci_repo_publish.py
Vogtinator: IMO harder to read and flake8 didn't complain :shrug:

Vogtinator:
> so if I understand this correctly. the only goal of the whole logic is to make the checkbox "red" rather than "green" in the botmaster listing/graph.

Yes.

> how about we change the method to return True/False on whether it actually did a publish or not and then simply convert that to an exit code of 0 / 1?
> then it would be red if it had something to publish but didn't publish (because tests failed or no matching run found or publish failed), and otherwise green?

This wouldn't be able to catch if there was no new build for a week, e.g. due to unresolvable or blocked.

> does the UI support tripple state? then we could encode the "still waiting" states with another color..

There's only orange for "pipeline is running"

dirkmueller:
> then it would be red if it had something to publish but didn't publish (because tests failed or no matching run found or publish failed), and otherwise green?
> This wouldn't be able to catch if there was no new build for a week, e.g. due to unresolvable or blocked.

well, unresolvable or build failed would lead to an exit 1, as well as a missing openqa run. unfinished openqa run or unclean build status would not. 

> > does the UI support tripple state? then we could encode the "still waiting" states with another color..
> There's only orange for "pipeline is running"

ok, too bad. 

Well, I don't have strong feelings either way. I think the 7 days to wait is arbitrary and feels a bit too long. maybe 3 days?
Other than that I still think trying to determine the status "this looks erratic" or "this can still be normal" and then just exiting 1/0 right away is the better approach. we don't want to wait for a week to notice that openqa has lost an event or the like. that is not gonna help anyone.

PR #2999, opened 2023-07-28
added QU 15 SP5

msmeissn:
add install checker for 15 SP5


Vogtinator (changes requested):
This file is autogenerated, edit checkers.suse.gocd.yaml.erb instead.

Also, this pipeline is broken for quite some time now because the used user account is blocked in IBS.

dirkmueller (approved):

PR #3002, opened 2023-08-02
factory-package-news: support newer python3 rpm bindings

g7:
rpm's python bindings changed in version 4.15 [0] so that they actually return utf-8 strings. Handle this case while keeping support for the older bindings such as the ones shipped with Leap 15.

[0] https://github.com/rpm-software-management/rpm/commit/84920f898315d09a57a3f1067433eaeb7de5e830

### DIFF
--- a/factory-package-news/factory-package-news.py
+++ b/factory-package-news/factory-package-news.py
@@ -18,6 +18,15 @@ data_version = 3
 changelog_max_lines = 100  # maximum number of changelog lines per package
 
 
+# rpm's python bindings changed in version 4.15 [0] so that they actually return
+# utf-8 strings. Leap 15 ships an older version, so this is needed to
+# keep this script working there too.
+#
+# [0] https://github.com/rpm-software-management/rpm/commit/84920f898315d09a57a3f1067433eaeb7de5e830
+def forcestr(content):
+    return str(content, 'utf-8') if isinstance(content, bytes) else content
+
+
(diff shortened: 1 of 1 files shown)


g7:
v2: addressed flake8 warnings :)

factory-package-news/factory-package-news.py
```diff
@@ -18,6 +18,15 @@
 changelog_max_lines = 100  # maximum number of changelog lines per package
 
 
+# rpm's python bindings changed in version 4.15 [0] so that they actually return
+# utf-8 strings. Leap 15 ships an older version, so this is needed to
+# keep this script working there too.
+#
+# [0] https://github.com/rpm-software-management/rpm/commit/84920f898315d09a57a3f1067433eaeb7de5e830
+def forcestr(content):
```
dirkmueller, line 26: I wonder if `utf8str` would be a slightly besser name.

dirkmueller (approved):
LGTM, small nit inline

dirkmueller:
test are passing, we can solve the nit anytime..

factory-package-news/factory-package-news.py
g7: Thanks, yeah definitely looks better than `forcestr` :) Should I make another PR that changes that?

dirkmueller: Yes please if you agree

PR #3007, opened 2023-08-18
add Leap 15.6 publish_distro config

lkocman:


lkocman:
copy and bump from 15.5

lkocman:
The failed linter check is unrelated

PR #3012, opened 2023-08-29
cleanup_rings: keep docker-buildx for QA tests

DimStar77:


Vogtinator (approved):

DimStar77:
Question actually: it being on the DVD is sufficient for cleanup_rings NOT to want to throw it away?

Vogtinator:
> Question actually: it being on the DVD is sufficient for cleanup_rings NOT to want to throw it away?

Yes, it should be

DimStar77:
PR not needed - docker in fact recommends docker-buildx - so the pure presence of docker-buildx in ring1 makes it be present on the DVD and there is a reason for the package to be in ring1

PR #3013, opened 2023-08-29
Don't crash when trying to fetch owner for 404 devel prjs

Vogtinator:
openh264 is very special.


DimStar77 (approved):

osclib/core.py
```diff
@@ -122,8 +122,7 @@ def convert_from_osc_et(xml):
 @memoize(session=True)
 def owner_fallback(apiurl, project, package):
     root = owner(apiurl, package, project=project)
-    entry = root.find('owner')
-    if not entry or project.startswith(entry.get('project')):
+    if not root or not root.find('owner') or project.startswith(entry.get('project')):
```
eduardoj, line 125:
```suggestion
    if not root or not root.find('owner') or project.startswith(root.find('owner').get('project')):
```
Maybe it will be better not to remove the assignment to `entry`... :thinking:

eduardoj (changes requested):

osclib/core.py
Vogtinator: That's actually what I did first, I just reverted to that

eduardoj (approved):

PR #3014, opened 2023-09-05
Add ALP installer pipelines

gyr:


gocd/alp.target.gocd.yaml
```diff
@@ -0,0 +1,225 @@
...
+        - script: |-
+            osc -A https://api.suse.de release SUSE:ALP:Products:Installer:1.0:ToTest
+            sleep 600
+            while (osc -A https://api.suse.de/ api "/build/SUSE:ALP:Products:Installer:1.0:PUBLISH/_result?view=summary&repository=images" | grep "result project" | grep -v 'code="published" state="published">'); do
+                echo PENDING
+                sleep 600
+            done
+            osc -A https://api.suse.de/ api "/build/SUSE:ALP:Products:Installer:1.0:PUBLISH/_result?view=summary&repository=images" | grep "result project" | grep 'code="published" state="published">' && echo PUBLISHED
+
+  Installer.ImagesManual:
```
Vogtinator, line 127: Isn't this covered by ^ already?

gocd/alp.target.gocd.yaml
```diff
@@ -0,0 +1,225 @@
...
+        resources:
+          - staging-bot
+        tasks:
+          - script: osc -A https://api.suse.de release SUSE:ALP:Products:Installer:1.0:ToTest
+
+  Installer.ReleaseScc:
+    group: ALP.Target
+    parameters:
+      product: installer-1.0
+      target: MyTarget
```
Vogtinator, line 183: Is this correct? Looks like a placeholder

gocd/alp.target.gocd.yaml
```diff
@@ -0,0 +1,225 @@
...
+        - script: |-
+            osc -A https://api.suse.de release SUSE:ALP:Products:Installer:1.0:ToTest
+            sleep 600
+            while (osc -A https://api.suse.de/ api "/build/SUSE:ALP:Products:Installer:1.0:PUBLISH/_result?view=summary&repository=images" | grep "result project" | grep -v 'code="published" state="published">'); do
+                echo PENDING
+                sleep 600
+            done
+            osc -A https://api.suse.de/ api "/build/SUSE:ALP:Products:Installer:1.0:PUBLISH/_result?view=summary&repository=images" | grep "result project" | grep 'code="published" state="published">' && echo PUBLISHED
+
+  Installer.Images.NEW:
```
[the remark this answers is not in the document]
nilxam: indeed, and I'm confused with `Installer.Images`, `Installer.Images.NEW` and `Installer.ImagesManual`, do they really have different meaningful purpose? this looks like just align with what **dolomite.target.gocd.yaml** had though

gocd/alp.target.gocd.yaml
nilxam: it looks correct, other alp pipelines have the similar params feed to scc

gocd/alp.target.gocd.yaml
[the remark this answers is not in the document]
gyr: Good point, I would like to remove `Installer.Images.NEW` and `Installer.ImagesManual` (also from SLE and Dolomite) as RelMgr've never used them and also to save agent power on botmaster (that is why I paused  the SLE ones). I will discuss it with RelMgr if they agreed on that. 
Just to give you a background about them:
 - `Installer.Images.NEW`: the only different is that it uses `sle-prjmgr-tools` on `Release.Images.To.Test`, instead of directly run `osc release`. IIRC this script should somehow update the Jira tickets but it is never put in production 
 - `Installer.ImagesManual`: the difference here is that `Release.To.Test` is automatically triggered after `Check.Images.Finish`, this new workflow was requested by RelMgr but never used

I will include `Installer.ReleaseScc` as it was never put in production, as well. For now, I will convert this PR to draft until I have a agreement with RelMgr.

gyr:
V2

PR #3016, opened 2023-09-13
Draft: check_tags_in_request: search on the linked Factory project as well

g7:
The IBS instance embeds the whole project space of build.opensuse.org under openSUSE.org:.

This change allows for the changelog checker to search there as well, thus fixing checkTagNotRequired() when running inside the IBS.

### DIFF
--- a/check_tags_in_requests.py
+++ b/check_tags_in_requests.py
@@ -25,7 +25,7 @@ class FactorySourceChecker(ReviewBot.ReviewBot):
 
     def __init__(self, *args, **kwargs):
         ReviewBot.ReviewBot.__init__(self, *args, **kwargs)
-        self.factory = ["openSUSE:Factory"]
+        self.factory = ["openSUSE:Factory", "openSUSE.org:openSUSE:Factory"]
         self.review_messages = {'accepted': 'ok', 'declined': 'the package needs to be accepted in Factory first'}
         self.history_limit = 5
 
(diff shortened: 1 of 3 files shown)


g7:
Drafting, checking the failing tests

nilxam:
1. I hope you read https://github.com/openSUSE/openSUSE-release-tools/blob/master/check_tags_in_requests.py#L16 already, although I'm not sure how many place in the SLE development process still use this tool
2. there is no `openSUSE.org` namespace on OBS, if it was hardcoded and without exception check then someone run this bot on OBS will just crashing, perhaps maintenance team still use this script on OBS, I'm not sure either
3. there is `--factory` param to replace the default Factory project, ReviewBot will have it at least, if TagChecker needs a specific Factory project then it can resuse it everywhere else in the check_tags_in_request.py
4. I don't get why do you need to check two Factory project, we don't have `openSUSE:Factory` project on IBS but an inherited `openSUSE.org:openSUSE:Factory`; vice versa, there is no `openSUSE.org:` namespace on OBS but a `openSUSE:Factory` exists..  even if you worked out the testdata and have had a passed tests, but it doesn't matches the reality

g7:
Hi, thanks for the review:

Yes, I think the only user of that would be check_tags_in_requests.py, missing projects are already handled gracefully.

But as for 3) indeed it works as well, got swayed by running it manually to debug another thing in the check_tags_in_request bot (totally overlooking `--factory`), looking at the gocd pipeline it already works as it should, so this is indeed useless :upside_down_face: 

Thanks again and sorry for the noise!

PR #3017, opened 2023-09-15
check_source.py: Don't skip various checks and reviews for non-spec packages

Vogtinator:
There was an early exit for packages which don't have a .spec file, but that is too early and skipped important parts like running the source_validator and download_files services and adding the review-team review.


DimStar77 (approved):
LGTM - and definitively makes sense

PR #3021, opened 2023-10-17
project-installcheck.py: Support arch-specific ignore-conflicts config

Vogtinator:
Sometimes installcheck failures should be ignored for a specific architecture only.


DimStar77 (approved):
seems reasonable

PR #3024, opened 2023-10-25
Temporary disable checks for ALP

gyr:
installcheck and packagelist checks were temporaly disabled for ALP Staging pipelines to allow openQA workflow test, while the issues reported by them are not fixed (they will be tracked on Staging H only)


nilxam:
Is it really needed? You can just go botmaster and paused the pipeline.

gyr:
Unfortunately yes. Agree that is not the ideal approach, but these 2 tests have issues that won't be fixed soon, they're blocking ALP to test the entire Staging workflow, botmaster (build checker and enable publish images) + openQA

PR #3026, opened 2023-10-26
Only check for products to be built successfully

dirkmueller:
The "images" repository contains both containers and ftp products. we only want to wait for the products, which are built in the "local" arch.


Vogtinator (approved):
Should work

PR #3028, opened 2023-10-26
gocd: alp-stagings: Re enable Update.000products step for all stagings.

gleidi-suse:
The step is actually required for openQA tests.

### DIFF
--- a/gocd/alp-stagings.gocd.yaml
+++ b/gocd/alp-stagings.gocd.yaml
@@ -195,6 +195,21 @@ pipelines:
                   ./report-status.py -A $STAGING_API -p $STAGING_PROJECT -n packagelists -r standard -s pending
                   ./verify-repo-built-successful.py -A $STAGING_API -p $STAGING_PROJECT -r standard
 
+    - Update.000product:
+        resources:
+          - repo-checker
+        tasks:
+          - script: |-
+              export PYTHONPATH=$PWD/scripts
+              cd scripts/gocd
+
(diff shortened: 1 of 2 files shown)


gyr (dismissed):

nilxam:
when https://github.com/openSUSE/openSUSE-release-tools/pull/3024 has merged I'm guessing you may need to re-enable update-000product again unless you want to do a packagelist calculation for each staging per the changes are in the each staging, IOW manual patch 000product per the changes, it's not a terrible thing though if you know what needs add to packagelist and what needs to be drop off from the list.

I don't know what happens in ALP staging but I suppose ALP staging is at a broken state already(except of H), if the default pattern in groups.yml can not be fulfilled then pkglistgen would not give a right list still, but post the wrong bits instead, like add ''FOO is uninstallable: nothing provides BAR'' to packagelist, something like that, just not to update 000product. In fact, if your staging is broken already, you should really fix your staging first: check-in the fix asap, or revert to a working version, make sure the requiring packages are installable. yes, you can workaround groups.yml to get a working packagelist by not adding broken package to packagelist, however you may get a non-working image, and lots of openqa failings. In short, you should fix  your staging first :-)

gyr:
I'm converting this to draft to avoid it be merge by mistake

dirkmueller:
@gyr still needed? can be closed?

gyr:
> @gyr still needed? can be closed?

No, closing it.

PR #3029, opened 2023-10-26
Release bci repositories per architecture independently

dirkmueller:
There is no real good reason to wait for all the architectures to pass testing, so if there is one architecture finished, release it.

### DIFF
--- a/gocd/bci_repo_publish.py
+++ b/gocd/bci_repo_publish.py
@@ -134,6 +134,7 @@ class BCIRepoPublisher(ToolBase.ToolBase):
             return
 
         # Check openQA results
+        mandatory_arches = ('aarch64', 'x86_64')
         for pkg in packages:
             passed = 0
             pending = 0
@@ -148,14 +149,13 @@ class BCIRepoPublisher(ToolBase.ToolBase):
                 else:
                     self.logger.warning(f'https://openqa.suse.de/tests/{job["id"]} failed')
                     failed += 1
-            if passed == 0 or pending > 0 or failed > 0:
+            if pending or failed:
                 self.logger.info(f'openQA did not (yet) pass for {pkg["name"]}: {passed}/{pending}/{failed}')
                 continue
+            if passed == 0 and pkg['arch'] in mandatory_arches:
+                self.logger.info('No positive result from openQA (yet)')
+                return
             openqa_passed_packages.append(pkg)
-
-        if not openqa_passed_packages:
-            self.logger.info(f'No positive result from openQA (yet)')
-            return
 
         # Trigger publishing
         if token is None:
--- a/gocd/bci_repo_publish.py
+++ b/gocd/bci_repo_publish.py
@@ -60,7 +60,7 @@ class BCIRepoPublisher(ToolBase.ToolBase):
         return self.openqa.openqa_request('GET', 'jobs', values)['jobs']
 
     def is_repo_published(self, project, repo, arch=None):
-        """Validates that the given prj/repo is fully published and all builds
+        """Validate that the given prj/repo is fully published and all builds
         have succeeded."""
         result_filter = {'view': 'summary', 'repository': repo}
         if arch:
@@ -197,8 +197,7 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
 
     @cmdln.option('--token', help='The token for publishing. Does a dry run if not given.')
     def do_run(self, subcmd, opts, project):
-        """${cmd_name}: run the BCI repo publisher for the specified project,
-        e.g. 15-SP3
+        """${cmd_name}: run BCI repo publisher for the project, e.g. 15-SP5.
 
         ${cmd_usage}
         ${cmd_option_list}
--- a/gocd/bci_repo_publish.py
+++ b/gocd/bci_repo_publish.py
@@ -19,6 +19,7 @@ import re
 from lxml import etree as ET
 from openqa_client.client import OpenQA_Client
 from osc.core import http_GET, makeurl
+from random import randint
 
 
 class BCIRepoPublisher(ToolBase.ToolBase):
@@ -178,7 +179,7 @@ class BCIRepoPublisher(ToolBase.ToolBase):
         for pkg in openqa_passed_packages:
             while not self.is_repo_published(pkg['publish_prj'], 'images'):
                 self.logger.debug(f'Waiting for {pkg["publish_prj"]}')
-                time.sleep(20)
+                time.sleep(randint(10, 30))
 
 
 class CommandLineInterface(ToolBase.CommandLineInterface):
--- a/gocd/bci_repo_publish.py
+++ b/gocd/bci_repo_publish.py
@@ -87,6 +87,8 @@ class BCIRepoPublisher(ToolBase.ToolBase):
 
         # Build the list of packages with metainfo
         packages = []
+        # List of packages that have passed openQA
+        openqa_passed_packages = []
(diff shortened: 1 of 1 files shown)


gocd/bci_repo_publish.py
```diff
@@ -131,35 +134,28 @@ def run(self, version, token=None):
...
                     self.logger.info(f'https://openqa.suse.de/tests/{job["id"]} pending')
-                    pending += 1
-                else:
+                    break
+                if job['result'] not in ('passed', 'softfailed'):
                     self.logger.warning(f'https://openqa.suse.de/tests/{job["id"]} failed')
-                    failed += 1
+                    break
+            else:
+                openqa_passed_packages.append(pkg)
```
dcermak, line 147: Please add a comment here how the `for - else` construct works. It is so seldomly used, that many have no idea what it does.

gocd/bci_repo_publish.py
```diff
@@ -173,10 +169,10 @@ def run(self, version, token=None):
                 raise RuntimeError(f'Releasing failed: {req.text}')
 
         self.logger.info('Waiting for publishing to finish')
-        for pkg in packages:
+        for pkg in openqa_passed_packages:
             while not self.is_repo_published(pkg['publish_prj'], 'images'):
                 self.logger.debug(f'Waiting for {pkg["publish_prj"]}')
-                time.sleep(20)
+                time.sleep(randint(10, 30))
```
Vogtinator, line 175: IMO either constant times or exponential backoff are ok, but waiting randomly is weird.

dcermak (approved):
LGTM overall, just one nit

gocd/bci_repo_publish.py
```diff
@@ -131,35 +134,28 @@ def run(self, version, token=None):
...
+                if job['result'] == 'none':
                     self.logger.info(f'https://openqa.suse.de/tests/{job["id"]} pending')
-                    pending += 1
-                else:
+                    break
+                if job['result'] not in ('passed', 'softfailed'):
                     self.logger.warning(f'https://openqa.suse.de/tests/{job["id"]} failed')
-                    failed += 1
+                    break
+            else:
```
Vogtinator, line 146: "No jobs" must not be treated as "passed", otherwise this would publish before openQA scheduled jobs

gocd/bci_repo_publish.py
```diff
@@ -131,35 +134,28 @@ def run(self, version, token=None):
...
-                elif job['result'] == 'none':
+                if job['result'] == 'none':
                     self.logger.info(f'https://openqa.suse.de/tests/{job["id"]} pending')
-                    pending += 1
-                else:
+                    break
+                if job['result'] not in ('passed', 'softfailed'):
                     self.logger.warning(f'https://openqa.suse.de/tests/{job["id"]} failed')
-                    failed += 1
+                    break
```
Vogtinator, line 145: It intentionally didn't break before so that the list of jobs blocking release is always complete

Vogtinator (changes requested):

gocd/bci_repo_publish.py
Vogtinator: +1. I had to double-check what the behaviour of `for - else` is.

Due to https://github.com/openSUSE/openSUSE-release-tools/pull/3029#discussion_r1374251819, the `break` should be removed which means the `for - else` way would no longer work anyway.

gocd/bci_repo_publish.py
dirkmueller: constant time means access spikes at regular intervals, we should avoid that. exponential means that we need to add an abort condition also. what do you suggest as maximum waiting time?

Vogtinator: > constant time means access spikes at regular intervals, we should avoid that

AFAIU that applies only if the regularity is at a certain time of day and this can accumulate. In this case the starting point is not that regular, so the checks are not regular either.

(Ideally GoCD gains builtin support for timer jitter like systemd.timer's `RandomizedDelaySec=`)

> exponential means that we need to add an abort condition also. what do you suggest as maximum waiting time?

I'd use an upper limit for the wait time instead and just wait indefinitely if necessary. The alternative is either to fail the job or ignore that publishing didn't complete, which doesn't help either.

Vogtinator: That said, I do like the simplicity of `randint(10, 30)` compared to some extra lines for linear or exponential backoff with a ceiling... So if regular intervals are an issue then randint is fine for me

dirkmueller: ok, then lets keep it at randint for now.. otherwise I'll just drop the commit.

gocd/bci_repo_publish.py
dirkmueller: changed back

gocd/bci_repo_publish.py
dirkmueller: ok, then we need a different approach on handling the missing s390x and ppc64le tests.

gocd/bci_repo_publish.py
```diff
@@ -131,13 +134,14 @@ def run(self, version, token=None):
             return
 
         # Check openQA results
-        openqa_passed = True
         for pkg in packages:
             passed = 0
             pending = 0
             failed = 0
+            openqa_tested = False
```
Vogtinator, line 141: Variable is unnecessary, there is already passed/pending/failed.

What about this:

```python3
         # Check openQA results
        for pkg in packages:
            passed = 0
            pending = 0
            failed = 0
            for job in self.openqa_jobs_for_product(arch=pkg['arch'], version=version,
                                                    build=pkg['built_version']):
                if job['result'] in ('passed', 'softfailed'):
                    passed += 1
                elif job['result'] == 'none':
                    self.logger.info(f'https://openqa.suse.de/tests/{job["id"]} pending')
                    pending += 1
                else:
                    self.logger.warning(f'https://openqa.suse.de/tests/{job["id"]} failed')
                    failed += 1

            if pkg['arch'] in ('ppc64le', 's390x') and failed == 0:
                openqa_passed_packages.append(pkg)
            elif passed > 0 and pending == 0 and failed == 0:
                openqa_passed_packages.append(pkg)
            else:
                self.logger.info(f'openQA did not (yet) pass for {pkg["name"]}: {passed}/{pending}/{failed}')
```

dirkmueller: only testing for "failed == 0" means that you can't distinguish between pending != 0 and pending == 0 (which means not yet scheduled). the boolean finds that there was at least one job. 

you can emulate the boolean by testing for `passed + failed + pending` if you think we need to save on variables.

gocd/bci_repo_publish.py
okurz: how about something like

```
for i in range(7):
    if self.is_repo_published(pkg['publish_prj'], 'images'):
        break
    …
    time.sleep(20 * i**2)
```

gocd/bci_repo_publish.py
Vogtinator: `if pkg['arch'] in ('ppc64le', 's390x') and failed == 0 and pending == 0:`?

dirkmueller: that detects "no job for ppc64 or s390x" but you don't detect whether its because openqa hasn't scheduled it yet or because openqa will never schedule it. 

thats why I inverted the logic to check for mandatory arches (defined as aarch64 and x86_64) and only move forward with releasing s390x and ppc64 if the mandatory arches have passed *and* there are no fails and no pending for ppc64 and s390x.

PR #3035, opened 2023-11-29
Rewrite bs_mirrorfull as osclib component

Vogtinator:
This has multiple benefits:
* Easier to interface with the python code that calls it
* Uses the osc python interface for connection and authentication handling
* Allows to drop the bundled OBS code in the bs_copy directory

Draft because not tested as part of pkglistgen and installcheck yet.


Vogtinator:
`project-installcheck.py`, `staging-installcheck.py` and `pkglistgen.py` work here

PR #3036, opened 2023-11-29
ttm: Allow bigger images for livecd-x11 on aarch64

ggardet:
since no CD will be used, but usb sticks

### DIFF
--- a/ttm/releaser.py
+++ b/ttm/releaser.py
@@ -81,15 +81,15 @@ class ToTestReleaser(ToTestManager):
         return self.iso_build_version(self.project.name, self.project.image_products[0].package,
                                       arch=self.project.image_products[0].archs[0])
 
-    def maxsize_for_package(self, package):
+    def maxsize_for_package(self, package, arch):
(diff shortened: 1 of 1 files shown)


ttm/releaser.py
```diff
@@ -81,15 +81,15 @@ def version_from_project(self):
...
+    def maxsize_for_package(self, package, arch):
         if re.match(r'.*-mini-.*', package):
             return 737280000  # a CD needs to match
 
         if re.match(r'.*-dvd5-.*', package):
             return 4700372992  # a DVD needs to match
 
-        if re.match(r'livecd-x11', package):
-            return 681574400  # not a full CD
+        if re.match(r'livecd-x11', package) and not re.match(r'aarch64', arch):
```
Vogtinator, line 91: At this point let's make it more explicit:

```suggestion
        if re.match(r'livecd-x11', package) and arch == 'x86_64':
```

ggardet: OK

Vogtinator (approved):

PR #3039, opened 2023-12-05
factory-package-news.py: Compress data files with zstd

Vogtinator:
It's less than 1/3rd of the uncompressed size.

### DIFF
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -57,6 +57,8 @@ Requires:       python3-lxml
 Requires:       python3-pycurl
 Requires:       python3-python-dateutil
 Requires:       python3-pyxdg
+# factory-package-news.py
+Requires:       python3-pyzstd
 Requires:       python3-requests
 # typing extensions are needed on SLE & Leap
 %if 0%{?suse_version} <= 1500
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -58,7 +58,7 @@ Requires:       python3-pycurl
 Requires:       python3-python-dateutil
 Requires:       python3-pyxdg
 # factory-package-news.py
-Requires:       python3-pyzstd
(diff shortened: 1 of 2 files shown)


dist/package/openSUSE-release-tools.spec
```diff
@@ -57,6 +57,8 @@ Requires:       python3-lxml
 Requires:       python3-pycurl
 Requires:       python3-python-dateutil
 Requires:       python3-pyxdg
+# factory-package-news.py
+Requires:       python3-pyzstd
```
DimStar77, line 61: this seems not to be available on ariel

factory-package-news/factory-package-news.py
```diff
@@ -27,6 +29,17 @@ def utf8str(content):
     return str(content, 'utf-8') if isinstance(content, bytes) else content
 
 
+# Older versions wrote uncompressed files, support both for now.
+def openZstdOrPlain(path: str):
+    f = open(path, 'rb')
+    magic = f.read(4)
```
Martchus, line 35: If it is impossible that a valid file is < 4 byte then this is ok. Otherwise I'd check the file size first.

factory-package-news/factory-package-news.py
```diff
@@ -27,6 +29,17 @@ def utf8str(content):
     return str(content, 'utf-8') if isinstance(content, bytes) else content
 
 
+# Older versions wrote uncompressed files, support both for now.
+def openZstdOrPlain(path: str):
```
Martchus, line 33: Not that I really care but I guess the coding style used here would be:

```suggestion
def open_zstd_or_plain(path: str):
```

factory-package-news/factory-package-news.py
Vogtinator: At least here it is

factory-package-news/factory-package-news.py
Vogtinator: Done

PR #3041, opened 2023-12-12
check_source.py: Use correct function for getting maintainer of the devel pkg

Vogtinator:
"maintainers_get" is a bit weird and uses /search/owner and not the given prj.

Noticed in https://build.opensuse.org/request/show/1132698.

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -896,7 +896,7 @@ class CommandLineInterface(cmdln.Cmdln):
         conf.get_config(override_apiurl=self.options.apiurl)
 
         if (self.options.osc_debug):
-            conf.config['debug'] = 1
+            conf.config['debug'] = True
 
         self.checker = self.setup_checker()
         if self.options.config:
--- a/check_source.py
+++ b/check_source.py
@@ -22,7 +22,7 @@ from osclib.core import entity_exists
 from osclib.core import group_members
 from osclib.core import package_kind
 from osclib.core import create_add_role_request
-from osclib.core import maintainers_get
+from osclib.core import package_role_expand
 from osc.core import show_package_meta, show_project_meta
 from osc.core import get_request_list
 from urllib.error import HTTPError
@@ -300,7 +300,7 @@ class CheckSource(ReviewBot.ReviewBot):
             devel_project, devel_package = devel_project_fallback(self.apiurl, target_project, target_package)
             if devel_project and devel_package:
                 submitter = self.request.creator
-                maintainers = set(maintainers_get(self.apiurl, devel_project, devel_package))
+                maintainers = set(package_role_expand(self.apiurl, devel_project, devel_package))
                 known_maintainer = False
                 if maintainers:
                     if submitter in maintainers:


DimStar77:
impressive this did not blow up much more

Vogtinator:
> `maintainers_get` seems to be doing what one would expect:

In which way?

DimStar77:
> > `maintainers_get` seems to be doing what one would expect:
> 
> In which way?

it reads the meta and finds the maintainer role groupid and userid

Vogtinator:
> > > `maintainers_get` seems to be doing what one would expect:
> > 
> > 
> > In which way?
> 
> it reads the meta and finds the maintainer role groupid and userid

It doesn't, it uses `/search/owner`

DimStar77:
See pasted code in https://github.com/openSUSE/openSUSE-release-tools/pull/3041#issuecomment-1852302091

there is no search/owner in there

DimStar77:
```
> grep maintainers_get . -r
./osclib/core.py:def maintainers_get(apiurl, project, package=None):
./check_maintenance_incidents.py:from osclib.core import maintainers_get
./check_maintenance_incidents.py:        maintainers = set(maintainers_get(self.apiurl, project, pkgname))
./devel-project.py:            userids = maintainers_get(apiurl, devel_project, devel_package)
./devel-project.py:def maintainers_get(apiurl, project, package=None):
./devel-project.py:        return maintainers_get(apiurl, project)
./devel-project.py:    userids = sorted(maintainers_get(apiurl, project, package))
./ReviewBot.py:from osclib.core import maintainers_get
./ReviewBot.py:        maintainers = set(maintainers_get(self.apiurl, project, package))
./check_source.py:from osclib.core import maintainers_get
./check_source.py:                maintainers = set(maintainers_get(self.apiurl, devel_project, devel_package))
```

Let's try to clean stuff up then... we have multiple consumers of mainers_get (and apparently also multiple provides.. yay)
I'd argue that if maintainers_get is wrong, it will likely be equally wrong in all cases and deserves to be fixed properly

Vogtinator:
> Let's try to clean stuff up then... we have multiple consumers of mainers_get (and apparently also multiple provides.. yay) I'd argue that if maintainers_get is wrong, it will likely be equally wrong in all cases and deserves to be fixed properly

I expect that for some cases it's not actually wrong but necessary and meant to get the actual "package maintainer" instead of the "technical devel package maintainer"... I don't really get the logic though, especially for those methods containing "fallback" in their names.

nilxam:
To be honest when I working on https://github.com/openSUSE/openSUSE-release-tools/commit/7a088e778459a1a27b3ba035cb5dac07bc43987f I've confused with maintainers_get/owner_fallback a bit, it sometimes gave a weird result, I spent time on debugging then move to other errands after have some findings, if my memory is right, the culprit is `owner()`, the default search mode is `binary` that means it has `binary=livecd-openSUSE` as an param of api call, if uses `mode=package` then it has `package=livecd-openSUSE` in query params. The right fix should be on `maintainers_get()` + `owner_fallback()` to support different mode and send a proper mode to `owner()`.

I'm not sure does default has `mode=binary` makes sense, but as it was there so long probably it works for other scenarios? perhaps maintenance review? if none uses binary mode we can force it to be `mode=package` in `owner_fallback()`

nilxam:
https://github.com/openSUSE/openSUSE-release-tools/pull/3046 can be a rather proper fix here IMO.

nilxam:
`osc maintainer PRJ PKG` triggers `/source/PRJ/PKG/_meta`, indeed use `package_role_expand()` is matches the behavior `osc maintainer` command does.

nilxam (approved):

Vogtinator:
I'll have a look at the other uses of maintainer_* and owner_*, maybe I can come up with something more consistent and less misleading.

PR #3044, opened 2024-01-03
publish bci repo right away on package set changes

dirkmueller:
When there was a source commit on the bci repo, then the set of packages changed, which likely means we need those changes to pass the tests on the container release pipelines. So don't wait for 24h but publish right away.

Also trigger on the oldest published state rather than the newest one so that we do not need to wait another 24h per openQA failure to resume publishing.

### DIFF
--- a/gocd/bci_repo_publish.py
+++ b/gocd/bci_repo_publish.py
@@ -141,9 +141,8 @@ class BCIRepoPublisher(ToolBase.ToolBase):
 
         # If the last published build is less than a day old, don't publish
         oldest_published_mtime = min([int(pkg['published_mtime']) for pkg in packages])
-        published_disturls = {pkg['published_srcmd5'] for pkg in packages}
         published_build_age_hours = int(time.time() - oldest_published_mtime) // (60 * 60)
-        pending_source_changes = published_disturls.difference({pkg['built_srcmd5']for pkg in packages})
+        pending_source_changes = [pkg['name'] for pkg in packages if pkg['built_srcmd5'] != pkg['published_srcmd5']]
         if pending_source_changes:
             self.logger.info(f"Pending source changes to published, skipping waiting for 24h: {pending_source_changes}")
             # If the source commit different than the published version, we
--- a/gocd/bci_repo_publish.py
+++ b/gocd/bci_repo_publish.py
@@ -39,6 +39,16 @@ class BCIRepoPublisher(ToolBase.ToolBase):
 
         raise RuntimeError(f"Failed to get version of {project}/{package}")
 
+    def srcmd5_of_product(self, project, package):
+        """Get the srcmd5 of the given source package."""
+        # convert to the base package (without multibuild flavor)
+        package = package.partition(':')[0]
+        url = makeurl(self.apiurl, ['source', project, package])
+        root = ET.parse(http_GET(url)).getroot()
+        if root.tag == 'directory' and 'srcmd5' in root.attrib:
+            return root.attrib['srcmd5']
+        raise RuntimeError(f"Failed to get srcmd5 of {project}/{package}")
+
     def mtime_of_product(self, project, package, repo, arch):
         """Get the build time stamp of the given product, based on _buildenv."""
         url = makeurl(self.apiurl, ['build', project, repo, arch, package])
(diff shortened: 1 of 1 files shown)


gocd/bci_repo_publish.py
```diff
@@ -126,11 +140,16 @@ def run(self, version, token=None):
...
         # If the last published build is less than a day old, don't publish
-        newest_published_mtime = max([int(pkg['published_mtime']) for pkg in packages])
-        published_build_age_hours = int(time.time() - newest_published_mtime) // (60 * 60)
-        if published_build_age_hours < 24:
-            self.logger.info('Current published build less than a day old '
-                             f'({published_build_age_hours}h).')
+        oldest_published_mtime = min([int(pkg['published_mtime']) for pkg in packages])
+        oldest_published_version = min([pkg['published_version'] for pkg in packages])
+        published_build_age_hours = int(time.time() - oldest_published_mtime) // (60 * 60)
+        if int(float(built_version)) != int(float(oldest_published_version)):
```
Vogtinator, line 146: Those are versions, not floats...

Vogtinator:
Can't you compare the srcmd5 instead of the version?

Vogtinator:
Meanwhile I also wonder whether the delay can be reduced or even removed entirely.

gocd/bci_repo_publish.py
dirkmueller: well, they seem to be in the format <CI_CNT>.<BCNT>, but I can make it use version comparison logic as well

dirkmueller:
> Meanwhile I also wonder whether the delay can be reduced or even removed entirely.

I think it still makes sense in general  for a customer to not receive too many changes (CDN caching etc)

dirkmueller:
> Can't you compare the srcmd5 instead of the version?

good idea. done in the new iteration.

gocd/bci_repo_publish.py
```diff
@@ -39,6 +39,21 @@ def version_of_product(self, project, package, repo, arch):
...
         raise RuntimeError(f"Failed to get version of {project}/{package}")
 
+    def disturl_of_product(self, project, package, repo, arch):
+        """Get the disturl of the given product build, based on the Media1 report."""
+        url = makeurl(self.apiurl, ['build', project, repo, arch, package])
+        root = ET.parse(http_GET(url)).getroot()
+        media1_report = None
+        for binary in root.findall('binary'):
+            if binary.get('filename').endswith('-Media1.report'):
+                media1_report = binary.get('filename')
```
Vogtinator, line 49:
```suggestion
                media1_report = binary.get('filename')
                break
```

Vogtinator:
> > Can't you compare the srcmd5 instead of the version?
> 
> good idea. done in the new iteration.

Does it not work to look at the actual /source/prj/000product (x)srcmd5? The release target is in arch specific projects, so it's arch-specific

gocd/bci_repo_publish.py
dirkmueller: merged

dirkmueller:
> Does it not work to look at the actual /source/prj/000product (x)srcmd5? The release target is in arch specific projects, so it's arch-specific

yes, that should work when looking up on the `/source` route. done now (can't test right now, waiting for builds to finish)

gocd/bci_repo_publish.py
```diff
@@ -126,11 +140,17 @@ def run(self, version, token=None):
...
         # If the last published build is less than a day old, don't publish
-        newest_published_mtime = max([int(pkg['published_mtime']) for pkg in packages])
-        published_build_age_hours = int(time.time() - newest_published_mtime) // (60 * 60)
-        if published_build_age_hours < 24:
-            self.logger.info('Current published build less than a day old '
-                             f'({published_build_age_hours}h).')
+        oldest_published_mtime = min([int(pkg['published_mtime']) for pkg in packages])
+        published_disturls = {pkg['published_srcmd5'] for pkg in packages}
+        published_build_age_hours = int(time.time() - oldest_published_mtime) // (60 * 60)
+        pending_source_changes = published_disturls.difference({pkg['built_srcmd5']for pkg in packages})
```
Vogtinator, line 146:
```suggestion
        oldest_published_mtime = min([int(pkg['published_mtime']) for pkg in packages])
        published_build_age_hours = int(time.time() - oldest_published_mtime) // (60 * 60)
        pending_source_changes = [pkg for pkg in packages if ['built_srcmd5'] != pkg['published_srcmd5']]
```

Vogtinator (approved):
+1 if it works in practice

dirkmueller:
> +1 if it works in practice

I'll test it and then merge it

PR #3046, opened 2024-01-09
Support package search in owner_fallback() and search package instead of binary in check_source

nilxam:
`owner()` supports two search mode: binary and package, the default one is `binary`, in some cases we should search `package` instead like for source checker.

The difference in API call is `/search/owner?package=xxx` vs. `/search/owner?binary=xxx`, try with livecd-openSUSE on binary search mode it gives empty collection since there is no binary matches livecd-openSUSE, with the package search mode, it gives the right result as searched by package name.

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -120,17 +120,19 @@ def convert_from_osc_et(xml):
 
 
 @memoize(session=True)
-def owner_fallback(apiurl, project, package):
-    root = owner(apiurl, package, project=project)
+def owner_fallback(apiurl, project, package, search_mode='binary'):
+    # binary is the default search mode in owner()
+    root = owner(apiurl, package, project=project, mode=search_mode)
     entry = root.find('owner') if root else None
     if not entry or project.startswith(entry.get('project')):
         # Fallback to global (ex Factory) maintainer.
-        root = owner(apiurl, package)
+        root = owner(apiurl, package, mode=search_mode)
     return convert_from_osc_et(root)
 
 
 @memoize(session=True)
-def maintainers_get(apiurl, project, package=None):
+def maintainers_get(apiurl, project, package=None, search_mode='binary'):
+    # owner() supports two search mode: binary and package
     if package is None:
         meta = ET.fromstringlist(show_project_meta(apiurl, project))
         maintainers = meta.xpath('//person[@role="maintainer"]/@userid')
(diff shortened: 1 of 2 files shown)


Vogtinator:
How does this relate to https://github.com/openSUSE/openSUSE-release-tools/pull/3041?

Vogtinator:
IMO even that is wrong in some cases. Given prj and pkg, there are three ways to get "maintainers":

1. `/source/prj/pkg/_meta` with fallbacks to prj, parent prj, ...
2. `/search/owner?package=pkg`
3. `/search/owner?binary=pkg` (does not make sense at all?)

And there's also `/search/owner?project=prj&package=pkg` which might or might not be equivalent to 1?

osclib/core.py
```diff
@@ -120,17 +120,19 @@ def convert_from_osc_et(xml):
...
         # Fallback to global (ex Factory) maintainer.
-        root = owner(apiurl, package)
+        root = owner(apiurl, package, mode=search_mode)
     return convert_from_osc_et(root)
 
 
 @memoize(session=True)
-def maintainers_get(apiurl, project, package=None):
+def maintainers_get(apiurl, project, package=None, search_mode='binary'):
+    # owner() supports two search mode: binary and package
```
eduardoj, line 135:
```suggestion
    # owner() supports two search modes: binary and package
```

nilxam:
> IMO even that is wrong in some cases. Given prj and pkg, there are three ways to get "maintainers":
> 
> 1. `/source/prj/pkg/_meta` with fallbacks to prj, parent prj, ...
> 2. `/search/owner?package=pkg`
> 3. `/search/owner?binary=pkg` (does not make sense at all?)

Perhaps it makes sense in other usecases, like maintenance to check does the binary is maintained. That is not the case for check_source though. This change doesn't change the behavior for other usecases, only check_sources going to use `mode=package`.

> 
> And there's also `/search/owner?project=prj&package=pkg` which might or might not be equivalent to 1?

It does. It give a result of prj, parents prj until the top one per the given `prj`.

Interesting that, `osc maintainer PKG` triggers `/search/owner?package=PKG`, however `osc maintainer PRJ PKG` triggers `/source/PRJ/PKG/_meta`, it sounds like if a particular PRJ was to give(like in check_sources we find devel project already) then use [`package_role_expand()`](https://github.com/openSUSE/openSUSE-release-tools/pull/3041) is a better option here.

nilxam:
lets move back to https://github.com/openSUSE/openSUSE-release-tools/pull/3041 then

Vogtinator:
The use case is too different between global package search, global binary search and prj/pkg lookup, so IMO separate methods would be clearer.

PR #3047, opened 2024-01-19
staging: add new option --new to list

ana:
Add option to staging list to only show unstaged packages that are not ignored.

I'd prefer that by default ignored packages are not shown, so I'm happy to change it.


Vogtinator (commented):
IMO `new` isn't quite clear and something like `exclude-ignored` is more explicit.

OTOH showing ignored SRs in a separate section might also be an option, but would require more code changes.

ana:
> OTOH showing ignored SRs in a separate section might also be an option, but would require more code changes.

I like more this idea, I'll submit a new PR showing first the ignored SR and later the output as usual.

PR #3052, opened 2024-02-05
Add IPv4 and IPv6 handling in one file

witekest:
Access logs for both protocls are now stored in a single file. The change refactors the code to handle it properly. 
InfluxDB client has been upgraded.

### DIFF
--- a/metrics/access/aggregate.php
+++ b/metrics/access/aggregate.php
@@ -1,11 +1,12 @@
 #!/usr/bin/php
 <?php
 
-use InfluxDB\Point;
-use InfluxDB\Database;
+use InfluxDB2\Client;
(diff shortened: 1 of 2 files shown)


metrics/access/aggregate.php
```diff
@@ -53,9 +56,11 @@ function cache_init()
     foreach (PROTOCOLS as $protocol) {
       mkdir("$CACHE_DIR/$protocol", 0755, true);
     }
+    mkdir("$CACHE_DIR/ipv4+6", 00755, true);
```
Vogtinator, line 59:
```suggestion
    mkdir("$CACHE_DIR/ipv4+6", 0755, true);
```

PR #3058, opened 2024-02-17
SLSA support

gyr:
Due to SLSA constrains pkglistgen.py cannot run on botmaster using gocd, as it must run on a separate and isolated server the following changes were added:
- create scripts that run pkgelistgen.py emulating the gocd pipelines (RelPkg and Update.000product)
- create systemd services to execute the scripts for multiple SLSA project
- add log support
- create a rpm package that will be used to deploy it on SLSA server

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -83,7 +83,7 @@ jobs:
         run: |
           for f in $(find . -maxdepth 1 -type f -executable -print); do
               # skip completely broken scripts or those without --help
-              [[ " ./checknewer.py ./repo2fileprovides.py ./openqa-maintenance.py ./docker_publisher.py ./publish_distro ./findfileconflicts ./write_repo_susetags_file.pl ./issue-diff.py " =~ "$f" ]] || "$f" --help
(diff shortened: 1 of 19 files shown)


nilxam:
the test is failed.

nilxam (approved):
I'm fine with the change. But we might need to wait https://github.com/openSUSE/openSUSE-release-tools/pull/3062 I assuming that, otherwise the packaging test can not be succesful if I understand that right...

gleidi-suse (approved):
lgtm

gyr:
> I'm fine with the change. But we might need to wait #3062 I assuming that, otherwise the packaging test can not be succesful if I understand that right...

All checks passed now, package built even with #3062 not being merged.

DimStar77:
somebody forgot about RPM 4.19:

```
can't install openSUSE-release-tools-slsa-build-service-20240315.b086253-1.1.noarch:
nothing provides group(osrt-slsa) needed by openSUSE-release-tools-slsa-build-service-20240315.b086253.noarch
nothing provides user(osrt-slsa) needed by openSUSE-release-tools-slsa-build-service-20240315.b086253.noarch
```

either the user is generated properly using sysusers.d or it is at least advertised in the pkg meta
https://en.opensuse.org/openSUSE:Packaging_guidelines#Users_and_Groups

PR #3059, opened 2024-02-19
build-fail-reminder: don't escape characters

tacerus:
osc uses the string literally, make the generated log entry more readable by using the correct characters instead of a character reference.

Previous behavior were  commit messages showing character references which are not useful for humans:

```
$ osc log -D openSUSE:Factory kopete |head -4
----------------------------------------------------------------------------
r134 | anag+factory | 2024-02-08 18:02:03 | 8b84f116348051059ea6ac6b364667a7 | 23.08.4 | rq1144747

[botdel] Package has had build problems for &gt;= 6 weeks
```

Change tested using:

```
r = osc.core.Request()
r.add_action('delete', tgt_project='home:crameleon:ignore', tgt_package='dummy2')
r.description = 'foo >= 1'
r.create('https://api.opensuse.org')

$ osc request list home:crameleon:ignore dummy2
1147567  State:new        By:crameleon    When:2024-02-19T12:18:04
        Created by: crameleon
        delete:                                                             home:crameleon:ignore/dummy2
        Descr: foo >= 1
 

$ osc request log 1147567
...
----------------------------------------------------------------------------
created | crameleon | 2024-02-19T12:18:04

foo >= 1

$  osc log -D home:crameleon:ignore dummy2|head -n4
----------------------------------------------------------------------------
r2 | crameleon | 2024-02-19 12:28:43 | 83b58530a64eab9b3792d6eec7fbae4c | unknown | rq1147567

foo >= 1
```

### DIFF
--- a/build-fail-reminder.py
+++ b/build-fail-reminder.py
@@ -242,7 +242,7 @@ def main(args):
                 # Package failed to build for 6 weeks - file a delete request
                 r = osc.core.Request()
                 r.add_action('delete', tgt_project=project, tgt_package=package)
-                r.description = "[botdel] Package has had build problems for &gt;= 6 weeks"
+                r.description = "[botdel] Package has had build problems for >= 6 weeks"
                 r.create(apiurl)
 
     if len(ProjectComplainList):


Vogtinator:
It's not clear whether the SR description is meant to be HTML/markdown or plain text.

Currently osc displays it as plain text while the Web UI displays it as HTML/MD. This means that it's going to be wrong in either the CLI or Web UI, it'll never look correct everywhere.

Currently it looks correct in the web UI which is probably what most users use, also because notification mails link to it, so IMO the best option for now.

However, there is a way to make it work for both!

> r.description = "[botdel] Package has failed to build for more than 6 weeks"

build-fail-reminder.py
```diff
@@ -242,7 +242,7 @@ def main(args):
                 # Package failed to build for 6 weeks - file a delete request
                 r = osc.core.Request()
                 r.add_action('delete', tgt_project=project, tgt_package=package)
-                r.description = "[botdel] Package has had build problems for &gt;= 6 weeks"
+                r.description = "[botdel] Package has had build problems for >= 6 weeks"
```
Vogtinator, line 245:
```suggestion
                r.description = "[botdel] Package has failed to build for at least 6 weeks"
```

Vogtinator (changes requested):

build-fail-reminder.py
DimStar77: We might even change the wording - as the bot also files del req for stuff 'failing to install' for 6 weeks (originally, the bot only did buildfail errors, but has evolved to also act on installcheck issues)

Vogtinator: It should probably also mention the reason in that case

tacerus:
@Vogtinator My variant looks fine in both, the shell as seen in my output, and the web UI as well: https://build.opensuse.org/request/show/1147567.

Vogtinator (approved):

tacerus:
Good with me. ;-)

PR #3060, opened 2024-02-19
Call rabbit-repoid.py for openSUSE:Slowroll

asdil12:
Ticket: https://progress.opensuse.org/issues/135731


asdil12:
@Vogtinator please have a look

Issue #1885, opened 2019-03-05
repo-checker does not handle indirect conflicts properly


Vogtinator:
`kile-lang` requires `kile`
`kile5-lang` requires `kile5`
`kile5` conflicts `kile`

This means `kile-lang` and `kile5-lang` can't be installed together on the same system.
repo-checker doesn't realize that and declines the `kile5` review as the `-lang` packages have file conflicts.

coolo:
Yeah. And it's unlikely to change.

lnussel:
so wontfix?

coolo:
Unless a complete rewrite of the fileconflicts script comes along, yes.

gonzoleeman:
so having everyone modify SPEC files is easier, for the company, than re-writing a script?

coolo:
Feel free to give it a try. and it's not 'everyone' - there have been 2 cases in 2 years.

gonzoleeman:
In that case, I assume there's a relatively easy method to override the check when it gives a false positive.

Vogtinator:
Shuld be fixed by https://github.com/openSUSE/openSUSE-release-tools/pull/3063

PR #3063, opened 2024-02-22
osclib/repochecks.py: Use libsolv for checking conflicts between packages

Vogtinator:
Don't rely just on findfileconflicts to check whether packages have
conflicting dependencies, but use libsolv to do a full check.

Fixes https://github.com/openSUSE/openSUSE-release-tools/issues/1885

### DIFF
--- a/findfileconflicts
+++ b/findfileconflicts
@@ -446,8 +446,8 @@ for my $tc (sort keys %tocheck) {
       my @sp1 = split(' ', $p1);
       my @sp2 = split(' ', $p2);
       print "- between:\n";
-      print "  - [$sp1[0], $sp1[1], $sp1[2], $sp1[3]]\n";
-      print "  - [$sp2[0], $sp2[1], $sp2[2], $sp2[3]]\n";
+      print "  - [$sp1[0], '$sp1[1]', '$sp1[2]', $sp1[3]]\n";
+      print "  - [$sp2[0], '$sp2[1]', '$sp2[2]', $sp2[3]]\n";
       print "  conflicts: |-\n";
       print "    $_\n" for (@con);
     }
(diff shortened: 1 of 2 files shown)


Vogtinator:
```
[15:23] <fvogt> DimStar: Please have a look at https://github.com/openSUSE/openSUSE-release-tools/pull/3063. Running it against adi:26 shows that only two actual conflicts remain
[15:23] <|Anna|> Github project openSUSE/openSUSE-release-tools pull request#3063: "osclib/repochecks.py: Use libsolv for checking conflicts between packages", created on 2024-02-22, status: open, unassigned, https://github.com/openSUSE/openSUSE-release-tools/pull/3063
[15:26] <DimStar> does it behave properly when it SHOULD detect conflicts? (e.g adi:27, adi:164), tested against Letter (with and without conflicts?)
[16:22] <fvogt> DimStar: I'll run it against both
[16:23] <fvogt> found conflict of ispell-sq-3.4.06-2.1.x86_64 with sequoia-sq-0.33.0-1.1.x86_64
[16:24] <fvogt> Do we have a letter with conflicts?
[16:24] <fvogt> If not, the PR is a noop as it only runs in the case where it would've failed previously
[16:25] <fvogt> found conflict of busybox-udhcpc-1.36.1-31.2.noarch with udhcp-0.9.8-1.1.x86_64
[16:25] <fvogt> So works as expected for both adis
[16:29] <DimStar> sounds good.. let's be brave and merge :)
[16:39] <fvogt> At first I assumed it would be too slow to handle it like this but it's not
[16:39] <fvogt> For adi:26 it takes an additional 2s
[16:41] <fvogt> Of which 1.3s are the initial setup and every conflict takes ~15ms
```

PR #3064, opened 2024-03-07
Create .changes files for reproducible builds

JanZerebecki:
release rpm files currently do not have .changes files, but these are needed to extract SOURCE_DATE_EPOCH for reproducible builds.

Using the time stamp of the last commit to the input, which is the 000package-groups source package, as the time for the changes entry would be reproducible, however there is resistance to that.

The other date that would is somewhat related is the time of the last source change for any input to the whole distribution, however AFAIK this date is not cheap enough to compute for OBS-VCS. This hopefully changes once we move to git. But I think the change date of 000package-groups makes more sense conceptually, because AFAIK that is the only information that can causes a change.

However to not delay making progress with reproducible builds in Factory, I propose this change which is non-reproducible, by using the current wall-clock time. We can revisit this later.

See https://reproducible-builds.org/ for more general information on this topic.

This is an alternative for: https://github.com/openSUSE/openSUSE-release-tools/pull/3062


DimStar77:
>The other date that would is somewhat related is the time of the last source change for any input to the whole distribution, however AFAIK this date is not cheap enough to compute for OBS-VCS. This hopefully changes once we move to git. But I think the change date of 000package-groups makes more sense conceptually, because AFAIK that is the only information that can causes a change.

* The 000package-releases contains the TW snapshot version - i.e is changed daily
* The weakremover.inc is updated by a bot and dynamically changes whenever a package is removed - without changes in 000package-groups
i.e the date of 000package-groups is in no way the only relevant change to the -releases package

Vogtinator (changes requested):
This is missing the "only if the output changes" part which is important otherwise it would commit changes on every run

JanZerebecki:
Now it should only change the changes files if there is something else to commit.

Note: after this is merged, the %changelog in the spec.in files in 000package-groups probably needs to be removed before the following run.

pkglistgen/tool.py
```diff
@@ -743,6 +745,22 @@ def update_and_solve_target(
 
         file_utils.multibuild_from_glob(release_dir, '*.spec')
         self.build_stub(release_dir, 'spec')
+
+        todo_spec_files = []
+        package = Package(release_dir)
+        if package.get_status():
```
Vogtinator, line 751: This looks a bit weird. Is that really the only way in the osc API to figure out if there are changes?

It might have to filter out `' '` for unchanged files.

JanZerebecki: There are many ways to see if there are changes. But this seemed a straight forward way to me. By my reading get_status returns an empty list [] if there are no changes, and otherwise one tuple of (status, filename) per changed file.

JanZerebecki: So where would there be a space to filer out?

Vogtinator: FWICT it returns `(' ', filename)` if unchanged and not part of `exclude_states`

Vogtinator: osc itself uses `p.get_status(False, ' ')` in various places

JanZerebecki: Ah. Changed.

Vogtinator (approved):
LGTM. Let's try this next week

PR #3075, opened 2024-04-29
gocd: microos: Completely switch to product composer.

gleidi-suse:


g7 (approved):
Changes are good to me, failing tests look unrelated

PR #3077, opened 2024-05-07
Use f-strings where possible

dirkmueller:


dirkmueller:
This was mechanically generated using flynt.

PR #3078, opened 2024-05-13
pkglistgen: tool: decouple summary dict creation

g7:
The summary dictionary used to create summary-{scope}.txt was
previously returned by solve_project() via write_all_groups().

After c46dd3e304d8e66e5f8f6a51bfa3ebda8cba8936, solve_project()
doesn't return anything anymore, as write_all_groups() is being
invoked just after solve_project().

Thus, the "summary" dictionary would always be None.

Since the summary dict creation is generic enough, move it
to its own function, and call it afterwards after project
solving.

This commit also drops the return statement both on write_all_groups()
and write_productcompose() (where it would always return an empty
dictionary), and moves the write_productcompose() call after
solve_project() as well to match the change with write_all_groups().

### DIFF
--- a/pkglistgen/tool.py
+++ b/pkglistgen/tool.py
@@ -138,7 +138,6 @@ class PkgListGen(ToolBase.ToolBase):
 
     def write_productcompose(self):
         self._check_supplements()
-        summary = dict()
         archs = ['*'] + self.all_architectures
         # a single file covering all builds via multibuild flavors
         with open(os.path.join(self.productcompose_dir, 'default.productcompose'), 'a') as opfh:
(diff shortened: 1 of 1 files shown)


Vogtinator:
`write_productcompose` also returns a summary, but this is lost by `solve_project`. Not sure how this is meant to behave in case only product composer or both are used.

g7:
On a side-note: I wonder if moving 

```
        if not self.skip_productcompose:
            self.write_productcompose()
```

from the bottom of `solve_project()` to the same place `self.write_all_groups()` is called would be okay just to keep things coherent.

Vogtinator:
> On a side-note: I wonder if moving
> 
> ```
>         if not self.skip_productcompose:
>             self.write_productcompose()
> ```
> 
> from the bottom of `solve_project()` to the same place `self.write_all_groups()` is called would be okay just to keep things coherent.

The core question is still

> `write_productcompose` also returns a summary, but this is lost by `solve_project`. Not sure how this is meant to behave in case only product composer or both are used.

i.e. should it take one over the other (maybe validate equality?) or keep both?

g7:
@Vogtinator looks like `write_productcompose()` would always return an empty dictionary (perhaps a leftover?) so I think that's not used right now.

Vogtinator:
> @Vogtinator looks like `write_productcompose()` would always return an empty dictionary (perhaps a leftover?) so I think that's not used right now.

@adrianschroeter ^?

gleidi-suse (approved):
I tried this for https://github.com/openSUSE/openSUSE-release-tools/pull/3079 and it worked for me

g7:
@Vogtinator are you ok if we merge this now in the meantime? It will unblock packagelist building

g7:
v2: after Fabian's suggestions, decoupled the summary dict creation from write_all_groups()/write_productcompose(). Please see the commit description for more details.

Vogtinator (approved):

PR #3049, opened 2024-01-23
Don't abort if weakremover.inc don't exist on source project

fcrozat:
For ALP, weakremover.inc might not be part of the staging parent project. Just ignore 404 error in that case.

PR #3083, opened 2024-05-15
Don't abort if weakremover.inc don't exist on source project

fcrozat:
For ALP, weakremover.inc might not be part of the staging parent project. Just ignore 404 error in that case.


fcrozat:
replace https://github.com/openSUSE/openSUSE-release-tools/pull/3049 (created from wrong branch)

Vogtinator (approved):

PR #3084, opened 2024-05-16
Support msgid domain override

tacerus:
The domain set by email.utils defaults to the one from the calling system. Systems having theirs set to "localhost" risk increased spam scoring. Allow the override of the domain on such systems using a configuration option.

Note: please test this yourself, I could not find a fitting YAML configuration file in this repository and don't know where you are executing this script from. I only confirmed the `domain` option to work by copying the message and SMTP lines into a short sample Python script, which executed and sent the mail as expected.

### DIFF
--- a/factory-package-news/announcer.py
+++ b/factory-package-news/announcer.py
@@ -130,7 +130,7 @@ msg['From'] = config['sender']
 msg['To'] = config['to']
 msg['Mail-Followup-To'] = config['to']
 msg['Date'] = email.utils.formatdate(localtime=1)
-msg['Message-ID'] = email.utils.make_msgid()
+msg['Message-ID'] = email.utils.make_msgid(domain=config.get(domain))
 
 if options.dry:
     print("sending ...")


Vogtinator:
According to the documentation, this uses the system's hostname by default. Apparently that's already broken.

Vogtinator:
> According to the documentation, this uses the system's hostname by default. Apparently that's already broken.

@sysrich can you check what's set in the used botmaster agent and maybe adjust it?

Vogtinator:
> > According to the documentation, this uses the system's hostname by default. Apparently that's already broken.
> 
> @sysrich can you check what's set in the used botmaster agent and maybe adjust it?

@sysrich ping ^

FWICT this should basically work OOTB, even with the default docker and podman container settings which just use an autogenerated hostname.

DimStar77:
https://github.com/openSUSE/openSUSE-release-tools/pull/3087 was merged and helped to get Richard's mails through the spam filter

PR #3089, opened 2024-05-22
pkglistgen: Support working on git branches

adrianschroeter:
Following scmsync syntax where a branch can be specified behind the URL fragment hash

### DIFF
--- a/pkglistgen/tool.py
+++ b/pkglistgen/tool.py
@@ -696,17 +696,24 @@ class PkgListGen(ToolBase.ToolBase):
                 logging.info(f'{project}/{product} build in progress')
                 return
         if git_url:
+            git_url_base, *fragment = git_url.split('#')
             if os.path.exists(cache_dir + "/.git"):
                 # reset and update existing clone
                 logging.debug(subprocess.check_output(
                     ['git', 'reset', '--hard'], cwd=cache_dir, encoding='utf-8'))
                 logging.debug(subprocess.check_output(
                     ['git', 'pull', '--rebase'], cwd=cache_dir, encoding='utf-8'))
+                if fragment:
+                    # FIXME: this is not takeing account default branches, which may change server side
+                    logging.debug(subprocess.check_output(
+                        ['git', 'switch', fragment[0]], cwd=cache_dir, encoding='utf-8'))
             else:
                 if os.path.exists(cache_dir):
                     shutil.rmtree(cache_dir)
-                logging.debug(subprocess.check_output(
-                    ['git', 'clone', '--recurse-submodules', git_url, cache_dir], encoding='utf-8'))
+                args = ['git', 'clone', '--recurse-submodules', git_url_base, cache_dir]
+                if fragment:
+                    args += ['--branch', fragment[0]]
+                logging.debug(subprocess.check_output(args, encoding='utf-8'))
             if os.path.exists(cache_dir + '/' + '000update-repos'):
                 logging.error('No support for 000update-repos in git projects atm')
                 return


pkglistgen/tool.py
```diff
@@ -703,8 +703,13 @@ def update_and_solve_target(
             else:
                 if os.path.exists(cache_dir):
                     shutil.rmtree(cache_dir)
-                logging.debug(subprocess.check_output(
-                    ['git', 'clone', '--recurse-submodules', git_url, cache_dir], encoding='utf-8'))
+                split = git_url.split('#')
+                git_url = split[0]
+                args = ['git', 'clone', '--recurse-submodules', git_url, cache_dir]
+                if len(split) == 2:
+                    args.append('--branch')
```
Vogtinator, line 710: `--single-branch` as well (unconditionally)?

adrianschroeter: we can do so ... added another commit for that as it is independend of the branch selection.

pkglistgen/tool.py
```diff
@@ -703,8 +703,13 @@ def update_and_solve_target(
...
                     shutil.rmtree(cache_dir)
-                logging.debug(subprocess.check_output(
-                    ['git', 'clone', '--recurse-submodules', git_url, cache_dir], encoding='utf-8'))
+                split = git_url.split('#')
+                git_url = split[0]
+                args = ['git', 'clone', '--single-branch', '--recurse-submodules', git_url, cache_dir]
+                if len(split) == 2:
+                    args.append('--branch')
+                    args.append(split[1])
+                logging.debug(subprocess.check_output(args, encoding='utf-8'))
```
Vogtinator, line 712:
```suggestion
                git_url_base, *fragment = git_url.split('#')
                args = ['git', 'clone', '--single-branch', '--recurse-submodules', git_url_base, cache_dir]
                if fragment:
                    args += ['--branch', fragment[0]]
                logging.debug(subprocess.check_output(args, encoding='utf-8'))
```

Vogtinator:
Should it also handle if the URL (any part) changes while the cache dir already exists?
Currently it'll pull and push to the previous location.

adrianschroeter:
yes, but that can get complex as also the default branch can change server side...

maybe we should do that with a second pull request, to avoid blocking Marco's request.

Vogtinator:
> yes, but that can get complex as also the default branch can change server side...

That only impacts the initial clone though, so I don't see how that would be complex?

If it's ok to discard the cache dir on mismatches, then `git remote get-url origin` and `git branch --show-current` could be used to check for mismatches.

Vogtinator:
Without support for cache invalidation it would currently keep pulling and pushing the wrong URL

adrianschroeter:
I added a 90% switch solution here, but it does not cover the situation when no branch is specified, but default branch switches server side.

I reverted the --single-branch clone option therefore to be able to switch later on.

pkglistgen/tool.py
```diff
@@ -696,17 +696,24 @@ def update_and_solve_target(
...
             if os.path.exists(cache_dir + "/.git"):
                 # reset and update existing clone
                 logging.debug(subprocess.check_output(
                     ['git', 'reset', '--hard'], cwd=cache_dir, encoding='utf-8'))
                 logging.debug(subprocess.check_output(
                     ['git', 'pull', '--rebase'], cwd=cache_dir, encoding='utf-8'))
+                if fragment:
+                    # FIXME: this is not takeing account default branches, which may change server side
+                    logging.debug(subprocess.check_output(
+                        ['git', 'switch', fragment], cwd=cache_dir, encoding='utf-8'))
```
Vogtinator, line 709: `git switch` is experimental and might not be available on ibs-slsa-botmaster

gleidi-suse:
@adrianschroeter @Vogtinator we need this for SLFO stagings. Is there any action to be taken or can this be merged?

pkglistgen/tool.py
lethliel: the git on botmaster knows git switch

lethliel: also git on ibs-slsa-botmaster knows it

adrianschroeter:
it is fine from my side.

Please speak up if something needs to be changed.

Vogtinator (approved):

PR #3091, opened 2024-05-23
pkglistgen: cli: Explicitly enable product-composer instead of detecting directories.

gleidi-suse:
This PR is based on https://github.com/openSUSE/openSUSE-release-tools/pull/3089

### DIFF
--- a/pkglistgen/cli.py
+++ b/pkglistgen/cli.py
@@ -12,6 +12,7 @@ import logging
 from osc import conf
 from osclib.conf import Config
 from osclib.stagingapi import StagingAPI
+from pkglistgen.engine import Engine, ENGINE_NAMES
 from pkglistgen.tool import PkgListGen, MismatchedRepoException
 from pkglistgen.update_repo_handler import update_project
 
@@ -47,6 +48,7 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
     @cmdln.option('-p', '--project', help='target project')
     @cmdln.option('-g', '--git-url', help='git repository for target project')
     @cmdln.option('-s', '--scope', help=f"scope on which to operate ({', '.join(SCOPES)}, staging:$letter)")
+    @cmdln.option('--engine', help=f"engine to be used. Available engines are: {', '.join(ENGINE_NAMES)}", default=Engine.legacy.name)
     @cmdln.option('--no-checkout', action='store_true', help='reuse checkout in cache')
     @cmdln.option('--stop-after-solve', action='store_true', help='only create group files')
     @cmdln.option('--staging', help='Only solve that one staging')
@@ -70,6 +72,12 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
 
         if not opts.scope:
             raise ValueError('--scope or --staging required')
+
+        if opts.engine not in ENGINE_NAMES:
+            raise ValueError(f"Illegal engine: {opts.engine} . Supported engines are {', '.join(ENGINE_NAMES)}.")
+        elif opts.engine == Engine.product_composer.name and opts.only_release_packages:
+            raise ValueError(f"--only-release-packages makes no sense with {Engine.product_composer.name} engine, as it does not handle "
+                             "000release-packages!")
 
         apiurl = conf.config['apiurl']
         Config(apiurl, opts.project)
@@ -95,6 +103,7 @@ class CommandLineInterface(ToolBase.CommandLineInterface):
                 self.tool.dry_run = self.options.dry
                 return self.tool.update_and_solve_target(api, target_project, target_config, main_repo,
                                                          git_url=opts.git_url, project=project, scope=scope,
+                                                         engine=Engine[opts.engine],
                                                          force=opts.force, no_checkout=opts.no_checkout,
                                                          only_release_packages=opts.only_release_packages,
                                                          stop_after_solve=opts.stop_after_solve,
--- a/pkglistgen/tool.py
+++ b/pkglistgen/tool.py
@@ -32,6 +32,7 @@ from osclib.repomirror import RepoMirror
 from urllib.parse import urlparse
 
 from pkglistgen import file_utils
+from pkglistgen.engine import Engine
 from pkglistgen.group import Group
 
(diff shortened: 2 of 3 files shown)


Vogtinator:
Why?

g7:
I've tested this in one of SLE 15 SP6, old behaviour ("legacy engine") works as expected.

gleidi-suse:
I know I said I'd work on a more structured proposal but we need this running on SLFO stagings ASAP.

In a perfect universe we should really get rid of all `if git_url` and  `if use_product_compose` by having different objects encapsulating git/svn and product-builder/product-composer operations exposing the same interface . In any case it is not required to deliver so I'll do that in the future, as they can be implemented in a backward compatible way now that we have `--engine` .

What do you think?

pkglistgen/cli.py
```diff
@@ -71,6 +73,11 @@ def do_update_and_solve(self, subcmd, opts):
         if not opts.scope:
             raise ValueError('--scope or --staging required')
 
+        if opts.engine not in ENGINE_NAMES:
+            raise ValueError(f"Illegal engine: {opts.engine} . Supported engines are {', '.join(ENGINE_NAMES)}.")
+        elif opts.engine == Engine.product_composer.name and opts.only_release_packages:
+            raise ValueError('--only-release-packages make no sense with product_composer engine, as it does not handle 000release-packages!')
```
g7, line 79: nitpick: `make` -> `makes`

pkglistgen/cli.py
```diff
@@ -95,6 +102,7 @@ def solve_project(project, scope: str):
                 self.tool.dry_run = self.options.dry
                 return self.tool.update_and_solve_target(api, target_project, target_config, main_repo,
                                                          git_url=opts.git_url, project=project, scope=scope,
+                                                         use_product_composer=(opts.engine == Engine.product_composer.name),
```
g7, line 105: I wonder if we should simply pass the engine rather than converting it to a fixed `use_product_composer` boolean. Doing comparisons against `Engine.product_composer` or `Engine.legacy` is more clear IMHO.

pkglistgen/tool.py
```diff
@@ -814,10 +818,9 @@ def update_and_solve_target(
                 locales_from=target_config.get('pkglistgen-locales-from')
             )
 
-            if not self.skip_productcompose:
+            if use_product_composer:
                 self.write_productcompose()
-
-            if product_dir:
+            else:
```
g7, line 823: group files aren't used by product-composer AFAIK but nevertheless they were still built in product-compose mode. I think it's nice to still have those around, but I'm fine either way. But still, this is a change of behaviour with what's currently on master. What do you think we should do here?

pkglistgen/tool.py
```diff
@@ -757,28 +759,32 @@ def update_and_solve_target(
                 checkout_package(api.apiurl, project, package, expand_link=True,
                                  prj_dir=cache_dir, outdir=os.path.join(cache_dir, package))
 
-        if not os.path.isdir(product_dir):
+        if use_product_composer:
```
g7, line 762: Since we now check against "product-composer" mode every time rather than looking at directories, I think we can drop this if block.

pkglistgen/tool.py
```diff
@@ -912,7 +915,7 @@ def update_and_solve_target(
         if git_url:
             if product_dir and os.path.isdir(product_dir):
```
g7, line 916: if the two modes are exclusive, I think this if block can be dropped as well.

g7 (commented):
Overall LGTM, with some nitpicks.

~~I want to highlight though the `write_all_groups()` method not being called anymore in product-compose mode. Which is fine to me, but it's currently different with what is in `master` right now.~~

Either way, tested against a SLE15SP6 staging (in "legacy" mode) and the output has been what I expected. Thanks!

pkglistgen/tool.py
gleidi-suse: thank you, agreed

pkglistgen/tool.py
g7: I was wrong on this, re-reading the current code again and the *current* generated artifacts, this looks good already and the group files aren't actually generated already in productcomposer mode. We had .group files in Micro 6.0 but they have been cleaned up. So please ignore my previous comment, sorry for that.

pkglistgen/tool.py
g7: yes, indeed, fine for me

g7 (commented):
Thanks for addressing the comments! :100: 

Aside from the one-line change in my latest comment (`self.product_dir` -> `product_dir`), it works as expected against a SLE15SP6 staging.

I will schedule a full target build so we can check that as well.

pkglistgen/tool.py
```diff
@@ -749,36 +752,38 @@ def update_and_solve_target(
         oldrepos_dir = os.path.join(cache_dir, oldrepos)
 
         self.input_dir = group_dir
-        self.output_dir = product_dir
+        if engine == Engine.product_composer:
+            self.output_dir = self.productcompose_dir
+        elif engine == Engine.legacy:
+            self.output_dir = self.product_dir
```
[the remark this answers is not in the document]
g7: (the replacement of `self.productcompose_dir` IMHO can also be made in a separate pull request, but mentioning it here so we don't forget)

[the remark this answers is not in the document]
gleidi-suse: Thank you for catching that. I should have addressed this , please let me know if it isn't working for you now.

Vogtinator:
> @Vogtinator Could you please give an in depth review to this PR? It was tested on SL Micro 6.1, SLFO Stagings and SLCS 6.0 and it is currently running for SLE 15 SP6 's target. Thank you

Can do, but earliest tomorrow

g7:
Tested on SLE15SP6 target (sorry, it took a rather long time), and the output is as expected.

gleidi-suse:
@Vogtinator were you able to take a look at this PR? Thank you

g7 (approved):
SP6 Staging and Targets work fine, my remarks were addressed, flake8 is happy: looks good to me

dirkmueller (approved):

PR #3097, opened 2024-05-30
gocd: Allow SLFO to have pkglists generated.

gleidi-suse:
This is a temporary change required to generate SLFO pkglists.


g7 (approved):
OK for me

PR #3099, opened 2024-05-31
support committing productcompose files to OBS and not just to git

dcermak:


Vogtinator:
I guess this fails if there is no 000productcompose in the target project?

dcermak:
Fabian Vogt ***@***.***> writes:

> I guess this fails if there is no 000productcompose in the target project?

Whoops, yes. Should be fixed now

dirkmueller:
@dcermak  needs rebase

dcermak:
This has already been implemented via https://github.com/openSUSE/openSUSE-release-tools/pull/3091

PR #3100, opened 2024-05-31
Adjust the staging configuration of ALP to maintenance needs

boiko:
This basically remove things only needed during product development:
* Updating of release packages
* re-generation of 000product content
* Selectively enabling the images repository

### DIFF
--- a/gocd/alp-stagings.gocd.yaml
+++ b/gocd/alp-stagings.gocd.yaml
@@ -1,98 +1,6 @@
 ---
 format_version: 3
 pipelines:
-  ALP.Source.Standard.1_0.Stagings.RelPkgs:
-    environment_variables:
-      OSC_CONFIG: /home/go/config/oscrc-staging-bot
-    group: ALP.Stagings
-    lock_behavior: unlockWhenFinished
-    timer:
-      spec: 0 0 * ? * *
-      only_on_changes: false
-    materials:
-      scripts:
-        git: https://github.com/openSUSE/openSUSE-release-tools.git
-    stages:
-    - Generate.Release.Package:
-        approval: manual
-        jobs:
-          ALP.Source.Standard.1_0.Staging.A:
-            resources:
-              - repo-checker
-            tasks:
-              - script: ./pkglistgen.py -A https://api.suse.de update_and_solve
-                  --staging SUSE:ALP:Source:Standard:1.0:Staging:A
-                  --only-release-packages --force
-          ALP.Source.Standard.1_0.Staging.B:
-            resources:
-              - repo-checker
-            tasks:
-              - script: ./pkglistgen.py -A https://api.suse.de update_and_solve
-                  --staging SUSE:ALP:Source:Standard:1.0:Staging:B
-                  --only-release-packages --force
-          ALP.Source.Standard.1_0.Staging.C:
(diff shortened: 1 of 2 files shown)


Vogtinator:
I'm a bit OOTL but do ALP stagings have openQA? If so, pkglistgen would still be needed if deps change, otherwise the DVD.

I also see that 000productcompose exists and builds, so the package list is still used. Is it managed differently now?

boiko:
Hi Fabian

> I'm a bit OOTL but do ALP stagings have openQA? If so, pkglistgen would still be needed if deps change, otherwise the DVD.

Stagings have openQA, a different set of tests than the ones used during product development. How is `pkglistgen` connected to openQA? As far as I could find out, the openQA tests and results are actually handle by a different bot/pipeline. If not, can you please point me to where pkglistgen and openQA cross their paths?
 
> I also see that 000productcompose exists and builds, so the package list is still used. Is it managed differently now?

Yes, it exists and it will still be used for picking binaries from maintenance updates. But it won't be auto-generated/updated anymore. Any change to that will be done via pull requests directly to the project. In some sense it is similar to what Maintenance does on SLE (the difference being that there are no maintenance channel files to be updated).

Vogtinator:
> Hi Fabian
> 
> > I'm a bit OOTL but do ALP stagings have openQA? If so, pkglistgen would still be needed if deps change, otherwise the DVD.
> 
> Stagings have openQA, a different set of tests than the ones used during product development. How is `pkglistgen` connected to openQA? As far as I could find out, the openQA tests and results are actually handle by a different bot/pipeline. If not, can you please point me to where pkglistgen and openQA cross their paths?

pkglistgen regenerates the package list in 000productcompose, used for building the DVD and repos. When successful, the image repo is build enabled. openQA waits for the repo to finish building and new DVDs and images getting published.

> > I also see that 000productcompose exists and builds, so the package list is still used. Is it managed differently now?
> 
> Yes, it exists and it will still be used for picking binaries from maintenance updates. But it won't be auto-generated/updated anymore. Any change to that will be done via pull requests directly to the project. In some sense it is similar to what Maintenance does on SLE (the difference being that there are no maintenance channel files to be updated).

Ok, so effectively pkglistgen itself is replaced by a manual job. That means you need some alternative for the image repo build enabling. You could enable the repo manually and remove the `packagelists` and `images:enabled` staging checks (see e.g. https://build.suse.de/staging_workflows/SUSE:ALP:Source:Standard:1.0/staging_projects/SUSE:ALP:Source:Standard:1.0:Staging:A)

Vogtinator (approved):

boiko:
Just adding some more context here (for the sake of documentation):

- openQA tests will not be image-based during the maintenance phase, they will do regular or transactional updates from a GA image, so the image repo is not needed on staging (that's why the logic to enable the repo got removed)
- dependency resolution for changes to the product package list will be handle by maintenance tooling and will be submitted as pull requests to the product git repo directly. How the testing of those will be done is still under discussion.
- maintenance tooling would never automatically remove packages from the product, so existing content is preserved. It would only suggest adding missing packages to resolve dependencies.
- the install check step was kept because our own checks are not yet ready to work with stagings, so for now it is helpful to have the original install check validating the packages.

PR #3102, opened 2024-06-10
gocd: sl-micro: Decouple publishing to :PUBLISH project, from publishing to :ToTest.

gleidi-suse:


Vogtinator (approved):

g7 (approved):
Thanks, looks good to me, and might also steal the idea for SP7 ;)

PR #3103, opened 2024-06-10
osclib: freeze_command: handle 'product' repositories if used 

g7:
Some projects (such as SLFO) are using product-composer, and
they build and publish in the 'product' repository.

If the 'product' repository is configured in the project meta,
ensure it gets configured when freezing the staging.

### DIFF
--- a/osclib/freeze_command.py
+++ b/osclib/freeze_command.py
@@ -154,6 +154,12 @@ class FreezeCommand(object):
         for lprj in links:
             ET.SubElement(root, 'link', {'project': lprj})
 
+        # 'images' and 'product' repositories are closely related,
+        # and they need to be handled the same way.
+        image_repositories = ['images']
+        if self.api.project_has_repo('product', self.prj):
+            image_repositories.append('product')
+
         # build flag
         f = ET.SubElement(root, 'build')
         # this one is the global toggle
@@ -161,12 +167,14 @@ class FreezeCommand(object):
         # this one stays
         ET.SubElement(f, 'disable', {'repository': 'bootstrap_copy'})
         # to be flipped by botmaster
-        ET.SubElement(f, 'disable', {'repository': 'images'})
+        for repository in image_repositories:
+            ET.SubElement(f, 'disable', {'repository': repository})
 
         # publish flag
         f = ET.SubElement(root, 'publish')
         ET.SubElement(f, 'disable')
-        ET.SubElement(f, 'enable', {'repository': 'images'})
+        for repository in image_repositories:
+            ET.SubElement(f, 'enable', {'repository': repository})
 
         # debuginfo flag
         f = ET.SubElement(root, 'debuginfo')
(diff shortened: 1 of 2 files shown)


osclib/stagingapi.py
```diff
@@ -144,9 +144,9 @@ def is_staging_manager(self):
             self._is_staging_manager = self.is_user_member_of(self.user, self.cstaging_group)
         return self._is_staging_manager
 
-    def project_has_repo(self, repo_name):
+    def project_has_repo(self, repo_name, project=None):
```
Vogtinator, line 147: I think this is unnecessary, `self.project` is the main project already

osclib/freeze_command.py
```diff
@@ -197,6 +203,22 @@ def prj_meta_for_bootstrap_copy(self):
...
+                a = ET.SubElement(r, 'arch')
+                a.text = 'local'
+
+            a = ET.SubElement(r, 'arch')
+            a.text = 'x86_64'
+
+            if 'ppc64le' in self.api.cstaging_archs:
+                a = ET.SubElement(r, 'arch')
+                a.text = 'ppc64le'
+
```
Vogtinator, line 220: One line too much?

osclib/stagingapi.py
g7: on "usual" projects it would have worked great, but SUSE:SLFO:Main unfortunately doesn't have any repo defined as the contents are built inside SUSE:SLFO:Main:Build, so we can only check on another project (like the staging itself)

osclib/freeze_command.py
```diff
@@ -197,6 +203,21 @@ def prj_meta_for_bootstrap_copy(self):
             a = ET.SubElement(r, 'arch')
             a.text = 'ppc64le'
 
+        if with_product:
+            r = ET.SubElement(root, 'repository', {'name': 'product', 'linkedbuild': 'all'})
+            ET.SubElement(r, 'path', {'project': self.prj, 'repository': 'standard'})
+
+            if self.prj.startswith('SUSE:'):
```
gleidi-suse, line 210: do you think it is worth to encapsulate this logic:

```
        r = ET.SubElement(root, 'repository', {'name': 'images', 'linkedbuild': 'all'})
        ET.SubElement(r, 'path', {'project': self.prj, 'repository': 'standard'})

        if self.prj.startswith('SUSE:'):
            a = ET.SubElement(r, 'arch')
            a.text = 'local'
        a = ET.SubElement(r, 'arch')
        a.text = 'x86_64'

        if 'ppc64le' in self.api.cstaging_archs:
            a = ET.SubElement(r, 'arch')
            a.text = 'ppc64le'
```

as it appears to be the same for both the images and product repositories?

g7: Fixed this in v3

g7:
v3:

- rebased on current master
- since both `images` and `product` repositories are closely related, try to reuse the same code to handle both in `prj_meta_for_bootstrap_copy`

gleidi-suse (approved):
lgtm

gleidi-suse:
@Vogtinator I'd merge this if you don't have any objection

gleidi-suse:
@Vogtinator I'm merging this please let us know if you find any regression, thank you very much

PR #3104, opened 2024-06-11
gocd: slfo-stagings: wait for product to be built before enabling images

g7:
During image testing, the artifacts built inside the 'product' repository are used. Let's ensure that the repository is fully built before attempting to enable images.

### DIFF
--- a/gocd/slfo-stagings.gocd.yaml
+++ b/gocd/slfo-stagings.gocd.yaml
@@ -59,6 +59,17 @@ pipelines:
                 exit 1
               fi
 
+    - Build.product:
+        resources:
+          - staging-bot
+        tasks:
+          - script: |-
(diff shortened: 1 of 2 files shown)


dirkmueller (approved):

DimStar77:
This is apparently the one that caused the SLFO pipelines to enter endless loupes.

BotMaster should be either used event-based or time-based, but do not run 'endless' jobs waiting 'forever'

The number of agents is limited and any job running 'forever' (longer than needed) takes agents from the pool away

g7:
argh, sorry about that.

It is a bug that a SLFO staging doesn't have any 'product' repository, but until #3103 is in and everyone runs with the up-to-date staging plugin freezing might drop the repository and cause this.

But it is a good point indeed - I will add a check for the repo existence, and perhaps a timeout for now. And I will think about some smarter way to do the same without waiting.

PR #3108, opened 2024-06-13
verify-build-and-generatelists: Adjust to use --engine and --git-url for SLFO based products.

gleidi-suse:


verify-build-and-generatelists
```diff
@@ -2,16 +2,32 @@
...
 [ ! -d "${LOG_DIR}" ] && mkdir ${LOG_DIR}
 
+
+ibs_to_gitea_product() {
+    printf '%s' "${GITEA_PRODUCTS_URL}$(echo "${PROJECT}" | sed "s/${IBS_PRODUCT_PREFIX}//" | sed -E 's/:/\#/g')"
+}
+
+
+pkglistgen_extra_args() {
+    if echo "${PROJECT}" | grep -Eq "${IBS_PRODUCT_PREFIX}.*" ; then
```
dirkmueller, line 16: technically if you want to check for a prefix you need to anchor the regex, like 
```suggestion
    if echo "${PROJECT}" | grep -Eq "^${IBS_PRODUCT_PREFIX}.*" ; then
```

dirkmueller: I think you could also do something like this which is imho a bit nicer to read: 

```bash

case "$PROJECT" in
    "${IBS_PRODUCT_PREFIX}"*)
        printf '%s' "--engine product_composer --git-url $(ibs_to_gitea_product)"
        ;;
     *)
       printf ''
        ;;
esac

```

Vogtinator: Even nicer:

```
if [[ $PROJECT == "${IBS_PRODUCT_PREFIX}"* ]]
```

Vogtinator: You can change the shebang to `#!/bin/bash` and use bash specific stuff. The host has bash already

gleidi-suse: @Vogtinator I should have addressed your comment, thank you.

@dirkmueller with `=~` from Bash I was able to match explicitly the prefix with an anchor as you suggested.

I didn't change the `if then else` to `case` since there is only one branch for now and I didn't see a real improvement in readability . What about switching once/if we have more than one conditional branch? Is this a blocker in your opinion?

Thank you both for your comments

dirkmueller (approved):

PR #3109, opened 2024-06-14
cleanup_rings: whitelist 'devscripts'

DimStar77:
devscripts has a sub-flavor for checknashisms that is required in Ring1 (for rpmlint)
The flavor "" is technically not required, but removing the package would make the
sub-flavor vanish as well.

### DIFF
--- a/osclib/cleanup_rings.py
+++ b/osclib/cleanup_rings.py
@@ -31,6 +31,8 @@ class CleanupRings(object):
             'vagrant',
             # https://github.com/openSUSE/open-build-service/issues/14129
             'snobol4',
+            # We need devscripts:checkbashism in Ring1, removing the main flavor is not possible
+            'devscripts',
         ]
 
     def perform(self):


DimStar77:
CC @ana - That should help cleanup_rings from wanting to break ring1

@Vogtinator Maybe we could teach the bot that package FOO:FLAVOR is needed, then removing FOO is a really bad idea ? :)

DimStar77:
At best the bot could mention that FOO is not needed and recommend to `excludebuild` it via prjconf - in some cases that might be a valid option (with the known risks of course)

ana:
wfm

Vogtinator:
> @Vogtinator Maybe we could teach the bot that package FOO:FLAVOR is needed, then removing FOO is a really bad idea ?

It should already do that: https://github.com/openSUSE/openSUSE-release-tools/blob/9d63fee9d5db1490c7748d735ebbb538ac73607a/osclib/cleanup_rings.py#L211

This might fix it:

```diff
-if src.startswith(f"{mainpkg}:"):
+if src == mainpkg or src.startswith(f"{mainpkg}:"):
```

> At best the bot could mention that FOO is not needed and recommend to excludebuild it via prjconf - in some cases that might be a valid option (with the known risks of course)

It does mention to exclude specific multibuild flavors for Ring 0 already.

dirkmueller (approved):

PR #3111, opened 2024-06-17
slfo-stagings: add a timeout for Build.product, and do not run if repo is missing

g7:
Ensure that the 'product' repo is configured in the project meta.

Unfortunately, since we're checking in the configuration of the staging
itself, rather than the parent project's, there are increased chances of
some human/tooling error that might drop the repository from the
configuration. This will make the job run an endless loop.

Also, add a timeout for 180 minutes for the job. It is enough for
the product build process, and if it isn't, it's better to cancel
and have human eyes looking at it.

This PR also changes `verify-repo-build-successful.py` so that it detects if the
repository is missing (and will exit with retcode `2`), and it also marks
the `--project` and `--repository` arguments required as the script expect those
to be set.

### DIFF
--- a/gocd/slfo-stagings.gocd.yaml
+++ b/gocd/slfo-stagings.gocd.yaml
@@ -59,13 +59,18 @@ pipelines:
               fi
 
     - Build.product:
+        timeout: 180
         resources:
           - staging-bot
         tasks:
           - script: |-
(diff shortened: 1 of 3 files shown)


gleidi-suse (approved):
LGTM

PR #3115, opened 2024-06-19
Automatic publishing of repos for LEAPM 6.0

lkocman:
I'd like to agree on how often gets the project republished.

The idea is that we have basically an unlocked project, where we keep on republishing repository and ensure that it contains all released updates for SLEM 6.0. This would be happening for 12 months and then would be disabled.

If there is a way to detect a change in underlying repository, that woud be perfect. Otherwise I think set it up for once a day would be sufficient.

Since we no longer have a regular install DVD, there is no corresponding test suite, we only have one for the :Images subproject/appliances. https://openqa.opensuse.org/group_overview/101

### DIFF
--- a/gocd/leapmicro.target.gocd.yaml
+++ b/gocd/leapmicro.target.gocd.yaml
@@ -36,3 +36,23 @@ pipelines:
         - repo-checker
         tasks:
           - script: ./pkglistgen.py --verbose -A https://api.opensuse.org update_and_solve --project openSUSE:Leap:Micro:6.0 --scope target --engine product_composer --force
+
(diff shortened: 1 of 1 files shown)


lkocman:
@boiko ^

lkocman:
@boiko mentioned that we should publish it at least on a daily basis

lkocman:
I did set pipeline to trigger publish twice day, at 0:50 and 12:50 every day.

lkocman:
@adrianschroeter I will merge in the evening this unless somebody says otherwise.

PR #3116, opened 2024-06-21
Also consider spdx.json an iso for the purpose of openQA

dirkmueller:
The SL Micro Maintenance effort has removed isos from staging, which means there was nothing to trigger openqa on anymore. trigger on the spdx description instead.


boiko:
The change looks good to me. I didn't try running it though (I am not that familiar with the infrastructure, always find it risky to try out and maybe break things)

dirkmueller:
it should only affect those where the OpenQA attribute matches, so that shouldn't be many

PR #3118, opened 2024-07-03
Install tools explicitly to avoid busybox variants being picked

dirkmueller:
zypper has the nasty habit of picking the alphabetically first package variant, which is not what we want


fridrich (approved):
Looks good to me and the errors in gocd/*xml don't look like caused by this one

PR #3119, opened 2024-07-04
Ignore patchinfo packages when checking for duplicated binaries

boiko:
In the staging areas used by maintenance, the patchinfo packages are staged together with the source packages, and because of the way they work, they collect binaries from the source packages around.

This causes binaries to be reported as duplicated, so ignore patchinfos.

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -243,7 +243,8 @@ def binary_list(apiurl, project, repository, arch, package=None):
 
 
 @memoize(session=True)
-def package_binary_list(apiurl, project, repository, arch, package=None, strip_multibuild=True, exclude_src_debug=False):
+def package_binary_list(apiurl, project, repository, arch, package=None, strip_multibuild=True, exclude_src_debug=False,
+                        exclude_patchinfo=False):
     path = ['build', project, repository, arch]
     if package:
         path.append(package)
@@ -256,6 +257,9 @@ def package_binary_list(apiurl, project, repository, arch, package=None, strip_m
         package = binary_list.get('package')
         if strip_multibuild:
             package = package.split(':', 1)[0]
+
+        if exclude_patchinfo and package.startswith("patchinfo."):
+            continue
 
         for binary in binary_list:
             filename = binary.get('name')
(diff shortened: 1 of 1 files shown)


staging-installcheck.py
```diff
@@ -267,10 +267,21 @@ def staging(self, project, force=False):
...
+        # remove white listed duplicates and ignore patchinfos
         for arch in list(duplicates):
             for binary in self.ignore_duplicated:
                 duplicates[arch].pop(binary, None)
+
+            # In maintenance stagings there are patchinfo packages which collect binaries from
+            # other source packages. This leads to duplicate binaries being reported. So remove
+            # any patchinfos from the comparison.
+            for package in list(duplicates[arch]):
+                filtered = [s for s in duplicates[arch][package] if not s.startswith("patchinfo")]
```
dirkmueller, line 279: do they start with `patchinfo:` ? I personally would prefer to include the dot here at the end.

adrianschroeter:
On Freitag, 5. Juli 2024, 10:16:47 CEST Dirk Mueller wrote:
> @dirkmueller commented on this pull request.
> 
> 
> 
> >          for arch in list(duplicates):
>              for binary in self.ignore_duplicated:
>                  duplicates[arch].pop(binary, None)
> +
> +            # In maintenance stagings there are patchinfo packages which collect binaries from
> +            # other source packages. This leads to duplicate binaries being reported. So remove
> +            # any patchinfos from the comparison.
> +            for package in list(duplicates[arch]):
> +                filtered = [s for s in duplicates[arch][package] if not s.startswith("patchinfo")]
> 
> do they start with `patchinfo:` ? I personally would prefer to include the dot here at the end.

we have at least also a patchinfo_ga package atm.

-- 

Adrian Schroeter ***@***.***>
Build Infrastructure Project Manager

SUSE Software Solutions Germany GmbH, Frankenstraße 146, 90461 Nürnberg, Germany
(HRB 36809, AG Nürnberg) Geschäftsführer: Ivo Totev

adrianschroeter:
sorry, got actually also renamed to patchinfo.ga meanwhile.

Vogtinator:
IMO `package_binary_list` could just get another parameter next to `exclude_src_debug=False` to exclude patchinfo. Makes the implementation simpler.

Vogtinator (approved):

PR #3120, opened 2024-07-08
gocd: changes to checkers, stagings and target for SLE15SP7

Volkano28:


g7 (approved):
LGTM. Tests failures are unrelated.

PR #3122, opened 2024-07-11
Drop pkglistgen for Leap 16. We no longer use 000package-groups 

lkocman:
No longer needed with adoption to __all__ by Adrian

### DIFF
--- a/gocd/pkglistgen.opensuse.gocd.yaml
+++ b/gocd/pkglistgen.opensuse.gocd.yaml
@@ -124,11 +124,6 @@ pipelines:
         approval:
           type: manual
         jobs:
-          openSUSE_Leap_16.0_target:
-            resources:
-            - repo-checker
-            tasks:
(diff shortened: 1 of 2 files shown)


gocd/pkglistgen.opensuse.gocd.yaml.erb
```diff
@@ -73,20 +73,6 @@ pipelines:
         approval:
           type: manual
         jobs:
-<% ['openSUSE:Leap:16.0/target'].each do |project|
```
nilxam, line 76: you should remove the entire Pkglistgen.openSUSE_Leap_16.0 section

gocd/pkglistgen.opensuse.gocd.yaml
```diff
@@ -124,11 +124,6 @@ pipelines:
         approval:
           type: manual
         jobs:
-          openSUSE_Leap_16.0_target:
```
nilxam, line 127: you should remove the entire Pkglistgen.openSUSE_Leap_16.0 section

nilxam (changes requested):

PR #3125, opened 2024-07-18
pkglistgen: add sort reasonings in product_composer mode

g7:
Reasonings are really useful to understand why a package has been
pulled in. This commit adds those back in product_composer mode
and it matches the legacy behaviour.

Note: this PR also moves `Group.tocompose()` to f-strings to match the rest of the file.


gleidi-suse (approved):
lgtm

PR #3127, opened 2024-07-30
gocd: sle15: Split publish pipelines

g7:
This allows to always publish from :TEST to :PUBLISH even without the :GA -> :TEST step.

Useful for when gocd cleans up old pipeline runs.

Pretty much the same thing as #3102.


gleidi-suse (approved):
lgtm

g7:
Test failures are unrelated.

PR #3128, opened 2024-08-02
Improve decline message (openSUSE/scm-staging#51)

dirkmueller:


check_source.py
```diff
@@ -180,7 +180,7 @@ def check_source_submission(
                     scm_pool_repository = f"https://src.opensuse.org/pool/{source_package}"
                     if not scm_sync.text.startswith(scm_pool_repository):
                         # devel project uses scm sync not from the trusted src location
-                        self.review_messages['declined'] = f"Expected a devel project scm sync from {scm_pool_repository}"
+                        self.review_messages['declined'] = f"devel project scmsync setting is {scm_sync}. Must be {scm_pool_repository} instead."
```
Vogtinator, line 183: `scm_sync` is a tag. Maybe `scm_sync.text` instead?

dirkmueller: good catch, fixed.

Vogtinator (approved):

PR #3129, opened 2024-08-05
Add extra builds to test for reproducible builds

bmwiedemann:
This helps slow down introduction of new issues into Factory.

See https://reproducible-builds.org/ for why this is good.

This needs added prjconf in a central place:
```
%if %_repository == "rb_future1y"
VMinstall: reproducible-faketools-futurepost
%endif

%if %_repository == "rb_j1"
VMinstall: reproducible-faketools-j1
BuildFlags: nochecks
%endif
```

The `nochecks` part is optional.


DimStar77:
`%if %_repository == "rb_j1"` tends to not work across project boundaries, a `<path>` element makes prjconf from the parent being included (good) - but it resets `%_repository` the the repository name of the `<path>` element, thus not having the effect you expect here.

This needs to be replicated to all staging projects (incl. the dynamically setup adi projects)

e.g. `osc buildconfig openSUSE:Factory:Staging:M standard | grep "%_repository"`

bmwiedemann:
I pushed an extra commit to add prjconf in the script.

osclib/stagingapi.py
```diff
@@ -1456,6 +1468,27 @@ def create_adi_project(self, name, use_frozenlinks=False, src_prj=None):
         # put twice because on first put, the API adds useless maintainer
         http_PUT(url, data=meta)
 
+        prjconf = textwrap.dedent("""
+            # prjconf for testing reproducible builds:
+            BuildFlags: nodisturl
+            Release: 1.1
+
+            %if %_repository == "rb_future1y"
+            VMinstall: reproducible-faketools-futurepost
```
DimStar77, line 1477: should probably be 'Required'` - not `VMInstall`

DimStar77: or even `support` - i.e do not trigger rebuilds down the line just becuase those packages might have changes

osclib/stagingapi.py
```diff
@@ -1425,6 +1425,17 @@ def create_adi_project(self, name, use_frozenlinks=False, src_prj=None):
                  <arch>x86_64</arch>
               </repository>"""
 
+        reproducible_builds_repos = f"""
+               <repository name="rb_future1y">
+                 <path project="{name}" repository="standard"/>
+                 <arch>x86_64</arch>
+               </repository>
+               <repository name="rb_j1">
```
dirkmueller, line 1433: so that's testing with `-j1` ? how much slowdown is that causing?

osclib/stagingapi.py
```diff
@@ -1425,6 +1425,17 @@ def create_adi_project(self, name, use_frozenlinks=False, src_prj=None):
                  <arch>x86_64</arch>
               </repository>"""
 
+        reproducible_builds_repos = f"""
+               <repository name="rb_future1y">
```
dirkmueller, line 1429: what would be the most descriptive name? maybe `standard_1yinfuture` ?

osclib/stagingapi.py
bmwiedemann: https://rb.zq1.de/compare.factory/reproducible.json has `duration` values for -j1 and -j4 .
For most smaller packages, it does not make much of a difference, because non-build times make up for the majority of time. 
For MozillaFirefox, it says
```
      "build_duration"  : 39489,
      "build_duration4" : 22782,
```
and for bash
```
      "build_duration"  : 363,
      "build_duration4" : 274,
```
However, nothing needs to block on this build to finish.

DimStar77:
As discussed this morning, the reproducible state should be used for informational purpose (to start with) and not be blocking. Later on we can probably let the bot accept on confirmed or ask a human for confirmation when not confirmed.

To not block when any of the reproducible builds fail (e.g timebomb in > 1 year) we discussed moving the reproducible builds out of the 'staging dashboard observed areas;

e.g openSUSE:Factory:Staging:adi:65:reproducible (redoing what adi:65 does)

This meta and prjconf snippets worked in my experiments:

# prj meta:

```
<project name="openSUSE:Factory:Staging:adi:65:reproducible">
  <title/>
  <description/>
  <link project="openSUSE:Factory:Staging:adi:65"/>
  <url>/staging_workflows/openSUSE:Factory/staging_projects/openSUSE:Factory:Staging:adi:65</url>
  <person userid="dimstar" role="maintainer"/>
  <group groupid="factory-staging" role="maintainer"/>
  <publish>
    <disable/>
  </publish>
  <debuginfo>
    <enable/>
  </debuginfo>
  <repository name="rb_j1" linkedbuild="all">
    <path project="openSUSE:Factory:Staging" repository="standard"/>
    <path project="openSUSE:Factory" repository="standard"/>
    <arch>x86_64</arch>
  </repository>
  <repository name="rb_future1y" linkedbuild="all">
    <path project="openSUSE:Factory:Staging" repository="standard"/>
    <path project="openSUSE:Factory" repository="standard"/>
    <arch>x86_64</arch>
  </repository>
</project>

```

# prjconf

```
%if "%_repository" == "rb_future1y"
Required: reproducible-faketools-futurepost
%endif

%if "%_repository" == "rb_j1"
Required: reproducible-faketools-j1
BuildFlags: nochecks
%endif

BuildFlags: nodisturl
Release: 1.1

Macros:
%distribution reproducible

```

PR #3131, opened 2024-08-06
Fix "option '__name__' in section 'remote' already exists"

dirkmueller:


Vogtinator:
This is just #3121, but now this hits the CI because osc 1.8.0 has the regression but without the fix (https://github.com/openSUSE/osc/pull/1601).

Vogtinator:
> This is just #3121, but now this hits the CI because osc 1.8.0 has the regression but without the fix ([openSUSE/osc#1601](https://github.com/openSUSE/osc/pull/1601)).

I asked for an osc update which includes the PR. If that doesn't happen soon, I resurrected #3121 as #3138.

PR #3133, opened 2024-08-09
metrics: Use https for openQA influxdb endpoint

nicksinger:
As per discussion in https://progress.opensuse.org/issues/159669


DimStar77 (approved):

okurz (changes requested):
But you stated in https://progress.opensuse.org/issues/159669#note-20 that http works for you so this wouldn't help anyway, right?

nicksinger:
Correct but I think it is the right approach anyway (uses one less request, https by default, etc)

okurz (approved):

PR #3134, opened 2024-08-12
Add TTM for openSUSE:Factory:RISCV

andreas-schwab:


Vogtinator:
LGTM, but let's enable it only once there's some build released so it won't just fail on every run.

Vogtinator:
> LGTM, but let's enable it only once there's some build released so it won't just fail on every run.

Looks like that's the case now, but the CI is broken because #3121 hasn't been merged yet

dirkmueller (approved):

PR #3137, opened 2024-08-15
Subpackage, service, timer for build-fail-reminder

tacerus:
This adds a build-fail-reminder subpackage, shipping the script along with a service user and unit files for daily execution via systemd, making for easier management and deployment in the openSUSE infrastructure.


Vogtinator (approved):
Ideally it would use systemd-tmpfiles for the directories in /etc and /var but it's not the case for the others either

tacerus:
Thanks for the quick approval.

This could do with ConfigurationDirectory/StateDirectory instead of tmpfiles as the directories don't need additional options. Happy to add those.

Do you know if my PR builds somewhere in OBS automagically or do I need to branch the package and copy paste my changes to test it?

Vogtinator:
It's part of the PR checks (see below). ATM it fails: https://build.opensuse.org/package/show/openSUSE:Tools:OSRT:TestGithub:openSUSE:openSUSE-release-tools:PR-3137/openSUSE-release-tools

tacerus:
Ah nice, thanks, I will check and repair it shortly.

tacerus:
Working now (tested as far as I got without a valid login), also added the /etc and /var improvement and moved the settings to some variables for easier adjustment.

Will rebase and mark as ready after https://github.com/openSUSE/openSUSE-release-tools/pull/3139.

PR #3139, opened 2024-08-15
Relax osclib dependencies

tacerus:
Allow installation of subpackages depending on osclib without the big main package to reduce their footprint.


tacerus:
Ah, I wasn't sure if they're needed for something else, then I will remove them from there.

tacerus:
Done, and I noticed pika was not listed before at all, but PubSubConsumer.py imports it - I added it as a Recommends as the other parts of osclib seem to work without it.

PR #3141, opened 2024-08-19
Remove Dolomite pipelines and add SLFO Snapshot pipeline

gyr:
- Remove ALP.Target group and all pipelanes under it as they were only used by Dolomite project
- Remove Dolomite pipelines from ALP.Checkers group
- Add pipeline to release SUSE:SLFO:Main:Build to SUSE:SLFO:Main:Build:Snapshot when standard repository finishes the build


g7 (approved):
Thank you! LGTM

PR #3143, opened 2024-08-20
legal-auto: Link to legal review via comment

kraih:
It's rather hard for packagers to find the matching legal review for their OBS/IBS request. A comment with link should help make legal information more widely accessible.

I'm not the maintainer of legal-auto, but this has been a rather frequent request i've been getting from LegalDB users.

### DIFF
--- a/legal-auto.py
+++ b/legal-auto.py
@@ -20,6 +20,7 @@ from lxml import etree as ET
 import osc.conf
 import osc.core
 from osclib.cache_manager import CacheManager
+from osclib.comments import CommentAPI
 import ReviewBot
 
 http_GET = osc.core.http_GET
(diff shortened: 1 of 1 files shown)


Vogtinator (approved):

Vogtinator:
Maybe let's do it only if there's no automatic accept?

DimStar77:
> Maybe let's do it only if there's no automatic accept?

I wonder for whom the message is: package submitters /packagers get no information from it, as they barely ever have access to it. Which means it's mostly targeting release managers who want a quick shortcut?

g7:
Perhaps the link to the legal review can be added directly in the decline message (rather than a comment)?

kraih:
We don't know if or when there will be an auto-accept on the LegalDB side. For Factory we could of course disable these comments, since all requests will be auto-accepted based on a timer.

kraih:
> Perhaps the link to the legal review can be added directly in the decline message (rather than a comment)?

That would not cover reviews stuck in the backlog. Which is currently one of the main reasons people contact me for more information on legal reviews.

kraih:
I think this OBS feature would probably solve all complaints, just have a bot flag and allow all bots to be muted: https://github.com/openSUSE/open-build-service/issues/13245

kraih:
> 
> In case of Factory: there is a timed auto-accept
> 
> I wonder for whom the message is: package submitters /packagers get no information from it, as they barely ever have access to it. Which means it's mostly targeting release managers who want a quick shortcut?

That raises the question of why we even have legal-auto as a reviewer on openSUSE:Factory. It gives the impression that there is a legal review happening, when there really isn't. Everything gets auto-accepted after a few hours unless there is a prior rejection present for that package from a project that actually does receive legal reviews. And this pretty much never happens.

The link to the legal review is meant for the human reviewers who have to sign off on the submission and take legal responsibility when doing so. That is you.

kraih:
I'm going to open another PR that excludes openSUSE:Factory from commenting. But i also want to change the way acceptance looks in the Request History. It needs to be more clear that legal responsibility lies with the openSUSE review team, and not legal-auto.

PR #3144, opened 2024-08-21
legal-auto: Exclude openSUSE:Factory from commenting

kraih:
Since there have been quite a few vocal complaints about comments with links to LegalDB reviews not being useful for openSUSE:Factory, we now exclude the project.

I've also rephrased review messages in a consistent format, and emphasised the fact that the opensuse-review-team is responsible for license compliance if the 2 hour timeout is reached without a legal review.

See also #3143.

### DIFF
--- a/legal-auto.py
+++ b/legal-auto.py
@@ -65,7 +65,7 @@ class LegalAuto(ReviewBot.ReviewBot):
             id = self.request.reqid
         return self.apinick + id
 
-    def create_db_entry(self, src_project, src_package, src_rev):
+    def create_db_entry(self, src_project, src_package, src_rev, target_project):
         params = {'api': self.apiurl, 'project': src_project, 'package': src_package,
                   'external_link': self.request_nick(),
                   'created': self.request.statehistory[0].when + ' UTC'}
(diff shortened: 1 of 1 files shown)


legal-auto.py
```diff
@@ -136,7 +138,7 @@ def check_source_submission(self, src_project, src_package, src_rev, target_proj
                 # reopen
                 return None
             if state == 'new' and self.valid_for_opensuse(target_project, report):
-                self.message = 'The legal review is accepted preliminary. The package may require actions later on.'
+                self.message = 'Legal review pending, opensuse-review-team is responsible for license compliance'
```
DimStar77, line 141: That statement is untrue and was never agreed in this form. The review team cannon be held responsible (as half of the team would not even have access to legaldb)

The plan was that the things are accepted preliminary - but stay open on legaldb to be reviewed and taken action whenever the legal team has the capacity to review the stuff. Nothing more, nothing less.

kraih: That might be a problem, because it's certainly not me either. Who from legal has agreed to take responsibility for preliminary accepted requests? Looking through the data, it appears since 2022 only 14 such reports have actually been reviewed after being auto-accepted.

kraih:
This appears to be a legal minefield. I'll not make any changes to the bot until i've got an opinion from legal.

PR #3145, opened 2024-08-22
Revert "legal-auto: Link to legal review via comment"

DimStar77:
Reverts openSUSE/openSUSE-release-tools#3143

The rework/rewording was canceled to 'not open a can of worms' - and no further changes to the legal-auto bot are to be done.

To adhere to that statement, let's go back to the state we had before (and not spam package maintainers with messages including links they cannot use)

### DIFF
--- a/legal-auto.py
+++ b/legal-auto.py
@@ -20,7 +20,6 @@ from lxml import etree as ET
 import osc.conf
 import osc.core
 from osclib.cache_manager import CacheManager
-from osclib.comments import CommentAPI
 import ReviewBot
 
 http_GET = osc.core.http_GET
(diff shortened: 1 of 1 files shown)


kraih:
I've specifically asked @DimStar77 not to disable it for IBS... that you just reverted everything is disappointing. 😞

DimStar77:
> I've specifically asked @DimStar77 not to disable it for IBS... that you just reverted everything is disappointing. 😞

The fact that you 'throw up your hands and not make any changes to the bot until you've got an opinion from legal.' just after adding a feature that was obviously not thought through is disappointing - I agree

kraih:
> The fact that you 'throw up your hands and not make any changes to the bot until you've got an opinion from legal.' just after adding a feature that was obviously not thought through is disappointing - I agree

The feature was fine, and a 2-line patch to exclude openSUSE:Factory already existed. I encouraged you to merge it without the message changes, and you chose not to.

I'm throwing up my hands because of your insistence on fake legal reviews for Factory. I've looked at the database and in the past two years out of thousands of relevant package submissions only 14 have actually been reviewed (one by me because of illegal code that made its way to Tumbleweed). Anyway, i've now asked SUSE legal for clarification about the status of the agreement.

PR #3146, opened 2024-08-26
gocd: slfo-stagings: Freeze the Stagings each Sunday.

gleidi-suse:


gocd/slfo-stagings.gocd.yaml
```diff
@@ -1,7 +1,30 @@
...
+        destination: scripts
+    stages:
+    - Freeze.stagings:
+        # 8 hours
+        timeout: 480
+        resources:
+          - staging-bot
+        tasks:
+          - script: |-
+              osc -A $STAGING_API staging freeze -p $SLFO_PROJECT A B C D E F G H S V Y
```
g7, line 27: What about checking if the rings are built first? Maybe in a preliminary stage

Something like:
* Check.Rings.Built job (runs every hour on Sundays, simple check and then exit)
* (If successful, and it has not been freezed in the past day) Freeze.stagings

If we're sure that the Ring is built I think we can also decrease the timeout (as supposedly it won't take that long to freeze the stagings)

gleidi-suse: Hi I should have addressed all your comments, thank you for your input! Let me know if you find anything else

g7 (approved):
LGTM, thanks!

Issue #3149, opened 2024-08-29
legal-auto needs to be updated for new LegalDB state


kraih:
In LegalDB we now have a state called `acceptable_by_lawyer` and legal-auto needs to be updated for it. I'm not too familiar with how legal-auto uses LegalDB states, but grepping showed that maybe only this line needs to be updated: https://github.com/openSUSE/openSUSE-release-tools/blob/master/legal-auto.py#L139

Until then some new OBS/IBS requests may get stuck in review because they are in that state.

kraih:
For all intends and puproses `acceptable_by_lawyer` is indetical to `correct` for now, if that help get this done more quickly.

PR #3151, opened 2024-08-30
Add state acceptable_by_lawyer

lkocman:
Related to Issue #3149
States are documented here https://github.com/openSUSE/cavil/blob/master/docs/Architecture.md#states

"accetpable" is reviewed by an expert, "acceptable_by_lawyer" reviewed by an actual lawyer

This new state gives us opportnity to handle situation differently in case that review is done by lawyer, vs best effort.

### DIFF
--- a/legal-auto.py
+++ b/legal-auto.py
@@ -2,6 +2,9 @@
 
 # SPDX-License-Identifier: MIT
 
+# States from legaldb.suse.de / Cavil are documented in
+# https://github.com/openSUSE/cavil/blob/master/docs/Architecture.md#states
+
 import os
 import os.path
 import sys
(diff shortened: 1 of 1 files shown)


lkocman:
Who's responsible for the tool. @darix suggested @gyr

kraih:
Quite a few implementation details of the LegalDB database schema are currently leaking into the legal-auto code. Which means even small changes can have very unintended consequences. It would be very helpful to have someone to discuss with how we can improve the situation in the future.

kraih (approved):
Not too familiar with the code, but this does look like the right solution to me.

gyr (commented):
LGTM

bugfinder (approved):

kraih:
Thanks, looks like it works. https://legaldb.suse.de/reviews/details/412946 -> https://build.suse.de/request/show/343414

PR #3154, opened 2024-09-04
gocd: slfo-stagings: Avoid busy waiting for 50 minutes, instead rely on scheduling from GoCD.

gleidi-suse:


g7 (approved):
:100: thanks, lgtm

PR #3155, opened 2024-09-04
gocd: SUSE.SLFO.Main.Staging.Weekly.Freeze: Add osc staging bot oscrc.

gleidi-suse:


g7 (approved):
Ah, yes. Makes sense to me.

PR #3158, opened 2024-09-10
Add slfo-packagelist-uploader and a adjusted check_source

nilxam:


nilxam:
I don't get the smoketest error:

`
import git
ModuleNotFoundError: No module named 'git'
Error: Process completed with exit code 1.
`

python3-GitPython is added, what should I do here?

check_source.py
```diff
@@ -147,11 +150,22 @@ def check_source_submission(
             self.review_messages['declined'] = f'May not modify a non-source package of type {kind}'
             return False
 
-        if not self.allow_source_in_sle and self.sle_project_to_check:
-            if entity_exists(self.apiurl, self.sle_project_to_check, target_package):
+        if not self.allow_source_in_sle:
+            if self.sle_project_to_check and entity_exists(self.apiurl, self.sle_project_to_check, target_package):
                 self.review_messages['declined'] = ("SLE-base package, please submit to the corresponding SLE project."
                                                     "Or let us know the reason why needs to rebuild SLE-base package.")
```
lkocman, line 156:
```suggestion
                self.review_messages['declined'] = ("SLFO package, please submit change to the corresponding project."
                                                    "Alternatively please provide a reason for rebuilding this SLFO package.")
```

lkocman: I think we could use SLFO as the package could be from SLES, SL Micro or similar.  Not sure if it helps the maintainer as SLFO is not exactly common term in community. Alternatively we could say just SLES as SLED is going away anyway.

lkocman (approved):
Generally aprooving, see my comment about the text

lkocman (approved):
looks good! thank you @nilxam

PR #3161, opened 2024-09-12
ttm: support Leap 16.0 development structure

nilxam:
The description of the Leap 16.0 setup
* **000productcompose** is the main product but it builds ftp only. Build on `product` repo.
* **agama-installer** need to have identical build number as 000productcompose had. Build on `images` repo.
* **000productcompose** would not set release number while releasing to `ToTest`, but **agama-installer** does.

Tested this change with **Build36.40**

### DIFF
--- a/ttm/releaser.py
+++ b/ttm/releaser.py
@@ -76,6 +76,11 @@ class ToTestReleaser(ToTestManager):
             return self.release_version()
 
         if len(self.project.main_products):
+            # 000productcompose has ftp built only and the build number
+            # agama-installer carry over build number from 000prodcutcompose
+            # but they are not from the same package container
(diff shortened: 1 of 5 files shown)


lkocman (approved):
Looks good Max! Thank you

ttm/releaser.py
```diff
@@ -296,7 +305,7 @@ def _release(self, set_release=None):
 
         for image in self.project.image_products:
             self.release_package(self.project.name, image.package, set_release=set_release,
-                                 repository=self.project.product_repo,
+                                 repository=getattr(self.project, 'totest_images_repo', self.project.product_repo),
```
nilxam, line 308: hmm...it's needed another solution, otherwises it might break Factory release from `images` to `appliances`..

PR #3051, opened 2024-02-02
Initial support for productcompose files

adrianschroeter:
It requires to have a "default.productcompose" file in 000package-groups to enable support for real

PR #3165, opened 2024-09-23
pkglistgen: add an ability to update weakremovers.inc only

nilxam:
Add an --only-update-weakremovers option to be able to update weakremovers.inc only. With this option, pkglistgen will checkout release package and update weakremovers.inc, nothing else, no needed to have other product package involved.

Since Leap 16.0 don't use pkglistgen anymore but productcomposer, we still need a tool to recalculate weakremovers.inc according to the binary pool from the current and the previous version, therefore lets reuse pkglistgen's code for weakremovers handling.

### DIFF
--- a/pkglistgen/tool.py
+++ b/pkglistgen/tool.py
@@ -665,6 +665,7 @@ class PkgListGen(ToolBase.ToolBase):
         force: bool,
         no_checkout: bool,
         only_release_packages: bool,
+        only_update_weakremovers: bool,
         stop_after_solve: bool,
         custom_cache_tag,
         engine: Engine,
@@ -690,6 +691,8 @@ class PkgListGen(ToolBase.ToolBase):
 
         drop_list = []
         checkout_list = [group, product_package, release]
+        if only_update_weakremovers:
+            checkout_list = [release]
         if not force:
             root = ET.fromstringlist(show_results_meta(api.apiurl, project, product_package,
                                                        repository=[main_repo], multibuild=True))
@@ -721,7 +724,7 @@ class PkgListGen(ToolBase.ToolBase):
         else:
             url = api.makeurl(['source', project])
             packages = ET.parse(http_GET(url)).getroot()
-            if packages.find(f'entry[@name="{product_package}"]') is None:
+            if packages.find(f'entry[@name="{product_package}"]') is None and not only_update_weakremovers:
                 if not self.dry_run:
                     undelete_package(api.apiurl, project, product_package, 'revive')
                 # TODO disable build.
@@ -753,7 +756,10 @@ class PkgListGen(ToolBase.ToolBase):
         if engine == Engine.product_composer:
             self.output_dir = productcompose_dir
         elif engine == Engine.legacy:
-            self.output_dir = product_dir
+            if not only_update_weakremovers:
+                self.output_dir = product_dir
+            else:
+                self.output_dir = release_dir
 
         if not no_checkout and not git_url:
             logging.debug(f'Skipping checkout of {project}')
@@ -761,19 +767,22 @@ class PkgListGen(ToolBase.ToolBase):
                 checkout_package(api.apiurl, project, package, expand_link=True,
                                  prj_dir=cache_dir, outdir=os.path.join(cache_dir, package))
 
-        if (engine == Engine.legacy) and (not only_release_packages):
+        if (engine == Engine.legacy) and (not only_release_packages) and not only_update_weakremovers:
             file_utils.unlink_all_except(self.output_dir)
-
-        ignore_list = ['supportstatus.txt', 'summary-staging.txt', 'package-groups.changes',
-                       'default.productcompose.in']
-        ignore_list += self.group_input_files()
 
         # old product-builder
         if engine == Engine.legacy:
-            file_utils.copy_directory_contents(group_dir, self.output_dir, ignore_list)
-            file_utils.change_extension(self.output_dir, '.spec.in', '.spec')
-            file_utils.change_extension(self.output_dir, '.product.in', '.product')
-            file_utils.unlink_all_except(release_dir, ['weakremovers.inc', '*.changes'])
+            if only_update_weakremovers:
+                # remove stub.spec before re-creating _multibuild per specfiles
+                file_utils.unlink_list(release_dir, ['stub.spec'])
+            else:
+                ignore_list = ['supportstatus.txt', 'summary-staging.txt', 'package-groups.changes',
+                               'default.productcompose.in']
+                ignore_list += self.group_input_files()
+                file_utils.copy_directory_contents(group_dir, self.output_dir, ignore_list)
+                file_utils.change_extension(self.output_dir, '.spec.in', '.spec')
+                file_utils.change_extension(self.output_dir, '.product.in', '.product')
+                file_utils.unlink_all_except(release_dir, ['weakremovers.inc', '*.changes'])
 
         # new product-composer
         fn = os.path.join(group_dir, 'default.productcompose.in')
@@ -804,7 +813,9 @@ class PkgListGen(ToolBase.ToolBase):
         self.filter_architectures(target_archs(api.apiurl, project, main_repo))
         self.update_repos(self.filtered_architectures)
 
-        if only_release_packages:
+        if only_update_weakremovers:
+            pass
+        elif only_release_packages:
             self.load_all_groups()
             self.write_group_stubs()
         else:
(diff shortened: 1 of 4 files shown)


Vogtinator:
> Since Leap 16.0 don't use pkglistgen anymore but productcomposer,

Doesn't pkglistgen write its output as productcompose files in that case?

nilxam:
> > Since Leap 16.0 don't use pkglistgen anymore but productcomposer,
> 
> Doesn't pkglistgen write its output as productcompose files in that case?

I'm not sure are you reference to productcomposer support in pkglistgen? or any other scenario please could you brief it a bit more?

Leap 16.0 doesn't use pkglistgen to generate any _product_ file(?), none for `000product`(sure thing), nor `000release-packages`, of course nor `000package-groups`, just have a pure and simple `000productcompose` to have [.productcompose](https://build.opensuse.org/package/show/openSUSE:Leap:16.0/000productcompose) file. There is no input files, and no pkglistgen, nothing going to update release-packages, therefor weakremovers.inc won't get updated.

Vogtinator:
> > > Since Leap 16.0 don't use pkglistgen anymore but productcomposer,
> > 
> > 
> > Doesn't pkglistgen write its output as productcompose files in that case?
> 
> I'm not sure are you reference to productcomposer support in pkglistgen? or any other scenario please could you brief it a bit more?

Yes: https://github.com/openSUSE/openSUSE-release-tools/pull/3051

You can use pkglistgen to manage 000productcompose just like 000product.

> Leap 16.0 doesn't use pkglistgen to generate any _product_ file(?), none for `000product`(sure thing), nor `000release-packages`, of course nor `000package-groups`, just have a pure and simple `000productcompose` to have [.productcompose](https://build.opensuse.org/package/show/openSUSE:Leap:16.0/000productcompose) file. There is no input files, and no pkglistgen, nothing going to update release-packages, therefor weakremovers.inc won't get updated.

i.e. 000productcompose in 16.0 currently only builds the FTP tree which doesn't really need solving because there's no DVD?

nilxam:
> Yes: #3051
> 
> You can use pkglistgen to manage 000productcompose just like 000product.

Yes, but we don't have input files in 000package-groups, packagelist calculation is not needed.

> i.e. 000productcompose in 16.0 currently only builds the FTP tree which doesn't really need solving because there's no DVD?

Yes. 000productcompose builds the ftp tree only, it uses `__all__` to include all binary RPM from the binary pool. Only medium we had is agama-installer image.

Vogtinator:
Ok, then it makes sense. Other approaches I can think of:

a) Detect if there is no solving needed and skip those steps automatically (-> no manual config needed)
b) Separate the weakremovers update from pkglistgen. Needs some work and synchronization though.

nilxam:
> Ok, then it makes sense. Other approaches I can think of:
> 
> a) Detect if there is no solving needed and skip those steps automatically (-> no manual config needed) b) Separate the weakremovers update from pkglistgen. Needs some work and synchronization though.

I initially go for option b, but pkglistgen has code over there already so I don't bother to, if at some point in the future we must to have a separate weakremovers updater then I'll do so.

pkglistgen/tool.py
```diff
@@ -721,15 +724,15 @@ def update_and_solve_target(
...
+            if packages.find(f'entry[@name="{product_package}"]') is None and not only_update_weakremovers:
                 if not self.dry_run:
                     undelete_package(api.apiurl, project, product_package, 'revive')
                 # TODO disable build.
                 logging.info(f'{product_package} undeleted from {project}, skip dvd until next cycle')
                 return
 
             drop_list = api.item_exists(project, oldrepos)
-            if drop_list and not only_release_packages:
+            if (drop_list and not only_release_packages) or (drop_list and only_update_weakremovers):
```
Vogtinator, line 735: I think this change is unnecessary - the first part is already true, isn't it?

pkglistgen/tool.py
```diff
@@ -753,27 +756,35 @@ def update_and_solve_target(
         if engine == Engine.product_composer:
             self.output_dir = productcompose_dir
         elif engine == Engine.legacy:
-            self.output_dir = product_dir
+            if not only_update_weakremovers:
```
Vogtinator, line 759: Shouldn't this be independent of the engine?

pkglistgen/tool.py
```diff
@@ -753,27 +756,35 @@ def update_and_solve_target(
...
                                  prj_dir=cache_dir, outdir=os.path.join(cache_dir, package))
 
-        if (engine == Engine.legacy) and (not only_release_packages):
+        if (engine == Engine.legacy) and (not only_release_packages) and not only_update_weakremovers:
             file_utils.unlink_all_except(self.output_dir)
 
-        ignore_list = ['supportstatus.txt', 'summary-staging.txt', 'package-groups.changes',
-                       'default.productcompose.in']
-        ignore_list += self.group_input_files()
+        if not only_update_weakremovers:
```
Vogtinator, line 773: I guess you can merge that with the `if engine == Engine.legacy` block below

pkglistgen/tool.py
nilxam: ah, good catch!

pkglistgen/tool.py
```diff
@@ -804,28 +815,29 @@ def update_and_solve_target(
...
-            self.load_all_groups()
-            self.write_group_stubs()
-        else:
-            self.solve_project(
-                ignore_unresolvable=str2bool(target_config.get('pkglistgen-ignore-unresolvable')),
-                ignore_recommended=str2bool(target_config.get('pkglistgen-ignore-recommended')),
-                locale=target_config.get('pkglistgen-locale'),
-                locales_from=target_config.get('pkglistgen-locales-from')
-            )
+        if not only_update_weakremovers:
```
Vogtinator, line 818: Can be simplified:

```
if only_update_weakremovers:
    pass  # done below
elif only_release_packages:
    ...
else:
    ...
```

pkglistgen/tool.py
```diff
@@ -804,28 +815,29 @@ def update_and_solve_target(
...
+                    self.write_all_groups()
 
-            summary = self.make_summary()
+                summary = self.make_summary()
 
         if stop_after_solve:
             return
 
-        if drop_list and not only_release_packages:
+        if (drop_list and not only_release_packages) or (drop_list and only_update_weakremovers):
```
Vogtinator, line 840: AFAICT unnecessary, like above

pkglistgen/tool.py
nilxam: I limits `--only-update-weakremovers` to legacy engine only, see changes in cli.py

pkglistgen/tool.py
nilxam: all done

pkglistgen/tool.py
nilxam: see https://github.com/openSUSE/openSUSE-release-tools/blob/master/pkglistgen/cli.py#L79 , if product_composer engine in pkglistgen would not handle 000release-packages, why should `--only-update-weakremovers` to be independednt of the engine?

Vogtinator: I'm not actually sure why, because even with product composer there is 000release-packages: https://src.suse.de/products/SLES/src/branch/16.0/000release-packages (note: internal link)

@adrianschroeter maybe you remember?

nilxam: Have a existing 000release-packages when pkglistgen with product composer is on that is fine, 000release-packages just a package name and it builds `-release` binary package, the point is who and how does the content of 000release-packages be generated, per 000release-pcakges in SLFO:Main:Staging:A it was not done by product composer, as well as not https://src.suse.de/products/SLES/src/branch/16.0/000release-packages, maybe I'm wrong but changelog says that. And you can not find `XXX-release.spec.in` in 000pacakge-groups in those projects.

Vogtinator: Right, so in that case it might just be managed manually, including weakremovers? Hm.

nilxam: That's true for Leap 16 atm, to manage release package manually, that's why I needed a better way here to handle weakremovers...

Vogtinator: Do you know what SLES is doing?

nilxam: > Do you know what SLES is doing?

I don't know will the thing gonna be changed in SLES16, but in SLE15, weakremovers.inc is not rely on 000update-repos, they have a manual maintained weakremovers.inc in 000packages-group, and copy it to 000release-pacakges by pkglistgen. See https://github.com/openSUSE/openSUSE-release-tools/blob/master/pkglistgen/tool.py#L863

g7: yes, indeed in SLES16/SL-Micro currently 000release-packages are managed manually

Vogtinator: Ok, so the productcompose + weakremovers.inc case is just not relevant yet.

Vogtinator (approved):

PR #3168, opened 2024-09-30
Move imports in osc plugins from top level to individual functions to reduce osc load times

dmach:


dirkmueller:
This is kinda ugly. can we maybe split the functions that need osc into a separate module that is then only loaded on demand? e.g. just one function-embedded load rather than all of them moved (which still then allows to pyflake check for missing imports imho, which is a nice property to keep).

dmach:
@dirkmueller good point, that sounds like a sane approach.
Unfortunately when I tried that, it improves some things while making others worse.
That new module would have to be imported on too many places anyway so it only adds more complexity to the code.
Maybe if you have an exact idea on how to do it, could you share it with me?

PR #3170, opened 2024-10-01
Update SLFO 1.1 stagings

gleidi-suse:


dirkmueller:
any rationale for why ?

gleidi-suse:
> any rationale for why ?

@dirkmueller  you mean why only two stagings? that is a good question, maybe we can get all 11 of them . let me verify

gleidi-suse:
I was able to create the 9 remaining Stagings

PR #3171, opened 2024-10-03
pcheck: fix creation of submit requests

DimStar77:
Use the proper API calls `<request><action type="submit">…` instead of
the no longer working `<request type="submit">…`

### DIFF
--- a/osc-pcheck.py
+++ b/osc-pcheck.py
@@ -114,7 +114,7 @@ class oscapi:
         query = {'cmd': 'create'}
         url = osc.core.makeurl(self.apiurl, ['request'], query=query)
 
-        data = '<request type="submit"><submit><source project="{project}" package="{package}" rev="{rev}"/>' \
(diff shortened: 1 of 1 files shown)


DimStar77:
The existing code resulted in :+1: 

```
> osc pcheck --push -m "Scripted push of GNOME 47" GNOME:Factory
Creating a request from GNOME:Factory/accerciser
Server returned an error: HTTP Error 400: Bad Request
request validation error: 1:0: ERROR: Expecting an element action, got nothing
1:0: ERROR: Invalid sequence in interleave
1:0: ERROR: Element request failed to validate content
```

Most likely a result of https://github.com/openSUSE/open-build-service/commit/fc6f1b4a5caa4c7d204e0a6cb9df9ffbb520af35

Vogtinator (commented):
There's also `osc.core.create_submit_request` which essentially replaces all of `oscapi.create`

Vogtinator (approved):

PR #3177, opened 2024-10-10
Fix broken relase to :PUBLISH pipeline for SLES16

gyr:
Only release SLES16 packages from :TEST to :PUBLISH, skiping containerfile repository content that breaks the pipeline


gleidi-suse (approved):
ok, thanks

PR #3185, opened 2024-10-31
osclib/list_command.py: Allow showing details for packages not in rings.

gleidi-suse:
This PR implements the ability to show for packages that do not belong to any ring the same details that are shown for packages that do live in rings.  Due to concerns about compatibility and volume of the output the feature is hidden behind a new --no-ring-detailed command line flag.

The following is an example of the change with the flag enabled:
image
and this is with the flag disabled
Screenshot From 2024-10-31 15-00-29

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -93,6 +93,8 @@ def clean_args(args):
               help='split the requests into individual groups')
 @cmdln.option('--supersede', action='store_true',
               help='replace staged requests when superseded')
+@cmdln.option('--adi-details', action='store_true',
+              help='show detailed summary for packages that are not in any ring')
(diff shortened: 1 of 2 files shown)


DimStar77:
`--no-ring-detailed` - I do like the fact that you create that new behaviour opt-in, but I fund that parameter really confusing...

`--no` implies somewhat of 'Do not do this` - so rather 'do not show details about rings`. What do you think about `--adi-details` ?

gleidi-suse:
Hi @DimStar77 , `--adi-details` is definitely better. I updated the code to match that

nilxam (approved):
LGTM

PR #3186, opened 2024-10-31
osclib/list_command.py: Allow overriding reviews filter from the command line.

gleidi-suse:
This patch implements overriding of the filter on reviews set up by the staging plugin. It is now possible to filter requests that have been reviewed by more than one group, for example; or to filter requests that have lack a request by a specific group or bot.

It allows queries like the following:

```
iosc staging -p SUSE:SLFO:Main list --match-filter "state/@name='review' and review[@by_group='factory-auto' and @state='new']"
```

or also

```
iosc staging -p SUSE:SLFO:Main list --match-filter "state/@name='review' and review[(@by_group='factory-auto' or @by_group='licensedigger') and @state='accepted']"
```

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -120,6 +120,7 @@ def clean_args(args):
 @cmdln.option('--merge', action='store_true', help='propose merge where applicable and store details to allow future merges')
 @cmdln.option('--try-strategies', action='store_true', default=False, help='apply strategies and keep any with desireable outcome')
 @cmdln.option('--strategy', help='apply a specific strategy')
+@cmdln.option('--match-filter', help='xpath by which to filter requests on the basis of the reviews they received', default=None)
 @cmdln.option('--no-color', action='store_true', help='strip colors from output (or add staging.color = 0 to the .oscrc general section')
 @cmdln.option('--remove-exclusion', action='store_true', help='unignore selected requests automatically', default=False)
 @cmdln.option('--save', action='store_true', help='save the result to the pseudometa package')
@@ -157,6 +158,14 @@ def do_staging(self, subcmd, opts, *args):
         changed from state new or review more than 3 days ago will be removed.
 
     "list" will list/supersede requests for ring packages or all if no rings.
+
+        By just calling list, the staging plugin will list all the request included
+        in the backlog section of the web UI. It is also possible to optionally limit
+        results with an XPATH filter. As an example, the following would list all
+        packages which have received a positive review from a member of the
+        licensedigger group or the factory-auto one
+
+        list --match-filter "state/@name='review' and review[(@by_group='factory-auto' or @by_group='licensedigger') and @state='accepted']"
 
     "lock" acquire a hold on the project in order to execute multiple commands
         and prevent others from interrupting. An example:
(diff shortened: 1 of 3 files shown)


osc-staging.py
```diff
@@ -157,6 +158,11 @@ def do_staging(self, subcmd, opts, *args):
         changed from state new or review more than 3 days ago will be removed.
 
     "list" will list/supersede requests for ring packages or all if no rings.
+        As an example, the following would list all packages which have received a
```
g7, line 161: I think we should make clear that this is not necessary (as `list` works fine without `--match-filter`).

osclib/stagingapi.py
```diff
@@ -587,10 +587,12 @@ def get_open_requests(self, query_extra=None):
         requests = []
 
         # xpath query, using the -m, -r, -s options
-        where = f"@by_group='{self.cstaging_group}' and @state='new'"
-        target = f"target[@project='{self.project}']"
+        if not match_filter:
+            target = f"target[@project='{self.project}']"
+            review_where = f"review[@by_group='{self.cstaging_group}' and @state='new']"
+            match_filter = f"state/@name='review' and {review_where} and {target}"
```
g7, line 593: should we expect the user to add the target project manually to the query when using the match_filter feature? Or can we populate it as well via `self.project`?  I ask since (IIRC) `self.project` is required anyway so the user might expect that (calling `staging -p <projectname> --match-filter "blabla"` might imply that the query should be run against the specified project)

osc-staging.py
```diff
@@ -157,6 +158,11 @@ def do_staging(self, subcmd, opts, *args):
...
+        positive review from a member of the licensedigger group or the factory-auto
+        one
+
+        list --match-filter "review[(@by_group='factory-auto' or @by_group='licensedigger') and @state='accepted']"
```
g7, line 165: Should we also add `state/@name='review'` to the filter here?

g7 (commented):
Thanks a lot for this nice feature! Just some minor nitpicks/questions

osc-staging.py
gleidi-suse: thanks. fixed

osclib/stagingapi.py
gleidi-suse: good catch, fixed.

osc-staging.py
gleidi-suse: sure, fixed. thank you!

osclib/stagingapi.py
```diff
@@ -585,12 +585,14 @@ def get_open_requests(self, query_extra=None):
...
 
         requests = []
+        target = f"target[@project='{self.project}']"
 
         # xpath query, using the -m, -r, -s options
-        where = f"@by_group='{self.cstaging_group}' and @state='new'"
-        target = f"target[@project='{self.project}']"
+        if not match_filter:
+            review_where = f"review[@by_group='{self.cstaging_group}' and @state='new']"
+            match_filter = f"state/@name='review' and {review_where}"
```
g7, line 593: I think we can fold `review_where` inside `match_filter`, but I don't have a strong opinion on that (and even previously it was split in `where` and `target` indeed).

g7 (approved):
Thanks a lot! I have a small comment but it's not really an issue :) Tested, works fine for me.

osclib/stagingapi.py
gleidi-suse: Thank you, I agree less bytes it is always better. Should be fixed now :)

PR #3187, opened 2024-11-01
Add Leap Micro 6.1 ttm and pkglist workflows

lkocman:
* keep it aside from 6.0 which is still being actively published to cover for all maintenance updates


gocd/leapmicro.target.gocd.yaml
```diff
@@ -56,3 +56,57 @@ pipelines:
...
+      OSC_CONFIG: /home/go/config/oscrc-staging-bot
+    stages:
+    - Create.Release.Packages:
+        approval: manual
+        resources:
+        - repo-checker
+        tasks:
+          - script: ./pkglistgen.py -A https://api.opensuse.org update_and_solve -p openSUSE:Leap:Micro:6.1 -s target --only-release-packages
+
+  LeapMicro61.Package.Lists:
```
lkocman, line 78: I think this would not be needed with __all__ see https://build.opensuse.org/requests/1221435

PR #3188, opened 2024-11-04
Added SUSE:SLFO prefix for Agama SLES 16

okynos:
We have added the prefix to the sync script. This will perform Agama SLES 16 synchronization in OSD.


Vogtinator:
LGTM, but I wonder how it worked for stuff inside SUSE:SLFO before this?

okynos:
Thank you @Vogtinator, in our case it never worked for `SUSE:SLFO:Products:SLES:16.0:TEST`

Vogtinator (approved):

PR #3189, opened 2024-11-07
Add 180 minutes timeout to SLFO product building pipelines.

gleidi-suse:


gocd/slfo-stagings.gocd.yaml
```diff
@@ -290,14 +304,21 @@ pipelines:
...
         tasks:
           - script: |-
+              minutes=$(expr 0)
               osc -A $STAGING_API api -X POST "/source/$STAGING_PROJECT?cmd=remove_flag&repository=product&flag=build"
               export PYTHONPATH=$PWD/scripts
               while ./scripts/gocd/verify-repo-built-successful.py -A $STAGING_API -p $STAGING_PROJECT -r product; ret=$?; [ ${ret} -ne 0 ]; do
                 if [ ${ret} -eq 2 ]; then
                     echo "product repository not found. Project configuration issue?" >&2
                     exit 1
+                elif [ ${ret} -ne 0 ] && ; then
```
DimStar77, line 314: `&&` is useless (if not even wrong) - there is no 2nd condition

DimStar77:
```
 if [ 0 -eq 0 ] && ; then
-bash: syntax error near unexpected token `;'
```

gocd/slfo-stagings.gocd.yaml.erb
```diff
@@ -274,8 +281,14 @@ pipelines:
                 if [ ${ret} -eq 2 ]; then
                     echo "product repository not found. Project configuration issue?" >&2
                     exit 1
+                elif [ ${ret} -ne 0 ]; then
```
g7, line 284: I think we can get rid of this and only keep the timeout check, as the exit status it's already checked before (line 280)

g7 (approved):
Looks good to me, thanks!

PR #3193, opened 2024-11-12
gocd: slfo-stagings: slfo-stagings: fail on broken/failed products (SLFO:Main)

g7:
Rather than using verify-repo-built-successful.py, check with the API directly whether the product is blocked, scheduled or building.

Since it might happen for a product being broken/failed (even after a successful pkglistgen run), just fail in that case.

This is #3191, but for SUSE:SLFO:Main rather than 1.1.


gleidi-suse (approved):
ok, thank you

PR #3197, opened 2024-12-05
publish Leap Micro 6.1 Vagrant boxes (Base and Default)

johanneskastl:


lkocman:
@johanneskastl seems like the linter pipeline did not pass. could you please look into it?

johanneskastl:
> @johanneskastl seems like the linter pipeline did not pass. could you please look into it?

Looking into it, hope the new commits fix the error.

johanneskastl:
@lkocman  Fixed!

gocd/vagrant-publisher.gocd.yaml
```diff
@@ -75,3 +75,15 @@ pipelines:
             tasks:
               - script: |
                   ruby obs-to-vagrantcloud.rb --url https://download.opensuse.org/distribution/leap/15.6/appliances/boxes/Leap-15.6.aarch64-libvirt_aarch64.json --organization opensuse --new-box-name Leap-15.6.aarch64
+          publish_Leap_Micro_6_1_Default_x86_64:
+            resources:
+            - staging-bot
+            tasks:
+              - script: |
+                  ruby obs-to-vagrantcloud.rb --url https://download.opensuse.org/distribution/leap-micro/6.1/appliances/openSUSE-Leap-Micro.x86_64-Default-Vagrant.json --organization opensuse --new-box-name Leap_Micro_6.1_Default.x86_64
```
Vogtinator, line 83: That URL is wrong, it points to some SBOM json file

Vogtinator (changes requested):

gocd/vagrant-publisher.gocd.yaml
johanneskastl: Ouch! You are of course correct. Looks like there is no `boxes` subdirectory for the Leap Micro 6.1 image.

johanneskastl: Could it be that we are missing an entry here:
https://build.opensuse.org/projects/openSUSE:Leap:Micro:6.1/prjconf

```
%if "%_repository" == "images"
Type: kiwi
Repotype: staticlinks
[...]
```

I think the `Repotype` should also contain `vagrant`, like [here](https://build.opensuse.org/projects/Virtualization:Appliances:Images:openSUSE-Leap-15.6/prjconf).

```
%if "%_repository" == "images"
Type: kiwi
Repotype: staticlinks vagrant
[...]
```

Vogtinator: Added. I noticed that totest-manager was misconfigured and it didn't find the openQA jobs, I fixed that in the config. The next published snapshot should have the proper .json files again.

(CC @lkocman)

Vogtinator: > (Fun question from when I was comparing prjconf's: Why is this not needed on devel:microos:images? My guess would be that there the images are published on another URL, hence everything is setup properly)

Good question. It's possible that this is not actually caused by `Repotype: vagrant` but the custom OBS -> d.o.o sync script ("publish_distro") does something weird here.

Vogtinator: Whenever openQA says so: https://openqa.opensuse.org/tests/overview?distri=leap-micro&version=6.1&build=5.9&groupid=102

kastl-ars: Thanks Fabian, I was not sure if the releases are frozen as there are so few updates for Leap Micro (something Lubos mentioned a while back)...

kastl-ars: Hmmm, OpenQA seems to be finished with the same result as 5.7, but I see no changed files. Apparently no new release was published.

johanneskastl: I totally forgot about this PR. But I just checked and I still cannot find any JSON files that are NOT sbom-related.

https://download.opensuse.org/distribution/leap-micro/6.1/appliances/

@lkocman  @Vogtinator Who would be the right person to ping?

Vogtinator: openQA still fails for Leap Micro 6.1, so nothing got published. Something for @lkocman.

PR #3203, opened 2025-01-13
gocd: checkers.suse: Drop Marble pipelines and add SLFO checkers.

gleidi-suse:
Since the switch to Git backed sources for SL-Micro 6.0, staging
workflow does not work anymore for SUSE:ALP:Products:Marble:6.0. This
patch drops all related pipelines. Additional it adds the same checkers
for the SUSE:SLFO:Main project.

### DIFF
--- a/gocd/checkers.suse.gocd.yaml.erb
+++ b/gocd/checkers.suse.gocd.yaml.erb
@@ -62,7 +62,7 @@ pipelines:
             - repo-checker
             tasks:
             - script: ./project-installcheck.py -A https://api.suse.de --debug check --store SUSE:SLE-15-SP7:GA:Staging/dashboard --no-rebuild SUSE:SLE-15-SP7:GA
-  SL-Micro.Project:
+  SUSE.SLFO.Main.Staging.Project:
     group: SLE.Checkers
     lock_behavior: unlockWhenFinished
     timer:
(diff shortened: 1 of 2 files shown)


gocd/checkers.suse.gocd.yaml.erb
```diff
@@ -78,11 +81,11 @@ pipelines:
         timeout: 30
         approval: manual
         jobs:
-          SL-Micro:
+          SLFO.Main:
             resources:
             - repo-checker
             tasks:
-            - script: ./project-installcheck.py -A https://api.suse.de --debug check --store SUSE:ALP:Products:Marble:6.0:Staging/dashboard --no-rebuild SUSE:ALP:Products:Marble:6.0
+            - script: ./project-installcheck.py -A $STAGING_API --debug check --store $SLFO_PROJECT:Staging/dashboard --no-rebuild $SLFO_PROJECT
```
g7, line 88: If I understood this correctly, the installcheck is ran on the main project, and not into the staging right? (the result is stored indeed in the "parent" Staging project -> dashboard).

If so, I think this needs to look at SUSE:SLFO:Main:Build instead

g7: (only for this script, the others should use SUSE:SLFO:Main)

gleidi-suse: good catch, fixed!

gocd/checkers.suse.gocd.yaml.erb
```diff
@@ -62,12 +62,15 @@ pipelines:
...
             tasks:
             - script: ./project-installcheck.py -A https://api.suse.de --debug check --store SUSE:SLE-15-SP7:GA:Staging/dashboard --no-rebuild SUSE:SLE-15-SP7:GA
-  SL-Micro.Project:
+  SUSE.SLFO.Main.Staging.Project:
     group: SLE.Checkers
     lock_behavior: unlockWhenFinished
     timer:
       spec: 0 0 * ? * *
       only_on_changes: false
+    environment_variables:
```
g7, line 71: `environment_variables:` is specified later at line 77 (catched by https://github.com/openSUSE/openSUSE-release-tools/actions/runs/12751048132/job/35537169000?pr=3203)

g7 (changes requested):
One comment, rest looks good to me, thanks!

gocd/checkers.suse.gocd.yaml.erb
gleidi-suse: I completely forgot to `./generate.sh`, thank you!

g7 (approved):
LGTM. Thanks!

PR #3208, opened 2025-03-10
Refactor check_source to work with Gitea Runners

YoukouTenhouin:
As a preliminary step for making bots work with Gitea Runners, this PR refactors the check_source bot to let it runs on Runners, while keep it original functionalities and the ability to work with OBS as-is.

A demo for this resides in https://src.opensuse.org/yudaike/release-tools-test


YoukouTenhouin:
Superseded by #3247

PR #3209, opened 2025-03-12
packaging: Do not reference python3-osc

DimStar77:
This is only a symbol provided by the actual osc package.

One of the main issues attempted to be addressed was ensuring that the
plugin matches the python version used by osc, but the symbol
'python3-*' does not guarantee that at all, as the symbol just moves between
python-flavors, based on primary python.

### DIFF
--- a/dist/package/openSUSE-release-tools.spec
+++ b/dist/package/openSUSE-release-tools.spec
@@ -35,7 +35,6 @@ BuildRequires:  python3-PyYAML
 BuildRequires:  python3-cmdln
 BuildRequires:  python3-colorama
 BuildRequires:  python3-lxml
-BuildRequires:  python3-osc
 BuildRequires:  python3-pycurl
 BuildRequires:  python3-python-dateutil
 BuildRequires:  python3-pyxdg
@@ -269,7 +268,6 @@ Group:          Development/Tools/Other
 # TODO Update requirements, but for now base deps.
 Requires:       %{name} = %{version}
 Requires:       osc >= 0.165.1
-Requires:       python3-osc
 BuildArch:      noarch
 
 %description -n osclib


DimStar77:
Lovely - that python3-influxdb is not caused by this SR, but rather due to the fact that python-influxdb does not build for python3.13 (%skip_python313) and is already the case in published TW snapshots

Vogtinator:
> This is only a symbol provided by the actual osc package.

i.e. this PR is a noop in any case?

> the symbol 'python3-*' does not guarantee that at all, as the symbol just moves between
python-flavors, based on primary python.

If the release tools use a generic `python3` shebang, then using the `python3-*` dependencies is AFAICT correct

Vogtinator:
> > > This is only a symbol provided by the actual osc package.
> > 
> > 
> > i.e. this PR is a noop in any case?
> 
> Sadly - no... there was an osc change to stricter be pythonxyz related; it no longer provides python3-osc, but only python313-osc (I'm also talking to the author of that change to possibly get the py3- symbol back too)

Ok, so that's just broken osc packaging IMO

DimStar77:
> Ok, so that's just broken osc packaging IMO

Don't disagree - but us requiring python3-osc which is essentially osc does not make sense neither

Vogtinator:
> Don't disagree - but us requiring python3-osc which is essentially osc does not make sense neither

It does: If osc for some reason uses some non-primary python and then only requires e.g. python312-osc, the osc module for primary python would be missing and release-tools wouldn't work despite osc being there.

dirkmueller:
> > Sadly - no... there was an osc change to stricter be pythonxyz related; it no longer provides python3-osc, but only python313-osc (I'm also talking to the author of that change to possibly get the py3- symbol back too)
> Ok, so that's just broken osc packaging IMO

I think that is actually correct... why would it be broken?

dirkmueller:
> It does: If osc for some reason uses some non-primary python and then only requires e.g. python312-osc, the osc module for primary python would be missing and release-tools wouldn't work despite osc being there.

can we cross that bridge when we get there? I can't predict a future where we have an osc package that doesn't work with primary python.

dirkmueller (approved):

DimStar77:
the 'large/remaining' issue which has come up is missing python3-influxdb (metrics) - that's out of osc's control, but makes one of the subpackages here uninstallable

Vogtinator:
> > > Sadly - no... there was an osc change to stricter be pythonxyz related; it no longer provides python3-osc, but only python313-osc (I'm also talking to the author of that change to possibly get the py3- symbol back too)
> > > Ok, so that's just broken osc packaging IMO
> 
> I think that is actually correct... why would it be broken?

Because the python3-osc symbol that needs to be provided by all python3 modules is missing?

dirkmueller:
> the 'large/remaining' issue which has come up is missing python3-influxdb (metrics) - that's out of osc's control, but makes one of the subpackages here uninstallable

it also makes the CI unbuildable so we need to do something about it. see https://github.com/openSUSE/openSUSE-release-tools/pull/3210

dirkmueller:
> Because the python3-osc symbol that needs to be provided by all python3 modules is missing?

I'd argue it isn't actually needed but we don't need to go there..

darix:
there is another way more important reason why you should not use python3-* as requires, which was the original reason why i touched osc.

lets say you patch your shebang lines with the macro to be version specific, as you should. if all your requires are against python3-* it means you might have a package that has /usr/bin/python311 as shebang line, but already pulls python313-* packages as dependency, because the python3-* provides moved to the python313 packages.

so a much cleaner approach is

```
%global pythons %{primary_python}

Requires: %{primary_python}-osc
```

that way your shebang line and the requires should actually target the same python version.

Vogtinator:
> lets say you patch your shebang lines with the macro to be version specific, as you should

As long as we don't, using `python3-*` is correct though.

darix:
> As long as we don't, using `python3-*` is correct though.

plus as long you don't install any files into the python library paths either.  because then your own python search path would be wrong.

dirkmueller:
(just a rebase)

dirkmueller:
@DimStar77 so I think we could move forward as is, or add the fix_shebang on top to address @darix input. wdyt?

PR #3167, opened 2024-09-26
Move metrics to InfluxDB v2 python client

bnavigator:
Fixes #3034

PR #3210, opened 2025-03-12
Remove the influxdb based metrics

dirkmueller:
this is dependant on influxdb 1.x which is end of life. the code would have to be conveted to influxdb-client-python instead, but it turns out no data has been collected the last 1+ years according to

  https://metrics.opensuse.org/d/osrt_staging/osrt-staging?orgId=1&from=now-1y&to=now&timezone=browser&var-project=000000001

so it appears we can just remove it.


Vogtinator:
There's https://github.com/openSUSE/openSUSE-release-tools/pull/3167

dirkmueller:
alright, lets do the upgrade then instead.

PR #3212, opened 2025-03-13
Add styles to openQA results

bmwiedemann:


DimStar77:
care to paste a screenshot of the changed design?

DimStar77 (approved):
Generally looking good - some color won't hurt us :)

DimStar77:
image

dashboard/templates/dashboard.html
```diff
@@ -13,6 +13,12 @@
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
 
   <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css" integrity="sha384-Vkoo8x4CGsO3+Hhxv8T/Q5PaXtkKtu6ug5TOeNV6gBiFeWPGFN9MuhOf23Q9Ifjh" crossorigin="anonymous">
+<style type="text/css">
```
Vogtinator, line 16: Some indentation would be nice

Vogtinator (approved):

dashboard/templates/dashboard.html
bmwiedemann: Done.

dashboard/templates/dashboard.html
```diff
@@ -102,8 +108,8 @@
         </td>
         <td>
         {% set openqa_summary = project.openqa_summary() %}
-        {% for state in openqa_summary.keys() %}
-          {{ state }}: {{ openqa_summary[state]|length }}
+        {% for state in openqa_summary.keys()|sort %}
```
DimStar77, line 111: alphabetical sort of the openQA states  is not in our interest

DimStar77 (changes requested):
Sorted states are newly added and not in our interest

dashboard/templates/dashboard.html
bmwiedemann: hmm... it looked like random order without this. In the screenshot above you can see `failed` before `softfailed` in the ARM line and the other way in the Factory line. Or is this order provided in some other way?

DimStar77: when ordering, then at a specified order which helps 'more' than alphabetical I'd say (passed, softfailed, failed, parallel_failed, cancelled, user_canceller)

'passed' as first is interestingly enough constant throughout the dashboard even without sorting

bmwiedemann:
One test fails because `No provider of 'python3-influxdb' found` because the `python-influxdb` source package only produced `python311-influxdb` even after the switch to python313.

dashboard/templates/dashboard.html
Vogtinator: IMO sorted in any order is an improvement over random

ana:
The colours looks good to me.

PR #3213, opened 2025-03-14
gocd: add a pipeline for the Leap 16 offline installer packagelist generating

nilxam:


nilxam:
python3-influxdb error is unrelated.

PR #3217, opened 2025-03-31
check_source.py: Devel project check for new packages

AdamMajer:
We are missing check for devel projects for new packages where the source project is in git and it's no longer searchable via API

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -16,7 +16,7 @@ from lxml import etree as ET
 import osc.conf
 import osc.core
 from osclib.conf import Config
-from osclib.core import devel_project_get
+from osclib.core import devel_project_get, factory_git_devel_project_mapping
 from osclib.core import devel_project_fallback
 from osclib.core import entity_exists
 from osclib.core import group_members
@@ -350,6 +350,10 @@ class CheckSource(ReviewBot.ReviewBot):
             return True
 
         # Allow any projects already used as devel projects for other packages.
+        if target_project.endswith('openSUSE:Factory'):
+            devel_pkgs = factory_git_devel_project_mapping(self.apiurl)
+            return True if source_project in devel_pkgs.values() else False
+
         search = {
             'package': f"@project='{target_project}' and devel/@project='{source_project}'",
         }
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -4,7 +4,7 @@ from dateutil.parser import parse as date_parse
 import re
 import socket
 import logging
-from typing import List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 try:
     from typing import Literal
 except ImportError:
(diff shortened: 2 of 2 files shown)


check_source.py
```diff
@@ -350,6 +350,10 @@ def is_devel_project(self, source_project, target_project):
             return True
 
         # Allow any projects already used as devel projects for other packages.
```
dirkmueller, line 352: so this comment was for the section below, can you please keep it there?

check_source.py
```diff
@@ -350,6 +350,10 @@ def is_devel_project(self, source_project, target_project):
...
+        if target_project.endswith('openSUSE:Factory'):
+            devel_pkgs = git_factory_devel_project_dict_get(self.apiurl)
+            return True if source_project in devel_pkgs.values() else False
```
dirkmueller, line 355: the "else False" disables the entire logic afterwards. are we planning to do this today rather than an opt in?

osclib/core.py
```diff
@@ -288,19 +288,33 @@ def binary_src_debug(binary):
 
 
 @memoize(session=True)
-def devel_project_get(apiurl: str, target_project: str, target_package: str) -> Union[Tuple[str, str], Tuple[Literal[None], Literal[None]]]:
-    """Fetch the devel project & package name of the supplied ``target_project``
-    and ``target_package`` and return the devel project and devel package name
-    as a tuple. If no devel project has been defined, then return ``None,
-    None``.
-
+def git_factory_devel_project_dict_get(apiurl: str) -> Dict[str, str]:
```
dirkmueller, line 291: I don't like the `_dict_get` suffix, the get if at all should be a verb at the beginning. 

how about `git_factory_pkg_project_mapping` ?

osclib/core.py
```diff
@@ -312,16 +326,10 @@ def devel_project_get(apiurl: str, target_project: str, target_package: str) ->
...
-                    pkg, _, prj = line.decode('utf-8').strip().partition(' ')
-                    if pkg == target_package:
-                        return prj, pkg
-        except HTTPError as e:
-            if e.code != 404:
-                raise e
+        devel_pkgs = git_factory_devel_project_dict_get(apiurl)
+        devel_prj = devel_pkgs.get(target_package)
+        if devel_prj is not None:
+            return devel_prj, target_package
```
dirkmueller, line 332: you could also do something like:

```suggestion
            if target_package in devel_pkgs:
                 return devel_pkgs[target_package], target_package
                 
```

osclib/core.py
AdamMajer: `factory_git_devel_projects_mapping` ?

check_source.py
AdamMajer: This is already a done deal. If OBS cannot store some of the package devel project info correctly, it may as well not store any of the information there anymore.

Furthermore, the attribute that sets devel projects in Factory is turned off for a few weeks as it's not possible to accept git-based new packages to Factory.

check_source.py
AdamMajer: This section is amended for Factory hence I put it above.

check_source.py
dirkmueller: well, it means that new package additions need to add their devel project here to this project first. did we communicate that?

AdamMajer: I see... my mistake. This is something else and you are correct, the `else False` is thus wrong here. We still want this attribute...

AdamMajer: So, I think there's nothing really to add here. New devel projects still need whitelisting before packages can pass this check. The whitelist is above.

For packages that come from devel projects for other packages, this is the change we want. We store the informatino on the "backend" in different place now than the XML of the package, that's all.

For non-openSUSE:Factory packages, this fallsthrough.

check_source.py
AdamMajer: Just to add, the new code is "special handling" of new Factory as we transition this mechanic to git. The others uses (if any) are not affected here.

dirkmueller (approved):

PR #3219, opened 2025-04-03
verify-build-and-generatelists: Fix pkglistgen command line arguments for SLFO based products.

gleidi-suse:
For SL-Micro is currently failing with:

```
[CHECKING] Checking standard repository from SUSE:SLFO:Products:SL-Micro:6.1
2025-03-27T00:15:00+01:00
usage: verify-repo-built-successful.py [-h] [--apiurl APIURL] -p PROJECT -r
                                       REPOSITORY
verify-repo-built-successful.py: error: unrecognized arguments: --engine product_composer --git-url https://src.suse.de/products/SL-Micro#6.1
```

PR #3220, opened 2025-04-03
verify-build-and-generate: For SLFO based products, wait for SUSE:SLFO:Main:Build to be done building.

gleidi-suse:
Depends on https://github.com/openSUSE/openSUSE-release-tools/pull/3219


g7 (approved):
Thanks, LGTM. For posterity (we already agreed with @gleidi-suse via chat): this will of course stop working once we will branch off Main for the actual product release

So we should either
1) disable the bot for the specific project on branching time (which is not a big deal, as we would need to do that anyway before handing it over, so it only anticipates the fact by some weeks)
2) modify the script so that it looks at the paths in the project meta <-- this might add complexity, so not sure how worth it is (but would be nice :smile: )

PR #3222, opened 2025-04-10
verify-build-and-generatelists: Update default Gitea URL to use SSH authentication

gleidi-suse:


g7 (approved):
Thanks, makes sense to me.

PR #3223, opened 2025-04-15
ttm: add a new match pattern for the package naming in Leap 16

nilxam:
pkglistgen with productcompose engine uses underscore `_` in most places rather than a dash symbol, like the flavor name is consist of {group name}_{arch}, along with that, flavor and packageset group has underscore symbol in general.

### DIFF
--- a/ttm/releaser.py
+++ b/ttm/releaser.py
@@ -87,13 +87,13 @@ class ToTestReleaser(ToTestManager):
                                       arch=self.project.image_products[0].archs[0])
 
     def maxsize_for_package(self, package, arch):
-        if re.match(r'.*-mini-.*', package):
+        if re.match(r'.*[-_]mini[-_].*', package):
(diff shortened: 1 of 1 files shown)


lkocman (approved):

ttm/releaser.py
```diff
@@ -87,13 +87,13 @@ def version_from_project(self):
                                       arch=self.project.image_products[0].archs[0])
 
     def maxsize_for_package(self, package, arch):
-        if re.match(r'.*-mini-.*', package):
+        if re.match(r'.*-mini-.*', package) or re.match(r'.*_mini_.*', package):
```
Vogtinator, line 90:
```suggestion
        if re.match(r'.*[-_]mini[-_].*', package):
```

and so on

Vogtinator (approved):

PR #3224, opened 2025-04-22
maintenance bot: add support for git-based devel projects

AdamMajer:
fixes maintenance requests like 1268814 where devel project information is only in git and the `/search/owner` API call is not very useful anymore.

https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/6ILLOYQ3W4IWTA47WTXBANZZGLHTFYBB/

### DIFF
--- a/check_maintenance_incidents.py
+++ b/check_maintenance_incidents.py
@@ -31,16 +31,19 @@ class MaintenanceChecker(ReviewBot.ReviewBot):
             a = req.actions[1]
         project = a.tgt_releaseproject if a.type == 'maintenance_incident' else req.actions[0].tgt_project
 
-        if project.startswith('openSUSE:'):
+        # First we look if devel project is defined and request review from it
+        if req.actions[0].tgt_project.startswith('openSUSE:Maintenance'):
             prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package)
-            if prj is None and package.find(".") > 0:
-                prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package[0:package.rfind(".")])
+            # period in package name as created by "osc mbranch"
+            if prj is None and '.' in package:
+                prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package.rpartition('.')[0])
             logging.debug(f'using devel project {prj}/{pkg}')
             if prj is not None:
                 msg = f'Submission for {pkg} by someone who is not maintainer in the devel project ({prj}). Please review'
                 self.add_review(req, by_project=prj, by_package=pkg, msg=msg)
                 return
 
+        # no devel project -- fallback to /search/owner?package -- OBS side "owner" prj/pkg search with pkg only fallback
         root = owner_fallback(self.apiurl, project, package)
 
         for p in root.findall('./owner'):
--- a/check_maintenance_incidents.py
+++ b/check_maintenance_incidents.py
@@ -1,5 +1,6 @@
 #!/usr/bin/python3
 
+import logging
 import sys
 
 import osc.conf
@@ -8,7 +9,7 @@ from urllib.error import HTTPError, URLError
 import yaml
 
 from osclib.memoize import memoize
-from osclib.core import action_is_patchinfo
+from osclib.core import action_is_patchinfo, devel_project_get
 from osclib.core import owner_fallback
 from osclib.core import maintainers_get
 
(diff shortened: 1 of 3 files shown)


check_maintenance_incidents.py
```diff
@@ -29,6 +30,17 @@ def add_devel_project_review(self, req, package):
         if action_is_patchinfo(a):
             a = req.actions[1]
         project = a.tgt_releaseproject if a.type == 'maintenance_incident' else req.actions[0].tgt_project
+
+        if project.startswith('openSUSE:'):
```
dirkmueller, line 34: In the request that was quoted in the mailthread this project should be `openSUSE:Maintenance` and that probably is always the case. why are we testing with startswith(openSUSE:) which is very broad rather than exactly the project name?

check_maintenance_incidents.py
```diff
@@ -29,6 +30,17 @@ def add_devel_project_review(self, req, package):
...
+            prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package)
+            if prj is None and package.find(".") > 0:
+                prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package[0:package.rfind(".")])
+            logging.debug(f'using devel project {prj}/{pkg}')
+            if prj is not None:
```
dirkmueller, line 39: that block seems to be a duplication of lines https://github.com/openSUSE/openSUSE-release-tools/pull/3224/files#diff-877efb0a9116edf680121175667b645eabb70f056d9d4c91b641377a39269380R49-R54 

why are we not implementing this logic in owner_fallback (by e.g. changing it to return proj, package instead of the entire xml tree..

check_maintenance_incidents.py
AdamMajer: Well, this is for maintenance check. The `/search/owner` backend here fails for every package that is in factory and doesn't have devel project set in OBS. And AFAIK, we do not have devel project configured for other projects than openSUSE:Factory. So, everything that is sent as maintenance update to any openSUSE: project that uses this bot will no longer work if openSUSE:Factory devel projects are not set.

If you would like to limit this to `openSUSE:Maintenance` only, that's fine and I can change this accordingly, but `/search/owner` will not really function against other projects (we have the attribute for it set in only one owning project, openSUSE:).

check_maintenance_incidents.py
AdamMajer:
```
check_maintenance_incidents.py:from osclib.core import owner_fallback
check_maintenance_incidents.py:        root = owner_fallback(self.apiurl, project, package)
osclib/core.py:def owner_fallback(apiurl, project, package):
osclib/core.py:    root = owner_fallback(apiurl, project, package)
```

The owner_fallback is basically only used here and it returns devel project and maintainer information, not not devel project info. Re-implementing OBS XML output here is not really right approach.

AdamMajer: And yes, it's duplication, because it's simplest solution without need to create additional complexity.

dirkmueller: I'd prefer if line 35-39 would be cacheable via memoize rather than repeated lookups. here's what we could do: 

* fold the existing owner_fallback implementation into `maintainers_get` (inline it). 
* change owner_fallback to just return a `project/package` tuple. 
* move line 35-39 into `owner_fallback`
*

AdamMajer: Then maybe we can rename owner_fallback to "find_devel_project"?

check_maintenance_incidents.py
```diff
@@ -29,6 +30,17 @@ def add_devel_project_review(self, req, package):
         if action_is_patchinfo(a):
             a = req.actions[1]
         project = a.tgt_releaseproject if a.type == 'maintenance_incident' else req.actions[0].tgt_project
+
+        if req.actions[0].tgt_project.startswith('openSUSE:Maintenance'):
+            prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package)
+            if prj is None and package.find(".") > 0:
```
dirkmueller, line 36:
```suggestion
            if prj is None and "." in package:
```

check_maintenance_incidents.py
```diff
@@ -29,6 +30,17 @@ def add_devel_project_review(self, req, package):
...
+                prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package[0:package.rfind(".")])
```
dirkmueller, line 37:
```suggestion
                prj, pkg = devel_project_get(self.apiurl, "openSUSE:Factory", package.rpartition('.')[0])
```

dirkmueller (approved):

PR #3228, opened 2025-04-30
Remove EOL Leap 15.4 and 15.5

Vogtinator:


okurz (approved):

gocd/totestmanager.gocd.yaml.erb
```diff
@@ -7,11 +7,7 @@ pipelines:
       openSUSE:Factory:PowerPC
       openSUSE:Factory:RISCV
       openSUSE:Factory:zSystems
-      openSUSE:Leap:15.5:Images
-      openSUSE:Leap:15.5:ARM
-      openSUSE:Leap:15.5:ARM:Images
       openSUSE:Leap:15.6:Images
-      openSUSE:Leap:15.6:ARM
```
foursixnine, line 14: Out of curiosity, was this also intentional?

Vogtinator: Yes. 15.6 was released a long time ago so those projects are no longer used.

PR #3229, opened 2025-04-30
buildids per arch can be different

lethliel:
don't take the buildid from the first arch.

### DIFF
--- a/publish_distro_conf/publish_leap-micro61.config
+++ b/publish_distro_conf/publish_leap-micro61.config
@@ -10,15 +10,12 @@ flavors=(x86_64 aarch64)
 
 # openSUSE-Leap-Micro-6.1-aarch64-Build13.1.iso is the filename by 000productcompose
 get_version() {
-	# get expected version from first flavor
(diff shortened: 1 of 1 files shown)


Vogtinator (commented):
Breaks intendation?

lethliel:
> Breaks intendation?

Yes.. I am working on this.. But I am not sure why intendation is a thing in a config file

lethliel:
Ok. Intendation fixed

Vogtinator:
> config file

*shell script disguised as config :P

Vogtinator:
IIRC it's meant to have synced buildids, so they should never deviate.

@lkocman ?

lethliel:
> IIRC it's meant to have synced buildids, so they should never deviate.
> 
> @lkocman ?

Yes this was the original plan. But they were different for 6.0 already.

lkocman:
Correct  16.0 was setup with a bind mount from ftp-stage so they're 1:1.

lethliel:
@lkocman So this PR is not needed anymore, when the build numbers are kept in sync?

PR #3231, opened 2025-05-05
check_source: Do not try to add maintainer review when maintainer not…

DimStar77:
… known

Especially in cases where the devel project has already removed the package, we cannot find information about the maintainer


DimStar77:
In the past, we never had an issue with that. When a devel package was removed, then the `<devel>` attribute on the package in Factory was also removed and the code thus did not have a developed-at link to follow. With the devel project list maintained in a text file in git, this does not happen anymore when q devel package is being removed. In turn, the bot still finds information about a devel project entry, tries to follow it just to not find the package there in the end.

ReviewBot.py
```diff
@@ -408,7 +408,11 @@ def devel_project_review_needed(self, request, project, package):
         if len(maintainers) == 0:
             devel_project, devel_package = devel_project_fallback(self.apiurl, project, package)
             if devel_package:
-                maintainers = set(package_role_expand(self.apiurl, devel_project, devel_package))
+                try:
+                    maintainers = set(package_role_expand(self.apiurl, devel_project, devel_package))
+                except HTTPError:
+                    self.logger.debug('Cannot find maintainer information')
+                    return False
```
Vogtinator, line 415: Is it a good idea to have `False` as a fallback here? Spurious failures would result in a missing review.

DimStar77: well, not returning false wants to add a maintainer which we cannot identify, as we have no devel project.

What could make sense is ensure we get a 404 - which is unlikely to happen when we get a server error (500 / 403)

Vogtinator (approved):

PR #3240, opened 2025-06-06
osclib: core: explicitly replace OBS instance prefix in devel_project_fallback

g7:
When getting the devel project via the build service API, and the source project is accessed via interconnect, the returned devel project contains the instance prefix.

However, when checking directly in the Factory git mapping, the interconnected prefix is not being added to the result.

As such, ensure we handle this case too - and just replace the instance name if found.

### DIFF
--- a/osclib/core.py
+++ b/osclib/core.py
@@ -345,7 +345,7 @@ def devel_project_fallback(apiurl, target_project, target_package):
             project, package = devel_project_get(apiurl, 'openSUSE.org:openSUSE:Factory', target_package)
             if project:
                 # Strip openSUSE.org: prefix since string since not used for lookup.
-                project = project.split(':', 1)[1]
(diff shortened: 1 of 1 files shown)


g7:
I'm not a big fan of having `devel_project_get` have two different behaviours, i.e. the prefix is added when looking at the devel project via the build service API, and it isn't when looking at `_meta/devel_packages`.

But I'm not sure if it's worth it to make `factory_git_devel_project_mapping()` aware of this rather than doing this change.

gleidi-suse:
failures of tests, look unrelated. I agree that `devel_project_get`'s behavior is very confusing but I would leave the implementation of something like your `factory_git_devel_project_mapping` proposal for a different PR. LGTM

gleidi-suse (approved):

dirkmueller (approved):

Issue #3241, opened 2025-06-09
check_source get a dated devel project/package data from git when that package was deleted in devel project


nilxam:
check_source fails to check openQA SR https://botmaster.suse.de/go/tab/build/detail/openSUSE.Source.Check/551271/Run/1/Run
`
[I] checking 1283749
[I] openSUSE:Factory/openQA@695 -> openSUSE:Leap:16.0/openQA
[I] Checking required maintainer from the source project ()
[I] target package does not exist openSUSE:Leap:16.0/openQA
Traceback (most recent call last):
  File "/godata/pipelines/openSUSE.Source.Check/ReviewBot.py", line 215, in check_requests
    good = self.check_one_request(req)
  File "/godata/pipelines/openSUSE.Source.Check/ReviewBot.py", line 470, in check_one_request
    ret = func(req, a)
  File "/godata/pipelines/openSUSE.Source.Check/ReviewBot.py", line 562, in check_action_submit
    return self.check_source_submission(a.src_project, a.src_package, a.src_rev, a.tgt_project, a.tgt_package)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/godata/pipelines/openSUSE.Source.Check/./check_source.py", line 316, in check_source_submission
    maintainers = set(package_role_expand(self.apiurl, devel_project, devel_package))
                      ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/godata/pipelines/openSUSE.Source.Check/osclib/memoize.py", line 159, in _fn
    value = fn(*args, **kwargs)
  File "/godata/pipelines/openSUSE.Source.Check/osclib/core.py", line 158, in package_role_expand
    meta = ET.fromstringlist(show_package_meta(apiurl, project, package))
                             ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/osc/core.py", line 1346, in show_package_meta
    f = http_GET(url)
  File "/usr/lib/python3.13/site-packages/osc/connection.py", line 405, in http_GET
    return http_request("GET", *args, **kwargs)
[...]

PR #3242, opened 2025-06-09
Return None if devel package does not exist on the OBS according to the stored data on the git

nilxam:
Fixes https://github.com/openSUSE/openSUSE-release-tools/issues/3241


nilxam:
None of these failed tests is because of the change here, I've opened issue ticket of the each failure.

PR #3249, opened 2025-07-17
staging-installcheck: allow checking random devel projects

DimStar77:
By calling the install checker as
> ./staging-installcheck.py -p home:dimstar:Factory --devel home:dimstar:Factory -r openSUSE_Factory

We can allow developers to use installcheck against devel projects

### DIFF
--- a/staging-installcheck.py
+++ b/staging-installcheck.py
@@ -173,10 +173,8 @@ class InstallChecker(object):
             args = args.replace(',', ' ').split(' ')
         return set(args)
 
-    def staging(self, project, force=False):
+    def staging(self, project, repository, force=False, devel=False):
         api = self.api
-
-        repository = self.api.cmain_repo
 
         # fetch the build ids at the beginning - mirroring takes a while
         buildids = {}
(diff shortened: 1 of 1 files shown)


Vogtinator:
Don't we have `project-installcheck.py` for this already?

DimStar77:
> Don't we have `project-installcheck.py` for this already?

doesn't do what one would expect - it's actually 'just' the bot running after missing installdeps and triggering rebuilds; I got mislead by that one too

Vogtinator:
> doesn't do what one would expect - it's actually 'just' the bot running after missing installdeps and triggering rebuilds; I got mislead by that one too

It does run installcheck and triggering rebuilds is optional

Vogtinator (commented):
One `)` too many, LGTM otherwise

dirkmueller:
ok, I created https://github.com/openSUSE/openSUSE-release-tools/pull/3252 with just the CI fixes also until I realized I no longer have access to the workflows. JFYI, it has a slightly different approach to fixing the metrics error.

PR #3250, opened 2025-07-18
TTM: do allow to override product_repo for ftp

nilxam:
Override `product_repo` for ftp product if the format follows **PRODUCT/product_repo:REPONAME**, eg. `0openSUSE/product_repo:product`

Ftp tree build will migrate to _productcomposer_ in Factory which designed to have individual repository for oss, and non-oss, and overall it can not use images repo since product-builder is still up for those products.


nilxam:
A draft for review...

nilxam:
Hmm... It's not work enough for publisher, will work on it later

PR #3251, opened 2025-07-22
check_maintenance_incidents.py: Expand devel pkg roles for maintainer check

Vogtinator:
Similar to #3041

https://build.opensuse.org/requests/1294764

> Submission for etcd by someone who is not maintainer in the devel project (devel:kubic). Please review

It uses `/search/owner` to determine the maintainers, but that does not expand roles recursively. If the submitter is not one of the owners, it adds a prj/pkg review which is recursive.

### DIFF
--- a/check_maintenance_incidents.py
+++ b/check_maintenance_incidents.py
@@ -12,6 +12,7 @@ from osclib.memoize import memoize
 from osclib.core import action_is_patchinfo, devel_project_get
 from osclib.core import owner_fallback
 from osclib.core import maintainers_get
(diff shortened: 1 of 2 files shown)


dirkmueller (approved):

dirkmueller:
https://github.com/openSUSE/openSUSE-release-tools/pull/3252 should be fixing the CI failures..

Vogtinator:
> I think by now the CI failure is actually related to the PR

Yeah, some more mocks had to be added.

That wasn't everything though: It turned out that it only ever passed by accident - I fixed that.

PR #3257, opened 2025-08-06
ttm/releaser.py: Fix check for config completeness

Vogtinator:
all_built_products_in_config got broken by the introduction of product_repo_overrides.

CC @DimStar77 @nilxam

With those, ttm complains as expected again:

```
ERROR:ttm.manager:agama-installer:openSUSE_PXE is built for s390x, but not mentioned as product
ERROR:ttm.manager:openSUSE-MicroOS:s390x-Cloud is built for s390x, but not mentioned as product
ERROR:ttm.manager:openSUSE-MicroOS:s390x-dasd is built for s390x, but not mentioned as product
```


DimStar77:
Thanks for catching that!

PR #3264, opened 2025-09-26
Refactor totest-manager for more flexibility

Vogtinator:
Remove some obsolete code and perform some simplifications (in separate commits) before refactoring ttm's core logic to be based on per-product attributes. The fixed classes (main, ftp, livecd, images, container, containerfile) specify default values for those, so this is backwards-compatible.

Draft because:

- [ ] Make linters happy
- [ ] Complete documentation


nilxam (approved):
I like the unified method to get product version, well done!

LGTM

PR #3266, opened 2025-09-29
gocd: Add SLES 16.1 pipelines.

gleidi-suse:


g7 (approved):
Thanks, changes look good to me :+1:

PR #3285, opened 2026-01-08
check_bugowner: Add Git workflow compatible bugowner checker.

gleidi-suse:
Implementation of the bugowner checker functionality for the Git workflow. The new functionality can be tested against a given PR with:

```
GITEA_ACCESS_TOKEN="YOUR-TOKEN" GITEA_ACCESS_USER="YOUR-user" python check_bugowner.py --user "$GITEA_BOT_USER" --scm git --platform gitea --git-base-url "https://$GITEA_BOT_USER:$GITEA_ACCESS_TOKEN@src.suse.de/" --gitea-url "https://src.suse.de" --verbose id $PR_OWNER:$PR_REPO:$PR_ID
```

Depends on https://github.com/openSUSE/openSUSE-release-tools/pull/3291

### DIFF
--- a/check_bugowner.py
+++ b/check_bugowner.py
@@ -2,9 +2,15 @@
 
 # SPDX-License-Identifier: MIT
 
+import json
+import os
+import shutil
 import sys
 import re
+from pathlib import Path
+from typing import Set
 
+from git.config import GitConfigParser
 from urllib.error import HTTPError
 
 import osc.conf
@@ -12,6 +18,7 @@ import osc.core
 import ReviewBot
 
 http_GET = osc.core.http_GET
+MAINTAINERSHIP_FILE = "_maintainership.json"
 
 
 class CheckerBugowner(ReviewBot.ReviewBot):
@@ -21,7 +28,7 @@ class CheckerBugowner(ReviewBot.ReviewBot):
         self.request_default_return = True
         self.override_allow = False
 
-    def check_source_submission(self, src_project, src_package, src_rev, target_project, target_package):
+    def _obs_check_source_submission(self, src_project, src_package, src_rev, target_project, target_package):
         self.logger.info("%s/%s@%s -> %s/%s" % (src_project,
                                                 src_package, src_rev, target_project, target_package))
         if src_package.startswith('patchinfo'):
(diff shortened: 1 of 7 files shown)


g7 (commented):
Thanks! Two nitpicks but your specific changes look good to me.

Perhaps something that we should discuss in the other PR would be if it would be possible to have an actual `PlatformRequest` class that can implement `tgt_rev` (and others) that can be used interchangeably by both platforms (in this case, my suggestion to rename `tgt_rev` to `tgt_git_rev` might be dropped, as it would be always present, but `None` in OBS).

I feel that might avoid some dances like checking for the attribute with hasattr. What do you think?

plat/gitea.py
```diff
@@ -0,0 +1,316 @@
...
+class RequestAction:
+    """Stub action class"""
+    def __init__(self, type, src_project, src_package, src_rev, tgt_project, tgt_package, tgt_rev):
+        self.type = type
+        self.src_project = src_project
+        self.src_package = src_package
+        self.src_rev = src_rev
+        self.tgt_project = tgt_project
+        self.tgt_package = tgt_package
+        self.tgt_rev = tgt_rev
```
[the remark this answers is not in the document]
g7: no, not a blocker for me, absolutely (and if we can get it shared between the platforms would be even better)

gleidi-suse:
done

gleidi-suse:
All requirements from https://confluence.suse.com/spaces/projectmanagement/pages/1941045642/Bugowner+Review+Bot+Integration+Workflow are implemented.

gleidi-suse:
superseded by #3306

plat/gitea.py
```diff
@@ -290,6 +296,11 @@ def get_request_list_with_history(
     def get_staging_api(self, project):
         return StagingAPI(project, self.api)
 
+    def get_user(self, name):
```
YoukouTenhouin, line 299: This PR added some additional interface, maybe we should also added some prototypes in `plat/base.py` and add a stub in `plat/obs.py`? (As the intention for the platform interfaces are a common interface across different platforms)

scm/git.py
```diff
@@ -17,6 +17,76 @@ def __init__(self, logger, base_url):
...
+            if revision_name not in [b.name for b in repo.branches]:
+                repo.git.checkout("-b", revision_name, f"{remote}/{revision}")
+            else:
+                repo.git.switch(revision_name)
+        else:
+            repo.git.checkout(revision, "--")
+
+        return repo
+
+    def package_url(self, target_project: str, target_package: str) -> str:
```
YoukouTenhouin, line 77: Is there a reason why we have to expose this function, instead of making it an internal one and pass those parameters from `submodule_diff()` or `clone_repository()`?

check_bugowner.py
```diff
@@ -71,13 +106,401 @@ def exists_in(self, project, package):
...
+
+        # Cleanup branches
+        self.scm.checkout_revision(repo, base_revision)
+        repo.git.branch("-D", head_revision_name if head_revision_name else head_revision)
+        repo.git.branch("-D", "-r", f"{head_remote_name}/{head_revision}")
+
+        return is_valid
+
+    def check_source_submission(self, src_project, src_package, src_rev, target_project, target_package):
+        if self.platform.name == "GITEA":
```
YoukouTenhouin, line 473: Since we are differentiating between platform anyways, I think it might be a better idea to directly implement those platform-specific functions as private functions here, instead of trying to add them as a part of platform interface.

plat/gitea.py
gleidi-suse: good point, fixed thanks

scm/git.py
gleidi-suse: EDIT: it is because the procedure is used in many places in `check_bugowner`

check_bugowner.py
gleidi-suse: you mean `get_user`, right? ok I'll move it in this class thanks

PR #3297, opened 2026-02-07
container_cleaner.py: keep the last 30 snapshots

johanneskastl:
Due to the rapid release of Tumbleweed, keeping the last 5 snapshots per architecture means we only have a week old image for x86_64 at most...

Keeping 30 snapshots should not be a big hit on storage, according to @dirkmueller 

Discussion see [this thread](https://lists.opensuse.org/archives/list/buildservice@lists.opensuse.org/thread/WDDABPDTXTUZRDWNIDIUO4GLLXHJZS7D/)

### DIFF
--- a/container_cleaner.py
+++ b/container_cleaner.py
@@ -86,7 +86,7 @@ class ContainerCleaner(ToolBase.ToolBase):
             for srccontainer in buckets[package]:
                 contributes = False
                 for arch in srccontainerarchs[srccontainer]:
-                    if archs_found[arch] < 5:
+                    if archs_found[arch] < 30:
                         archs_found[arch] += 1
                         contributes = True
 


Vogtinator:
> Keeping 30 snapshots should not be a big hit on storage, according to @dirkmueller

I think it's a very big hit, as it needs 6x the storage and likely processing power as well: For every additional image in the project, publishing gets a bit slower. There's no deduplication, as most images contain the TW snapshot number and get rebuilt for every new snapshot.

Currently with 5 snapshots it needs about 608GB:

```
> osc api /build/openSUSE:Containers:Tumbleweed/_result?view=binarylist | xmllint - --xpath "/resultlist/result/binarylist/binary[not(contains(@filename, '.tar'))]/@size" | awk -F'"' 'BEGIN { total=0 } /size/ { total += $2 } END { print total " B" }' | qalc -f -
608016158483 bytes ≈ 608,0161585 GB
```

With 30 snapshots it would be over 3.5TB...

Vogtinator:
> Discussion see [this thread](https://lists.opensuse.org/archives/list/buildservice@lists.opensuse.org/thread/WDDABPDTXTUZRDWNIDIUO4GLLXHJZS7D/)

For Tumbleweed, using `:latest` is the most sensible option: You don't get any security or bug fixes in previous images, and if you install packages, you always get the latest snapshot (unless you try to use the https://download.opensuse.org/history/ repo).

For Leap, container images are (for now) kept indefinitely.

Vogtinator:
> Currently with 5 snapshots it needs about 608GB:

I asked Adrian to have a look on the backend server directly and apparently it's about 389GiB.

PR #3301, opened 2026-02-18
check-source: add systemd service and timer for running on gitea

eliroca:


eliroca:
@Vogtinator what about this one? ^^

systemd/osrt-check-source-gitea@.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=service for osrt-check_source on gitea running as %i
+After=network-online.target
+
+[Service]
+Type=oneshot
+Group=users
+User=osrt-check-source
+EnvironmentFile=/etc/default/osrt-check-source-gitea-%i.env
+ExecStart=/usr/bin/osrt-check_source --scm=git --platform=gitea --gitea-url="https://${GITEA_DOMAIN}"  --git-base-url="https://${GITEA_BOT_USER}:${GITEA_ACCESS_TOKEN}@${GITEA_DOMAIN}" --user=%i review
```
Vogtinator: Nit:

```suggestion
ExecStart=/usr/bin/osrt-check_source --scm=git --platform=gitea --gitea-url="https://${GITEA_DOMAIN}" --git-base-url="https://${GITEA_BOT_USER}:${GITEA_ACCESS_TOKEN}@${GITEA_DOMAIN}" --user=%i review
```

systemd/osrt-check-source-gitea@.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=timer for osrt-check_source on gitea running as %i
+
+[Timer]
+OnBootSec=2min
+OnUnitActiveSec=5min
+Persistent=true
```
Vogtinator: The other timer units don't use Persistent=true, any reason to use it here? IMO if it runs all 5 min it makes no difference

eliroca: Agreed, no reason since we run the service very often

Vogtinator (approved):
LGTM.

I'm not sure if this needs %service_add_pre etc. macros? Those may not work anyway because it's a template unit...

eliroca:
> LGTM.
> 
> I'm not sure if this needs %service_add_pre etc. macros? Those may not work anyway because it's a template unit...

Thanks! I don't think the macros are useful with template units either, but I only looked briefly into it.

eliroca:
Anything missing here? @Vogtinator

eliroca:
Thanks!

PR #3302, opened 2026-02-24
ReviewBot: consume the generator when using search_review(), and ensure the behaviour between platform is consistent

g7:
~Users of search_review() expect a list, not a generator.~

### DIFF
--- a/ReviewBot.py
+++ b/ReviewBot.py
@@ -691,7 +691,7 @@ class ReviewBot(object):
         return self.platform.can_accept_review(req, review_user=self.review_user, review_group=self.review_group)
 
     def set_request_ids_search_review(self):
-        self.requests = self.platform.search_review(review_user=self.review_user, review_group=self.review_group)
+        self.requests = list(self.platform.search_review(review_user=self.review_user, review_group=self.review_group))
 
     # also used by openqabot
     def ids_project(self, project, typename):
--- a/docs/bot-framework/bot_framework.md
+++ b/docs/bot-framework/bot_framework.md
@@ -109,7 +109,7 @@ The Platform layer provides an abstraction for interacting with different collab
         Common kwargs:
             review_user: Filter by reviewer username
             review_group: Filter by reviewer group
-        Returns: List/generator of Request objects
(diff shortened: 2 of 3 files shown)


g7:
Alternatively, since the only (known) user is ReviewBot.py, we can force a list there.

g7:
v2:

as agreed in Slack - let's turn it the other way: turn obs's search_platform() a generator too, and ensure that ReviewBot consumes it and stores it as a list.

ReviewBot.py
```diff
@@ -691,7 +691,7 @@ def can_accept_review(self, req):
         return self.platform.can_accept_review(req, review_user=self.review_user, review_group=self.review_group)
 
     def set_request_ids_search_review(self):
-        self.requests = self.platform.search_review(review_user=self.review_user, review_group=self.review_group)
+        self.requests = list(self.platform.search_review(review_user=self.review_user, review_group=self.review_group))
```
YoukouTenhouin, line 694: That's reasonable, but is there really a hard dependency for `self.requests` being a list at the moment tho? If not I think maybe we can take this chance to change this to a generator and offer a chance for a small optimization. What do you think?

g7: I think it would be nice, just worried about the usage elsewhere. From a rapid skimming through ReviewBot.py:

* ReviewBot.ids_project should be moved to a generator: https://github.com/g7/openSUSE-release-tools/blob/28b165e775a5371ee3a866058b74907293219696/ReviewBot.py#L697
* ReviewBot.set_requests_id should be moved too: https://github.com/g7/openSUSE-release-tools/blob/28b165e775a5371ee3a866058b74907293219696/ReviewBot.py#L236
* ReviewBot.set_request_ids_search_review should be moved as well (this line)


Doesn't look that bad to be honest, but perhaps that can be done later in another PR, what do you think?

YoukouTenhouin: A follow up pull request later sounds good. Thanks for the investigation!

YoukouTenhouin:
LGTM

g7:
Thanks! CI Failures unrelated, merging

PR #3305, opened 2026-03-11
fix: pass in branch name in checkout_package

YoukouTenhouin:
on Gitea the submission target might contain different branches as well. Update checkout_package to also pass in target revision.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -135,16 +135,17 @@ class CheckSource(ReviewBot.ReviewBot):
 
         return ret
 
-    def check_source_submission(
+    def check_source_submission_v2(
             self,
             source_project: str,
             source_package: str,
             source_revision: str,
             target_project: str,
-            target_package: str
+            target_package: str,
+            target_rev: Optional[str]
     ) -> bool:
-        super(CheckSource, self).check_source_submission(source_project,
-                                                         source_package, source_revision, target_project, target_package)
+        super(CheckSource, self).check_source_submission_v2(source_project,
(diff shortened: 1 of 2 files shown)


check_source.py
```diff
@@ -131,16 +131,17 @@ def _package_source_parse_obs(self, project, package, revision=None, target_pack
...
-    def check_source_submission(
+    def check_source_submission_v2(
             self,
             source_project: str,
             source_package: str,
             source_revision: str,
             target_project: str,
-            target_package: str
+            target_package: str,
+            target_rev: str | None
```
eliroca: Please change to work on older python versions:

```suggestion
            target_rev: Optional[str]
```

YoukouTenhouin: Nice catch, I updated the notation in `ReviewBot.py` but forgot about this one. Thanks for pointing this out!

eliroca (approved):
LGTM - thank you!

eliroca:
@g7 I'd like this one merged as well, please take a look when you can

g7 (approved):
Sounds good to me. We should also change the documentation in `docs/bot-framework/` but can be done later. Thanks!

For reference, these are the current users:
```
origin-manager.py
legal-auto.py
git-installcheck.py
git-pkglistgen.py
check_tags_in_requests.py
check_bugowner.py
abichecker/abichecker.py
```

This is also going to be very useful for git-installcheck, git-pkglistgen and check_bugowner

PR #3306, opened 2026-03-19
 check_bugowner: Add Git support and systemd unit file.

gleidi-suse:
Supersedes #3285

### DIFF
--- a/check_bugowner.py
+++ b/check_bugowner.py
@@ -2,9 +2,15 @@
 
 # SPDX-License-Identifier: MIT
 
+import json
+import os
+import shutil
 import sys
 import re
+from pathlib import Path
+from typing import Set
 
+from git.config import GitConfigParser
 from urllib.error import HTTPError
 
 import osc.conf
@@ -12,6 +18,7 @@ import osc.core
 import ReviewBot
 
 http_GET = osc.core.http_GET
+MAINTAINERSHIP_FILE = "_maintainership.json"
 
 
 class CheckerBugowner(ReviewBot.ReviewBot):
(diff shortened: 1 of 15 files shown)


check_bugowner.py
```diff
@@ -71,13 +107,411 @@ def exists_in(self, project, package):
...
+        self.scm.checkout_revision(repo, base_revision)
+        repo.git.branch("-D", head_revision_name)
+        repo.git.branch("-D", "-r", f"{head_remote_name}/{head_revision}")
+
+        return is_valid
+
+    def check_source_submission(self, src_project, src_package, src_rev, target_project, target_package):
+        if self.platform.name == "GITEA":
+            return self._gitea_check_source_submission(src_project, src_package, src_rev, target_project, target_package)
+        else:
```
g7: what do you think about


```suggestion
        elif self.platform.name == "OBS":
            return self._obs_check_source_submission(src_project, src_package, src_rev, target_project, target_package)
        else:
           raise Exception("Unsupported platform")
```

g7 (approved):
one nitpick that can also be addressed later, otherwise looks good to me, thanks!

check_bugowner.py
gleidi-suse: looks like a good idea, thank you!

g7 (approved):
LGTM, thanks!

gleidi-suse:
https://github.com/openSUSE/openSUSE-release-tools/actions/runs/24332331280/job/71040853429?pr=3306 will fail until OCI images over in OBS are rebuilt with the updated specfile

PR #3307, opened 2026-03-19
platform: gitea: support Reviews and Labels

g7:
Moved from #3287.

### DIFF
--- a/plat/gitea.py
+++ b/plat/gitea.py
@@ -235,6 +235,7 @@ class Request:
         self._issues = None
 
         # Gitea specific attributes
+        self._labels = set()
         self._owner = None
         self._repo = None
         self._pr_id = None
@@ -264,6 +265,9 @@ class Request:
             for review in reviews_json:
                 if not review.get("dismissed", False):
                     self.reviews.append(Request.format_review(review))
+
+        self._labels = {x["name"] for x in json["labels"]}
+
 
 class ProjectConfig:
     """Project Config implemented for Gitea"""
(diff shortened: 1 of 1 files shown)


gleidi-suse (approved):
lgtm

plat/gitea.py
```diff
@@ -197,30 +235,37 @@ def _init_attributes(self):
         self._issues = None
 
         # Gitea specific attributes
+        self._labels = set()
         self._owner = None
         self._repo = None
         self._pr_id = None
 
-    def read(self, json, owner, repo):
+    def read(self, request_json, reviews_json, owner, repo):
```
YoukouTenhouin: I'm thinking whether we should keep the signature of this function match `osc.core.Request.read()`. On the one hand in principle this `Request` is supposed to mimic `osc.core.Request`; on the other hand this specific function is unlikely to be directly replaceable anyways, and forcing the signature to be the same will mean that we need to do a HTTP request inside "read", which semantically is kind of wrong... Opinions?

YoukouTenhouin:
Dropped a comment; LGTM otherwise. Thanks!

plat/gitea.py
gleidi-suse: it could be an optional keyword argument, so that the default signature stays the same for consistency of the API but the platform specific behavior can still be conditionally implemented

YoukouTenhouin: An optional keyword argument sounds good, thanks for the idea!

g7: thanks for the review and the tip! Sounds good to me too. Implemented

g7:
Thanks!

g7:
CI failure is unrelated, merging

PR #3310, opened 2026-04-13
check_source.py: Allow forks to be named differently

eliroca:


eliroca:
Ready to be reviewed, got a LGTM from @YoukouTenhouin

check_source.py
```diff
@@ -158,6 +158,10 @@ def check_source_submission(
             # XXX stub
             kind = 'source'
 
+        # allow forks to be named differently
+        if self.platform_type == "GITEA":
+            self.in_air_rename_allow = True
```
g7: The change makes sense to me. Thanks!

However, what about moving it inside `target_project_config()` further up? Looks like we are already overriding that setting on maintenance_incident there, so perhaps we can add this as well?

Functionally nothing changes, but looks like a better place to me to have the override.

eliroca: Agreed, moved further up. Thanks!

g7 (approved):
Thanks!

PR #3311, opened 2026-04-13
Gitea review bots: pin RPM requirements to same version

eliroca:


dist/package/openSUSE-release-tools.spec
```diff
@@ -142,8 +142,9 @@ Requires:       obs-service-download_files
 Requires:       obs-service-source_validator
 Requires:       osclib = %{version}
 Requires:       perl-Text-Diff
-Requires:       %{name}-plat
-Requires:       %{name}-scm
+Requires:       %{name}-plat = %{version}
```
Vogtinator: The `= %{version}` isn't necessary here, if the subpackages have a `Requires:       %{name} = %{version}`

eliroca: Thanks for pointing that out!

Vogtinator: FWICT they actually don't, so we'd need the `= %{version}` for the -plat and -scm requires but the `Requires: %{name}` isn't necessary?

eliroca: It's needed for pulling ReviewBot.py shipped by openSUSE-release-tools package

eliroca: Added version requirement back

Vogtinator (approved):
At some point, splitting the packages stops making sense, let's see

PR #3312, opened 2026-04-14
freeze: Do not resign binaries when aggregating from Ring1

DimStar77:
Rings:0-Bootstrap binaries are copied by means of _aggregate into
the letter stagings. Rings:* and Stagings use the same signing key,
we can thus avoid resigning all RPMs on aggregate and substantially
speed up `osc staging freeze`


g7 (approved):
Looks good to me, and the CI failure is unrelated.

Vogtinator (approved):

PR #3313, opened 2026-04-15
fix: check_bugowner: Fast forward when fetching non current local branches.

gleidi-suse:


g7 (approved):
Looks good to me, thanks!

PR #3317, opened 2026-04-22
check_source: normalize names between git repoisitory and spec file

DimStar77:
In git (src.opensuse.org), repository names are forced to replace `+` by `_`
Which results in some packages like libsigc++40 being named libsigc__40 in git

On the OBS side (openSUSE:Factory) for now we are still tracking the original name
(until migrating to git). Accept this 'in-air rename' when submitting from a git
devel project (e.g GNOME:Factory) to openSUSE:Factory, as well as the spec file
having the proper libsigc++40 name instead of the git re-written libsigc__40.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -93,6 +93,10 @@ class CheckSource(ReviewBot.ReviewBot):
 
         return target_package == package
 
+    # Helper to treat + and _ as equivalent for comparison
+    def git_normalize(self, name):
+        return name.replace('+', '_')
+
     def package_source_parse(self, project, package, revision=None, target_package=None):
         # XXX should we refactor this out?
         if self.platform_type == "OBS":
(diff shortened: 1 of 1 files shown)


Vogtinator (approved):
FTR, the regex appears to be `[^-.\w]` but in oS:F, only + currently matches that

DimStar77:
> FTR, the regex appears to be `[^-.\w]` but in oS:F, only + currently matches that

* `-` seems not to be an issue (?): https://src.opensuse.org/GNOME/gnome-2048
* `.` seems also to be fine: https://src.opensuse.org/GNOME/gnome-next.x86_64
* `\w` is probably an issue - but at least so far we never accepted whitespaces in names; so that's not new at least

Vogtinator:
> > FTR, the regex appears to be `[^-.\w]` but in oS:F, only + currently matches that
> 
> * `-` seems not to be an issue (?): https://src.opensuse.org/GNOME/gnome-2048
> * `.` seems also to be fine: https://src.opensuse.org/GNOME/gnome-next.x86_64
> * `\w` is probably an issue - but at least so far we never accepted whitespaces in names; so that's not new at least

I think you read it wrong: Everything except `-`, `.` and word characters are replaced by `_`

DimStar77:
> I think you read it wrong: Everything except `-`, `.` and word characters are replaced by `_`

image

PR #3319, opened 2026-04-24
check_bugowner: Add support for new _maintainership.json format.

gleidi-suse:
In accord with https://src.opensuse.org/git-workflow/autogits/src/branch/main/workflow-pr/README.md#new-format

### DIFF
--- a/check_bugowner.py
+++ b/check_bugowner.py
@@ -7,8 +7,9 @@ import json
 import os
 import sys
 import re
+from enum import IntEnum
 from pathlib import Path
-from typing import List, Set
+from typing import List, Set, Dict
 from cmdln import CmdlnOptionParser
 
 try:
@@ -30,6 +31,11 @@ MAINTAINERSHIP_FILE = "_maintainership.json"
 WHITELIST_FILE = "whitelist_maintainership.json"
 
 
+class MaintainershipFormat(IntEnum):
+    unversioned = 0
+    one_dot_o = 1
+
+
 class CheckerBugowner(ReviewBot.ReviewBot):
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


check_bugowner.py
```diff
@@ -30,13 +31,19 @@
 WHITELIST_FILE = "whitelist_maintainership.json"
 
 
+class MaintainershipFormat(Enum):
```
g7: super mega nitpick: what about an `IntEnum`? Should make comparisons easier in future (we can compare against integers)

check_bugowner.py
```diff
@@ -361,54 +390,75 @@ def _gitea_email(self, owner):
...
+                users_message = [f"Gitea users: {users}.{ldap_status}"]
+            else:
+                users_message = []
+
+            groups_message = []
+            if self.maintainership_format_version == MaintainershipFormat.one_dot_o:
+                groups = self.maintained[package].get("groups", False)
+                if groups:
+                    groups = ', '.join('[`' + g + '`](https://build.suse.de/groups/' + g + ')' for g in groups)
+                    groups_message = [f"OBS groups: {groups}."]
```
g7, line 455: do we need to check for group existence? We have an export here https://src.suse.de/obs/groups/src/branch/main which can avoid having to query IBS

g7 (commented):
Thanks! Changes look good to me, I have one super-mega-nitpick and mostly a question about the scope (should we expand the group and check inside?)

check_bugowner.py
gleidi-suse: good point, fixed.

check_bugowner.py
gleidi-suse: yes, I think we should check both users and groups and also decline the review in case either:

- a user does not exist on Gitea
- a user is not active anymore on LDAP
- a group does not exist on IBS. when we query for the group existance we could get back the list of members and display it.
- maybe check that the group exists in Bugzilla?

but I would do this in a separate change as it's more of broader new functionality. what do you think?

g7: yes, agreed, thanks!

g7 (approved):
Thanks!

PR #3321, opened 2026-04-27
osclib: repochecks: maintain compatibility with Python 3.6

g7:
The "text" kwarg has been added to subprocess.run in Python 3.7, as an alias of "universal_newlines".

Use the latter so that we can keep supporting Python 3.6, which is the system interpreter in SLE15 and we're still using.


gleidi-suse (approved):
lgtm thank you!

PR #3324, opened 2026-05-04
check_bugowner: Allow overriding OSRT_BUGOWNER_CACHE_HOME for the systemd service.

gleidi-suse:
This allows to set the cache directory to mountpoints outside `/var`


g7 (approved):
Makes sense to me, thanks!

PR #3326, opened 2026-05-08
check_bugowner: Fix rendering of users.

gleidi-suse:


scm/git.py
```diff
@@ -58,6 +59,14 @@ def submodule_diff(
     def is_repo(repo):
         if isinstance(repo, git.Repo):
             return True
+        elif isinstance(repo, str) and not Path(repo).exists():
```
g7: nit: how about creating a path object if it is not supplied and just check that one?


```suggestion
        path = repo if isinstance(repo, Path) else Path(repo)
        return path.is_dir()
```

also: if the path is a directory (as expected), shouldn't we still check if it is an actual git repo?

```
        def is_repo(repo):
            if isinstance(repo, git.Repo):
                return True
            else:
                path = repo if isinstance(repo, Path) else Path(repo)
                if not path.is_dir():
                    return False

                try:
                    git.Repo(path)
                except git.exc.InvalidGitRepositoryError:
                    return False
                else:
                    return True
```

g7 (commented):
I'm unsure on the is_repo() changes, rest looks good to me.

scm/git.py
gleidi-suse: Good point, thank you. I think I fixed this, let me know in case you have further feedback!

g7 (approved):
looks good to me, thanks!

PR #3328, opened 2026-05-29
Avoid division by zero.

ana:


Vogtinator:
Hm, maybe 100% would be more appropriate as a repo with nothing to build is effectively done? Or `None`, so that the caller can interpret it?

ana:
> Hm, maybe 100% would be more appropriate as a repo with nothing to build is effectively done? Or `None`, so that the caller can interpret it?

`None` seems better.

Vogtinator (commented):
I don't see any callers in this repo, so 🤷

Vogtinator (approved):

ana:
> I don't see any callers in this repo, so 🤷

That's probably why the issue lasted for so long. I used this function in my local dashboard and that's who I ran into the issue.

PR #3330, opened 2026-06-10
Add new command syncnewpackages to devel_update.sh

adrianschroeter:
This should be used to add new packages in open submit requests
already to our devel package definitions.
    
This is supposed to be called before a new package gets
accepted to Factory to help Dirks git import script for
OBS native packages.

### DIFF
--- a/devel_update.sh
+++ b/devel_update.sh
@@ -7,6 +7,7 @@
 #      set <prj> <pkg>
 #      rm  <prj> <pkg>
 #      sync
+#      syncnewpackages
 #
 #  DEVEL_PACKAGES env should point to the devel_packages clone from
 #  repo above, otherwise will look in CWD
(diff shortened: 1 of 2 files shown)


devel_update.sh
```diff
@@ -43,6 +44,24 @@ function rmdevel {
...
+    declare -A packages
+    while IFS="=" read -r key value; do
+        [[ -z "$key" || "$key" =~ ^# ]] && continue
+        packages["$key"]=1
+    done < <(osc ls openSUSE:Factory 2> /dev/null | grep -v :)
+
+    echo Check for new packages in submit requests..
+    osc api '/request?view=collection&project=openSUSE%3AFactory&states=new,review' | \
+      xmlstarlet sel -t -m "//request/action" -v "target/@package" -o " " -v "source/@project" -n | while read target project; do
+        [ "$target" == "openSUSE:Factory" ] && continue
```
DimStar77: shouldn't that be 
`[ "$project" == "openSUSE:Factory" ] && continue` ? `$target` is the package name if I read this right

DimStar77 (changes requested):

devel_update.sh
adrianschroeter: unfortunatly it is correct this way, because when you have maintenance requests with factory as source, there is no package in the target. but let me do a better implementation nevertheless ...

adrianschroeter:
hopefully easier to read now

DimStar77 (approved):

ana:
Hi @adrianschroeter 
 
> This is supposed to be called before a new package gets accepted to Factory to help Dirks git import script for OBS native packages.

If I understand correctly, you're expecting Factory release managers to run this script whenever a new package is accepted.
Adding another manual step to the workflow doesn't sound ideal. Is there any reason this can't be handled automatically within the Git import script you from Dirk?

PR #3338, opened 2026-07-06
Add systemd unit for legal-auto checker.

gleidi-suse:


systemd/osrt-legal-auto@.service
```diff
@@ -0,0 +1,11 @@
...
+Description=service for osrt_legal-auto running as %i
+After=network-online.target
+
+[Service]
+Type=oneshot
+User=osrt-staging
+EnvironmentFile=/etc/default/osrt-legal-auto-%i.env
+WorkingDirectory=/var/lib/osrt-staging/legal-auto
+ExecStartPre=/bin/bash -xc '/usr/bin/systemctl is-active --quiet osrt-legal-auto@%i.service && exit 1 || exit 0'
+ExecStart=/bin/bash -xc '/usr/bin/legal-auto -A ${LEGAL_AUTO_OBS_API} --debug --legaldb "${LEGAL_AUTO_CAVIL_URL}" --group=legal-auto review && /usr/bin/legal-auto -A "${LEGAL_AUTO_OBS_API}" --debug --legaldb "${LEGAL_AUTO_CAVIL_URL}" --user=licensedigger review'
```
g7: with `oneshot` services we can use multiple `ExecStart=` entries - this way we can drop the `&&` and have better feedback if one of those two commands fails

also I think it would be cool if we can configure the group and/or user to check for in the configuration file

g7:
Thanks! A really minor comment but rest looks good to me

systemd/osrt-legal-auto@.service
gleidi-suse: I was not aware, thanks! I should have addressed both your points. Let me know if you have further feedback.

systemd/osrt-legal-auto@.service
```diff
@@ -0,0 +1,12 @@
...
+Description=service for osrt_legal-auto running as %i
+After=network-online.target
+
+[Service]
+Type=oneshot
+User=osrt-staging
+EnvironmentFile=/etc/default/osrt-legal-auto-%i.env
+WorkingDirectory=/var/lib/osrt-staging/legal-auto
+ExecStartPre=/bin/bash -xc '/usr/bin/systemctl is-active --quiet osrt-legal-auto@%i.service && exit 1 || exit 0'
+ExecStart=/usr/bin/legal-auto -A "${LEGAL_AUTO_OBS_API}" --debug --legaldb "${LEGAL_AUTO_CAVIL_URL}" --group="${LEGAL_AUTO_REVIEW_GROUP}" review
```
g7: can we run this only if LEGAL_AUTO_REVIEW_GROUP is set?

systemd/osrt-legal-auto@.service
```diff
@@ -0,0 +1,12 @@
...
+ExecStart=/usr/bin/legal-auto -A "${LEGAL_AUTO_OBS_API}" --debug --legaldb "${LEGAL_AUTO_CAVIL_URL}" --user="${LEGAL_AUTO_REVIEW_USER}" review
```
g7: Likewise, can we run it only if LEGAL_AUTO_REVIEW_USER is set?

g7 (commented):
Awesome, just another minor comment rest looks good, thanks!

systemd/osrt-legal-auto@.service
gleidi-suse: good point, should be fixed now . thanks!

systemd/osrt-legal-auto@.service
gleidi-suse: same, please let me know if it is what you meant. thanks

g7 (approved):
Thanks a lot! Looks good to me.

dist/package/openSUSE-release-tools.spec
```diff
@@ -19,7 +19,7 @@
 %global __provides_exclude ^perl.*
 %define source_dir openSUSE-release-tools
 %define announcer_filename factory-package-news
-%define services osrt-slsa.target osrt-check-bugowner-gitea@.service osrt-git-installcheck@.service osrt-relpkggen@.timer osrt-relpkggen@.service osrt-pkglistgen@.timer osrt-pkglistgen@.service
+%define services osrt-slsa.target osrt-check-bugowner-gitea@.service osrt-git-installcheck@.service osrt-legal-auto@.service osrt-legal-auto-import@.service osrt-relpkggen@.timer osrt-relpkggen@.service osrt-pkglistgen@.timer osrt-pkglistgen@.service
```
DimStar77, line 22: %services is used in  %post to enable/trigger/restart the services

this does not work with  @-template services without instance-names

PR #3340, opened 2026-07-17
devel_update: Use POSIX locale for sorting

DimStar77:
We need to be sure that everybody uses the same LANG to sort the file otherwise the file just keeps re-ordering depending on who runs the script on what machine.

### DIFF
--- a/devel_update.sh
+++ b/devel_update.sh
@@ -28,7 +28,7 @@ function setdevel {
         exit 10
     fi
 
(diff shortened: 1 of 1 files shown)


DimStar77:
@AdamMajer - any objection?

Without this, things like https://src.opensuse.org/openSUSE/Factory/commit/964f98740e9eee6f809d7034b9484c1b5266d7294eb694d0dbfa0b52955adfef are going to happen more often

AdamMajer:
No objections. Could also use LANG=C

PR #3343, opened 2026-07-24
gocd: run SLE.Legal.Import at 02:00 instead of midnight

Volcano28:
Shift the daily legal import timer from 00:00 to 02:00 to avoid the midnight run window.


Vogtinator (approved):

Vogtinator (changes requested):
LGTM but needs an erb run

Vogtinator (approved):

Vogtinator:
02:00 may also cause issues with DST

PR #3344, opened 2026-07-31
gocd: enable shallow_clone for the botmaster suse-repos and opensuse-repos materials

Volcano28:
suse-repos.git is 1.37 GiB / 893k commits and opensuse-repos.git is 822 MiB / 939k commits, but together they only hold ~7 MB of actual content, so a cold checkout costs up to 17 minutes; shallow_clone brings the agent-side checkout down to ~12 MB and makes it near-instant.


Vogtinator (approved):

DimStar77 (approved):
Great find!

PR #3350, opened 2026-08-11
staging: update frozenage to show all letter projects

ana:
Also show the experimental Gcc7 and Glibc in the list. Skip all the ADI projects since now we rely on api.get_staging_projects

Use f-strings to make it easier to format the output

### DIFF
--- a/osc-staging.py
+++ b/osc-staging.py
@@ -429,13 +429,13 @@ def do_staging(self, subcmd, opts, *args):
                 print(Fore.YELLOW + prj)
                 FreezeCommand(api).perform(prj, copy_bootstrap=opts.bootstrap)
         elif cmd == 'frozenage':
-            projects = api.get_staging_projects_short() if len(args) == 1 else args[1:]
+            projects = api.get_staging_projects() if len(args) == 1 else args[1:]
+            projects = [prj for prj in projects if not api.is_adi_project(prj)]
             for prj in projects:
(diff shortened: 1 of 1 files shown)


osc-staging.py
```diff
@@ -429,13 +429,12 @@ def do_staging(self, subcmd, opts, *args):
...
+            projects = api.get_staging_projects() if len(args) == 1 else args[1:]
+            projects = [prj for prj in projects if not api.is_adi_project(prj)]
             for prj in projects:
-                prj = api.prj_from_letter(prj)
-                print('{} last frozen {}{:.1f} days ago'.format(
-                    Fore.YELLOW + prj + Fore.RESET,
-                    Fore.GREEN if api.prj_frozen_enough(prj) else Fore.RED,
-                    api.days_since_last_freeze(prj)))
+                print(f'{Fore.YELLOW}{api.prj_from_letter(prj):<30}{Fore.RESET} '
+                      f'last frozen {Fore.GREEN if api.prj_frozen_enough(prj) else Fore.RED}'
```
Vogtinator, line 437: That may need `api.prj_from_letter(prj)` in case args are used

Vogtinator (changes requested):

osc-staging.py
ana: Thank you, the line is back.

Vogtinator (approved):

PR #3351, opened 2026-08-13
pkglistgen: Force 2nd run for prodcomposer

DimStar77:
For Tumbleweed, we currently calculate two media, the legacy yast dvd and the modern agama offline image. Once the release/yast dvd is calculated and checked in, the repo is 'dirty', which stops the calculation of the prodcomposer based product. As we know at this time that the repo was effectively clean, and we polluted it, we can just force the calculation of the new DVD as well instead of waiting a whole cycle to catch a repo-clean state again.

### DIFF
--- a/gocd/pkglistgen.opensuse.gocd.yaml.erb
+++ b/gocd/pkglistgen.opensuse.gocd.yaml.erb
@@ -33,7 +33,7 @@ pipelines:
               - script: |
                   python3 -u ./pkglistgen.py -d -A https://api.opensuse.org update_and_solve -p <%= project[0] %><%= options %> --only-release-packages --force
                   python3 -u ./pkglistgen.py -d -A https://api.opensuse.org update_and_solve -p <%= project[0] %><%= options %>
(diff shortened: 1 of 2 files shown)


nilxam (approved):

gocd/pkglistgen.opensuse.gocd.yaml.erb
```diff
@@ -33,7 +33,7 @@ pipelines:
               - script: |
                   python3 -u ./pkglistgen.py -d -A https://api.opensuse.org update_and_solve -p <%= project[0] %><%= options %> --only-release-packages --force
                   python3 -u ./pkglistgen.py -d -A https://api.opensuse.org update_and_solve -p <%= project[0] %><%= options %>
-                  python3 -u ./pkglistgen.py -d -A https://api.opensuse.org update_and_solve -p <%= project[0] %><%= options %> --engine=product_composer
```
Vogtinator, line 36: IMO add a comment

PR #3352, opened 2026-08-18
Freeze: also setup product repository for Agama offline image

DimStar77:
The code detects if the staging to be frozen has a product repo, which matches our usecase very well. But the repo needs to be built against images (for agama-baseiso) and bootstrap_copy) as product compose only takes packages from referenced paths.

### DIFF
--- a/osclib/freeze_command.py
+++ b/osclib/freeze_command.py
@@ -200,6 +200,13 @@ class FreezeCommand(object):
                 a = ET.SubElement(r, 'arch')
                 a.text = 'local'
 
+            # The 'product' repo needs to build against 'images' and 'bootstrap-copy'
+            # images is used for agama-baseiso (baseiso-containment of agama installer)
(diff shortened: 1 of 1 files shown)


osclib/freeze_command.py
```diff
@@ -200,6 +200,13 @@ def prj_meta_for_bootstrap_copy(self):
                 a = ET.SubElement(r, 'arch')
                 a.text = 'local'
 
+            # The 'product' repo needs to build against 'images' and 'bootstrap-copy'
+            # images is used for agama-baseiso (baseiso-containment of agama installer)
+            # bootstrapc-copy is used for Ring0 packages to be availale for the product
```
ana: Two typo in the comment: *bootstrap_copy* is used for Ring0 packages to be *available* for the product

Otherwise looks good to me!

ana (changes requested):
ok

ana (approved):

PR #3356, opened 2026-09-02
check_source.py: Do not skip unchanged URLs

Vogtinator:
- The URLs may contain macros like %{version}
- The file may have changed but the URL stayed the same
- The URL may be dead meanwhile

https://build.opensuse.org/request/show/1374395 was incorrectly accepted.

### DIFF
--- a/check_source.py
+++ b/check_source.py
@@ -702,26 +702,7 @@ class CheckSource(ReviewBot.ReviewBot):
 
         return True
 
-    def _snipe_out_existing_urls(self, old, directory, specs):
-        if not os.path.isdir(old):
-            return
-        oldsources = self._mentioned_sources(old, specs)
-        for spec in specs:
-            specfn = os.path.join(directory, spec)
-            nspecfn = specfn + '.new'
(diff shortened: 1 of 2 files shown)


DimStar77:
Was that part of the anti-frustration-task when submitting from Factory to Leap when URLs happened to have vanished?

Other than that: how did we survive this long? :)

Vogtinator:
> Was that part of the anti-frustration-task when submitting from Factory to Leap when URLs happened to have vanished?

Possibly.

> Other than that: how did we survive this long? :)

Luck? I wonder how many URLs are currently broken.

DimStar77:
I had seen declines for invalid tarballs in the recent times; not sure how they came to be; will be a fun excercise to run through all 17k packages to check them

Vogtinator:
If this turns out to be too annoying, we could possibly weaken it again by skipping the check only if the file is unchanged and the URL contains no macros. Though I doubt that that helps much, source URLs likely mention %{version}.

DimStar77 (approved):

PR #3359, opened 2026-09-10
gocd: Add configuration for openSUSE:Factory:LegacyARM

ggardet:


Vogtinator (changes requested):
Is TTM already set up? Locally it fails with `Exception: Failed to read ToTestManagerConfig`

Vogtinator (approved):

PR #3362, opened 2026-09-16
check-source: ignore .obs workflows dir

eliroca:


eliroca:
@DimStar77: not entirely sure we should ignore the warning, please take a look and let me know.

DimStar77:
That should have been fixed by https://github.com/openSUSE/obs-service-source_validator/pull/163

Filtering this out here sounds kinda wrong

eliroca:
Confirmed, I was running an older version. Thanks!

